PR #6, opened 2017-08-02
Implement maxDeltaY for PanGestureHandler on (for iOS only currently)

brentvatne:
No need to use this exact implementation, it was quick to just get a proof of concept working. It does seem like a valuable prop to add, however. I was able to implement SwipeActions like this:

```javascript
render() {
  const translateX = this.state._translateX;

  return (
    <View>
      <PanGestureHandler
        id={this.props.gestureId}
        minDeltaX={50}
        maxDeltaY={5}
        onGestureEvent={this._onGestureEvent}
        onHandlerStateChange={this._onHandlerStateChange}>
        <Animated.View
          style={{ transform: [{ translateX: translateX }] }}
          onLayout={this._onLayoutChildren}>
          {this.props.children}
        </Animated.View>
      </PanGestureHandler>

      {this._renderButtons()}
    </View>
  );
}
```

And on the parent `ScrollView` I wait for the pan gestures to fail.

### DIFF
--- a/ios/Handlers/RNGestureHandler.m
+++ b/ios/Handlers/RNGestureHandler.m
@@ -11,6 +11,9 @@
 #define TEST_MIN_IF_NOT_NAN(value, limit) \
     (!isnan(limit) && ((limit < 0 && value <= limit) || (limit >= 0 && value >= limit)))
 
+#define TEST_MAX_IF_NOT_NAN(value, max) \
+    (!isnan(max) && ((max < 0 && value < max) || (max >= 0 && value > max)))
+
 @interface RNDummyGestureRecognizer : UIGestureRecognizer
 @end
 
@@ -287,6 +290,7 @@ shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherG
 @property (nonatomic) CGFloat minVelocityX;
 @property (nonatomic) CGFloat minVelocityY;
 @property (nonatomic) CGFloat minVelocitySq;
+@property (nonatomic) CGFloat maxDeltaY;
 
 - (id)initWithGestureHandler:(RNGestureHandler*)gestureHandler;
 - (BOOL)shouldActivate;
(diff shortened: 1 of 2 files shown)


ios/Handlers/RNGestureHandler.m
```diff
@@ -356,12 +371,25 @@ - (BOOL)hasCustomCriteria
 {
     return !isnan(_minDistSq) || !isnan(_minDeltaX) || !isnan(_minDeltaY)
         || !isnan(_minOffsetX) || !isnan(_minOffsetY)
-        || !isnan(_minVelocityX) || !isnan(_minVelocityY) || !isnan(_minVelocitySq);
+        || !isnan(_minVelocityX) || !isnan(_minVelocityY) || !isnan(_minVelocitySq)
+        || !isnan(_maxDeltaY);
```
kmagiera: I think setting "max" bounds shouldn't affect `_hasCustomCriteria` as when you for example only set "max" but no min, then gesture will never fire. The purpose of `_hasCustomCriteria` was so that we can use platform default settings of `UIPanGestureHandler` if no other criteria are set, but the way it works is that if some "min" constrains are set we have a custom logic that is responsible for testing if handler should activate. So if there are no "min" props set and `hasCustomCriteria` is true the gesture will never activate.

ios/Handlers/RNGestureHandler.m
```diff
@@ -325,6 +330,10 @@ - (void)setMinimumNumberOfTouches:(NSUInteger)minimumNumberOfTouches
 
 - (void)touchesBegan:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
 {
+    // NOTE(brent): Probably there is a better place to update this cached value, didn't
+    // think too much about it.
+    _hasCustomCriteria = [self hasCustomCriteria];
```
kmagiera: This should go in the configure step of gesture handler subclass

kmagiera:
There are few small things that require fixing but I want to get this in so will fix this on my own. ty

PR #7, opened 2017-08-02
Don't warn when user passes in Animated.event with useNativeDriver: true to onGestureEvent

brentvatne:
I thought about validating the shape but I'm not sure if we want to depend on internal properties in the object shape, it feels brittle, so just used PropTypes.object instead.


kmagiera:
💯

PR #10, opened 2017-08-05
Support updating gesture handler config for PanGestureHandler on iOS

brentvatne:
... but only on PanGesutreHandler right now. Perhaps not super useful for you, just used this so I could get the pan/scroll example working (far from perfect, but working).


kmagiera:
❤️

PR #12, opened 2017-08-11
Add example for normal scrollview

brentvatne:


kmagiera:
Added more examples along with a master->detail navigation here https://github.com/kmagiera/react-native-gesture-handler/commit/c0c3ea1c97cccd77775cb950f8bc575e7e6e792e

But thanks anyways

PR #19, opened 2017-09-08
Change "react" peerDep version from >= 16 to > 15 to allow prereleases

ide:
`>= 16` doesn't include `16.0.0-alpha.12`, for example, while `> 15` does (but excludes 15.x releases).


kmagiera:
TY @ide !

PR #20, opened 2017-09-12
Fix replaceState and isMounted warnings

janicduplessis:
Those warning are probably implemented using getters so just getting a ref of those methods triggers it. Just bail out early for the 2 methods.


kmagiera:
Thanks @janicduplessis, I though the warning were caused by something else. Thanks for digging into it!

PR #21, opened 2017-09-12
Use prop-types instead of React.PropTypes

janicduplessis:
PropTypes is removed from React 16


kmagiera:
Cool @janicduplessis 
Thanks to you we are now clear from warning!

PR #23, opened 2017-09-13
Remove extra wrapper view for BorderlessButton on iOS

janicduplessis:
This removes an extra wrapper view on iOS that made styling of BorderlessButton's content hard. I hit this problem when I had some layout that required `flex: 1` on the button child to match the button size but that extra wrapper view caused issue.

It's actually possible to just remove it by setting the opacity on the actual button view.

Tested that it works well in an app I'm working on and also tested the demo app.

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -423,6 +423,8 @@ class BaseButton extends React.Component {
   }
 }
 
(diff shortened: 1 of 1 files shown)


kmagiera:
This looks good! Thanks

PR #25, opened 2017-09-13
Fix PropTypes for waitFor and simultaneousHandlers

brentvatne:
This fixes a couple of warnings: "Warning: Invalid argument supplied to oneOf, expected an instance of array."


kmagiera:
💯

PR #32, opened 2017-09-23
Fix shouldCancelWhenOutside with iOS PanGestureHandler

### LINKED ISSUE #30: PanGestureHandler does not cancel when outside.
Hey there, love what this project is trying to accomplish and I was hoping that it could assist with an use case in our app. What we have is a scroll view with a nested view that responds to both tap and pan gestures. Once activated with tap or pan we want the view to cancel the gesture and make the ScrollView the responder.

I've found that `shouldCancelWhenOutside` doesn't work with `PanGestureHandler`. I have [an example project here](https://github.com/wmcbain/gesture-handler-pan-issue-example/blob/master/main.js) that shows the issue. Is this expected? Related gif below as well.

Thanks for your help and appreciate the work you're doing. 👍🏻

example

**Edit:** Jumped into the source code and did a bit of debugging and I think I could handle the PR for this if that's alright. :smile:

wmcbain:
This ended up only being an issue on iOS. I found that a cancelled PanGestureHandler doesn't automatically make the ScrollView the responder. This seems like default behavior as it's the same on Android. Having the cancelled state come through though is what's important and that is taking place now.

Fixes #30


henrikra:
Protip: if you write `Fixes #30` in you description the issue will auto close when this gets merged ;)

ios/RNGestureHandler.m
```diff
@@ -370,6 +373,16 @@ - (void)touchesMoved:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
         self.state = UIGestureRecognizerStateFailed;
         return;
     }
+    if ((self.state == UIGestureRecognizerStatePossible || self.state == UIGestureRecognizerStateChanged) && _gestureHandler.shouldCancelWhenOutside) {
+        CGPoint pt = [self locationInView:self.view];
+        if (TEST_IS_OUT_OF_BOUNDS(pt, self.view.frame.size)) {
+            self.state = self.state == UIGestureRecognizerStatePossible
```
kmagiera: What's the reason you set `Cancelled` if the state was `Changed`? Wouldn't setting `Failed` work regardless? Also calling `reset` shouldn't be necessary (e.g. it is not being called a few lines earlier)

BTW:
One `self.state = ` should be enough :)

Example/panAndScroll/index.js
```diff
@@ -0,0 +1,121 @@
...
+const windowWidth = Dimensions.get('window').width;
+const circleRadius = 30;
+
+export class TapOrPan extends Component {
+  constructor(props) {
+    super(props);
+    this._translateX = new Animated.Value(windowWidth / 2 - circleRadius);
+    this._lastX = 0;
+    this._onGestureEvent =
+      Platform.OS === 'ios'
```
kmagiera: Is this really necessary? I prefer for all the examples to use animated and preferably with native driver which is one of the goals of the project to make that possible.

kmagiera:
@wmcbain thanks so much for sending this.

I've left a few inline comments, would be great if you could find some time to respond to those. Also would you mind getting rid of package lock files? I decided for this project to use yarn, maintaining two lock files overtime seems like an unnecessary hassle to me.

Thanks!

ios/RNGestureHandler.m
wmcbain: I'm just being more explicit about the behavior. From what I can tell it isn't possible to go from Changed > Failed in UIGestureRecognizer's state machine. Even if you set Failed here the next state will always be Cancelled if the last state was Changed.

Example/panAndScroll/index.js
wmcbain: Unsure if this would be better in a separate PR but mapping translationX on Android with a PanGestureHandler doesn't work as expected, it always seems to be a 10-20 pixels behind the touch point.

wmcbain:
Responded and will address the comments.

Example/panAndScroll/index.js
kmagiera: I think I still would prefer to have the same code running on ios and android that would serve as a test and perhaps can be fixed by someone in the future.

Can you please change this so that we don't have a `Platform` check in this example and maybe add a short comment note about what needs to be fixed on android?

ios/RNGestureHandler.m
```diff
@@ -370,6 +373,16 @@ - (void)touchesMoved:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
         self.state = UIGestureRecognizerStateFailed;
         return;
     }
+    if ((self.state == UIGestureRecognizerStatePossible || self.state == UIGestureRecognizerStateChanged) && _gestureHandler.shouldCancelWhenOutside) {
+        CGPoint pt = [self locationInView:self.view];
+        if (TEST_IS_OUT_OF_BOUNDS(pt, self.view.frame.size)) {
```
kmagiera: This can be just:
```
CGRectContainsPoint(self.view.bounds, pt)
```

And same in line :581

kmagiera:
This looks good. Just added two final comments and when these are fixed this is ready to be merged.

Thanks again @wmcbain for working on this!

Example/panAndScroll/index.js
wmcbain: Sounds good to me, this saves me tine I was about to make a gif to illustrate. I think your approach is the way to go that way the issue is at least apparent to people using the demo.

wmcbain:
Should be good to go!

PR #33, opened 2017-09-26
Support 'enabled' prop for gesture handlers and buttons

kmagiera:
Setting `enabled={false}` will disable interaction with a given gesture handler or button. It does not have any visual effect on button implementation.


alexghi:
Hello,

Could you please also update the docs for this property?
I could also submit a PR for this, but it's simply just a text update.

PR #34, opened 2017-09-26
Unify events sent for TapGestureHandlers

kmagiera:


kmagiera:
Hey @wmcbain!

I looked into the example you've added and it wasn't clear to me what were you trying to achieve there. So I restructured it a bit in a way I think you wanted it to work:
 - when you pan over the pink box the circle will follow gestures and stop at the place where you stop
 - when you tap in some place in pink box the circle will jump to that place (much like it happens with video scrubber bars)
 - when you start gesture inside the pink box but you intend to scroll the underlying scrollview will take over
 - when you pan horizontally inside of the pink box the scrollview should cancel, and when you drag your finger outside of the pink area, panning should stop


Hope I captured that correctly.

Nevertheless I think some of the issues you were having could be due to the fact that on android tap handler wasn't providing x/y coords. On top of some changes to your example app this PR also introduces changes that add missing x/y coords as a part of the event + also unifies it with other similar events by adding `absoluteX` and `absoluteY` both on android and iOS.

@wmcbain I will be grateful if you could find some time and take a look at the changes I made to the example app and if they match your expectations.

wmcbain:
Hey awesome! So just took a quick look and this is close but not exactly what I was looking for. I want the circle to be centered at the gesture point rather than just moving with the offset as so:

example

The tap and scroll behavior are on point. Really appreciate you spending time with this, will be watching this PR and here to help if need be.

wmcbain:
Are your DMs open on Twitter? What's a good way to communicate?

kmagiera:
@wmcbain I prefer slack, I'm in expo-developers and react-native core teams. hit me up there. Also my DMs are open but I rarely check there

PR #41, opened 2017-10-13
Add useRippleEffect (default: true) prop to be able to disable the ri…

sjchmiela:
…pple effect


henrikra:
Do you have example gif or something for this?

kmagiera:
I'll soon be woring on making ripple effect more configurable to match what is already available in RN (with TouchableNativeFeedback). Apparently we can't just extend ReactViewGroup for the button component as it breaks native android event flow (that's what we try to avoid in RNGH). Once the changes are done this change may no longer be necessary

kmagiera:
I'll soon be woring on making ripple effect more configurable to match what is already available in RN (with TouchableNativeFeedback). Apparently we can't just extend ReactViewGroup for the button component as it breaks native android event flow (that's what we try to avoid in RNGH). Once the changes are done this change may no longer be necessary

kmagiera:
Closing this as it got out of sync with the current version. And as we plan some bigger refactors of the referenced code it would be easier to address that when its ready

PR #44, opened 2017-10-15
Event system fixes for compatibility with react-native 0.50

janicduplessis:
The DummyViewManager technique to register event types doesn't work anymore because events now registered when the component class is used. We actually still need DummyViewManager on Android so native animated knows about the custom event name mapping. iOS has some magic function to convert from topX to onX so it's not needed anymore.

We can now use `ReactNativeBridgeEventPlugin` to register custom event types instead. Event names should also be topX instead of onX on Android so we can have the same event types as iOS.

Tested a bunch of examples in the example app with RN 0.50.0

This is in no way backward compatible with previous RN versions since `ReactNativeBridgeEventPlugin` does not exist so I haven't made any effort to make other changes backwards compatible.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerPackage.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerPackage.java
@@ -7,6 +7,7 @@ import com.facebook.react.bridge.NativeModule;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.common.MapBuilder;
 import com.facebook.react.uimanager.ReactShadowNode;
+import com.facebook.react.uimanager.ReactShadowNodeImpl;
 import com.facebook.react.uimanager.ThemedReactContext;
 import com.facebook.react.uimanager.ViewManager;
 
@@ -35,7 +36,7 @@ public class RNGestureHandlerPackage implements ReactPackage {
 
     @Override
     public Class getShadowNodeClass() {
-      return ReactShadowNode.class;
+      return ReactShadowNodeImpl.class;
     }
 
     @Override
@@ -51,9 +52,9 @@ public class RNGestureHandlerPackage implements ReactPackage {
     public @Nullable Map getExportedCustomDirectEventTypeConstants() {
       return MapBuilder.of(
               RNGestureHandlerEvent.EVENT_NAME,
-              MapBuilder.of("registrationName", RNGestureHandlerEvent.EVENT_NAME),
+              MapBuilder.of("registrationName", "onGestureHandlerEvent"),
               RNGestureHandlerStateChangeEvent.EVENT_NAME,
-              MapBuilder.of("registrationName", RNGestureHandlerStateChangeEvent.EVENT_NAME));
+              MapBuilder.of("registrationName", "onGestureHandlerStateChange"));
     }
   }
 
--- a/Example/package.json
+++ b/Example/package.json
@@ -7,8 +7,8 @@
     "test": "jest"
   },
   "dependencies": {
-    "react": "16.0.0-alpha.12",
-    "react-native": "^0.49.3",
+    "react": "16.0.0",
+    "react-native": "^0.50.0",
     "react-native-gesture-handler": "file:../",
     "react-native-vector-icons": "^4.4.2",
     "react-navigation": "^1.0.0-beta.13"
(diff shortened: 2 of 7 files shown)


janicduplessis:
If we really want to make this backwards compatible I think we could use `requireNativeComponent` to register the events from the DummyView but I feed like this is a cleaner solution since we can get rid of that.

kmagiera:
Thanks @janicduplessis for being on top of this! TBH I believe this library is at the stage when users wouldn't be surprise when they need to deal with breaking changes while upgrading. Yet, since I expect there are still some bugs that may be blocking ppl using this lib for their product, I'd like to wait with updating the core dependency to an "unstable" branch (even though RN rc's are perhaps more "stable" than this library).

I'll be more than happy to merge this in as soon as 0.50 is released. I also like this approach more considering we can get rid of the "DummyView" hack. But if you could offer a workaround that would make it work with 0.50-rc and 0.49 that I'd very much appreciated it. In which case we can also leave this PR open and revisit when for 0.50 release.

janicduplessis:
@kmagiera 0.50 is out, I don't really have the time to investigate a backwards compatible solution at the moment sadly.

janicduplessis:
Just tried this PR again and noticed an issue with Swipeable + native animations, not sure if it's related to those changes but I'll have a look.

kmagiera:
Thanks for checking @janicduplessis. Let me know what your findings are. I'm fine with merging this in now that 0.50 is out

kmagiera:
Seems I figured it out. The problem was that native driver would register event under "onGestureHandler" name but was actually getting "topGestureHandler" as the event name and couldn't map that to any value ([here](https://github.com/facebook/react-native/blob/2b4ff6ea19dc674cf035ee419daa132cde8d1f5e/ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedNodesManager.java#L391) is the relevant code in RN)

So since I don't understand the recent changes in RN core regarding events I have two possible solutions for the problem:

1) keep DummyViewManager and update its exported direct event constants so that the registration name for "topGestureHandlerEvent" is "onGestureHandlerEvent" (similarly to how was it done for scroll events here: https://github.com/facebook/react-native/commit/ca834f96afcb99a7f1f4eec0a864f4f5b33a632b)
2) In `ReactNativeBridgeEventPlugin.processEventTypes` use names starting with "on" instead of "top" and update `EVENT_NAME` in native code to also use names starting with "on"

@janicduplessis If you understand what the events changes are all about I would trust your judgment to select the correct approach. If you are not certain I can ask around or research it on my own.

Otherwise this is great stuff. Thanks!

janicduplessis:
@kmagiera I have figured out something that works, from what I understand the main issue is that events are now registered lazily the first time the component is rendered which means it will never happen for `DummyViewManager`. The best solution seem to be to call `ReactNativeBridgeEventPlugin.processEventTypes` manually but on Android we also need to keep the `DummyViewManager` since native animated uses the constants it provides to map event names. This is not needed on iOS because of some magic code to convert from topX to onX.

I did try your second solution but it does not work on iOS because of the magic topX to onX conversion. I could platform check but I feel this is better.

Also need to update getShadowNodeClass to return ReactShadowNodeImpl instead.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerPackage.java
```diff
@@ -35,7 +36,7 @@ public ReactShadowNode createShadowNodeInstance() {
 
     @Override
     public Class getShadowNodeClass() {
-      return ReactShadowNode.class;
+      return ReactShadowNodeImpl.class;
```
kmagiera: Instead of doing this I'd prefer to change the baseclass of ViewManager and use ReactViewManager instead. It doesn't make much of a difference because views won't be instantiated anyways but may prevent similar incompatibility issues in the future

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerPackage.java
```diff
@@ -51,9 +52,9 @@ public void updateExtraData(View root, Object extraData) {
     public @Nullable Map getExportedCustomDirectEventTypeConstants() {
       return MapBuilder.of(
               RNGestureHandlerEvent.EVENT_NAME,
-              MapBuilder.of("registrationName", RNGestureHandlerEvent.EVENT_NAME),
+              MapBuilder.of("registrationName", "onGestureHandlerEvent"),
```
kmagiera: Can you put both "topGestureHandlerEvent" and "onGestureHandlerEvent" as strings here instead of using constant? Or maybe define "onXXX" as a constant in the event class? I think it will make it easier to understand how events are translated in the future as well as may help with spotting typos etc

(same below)

kmagiera:
I tested this and seems to be working fine. I'm going to merge it and fix the small comments on my own. Thanks for working on this @janicduplessis !

Issue #48, opened 2017-10-28
Swipeable on Android throws exception if component is removed


tallenh:
I have a FlatList with Swipeable items.  I have a handler onSwipeableLeftOpen that will remove the item from the list.  Each item also has a checkbox that when clicked will remove the item from the list.  If the only way you remove items is to swipe them away everything works great.  If you click the checkbox the item is removed, but the next time you touch the list it will throw an exception.

  I have only tested this on Android.

  I tried to debug but I haven't had much luck tracking it down.  I can say the exception is thrown in GestureHandlerOrchestrator.java line 318:

      throw new IllegalArgumentException("Parent is null? View is no longer in the tree");

I have a small example of this at
https://gist.github.com/tallenh/dfebf73859cca956bf8d6f31ccc4bf7c

kmagiera:
Thanks for reporting @tallenh I'm going to take a look at this soon

osdnk:
I also came across this issue and made a PR. It works in my case

kmagiera:
@osdnk have you managed to repro this issue on the example app @tallenh provided: https://gist.github.com/tallenh/dfebf73859cca956bf8d6f31ccc4bf7c

osdnk:
Yes. I've just pasted this gist into app 
GenyMotion, Nexus 5, API 23

kmagiera:
Thanks a lot @tallenh for reporting this issue. I allowed me to catch two serious problems with Android implementation. These are fixed in #53 which will go out in the next release today or tomorrow

PR #49, opened 2017-11-04
Fix android "parent null" issue and scrolling in example app.

osdnk:
Remove parent null exception on Android (throw when component was removed 'externally') and change deprecated way of scrolling in example app

### DIFF
--- a/Example/combo/index.js
+++ b/Example/combo/index.js
@@ -108,7 +108,7 @@ class ControlledSwitch extends React.Component {
 class Combo extends Component {
   _onClick = () => {
     Alert.alert("I'm so touched");
-    this._scrollView.scrollTo(200);
+    this._scrollView.scrollTo({ y: 200 });
   };
   render() {
     const { ScrollViewComponent } = this.props;
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
@@ -314,8 +314,11 @@ public class GestureHandlerOrchestrator {
       outputCoords[1] = event.getY();
       return;
     }
-    if (view == null || !(view.getParent() instanceof ViewGroup)) {
-      throw new IllegalArgumentException("Parent is null? View is no longer in the tree");
+    if (!(view.getParent() instanceof ViewGroup)) {
+      return;
+    }
+    if (view == null) {
+      throw new IllegalArgumentException("View is no longer in the tree");
     }
     ViewGroup parent = (ViewGroup) view.getParent();
     extractCoordsForView(parent, event, outputCoords);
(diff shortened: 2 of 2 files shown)


kmagiera:
Thanks @osdnk for working on this! Before it gets merge I'd like to get better sense on what was the issue you were hitting to be able to tell if that's the right approach. Can you provide a better description on what do you mean by "parent null issue" as well as what are you referring to by saying that "component has been removed 'externally'"

Thanks!

osdnk:
I just want to get rid of the issue mentioned here:
https://github.com/kmagiera/react-native-gesture-handler/issues/48
Component here is being removed by changing state of parent component

kmagiera:
@osdnk so I looked at #48 and think this is not a right way to solve it. The issue seems to be related to the fact that some views with connected gesture handler are being removed while their handlers remains in the list of "active" handlers (`GestureHandlerOrchestrator.mGestureHandlers`). This is no good and your change doesn't prevent it from happening.

My suggestion would be to allow GestureHandler class to do some cleanup when it gets torn down. For example we could call `cancel` on `GestureHandler` instances that gets removed (e.g. at the end of `RNGestureHandlerRegistry.detachHandler`). This would allow gesture handler to notify orchestrator (if it's in active state) that it should be removed from the active handlers list.

I'd also prefer for the `if` condition you're changing to stay the way it was. It helps with tracking issues like #48 which wouldn't be possible if we were silently returning there.

@osdnk do you have time / would you like to update your PR following this suggestions?

Thanks again for investigating the problem, that was very helpful in understanding on what was going on!

osdnk:
@kmagiera, Ok, I will try to update my PR

osdnk:
@kmagiera, Hi there,
I delve into your library and made a research about this issue and found out that it seems to be a quite specific case:
I was counting `GestureHandlers` in `GestureHandlerOrchestrator`. Every tap on one of checkbox was surprisingly increasing number of `GestureHandlers` by one (even after cleanup made by `GestureHandlerOrchestrator.cleanupFinishedHandlers`). The reason was that every tap on any checkbox was creating a new GestureHandler and it was left orphaned by `View` but not removed by cleanup method. The reason was that it wasn't finished (method `isFinished` was returning `false`, because state was `GestureHandler.STATE_UNDETERMINED` - `PanGestureHadler` make GestureHandler active only when MinDeltaX is reached). 
So I decided to cleanup not only finished gestures, but also those in undetermined state. I suppose that's not a perfect solution, but it's a result of inevitable gesture conflicts in JS thread (checkbox) and native one. 
Maybe that's just not a good way to wrap components like checkboxes into `<Swipeable>`

kmagiera:
@osdnk it is perfectly fine to wrap any components in Swipeable. The problem you are mentioning is the result of what I have already posted in my previous comment: handlers attached to views that get removed does not get cleared away from `GestureHandlerOrchestrator.mGestureHandlers`.

If you would like to work on this issue please follow my suggestion from the previous comment. If you'd cancelled "active" gesture handlers in `RNGestureHandlerRegistry.detachHandler` they'd be in the correct "finished" state when the cleanup method gets called. Treating "undetermined" state as finished will likely break things like multitouch PanHandler that goes from "undetermined" to "begin" state only when it registers sufficient number of pointers on the screen.

Also, please revet the change with `if (view == null) {`. It is redundant if few lines earlier you call `view.getParent` because if `view` was null that previous call would throw a `NullPointerException`. As I mentioned in my previous comment, this check was there for purpose and as you can see thanks to it we have discovered the issue with handlers not getting cleared up properly. If we were silently returning there w/o an exception we would have never found this problem.

kmagiera:
Hey @osdnk ! Thanks for your work and investigating this issue. As I had to come up with a fix quickly I decided to submit and merge #53

My changes fixes #48 so I am closing this PR. Anyways thanks for your time put in preparing a fix and discussing it, that was really helpful!

PR #52, opened 2017-11-12
Fix Flow types

Ashoat:
Right now, DrawerLayout.cs and Swipeable.js declare Flow types, but they aren't actually correct. This package doesn't seem to have Flow set up, so it probably doesn't matter to you, but unfortunately this means that anybody who *is* using Flow that installs this package will see Flow errors.

This PR fixes the Flow types. There should be no changes in functionality or behavior.

### DIFF
--- a/Swipeable.js
+++ b/Swipeable.js
@@ -6,6 +6,7 @@
 
 import React, { Component } from 'react';
 import { Animated, StyleSheet, View } from 'react-native';
+import { AnimatedEvent } from 'react-native/Libraries/Animated/src/AnimatedEvent';
 
 import {
   PanGestureHandler,
@@ -17,7 +18,7 @@ const DRAG_TOSS = 0.05;
 
 export type PropType = {
   children: any,
-  friction?: number,
+  friction: number,
   leftThreshold?: number,
   rightThreshold?: number,
   overshootLeft?: boolean,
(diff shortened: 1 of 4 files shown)


Swipeable.js
```diff
@@ -200,8 +226,12 @@ export default class Swipeable extends Component {
   };
 
   _currentOffset = () => {
-    const { leftWidth = 0, rowWidth = 0, rowState } = this.state;
-    const { rightOffset = rowWidth } = this.state;
+    const { rowState } = this.state;
+    const leftWidth = this.state.leftWidth ? this.state.leftWidth : 0;
+    const rowWidth = this.state.rowWidth ? this.state.rowWidth : 0;
+    const rightOffset = this.state.rightOffset
```
kmagiera: Why changing this? Seems like flow can deal with this syntax, can it?

Also: this is not completely accurate with the way this code used to work. Instead of doing:
```this.state.foo ? this.state.foo : bar```
this should rather be: 
```this.state.foo === undefined ? this.state.foo : bar```

For most of the cases it won't matter but it will change behavior in the case when `this.state.rightOffset = 0`. Then, instead of having `rightOffset = 0` we will get `rightOffset = rowWidth`.

Ashoat: You're absolutely right, this is broken for the 0 case. Good catch, sorry I missed that.

The reason I changed it in the first place was because I was typing `leftWidth`, `rowWidth`, etc. as nullable, which meant they could be either `null` or `undefined`, but the "destructuring assignment with default value" syntax only handles `undefined`.

I'll change the types to be `number | typeof undefined` instead of `?number` and switch the logic back to the destructuring assignment.

Ashoat:
> Since we don't have CI set up for this project would it make sense if we add "flow" to be run by "lint-staged" in pre-commit phase?

Good call! I'll make that change.

kmagiera:
Thanks a lot @Ashoat ! I'll make a release tomorrow as there are a few things I'd like to include in it

PR #70, opened 2017-12-07
Fix order or steps

henrikra:
This will avoid confusion about skipping steps when you have native navigation library :)


kmagiera:
Makes sense. Thanks a lot!

PR #77, opened 2017-12-13
Use PureComponent instead of Component

### LINKED ISSUE #76: Use PureComponent instead of Component
I'm using https://github.com/maicki/why-did-you-update to avoid unnecessary rerenders and it's complaining about Swipeable:

image

brunolemos:
Closes #76 

It's not ideal to use PureComponent everywhere, but should be fine.


brunolemos:
Closed in favor of #83

Issue #81, opened 2017-12-16
[Swipeable] Not working on older iOS versions


brunolemos:
Tried the `Swipeable` component on iOS 8.1 and it throws this exception:

[Swipeable.js:195:40](https://github.com/kmagiera/react-native-gesture-handler/blob/master/Swipeable.js#L195) `undefined is not a function (evaluating 'Math.sign(toValue)')` 
Fix: #82

On iOS 9.0 the swipe gesture works but it doesn't show the buttons.

kmagiera:
Not sure why github didn't close this automatically. Anyways, closing it now, thanks @brunolemos again for your fix!

brunolemos:
@kmagiera it's because I only fixed the crash, but the swipeable buttons still doesn't show up on older ios versions. Please reopen :)

pfyod:
@kmagiera still shows weird artifacts instead of buttons on older iOS versions (at least 9.3.5).

PR #82, opened 2017-12-16
[Swipeable] Fix crash on iOS 8.x

brunolemos:
Math.sign polyfill from https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Math/sign

PS: This fixes a crash but Swipeable feature is still not fully working
Related: #81


kmagiera:
Sweet, thanks @brunolemos !

Issue #78, opened 2017-12-13
Typescript definitions


brunolemos:
Currently this is the only package in my dependencies that doesn't have type definitions.
EDIT: Now this and react-native-navigation

Tried to use https://github.com/Microsoft/dts-gen but it doesn't support es6 modules (Microsoft/dts-gen#42) which this repo uses.

I'm new to Typescript so I'll not jump into a PR right away.

kmagiera:
cc @krzysztofciombor who was interested in working on providing TS definitions

kmagiera:
Closing as per https://github.com/kmagiera/react-native-gesture-handler/pull/86

PR #86, opened 2017-12-19
Add TS types definitions

krzysztofciombor:
## Description
Adds TypeScript types definitions.
See also: #78 

## Test plan
The TS types are correctly picked up when importing Gesture Handlers inside `*.ts` file:
zrzut ekranu 2017-12-19 o 17 17 25
zrzut ekranu 2017-12-19 o 17 17 34

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "Swipeable.js",
     "DrawerLayout.js",
     "gestureHandlerRootHOC.android.js",
-    "gestureHandlerRootHOC.ios.js"
+    "gestureHandlerRootHOC.ios.js",
+    "react-native-gesture-handler.d.ts"
   ],
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,14 +1,14 @@
 {
   "name": "react-native-gesture-handler",
   "version": "1.0.0-alpha.35",
-  "description":
-    "Experimental implementation of a new declarative API for gesture handling in react-native",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,14 +1,14 @@
...
-  "description":
-    "Experimental implementation of a new declarative API for gesture handling in react-native",
+  "description": "Experimental implementation of a new declarative API for gesture handling in react-native",
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
     "test": "jest",
     "precommit": "lint-staged"
   },
   "main": "GestureHandler.js",
+  "types": "react-native-gesture-handler.d.ts",
```
kmagiera: I think this needs to be added to "files" section too

brunolemos:
It works if I import like this: `import { Swipeable } from 'react-native-gesture-handler'`
But not if I import like this: `import Swipeable from 'react-native-gesture-handler/Swipeable'`

brunolemos:
#87 should fix it

brunolemos:
I get this on `RectButton` and don't know why:

```
[ts] Property 'style' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<RectButton> & Readonly<{ children?: ReactNode; }> ...'.
(JSX attribute) style: ((number & {
    __registeredStyleBrand: ViewStyle | TextStyle | ImageStyle;
}) | {
    backgroundColor: string;
    width: number | undefined;
})[]
```

image

PR #87, opened 2017-12-23
Add class methods typings

brunolemos:


brunolemos:
image
image

It's auto-completing nicely but it's not showing an error for me when I pass a string to velocity. 
@krzysztofciombor can you please review?

kmagiera:
Thanks @brunolemos this looks good

brunolemos:
I just noticed the openDrawer options is not optional in this pr but should be, you may quick fix that

PR #88, opened 2017-12-23
[TYPING] Use ReactNode instead of JSX.Element to match official react typings

brunolemos:
I was getting an incompatibility type error and noticed this.

Also, ReactNode already includes these null | undefined | false conditions


kmagiera:
Sweet @brunolemos, we've had a discussion with @krzysztofciombor about this because neither of us is a typescript expert. Glad that you have found this and pointed the right solution

PR #96, opened 2018-01-06
Pass navigatorButtons static component so it can be accessed by ...

joserocha3:
navigation library

### DIFF
--- a/gestureHandlerRootHOC.android.js
+++ b/gestureHandlerRootHOC.android.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { requireNativeComponent, View, StyleSheet } from 'react-native';
+import hoistNonReactStatics from 'hoist-non-react-statics';
 
 var iface = {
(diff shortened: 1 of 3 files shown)


joserocha3:
This change allows for the functionality found [here](https://wix.github.io/react-native-navigation/#/adding-buttons-to-the-navigator). Without it, buttons declared within the component class using `static navigatorButtons:  {...}` do not show up.

kmagiera:
Thanks for working on this @joserocha3 

I think it would be better to use a more generic solution as there are perhaps some other static fields that deserve to be copied. I recommend that we use this instead: https://github.com/mridgway/hoist-non-react-statics

Would you be open to updating your pull request with my suggestion?

joserocha3:
Great idea @kmagiera. I have updated it.

kmagiera:
Sweet! Thanks a ton @joserocha3 Will push it out to NPM with the next alpha release

PR #101, opened 2018-01-12
Rename uiManagerWillFlushUIBlocks to uiManagerWillPerformMounting

### LINKED ISSUE #100: iOS: Touch/pan not working on RN 0.52.0
Seems like all touch, tap and pan are not working on latest React Native on iOS.

A simple component:

```
<BaseButton onPress={() => alert('Clicked!')}>
   <Text>Click me</Text>
</BaseButton>
```

Tapping it on Android yields and alert while iOS doesn't.

brentvatne:
This makes rngh compatible with react-native 0.52. It resolves #100.

This is necessary due to a breaking change in react-native 0.52: https://github.com/facebook/react-native/commit/0a8721c

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,7 @@ Before instaling please check the following table to find which library version 
 
  | React Native version | React Native Gesture Handler version | Notes |
  | --- | --- | --- |
- | v0.52+ | **latest** | Breaking changes explained and fixed in [#101](https://github.com/kmagiera/react-native-gesture-handler/pull/44) |
- | v0.50 - v0.51 | 1.0.0-alpha.39 | |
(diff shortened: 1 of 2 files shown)


pfeiffer:
Awesome!

kmagiera:
Thanks @brentvatne! I was thinking maybe we can keep the old method and just make it call the new one, this way we avoid introducing changes that introduce yet another compatibility break. I am aware that this will not always be possible but for this particular case it seems like we can keep compatibility for 51 without introducing much complexity to the codebase. What do you think?

kmagiera:
🥇 🥇 🥇 🥇 🥇 🥇

PR #104, opened 2018-01-14
[TYPING] Fix RectButton props

brunolemos:
Fix errors like these:

`[ts] Property 'onPress' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<RectButton> & Readonly<{ children?: ReactNode; }> ...'.`

`[ts] Property 'style' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<RectButton> & Readonly<{ children?: ReactNode; }> ...'.`

when using `<RectButton onPress={...} style={...} />`


kmagiera:
Nice catch, thanks @brunolemos !

PR #105, opened 2018-01-15
Update build.gradle

mahomahoxd:
provided is deprecated, in the new gradle its compileOnly. I've also updated the sdk and buildtools version. If its not conflicting with what you want to do, you can merge it.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,8 +1,8 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion "25.0.2"
+    compileSdkVersion 27
+    buildToolsVersion "27.0.3"
 
     defaultConfig {
         minSdkVersion 16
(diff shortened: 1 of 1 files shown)


kmagiera:
Thanks for submitting this @mahomahoxd. i'm just going to need to try it on my own before merging if it doesn't break my workflow. Should be able to merge it in few days.

mahomahoxd:
In case of your Example directory, I've made changes to the library there. I've also tested it and it seems to be working just like it'd regularly. I've also noticed that **if** you use gradle 3, it'll look for the API of android tools to be greater than 26, if it can't find any, it'll download and install it either way. I assume this doesn't happen on gradle 2.x, but you'd not want to be stuck on the same dependencies for long as it might introduce bugs in the future.

osdnk:
@mahomahoxd 
I did made some changes as in new Android Gradle Plugin `buildToolsVersion` is not required at all since "each version of the Android Gradle Plugin now has a default version of the build tools". 
Cannot see the reason why to update `compileSdkVersion`

PR #126, opened 2018-02-28
Update pod spec

miighty:
Add homepage string to pod file to allow latest version of cocoa pods to build the project.


kmagiera:
Thanks @miighty !

PR #140, opened 2018-04-11
Swipe handler

osdnk:


android/lib/src/main/java/com/swmansion/gesturehandler/FlingGestureHandler.java
```diff
@@ -0,0 +1,97 @@
...
+  private int mMaxNumberOfTouchesSimultaneously;
+
+  private final Runnable mFailDelayed = new Runnable() {
+    @Override
+    public void run() {
+      fail();
+    }
+  };
+
+  public void setNumberOfTouchesRequired(int mNumberOfTouchesRequired) {
```
kmagiera: just use `numberOfTouchesRequired` as a parameter name (no `m` prefix)

android/lib/src/main/java/com/swmansion/gesturehandler/FlingGestureHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    this.mNumberOfTouchesRequired = mNumberOfTouchesRequired;
```
kmagiera: Then you will also just write:
`mNumberOfTouchesRequired = numberOfTouchesRequired`

(no `this.`)

android/lib/src/main/java/com/swmansion/gesturehandler/FlingGestureHandler.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  public void setDirection(int direction) {
+    this.direction = direction;
```
kmagiera: Same here. We prefix member instances with `m`

https://source.android.com/setup/contribute/code-style#follow-field-naming-conventions

android/lib/src/main/java/com/swmansion/gesturehandler/FlingGestureHandler.java
```diff
@@ -0,0 +1,97 @@
...
+  @Override
+  protected void onHandle(MotionEvent event) {
+    int state = getState();
+
+    if (state == STATE_UNDETERMINED) {
+      mStartX = event.getRawX();
+      mStartY = event.getRawY();
+      begin();
+      mMaxNumberOfTouchesSimultaneously = 1;
+      startFling();
```
kmagiera: A bit confused of why the few above lines are not the part of `startFling` method while `endFling` seems to be calling state change methods and also using `mStartX` etc. Wouldn't it be more consistent to move that logic into `startFling()`

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
```diff
@@ -70,6 +71,8 @@
   private static final String KEY_PAN_MIN_POINTERS = "minPointers";
   private static final String KEY_PAN_MAX_POINTERS = "maxPointers";
   private static final String KEY_PAN_AVG_TOUCHES = "avgTouches";
+  private static final String KEY_NUMBER_OF_TOUCHES= "numberOfTouches";
```
kmagiera: space before `=` here and in the next line

Also we've been using a different naming before – for PanHandler we use `minPointers` so would be good to make it consistent and call this attribute the same way

ios/Handlers/RNFlingHandler.h
```diff
@@ -0,0 +1,12 @@
+//
```
kmagiera: Can you please delete all this comment. We don't have any file headers in the project for now (may want to add some in the future).

ios/Handlers/RNFlingHandler.m
```diff
@@ -0,0 +1,58 @@
+//
```
kmagiera: Same here

ios/Handlers/RNFlingHandler.m
```diff
@@ -0,0 +1,58 @@
...
+    }
+    return self;
+}
+
+- (void)configure:(NSDictionary *)config
+{
+    [super configure:config];
+    UISwipeGestureRecognizer *recognizer = (UISwipeGestureRecognizer *)_recognizer;
+
+
```
kmagiera: delete this line (try to avoid two consecutive empty lines in method bodies)

ios/Handlers/RNFlingHandler.m
```diff
@@ -0,0 +1,58 @@
...
+- (void)configure:(NSDictionary *)config
+{
+    [super configure:config];
+    UISwipeGestureRecognizer *recognizer = (UISwipeGestureRecognizer *)_recognizer;
+
+    
+    id prop = config[@"direction"];
+    if (prop != nil) {
+        
+        NSInteger direction = [RCTConvert NSInteger:prop];
```
kmagiera: Would be nice to use RCTConvert category for `UISwipeGestureRecognizerDirection` and put that logic in there, see something similar being done for example here:
https://github.com/wix/react-native-notifications/blob/master/RNNotifications/RNNotifications.m#L28

ios/RNGestureHandler.xcodeproj/project.pbxproj
```diff
@@ -34,6 +34,8 @@
 		44AEC73A1F8FA1270086889F /* RNTapHandler.m in Sources */ = {isa = PBXBuildFile; fileRef = 44AEC72E1F8FA1270086889F /* RNTapHandler.m */; };
 		44BE34481F1E1AAA008679D1 /* RNGestureHandlerManager.m in Sources */ = {isa = PBXBuildFile; fileRef = 44BE34471F1E1AAA008679D1 /* RNGestureHandlerManager.m */; };
 		44BE344A1F1E1ABA008679D1 /* RNGestureHandlerManager.h in Headers */ = {isa = PBXBuildFile; fileRef = 44BE34491F1E1ABA008679D1 /* RNGestureHandlerManager.h */; };
+		66A291D5207D032400809C27 /* RNFlingHandler.m in Sources */ = {isa = PBXBuildFile; fileRef = 66A291D3207D032400809C27 /* RNFlingHandler.m */; };
```
kmagiera: This changes in proj file are ok cause they are here to include new files you added in the project

ios/RNGestureHandlerEvents.h
```diff
@@ -21,7 +21,9 @@
 + (RNGestureHandlerEventExtraData *)forPinch:(CGFloat)scale
                               withFocalPoint:(CGPoint)focalPoint
                                 withVelocity:(CGFloat)velocity
-                            withNumberOfTouches:(NSUInteger)numberOfTouches;
+                         withNumberOfTouches:(NSUInteger)numberOfTouches;
++ (RNGestureHandlerEventExtraData *) withNumberOfTouches:(NSUInteger)numberOfTouches
```
kmagiera: no space after `)`

ios/RNGestureHandlerEvents.m
```diff
@@ -23,6 +23,16 @@ + (RNGestureHandlerEventExtraData *)forPosition:(CGPoint)position
                            @"absoluteY": @(absolutePosition.y)}];
 }
 
++ (RNGestureHandlerEventExtraData *) withNumberOfTouches:(NSUInteger)numberOfTouches
```
kmagiera: no space after `)`

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
osdnk: That might be confusing. Because it's not just minimal number of touches, but _exact_ number o them. If you try to fling with more than given number of fingers, it won't work

ios/Handlers/RNFlingHandler.h
osdnk: Ok, I was just trying to be coherent with another files in the project

PR #141, opened 2018-04-11
Update RN 

osdnk:


kmagiera:
@osdnk there should also be some changes in yarn.lock, can you take a look?

PR #143, opened 2018-04-17
Bottom sheet example and TapGestureHandler improvements

osdnk:
## Motivation
https://github.com/brentvatne/bottom-sheet-gesture-handler-experiment
Got inspired by @brentvatne code and made it possibile to implement bottom sheet in a bit hacky way using very long TapHandler. The problem was to to trigger scrolling scrollview at the end of extending a main card and not to trigger it when card is not opened fully.
Also then it became extremely easy to resolve another issue: https://github.com/kmagiera/react-native-gesture-handler/issues/146

## Changes
Initially I was trying to made it possible to track route (and maxDeltas) using 'normal' TapGestureHandler. Then I decided to implement new `TapGestureHandler` as quite unobvious extension of `UIPanGestureRecognizer` which gives us many ways of parameterization.
However, I wasn't really into extending out `PanGestureHandler` which seems to be unnecessary.
It gives a possibility to fail gesture whet its position pass `maxDeltaY` and then it allows scroll gesture to be triggered then. If it won't fail, scroll won't be scrolled.

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
@@ -7,8 +7,8 @@ public class TapGestureHandler extends GestureHandler<TapGestureHandler> {
   private static float MAX_VALUE_IGNORE = Float.MIN_VALUE;
   private static final long DEFAULT_MAX_DURATION_MS = 500;
   private static final long DEFAULT_MAX_DELAY_MS = 500;
-  private static final int DEFAULT_NUMBER_OF_TAPS= 1;
-  private static final int DEFAULT_MIN_NUMBER_OF_POINTERS= 1;
+  private static final int DEFAULT_NUMBER_OF_TAPS = 1;
+  private static final int DEFAULT_MIN_NUMBER_OF_POINTERS = 1;
 
   private float mMaxDeltaX = MAX_VALUE_IGNORE;
   private float mMaxDeltaY = MAX_VALUE_IGNORE;
@@ -137,8 +137,10 @@ public class TapGestureHandler extends GestureHandler<TapGestureHandler> {
     mLastX = event.getRawX();
     mLastY = event.getRawY();
 
-    if (mNumberOfPointers < event.getPointerCount()) {
-      mNumberOfPointers = event.getPointerCount();
+    if (action == MotionEvent.ACTION_MOVE) {
+      if (mNumberOfPointers < event.getPointerCount()) {
+        mNumberOfPointers = event.getPointerCount();
+      }
     }
 
     if (action == MotionEvent.ACTION_DOWN) {
@@ -146,9 +148,7 @@ public class TapGestureHandler extends GestureHandler<TapGestureHandler> {
         begin();
       }
       startTap();
-    }
-
-    if (shouldFail()) {
+    } else if (shouldFail()) {
       fail();
     }
     if (action == MotionEvent.ACTION_UP) {
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
@@ -4,15 +4,21 @@ import android.os.Handler;
 import android.view.MotionEvent;
 
 public class TapGestureHandler extends GestureHandler<TapGestureHandler> {
-
+  private static float MAX_VALUE_IGNORE = Float.MIN_VALUE;
   private static final long DEFAULT_MAX_DURATION_MS = 500;
   private static final long DEFAULT_MAX_DELAY_MS = 500;
   private static final int DEFAULT_NUMBER_OF_TAPS= 1;
+
+  private float mMaxDeltaX = MAX_VALUE_IGNORE;
+  private float mMaxDeltaY = MAX_VALUE_IGNORE;
+  private float mMaxDist = MAX_VALUE_IGNORE;
 
   private long mMaxDurationMs = DEFAULT_MAX_DURATION_MS;
   private long mMaxDelayMs = DEFAULT_MAX_DELAY_MS;
   private int mNumberOfTaps = DEFAULT_NUMBER_OF_TAPS;
 
+  private float mStartX, mStartY;
+  private float mOffsetX, mOffsetY;
   private float mLastX, mLastY;
   private float mLastEventOffsetX, mLastEventOffsetY;
 
@@ -38,6 +44,20 @@ public class TapGestureHandler extends GestureHandler<TapGestureHandler> {
 
   public TapGestureHandler setMaxDurationMs(long maxDurationMs) {
     mMaxDurationMs = maxDurationMs;
+    return this;
+  }
+
+  public TapGestureHandler setMaxDx(float deltaX) {
+    mMaxDeltaX = deltaX;
+    return this;
+  }
+
+  public TapGestureHandler setMaxDy(float deltaY) {
+    mMaxDeltaY = deltaY;
+    return this;
+  }
+  public TapGestureHandler setMaxDist(float maxDist) {
+    mMaxDeltaY = maxDist;
     return this;
   }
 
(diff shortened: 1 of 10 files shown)


android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
```diff
@@ -4,15 +4,21 @@
...
 public class TapGestureHandler extends GestureHandler<TapGestureHandler> {
-
+  private static float MAX_VALUE_IGNORE = Float.MIN_VALUE;
   private static final long DEFAULT_MAX_DURATION_MS = 500;
   private static final long DEFAULT_MAX_DELAY_MS = 500;
   private static final int DEFAULT_NUMBER_OF_TAPS= 1;
 
+  private float mMaxDeltaX = MAX_VALUE_IGNORE;
+  private float mMaxDeltaY = MAX_VALUE_IGNORE;
+  private float mMaxDist = MAX_VALUE_IGNORE;
```
kmagiera: In other handlers we store square of the dist (`mMaxDistSq`) would be nice to change it here for consistency.

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
```diff
@@ -68,27 +88,74 @@ private void endTap() {
     }
   }
 
+  private static float getLastPointer(MotionEvent event, boolean isX) {
```
kmagiera: Is this the same code that we have in `PanGestureHandler`? If so maybe worth extracting into some "utils" class. Please also include the comment I've put on `PanGestureHandler.getsLastPointerX` that explains what its for

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
```diff
@@ -68,27 +88,74 @@ private void endTap() {
...
+      return true;
+    }
+
+    float dy = mLastY - mStartY + mOffsetY;
+    if (mMaxDeltaY != MAX_VALUE_IGNORE && Math.abs(dy) > mMaxDeltaY) {
+      return true;
+    }
+
+    float dist = dy * dy + dx * dx;
+    return mMaxDist != MAX_VALUE_IGNORE && Math.abs(dist) > mMaxDist;
```
kmagiera: `abs` is not needed here, also we should have `mMaxDistSq` (square of the distance). Please take a look on how its done in Pan handler (e.g. note that we keep `mMinDistSq` but have a setter for just minDist) and lets use the same patter here

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
```diff
@@ -68,27 +88,74 @@ private void endTap() {
...
+    }
 
     mLastEventOffsetX = event.getRawX() - event.getX();
     mLastEventOffsetY = event.getRawY() - event.getY();
     mLastX = event.getRawX();
     mLastY = event.getRawY();
 
-    if (state == STATE_UNDETERMINED) {
-      begin();
-      mTapsSoFar = 0;
```
kmagiera: shouldn't we keep this? Have you tested multi-tap?

ios/Handlers/RNTapHandler.m
```diff
@@ -12,11 +12,22 @@
 
 #import <React/RCTConvert.h>
 
-@interface RNBetterTapGestureRecognizer : UIGestureRecognizer
+@interface RNBetterTapGestureRecognizer : UIPanGestureRecognizer
+// We decided to extend UIPanGestureRecognizer
```
kmagiera: This comment would look better in an impersonal form:

> `RNBetterTapGestureRecognizer` extends `UIPanGestureRecognizer` instead of `UITapGestureRecognizer` so that translationInView can be used to make the recognizer respect max deltas constraint. Also parameters like `maxDelay` and `maxDuration` are not configurable in `UITapGestureRecognizer` and therefore require custom implementation.

Also would be better to put that comment above the interface definition

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
osdnk: Yes, please take a look on onTap method and way of invoking this method

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
```diff
@@ -60,35 +88,70 @@ private void endTap() {
...
+      mLastEventOffsetX = event.getRawX() - event.getX();
+      mLastEventOffsetY = event.getRawY() - event.getY();
+    }
 
     mLastEventOffsetX = event.getRawX() - event.getX();
     mLastEventOffsetY = event.getRawY() - event.getY();
     mLastX = event.getRawX();
     mLastY = event.getRawY();
+    if (mNumberOfPointers < event.getPointerCount()) {
+      mNumberOfPointers = event.getPointerCount();
```
kmagiera: `mNumberOfPointers` never get reset

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
```diff
@@ -60,35 +88,70 @@ private void endTap() {
...
-      begin();
-      mTapsSoFar = 0;
+    if (action == MotionEvent.ACTION_DOWN) {
+      if (state == STATE_UNDETERMINED) {
+        mOffsetX = 0;
+        mOffsetY = 0;
+        begin();
+      }
       startTap();
+    } else if (shouldFail()) {
```
kmagiera: This seems to assume that if action is ACTION_DOWN handler would never want to fail. But that seems like an incorrect assumption. E.g. wha if we have numberOfTaps=2 and maxDist set to sth small, then we do double tap but the second tap is very far from the first tap?

I mentioned it earlier but I prefer that we structure this by first checking the current state and then deciding what action to take. This would make it easier to verify that it behaves correctly in many edge cases like the one above.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
```diff
@@ -164,6 +168,18 @@ public void configure(TapGestureHandler handler, ReadableMap config) {
       if (config.hasKey(KEY_TAP_MAX_DELAY_MS)) {
         handler.setMaxDelayMs(config.getInt(KEY_TAP_MAX_DELAY_MS));
       }
+      if (config.hasKey(KEY_TAP_MAX_DELTA_X)) {
+        handler.setMaxDx(config.getInt(KEY_TAP_MAX_DELTA_X));
```
kmagiera: For params that refer to distances we need to transform from DP units that react is using into pixels. Please take a look few lines down where we extract `KEY_LONG_PRESS_MAX_DIST` for the long press handler:

```
if (config.hasKey(KEY_LONG_PRESS_MAX_DIST)) {
    handler.setMaxDist(PixelUtil.toPixelFromDIP(config.getDouble(KEY_LONG_PRESS_MAX_DIST)));
}
```

Also note that we should use `getDouble` not `getInt`

ios/Handlers/RNTapHandler.m
```diff
@@ -81,14 +99,35 @@ - (void)touchesMoved:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
...
+        return YES;
+    }
+    return NO;
+}
+
 - (void)touchesEnded:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
 {
-    [super touchesEnded:touches withEvent:event];
-    if (_numberOfTaps == _tapsSoFar) {
+    if (_numberOfTaps == _tapsSoFar && self.numberOfTouches >= _minPointers) {
```
kmagiera: From my understanding `self.numberOfTouches` refers to the current number of pointers placed on the screen. This is inconsistent with Android implementation where we take maximum number of touches over the duration of gesture.

ios/Handlers/RNTapHandler.m
```diff
@@ -51,16 +69,16 @@ - (void)cancel
 
 - (void)touchesBegan:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
 {
-    [super touchesBegan:touches withEvent:event];
+    super.minimumNumberOfTouches = 20;
```
kmagiera: move this to `init`

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
osdnk: I see, I think that failing shoudl be checked in every cases

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
osdnk: now it gest

brentvatne:
tried this on ios and i'm unable to scroll inside of the scrollview, otherwise it feels great though :)

scroll

osdnk:
@brentvatne 
It seems to be quite weird. I just opened it and everything seems to be fine. Are you sure you compiled it with my native changes?

brentvatne:
@osdnk - ah, I cleaned and then built again and it worked this time :)

ios/Handlers/RNTapHandler.m
```diff
@@ -81,14 +108,35 @@ - (void)touchesMoved:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
         }
     }
 
+    if ([self shouldFailUnderCustomCriteria]) {
+        self.state = UIGestureRecognizerStateFailed;
```
kmagiera: add:

```
[self triggerAction];
[self reset];
```

Move block above to `shouldFailUnderCustomCriteria`:

```
- (BOOL)shouldFailUnderCustomCriteria
{
   if (_gestureHandler.shouldCancelWhenOutside) {
      CGPoint pt = [self locationInView:self.view];
      if (!CGRectContainsPoint(self.view.bounds, pt)) {
        return YES;
      }
   }
   ....
```

android/lib/src/main/java/com/swmansion/gesturehandler/TapGestureHandler.java
```diff
@@ -60,35 +88,73 @@ private void endTap() {
...
-      } else if (action == MotionEvent.ACTION_DOWN) {
-        startTap();
-      }
+
+    if (shouldFail()) {
+      fail();
+    }
+
+    if (action == MotionEvent.ACTION_UP) {
+      endTap();
```
kmagiera: Make sure `endTap` is only called in BEGAN state. My suggestion is to keep the structure of this code as it was:
```
if (shouldFail()) {
   ...
} else if (state == STATE_UNDETERMINED) {
   ...
} else if (state == STATE_BEGAN) {
  ...
}
```

PR #150, opened 2018-04-20
Match types with react-native-navigation library

henrikra:
This fixes mismatch in types with [react-native-navigation](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native-navigation/index.d.ts#L10)

As you can see from the link right type is `React.ComponentType<any>` since `gestureHandlerRootHOC` is only meant to be used with `registerComponent` from `react-native-navigation` :)


kmagiera:
Makes sense, thanks a lot @henrikra

PR #151, opened 2018-04-23
Button border-related improvements

osdnk:
## Motivation
This PR is my answer to https://github.com/kmagiera/react-native-gesture-handler/issues/59
This solution is not a perfect one what was explained in the code. However, it seems to be fairly fine in most use cases
## Changes
Changes in button-related files on both iOS and Android side. 
iOS fix is only changing one flag in order to force bounding highlight layer
Android is more complex and involved adding extra layer. I didn't wish just to copy code from RN core and I suppose it calls for some changes in RN code.

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -522,7 +522,6 @@ class RectButton extends React.Component {
   static defaultProps = {
     activeOpacity: 0.105,
     underlayColor: 'black',
-    overflow: 'hidden',
   };
 
   constructor(props) {
@@ -565,7 +564,6 @@ class BorderlessButton extends React.Component {
   static defaultProps = {
     activeOpacity: 0.3,
     borderless: true,
-    overlay: 'hidden',
   };
 
   constructor(props) {
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -450,6 +450,10 @@ class BaseButton extends React.Component {
     onActiveStateChange: PropTypes.func,
   };
 
+  static defaultProps = {
+    overflow: 'hidden',
+  };
+
   constructor(props) {
     super(props);
     this._lastActive = false;
@@ -518,6 +522,7 @@ class RectButton extends React.Component {
   static defaultProps = {
     activeOpacity: 0.105,
     underlayColor: 'black',
+    overflow: 'hidden',
   };
 
   constructor(props) {
@@ -560,6 +565,7 @@ class BorderlessButton extends React.Component {
   static defaultProps = {
     activeOpacity: 0.3,
     borderless: true,
+    overlay: 'hidden',
   };
 
   constructor(props) {
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -450,10 +450,6 @@ class BaseButton extends React.Component {
     onActiveStateChange: PropTypes.func,
   };
 
-  static defaultProps = {
-    overflow: 'hidden',
-  };
-
   constructor(props) {
     super(props);
     this._lastActive = false;
@@ -494,9 +490,12 @@ class BaseButton extends React.Component {
   };
 
   render() {
+    const { style, ...rest } = this.props;
+
     return (
       <RawButton
-        {...this.props}
+        style={[{ overflow: 'hidden' }, style]}
+        {...rest}
         onGestureEvent={this._onGestureEvent}
         onHandlerStateChange={this._onHandlerStateChange}
       />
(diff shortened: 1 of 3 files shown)


ios/RNGestureHandlerButton.m
```diff
@@ -45,5 +45,11 @@ - (UIView *)hitTest:(CGPoint)point withEvent:(UIEvent *)event
     return inner;
 }
 
+- (void)layoutSubviews
+{
+    self.clipsToBounds = YES;
```
kmagiera: Can you explain what this change is for? Unless there is some weird set up it is not the best idea to do it in layoutSubviews, would be much better to just do it one in init or somewhere

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -168,6 +190,11 @@ public void setEnabled(ButtonViewGroup view, boolean enabled) {
     view.setEnabled(enabled);
   }
 
+  @ReactProp(name = ViewProps.BORDER_RADIUS)
+  public void setBorderRadius(ButtonViewGroup view, float borderRadius) {
+    view.setBorderRadius(borderRadius);
```
kmagiera: don't we need to set `mNeedBackgroundUpdate = true` here?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -79,9 +86,24 @@ private void updateBackground() {
...
+        // as a temporary solution. It do not allow to set
+        // different radius on each corner. However, I suppose it's fairly
+        // fine for button-related use cases.
+        // Therefore it might be used as long as:
+        // 1. ReactViewManager is not a generic class with a possibility to handle another ViewGroup
+        // 2. There's no way to force native behavior of ReactViewGroup's superclass's onTouchEvent
+        colorDrawable.setCornerRadius(mBorderRadius);
+        if (selectable instanceof RippleDrawable
+                && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+          PaintDrawable mask = new PaintDrawable(Color.WHITE);
```
kmagiera: would it make sense to only do it when radius is set?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
osdnk: yep, we need

ios/RNGestureHandlerButton.m
osdnk: If you do set some value of radius button on iOS, u will notice that highlight does not clip to bounds. It does change it

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
osdnk: maybe not here but in the `setBorderRadius` method

osdnk:
@kmagiera 
Do you really think so? I was searching for it for a long while and found out it as a quite normal solution and see no reasons why not to use it

GestureHandler.js
```diff
@@ -518,6 +522,7 @@ class RectButton extends React.Component {
   static defaultProps = {
     activeOpacity: 0.105,
     underlayColor: 'black',
+    overflow: 'hidden',
```
kmagiera: this may not be necessary as we render BaseButton here that already has it set as default

GestureHandler.js
```diff
@@ -560,6 +565,7 @@ class BorderlessButton extends React.Component {
   static defaultProps = {
     activeOpacity: 0.3,
     borderless: true,
+    overlay: 'hidden',
```
kmagiera: same here

GestureHandler.js
```diff
@@ -450,6 +450,10 @@ class BaseButton extends React.Component {
     onActiveStateChange: PropTypes.func,
   };
 
+  static defaultProps = {
+    overflow: 'hidden',
```
kmagiera: have you tested this? From what I remember overflow is a style property so it should go under `style` section. Can you also please verify that this setting can be overwritten (e.g. we could do `<RectButton style={{overflow: 'visible'}} />`)

osdnk: https://react-native.canny.io/feature-requests/p/style-defining-a-default-style-for-components
Default styles is not allowed currently ion RN and I believe it will be available soon.
Now it works but it's impossible to override `overflow`

kmagiera:
```
this { style, ...rest } = this.props;

<RawButton style={[{overflow: 'hidden' }, style]} {...rest} onGesture.... />
```

PR #152, opened 2018-04-23
Add number of touches on Android and iOS if needed

osdnk:
## Motivation
https://github.com/kmagiera/react-native-gesture-handler/issues/134
This PR refers to changes made someday by @tsapeta involving passing natively number of touches given out-of-box on iOS  
I just provide same mechanism on Android which seems to be quite useful.
## Changes
Added addditional field in `GestureHandler.java`, getter and few lines to update this field. Now it's also exported to js bridge

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
@@ -35,7 +35,7 @@ public class GestureHandler<T extends GestureHandler> {
   private float mHitSlop[];
 
   private boolean mShouldCancelWhenOutside;
-  private int mNumberOfPointers = 1;
+  private int mNumberOfPointers = 0;
 
   private GestureHandlerOrchestrator mOrchestrator;
   private OnTouchEventListener<T> mListener;
@@ -157,6 +157,8 @@ public class GestureHandler<T extends GestureHandler> {
     }
     mX = event.getX();
     mY = event.getY();
+    mNumberOfPointers = event.getPointerCount();
+
     mWithinBounds = isWithinBounds(mView, mX, mY);
     if (mShouldCancelWhenOutside && !mWithinBounds) {
       if (mState == STATE_ACTIVE) {
@@ -165,9 +167,6 @@ public class GestureHandler<T extends GestureHandler> {
         fail();
       }
       return;
-    }
-    if (mNumberOfPointers != event.getPointerCount()) {
-      mNumberOfPointers = event.getPointerCount();
     }
     onHandle(event);
   }
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
@@ -35,6 +35,7 @@ public class GestureHandler<T extends GestureHandler> {
   private float mHitSlop[];
 
   private boolean mShouldCancelWhenOutside;
+  private int mMaxNumberOfPointers = 1;
 
   private GestureHandlerOrchestrator mOrchestrator;
   private OnTouchEventListener<T> mListener;
@@ -132,6 +133,10 @@ public class GestureHandler<T extends GestureHandler> {
     return mY;
   }
 
+  public int getMaxNumberOfPointers() {
+    return mMaxNumberOfPointers;
+  }
+
   public boolean isWithinBounds() {
     return mWithinBounds;
   }
@@ -160,6 +165,9 @@ public class GestureHandler<T extends GestureHandler> {
         fail();
       }
       return;
+    }
+    if (mMaxNumberOfPointers < event.getPointerCount()) {
+      mMaxNumberOfPointers = event.getPointerCount();
     }
     onHandle(event);
   }
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
@@ -35,7 +35,7 @@ public class GestureHandler<T extends GestureHandler> {
   private float mHitSlop[];
 
   private boolean mShouldCancelWhenOutside;
-  private int mMaxNumberOfPointers = 1;
+  private int mNumberOfPointers = 1;
 
   private GestureHandlerOrchestrator mOrchestrator;
   private OnTouchEventListener<T> mListener;
@@ -133,8 +133,8 @@ public class GestureHandler<T extends GestureHandler> {
     return mY;
   }
 
-  public int getMaxNumberOfPointers() {
-    return mMaxNumberOfPointers;
+  public int getNumberOfPointers() {
+    return mNumberOfPointers;
   }
 
   public boolean isWithinBounds() {
@@ -166,8 +166,8 @@ public class GestureHandler<T extends GestureHandler> {
       }
       return;
     }
-    if (mMaxNumberOfPointers < event.getPointerCount()) {
-      mMaxNumberOfPointers = event.getPointerCount();
+    if (mNumberOfPointers != event.getPointerCount()) {
+      mNumberOfPointers = event.getPointerCount();
     }
     onHandle(event);
   }
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
@@ -133,7 +133,7 @@ public class RNGestureHandlerModule extends ReactContextBaseJavaModule {
     @Override
     public void extractEventData(NativeViewGestureHandler handler, WritableMap eventData) {
       eventData.putBoolean("pointerInside", handler.isWithinBounds());
-      eventData.putDouble("numberOfTouches", handler.getMaxNumberOfPointers());
+      eventData.putDouble("numberOfTouches", handler.getNumberOfPointers());
     }
   }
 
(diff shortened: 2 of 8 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
```diff
@@ -133,6 +133,7 @@ public void configure(NativeViewGestureHandler handler, ReadableMap config) {
     @Override
     public void extractEventData(NativeViewGestureHandler handler, WritableMap eventData) {
       eventData.putBoolean("pointerInside", handler.isWithinBounds());
+      eventData.putDouble("numberOfTouches", handler.getMaxNumberOfPointers());
```
kmagiera: Not sure if it was intended to provide a max number of pointers here but my intuition for this particular event param would be that it represents the current number of touches (much like pointerInside refers to the current state of the handler).
Also consider the edge case when this method is used for the state change event and make sure this is consistent with the behaviour on iOS.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
```diff
@@ -161,6 +166,9 @@ public final void handle(MotionEvent event) {
       }
       return;
     }
+    if (mNumberOfPointers != event.getPointerCount()) {
```
kmagiera: this check doesn't save much, can you get rid of it?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
```diff
@@ -133,6 +133,7 @@ public void configure(NativeViewGestureHandler handler, ReadableMap config) {
     @Override
     public void extractEventData(NativeViewGestureHandler handler, WritableMap eventData) {
       eventData.putBoolean("pointerInside", handler.isWithinBounds());
+      eventData.putDouble("numberOfTouches", handler.getNumberOfPointers());
```
kmagiera: OMG. I don't know how this slipped into here under this name. Do you mind renaming to `numberOfPointers` here and on iOS for consistency?

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
```diff
@@ -35,6 +35,7 @@
   private float mHitSlop[];
 
   private boolean mShouldCancelWhenOutside;
+  private int mNumberOfPointers = 1;
```
kmagiera: Doesn't matter too much but 0 would be a more appropriate initial value as there are no pointers when handler gets instantiated

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
```diff
@@ -161,6 +166,9 @@ public final void handle(MotionEvent event) {
       }
       return;
     }
+    if (mNumberOfPointers != event.getPointerCount()) {
+      mNumberOfPointers = event.getPointerCount();
```
kmagiera: You may want to move it up a bit as there is some logic that may cause the gesture to fail in which case it would emit an event with an incorrect number of pointers

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
tsapeta: Renaming is not a good idea since some people are already using this. I just borrowed the name that is used on iOS 🤔 If you really want to change it, it's better to set a getter for `numberOfTouches` in `nativeEvent` that returns `numberOfPointers` and warns that it's deprecated.

osdnk: @tsapeta, I do not consider it as good motivation. Well, I don't we need to care about field which are not documented and appears on only one platform

tsapeta: Fields returned in the event are not documented at all, so certainly people are checking what it has and trying on their own. If someone has reported a problem, it means that someone is using it.
Also, we don't have a changelog where we could explain that change.

kmagiera: the lib is still in alpha so breaking changes is just sth people will have to deal with when upgrading. Let's make sure to communicate this change to people reporting the issue and also when the next release comes out. We don't have capacity at the moment to support backward compatibility.

Also, we now publish change logs in the releases section on GH

kmagiera: Also, we have event field definition in typescript def file, so lets make sure to update it there too

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
```diff
@@ -370,6 +378,8 @@ public void extractEventData(RotationGestureHandler handler, WritableMap eventDa
       eventData.putDouble("anchorX", PixelUtil.toDIPFromPixel(handler.getAnchorX()));
       eventData.putDouble("anchorY", PixelUtil.toDIPFromPixel(handler.getAnchorY()));
       eventData.putDouble("velocity", handler.getVelocity());
+      eventData.putDouble("numberOfTouches", handler.getNumberOfPointers());
```
kmagiera: Also, I'm curious why did you skip fling, tap and long press handlers when adding this new parameter? Looks like we can add it directly to the base class's `HandlerFactory.extractEventData` and call super in all subclasses (you can use `https://developer.android.com/reference/android/support/annotation/CallSuper.html` to ensure that it gets called)

osdnk: @kmagiera , 
Well, I wish to follow current iOS state and not to add this field where it was not before.
I found it also quite useles on fling as long as iOS requires exact number of pointers.
However, I will add it everywhere for consciousness

osdnk:
I will add corresponing numberOfPointers on Tap on iOS when PR related to changes in Tap will be merged

ios/Handlers/RNFlingHandler.m
```diff
@@ -27,5 +27,11 @@ - (void)configure:(NSDictionary *)config
     }
 }
 
+- (RNGestureHandlerEventExtraData *)eventExtraData:(UISwipeGestureRecognizer *)recognizer
```
kmagiera: Note that this changes behavior for Fling (same for long press). They've previously been using base class implementation from `RNGestureHandler.m`:
image

It sends position and absolute position. You perhaps want to add "withNumberOfTouches" there instead

ios/Handlers/RNNativeViewHandler.m
```diff
@@ -99,21 +99,21 @@ - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
 
     [self sendEventsInState:RNGestureHandlerStateActive
              forViewWithTag:sender.reactTag
-              withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES]];
+              withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES withNumberOfTouches: _recognizer.numberOfTouches]];
```
kmagiera: nit:
` withNumberOfTouches: _recognizer.numberOfTouches]];`

should be:
`withNumberOfTouches:_recognizer.numberOfTouches]];`

(no space after colon)

Same applies below

ios/RNGestureHandlerEvents.h
```diff
@@ -28,7 +28,9 @@
                                 withAnchorPoint:(CGPoint)anchorPoint
                                    withVelocity:(CGFloat)velocity
                             withNumberOfTouches:(NSUInteger)numberOfTouches;
-+ (RNGestureHandlerEventExtraData *)forPointerInside:(BOOL)pointerInside;
++ (RNGestureHandlerEventExtraData *)forPointerInside:(BOOL)pointerInside
+                                 withNumberOfTouches:(NSUInteger)numberOfTouches;
++ (RNGestureHandlerEventExtraData *)withNumberOfTouches:(NSUInteger)numberOfTouches;
```
kmagiera: This won't be necessary if you add `withNumberOfTouches` to the very first method and use it from base class `RNGestureHandler`

ios/Handlers/RNNativeViewHandler.m
```diff
@@ -99,21 +99,21 @@ - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
 
     [self sendEventsInState:RNGestureHandlerStateActive
              forViewWithTag:sender.reactTag
-              withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES]];
+              withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES withNumberOfTouches:_recognizer.numberOfTouches]];
```
kmagiera: I believe numberOfTouches should be extracted from `event` not `_recognizer` in this case

ios/RNGestureHandlerEvents.m
```diff
@@ -66,12 +68,21 @@ + (RNGestureHandlerEventExtraData *)forRotation:(CGFloat)rotation
...
++ (RNGestureHandlerEventExtraData *)forPointerInside:(BOOL)pointerInside
+                                 withNumberOfTouches:(NSUInteger)numberOfTouches
 {
-    return [[RNGestureHandlerEventExtraData alloc] initWithData:@{@"pointerInside": @(pointerInside)}];
+    return [[RNGestureHandlerEventExtraData alloc]
+            initWithData:@{@"pointerInside": @(pointerInside),
+                           @"numberOfPointers": @(numberOfTouches)}];
+}
+
++ (RNGestureHandlerEventExtraData *)withNumberOfTouches:(NSUInteger)numberOfTouches
```
kmagiera: not used

ios/Handlers/RNNativeViewHandler.m
kmagiera: Or just stop sending numberOfPoiners for UIControl subclasses as it may not be that important

PR #157, opened 2018-04-25
Add support for references in React 16.03

osdnk:
## Motivation
Using string IDs for linking handlers has been a huge pain. The reason is that the IDs has to be globally unique to make your components reusable. This change makes it possible to replace string identifiers with a new mechanism for creating refs introduced in React 16.3 (https://reactjs.org/docs/refs-and-the-dom.html)

## Changes
When processing `waitFor` or `simultHandlers` lists we try to figure out if the provided object is a React ref (by testing if `.current` field is set). If so then we extract handler tag directly from it.

### DIFF
--- a/Example/multitap/index.js
+++ b/Example/multitap/index.js
@@ -26,6 +26,7 @@ export class PressBox extends Component {
       Alert.alert('D0able tap, good job!');
     }
   };
+  doubleTapRef = React.createRef();
   render() {
     return (
       <LongPressGestureHandler
@@ -33,9 +34,9 @@ export class PressBox extends Component {
         minDurationMs={800}>
         <TapGestureHandler
           onHandlerStateChange={this._onSingleTap}
-          waitFor="double_tap">
+          waitFor={this.doubleTapRef}>
           <TapGestureHandler
-            id="double_tap"
+            ref={this.doubleTapRef}
             onHandlerStateChange={this._onDoubleTap}
             numberOfTaps={2}>
             <View style={styles.box} />
@@ -49,9 +50,7 @@ export class PressBox extends Component {
 export default class Example extends Component {
   render() {
     return (
-      <ScrollView
-        waitFor={['image_pinch', 'image_rotation', 'image_tilt']}
-        style={styles.scrollView}>
+      <ScrollView style={styles.scrollView}>
         <LoremIpsum words={40} />
         <PressBox />
         <LoremIpsum />
--- a/Example/multitap/index.js
+++ b/Example/multitap/index.js
@@ -11,6 +11,7 @@ import {
 import { LoremIpsum } from '../common';
 
 export class PressBox extends Component {
+  doubleTapRef = React.createRef();
   _onHandlerStateChange = event => {
     if (event.nativeEvent.state === State.ACTIVE) {
       Alert.alert("I'm being pressed for so long");
@@ -26,7 +27,6 @@ export class PressBox extends Component {
       Alert.alert('D0able tap, good job!');
     }
   };
-  doubleTapRef = React.createRef();
   render() {
     return (
       <LongPressGestureHandler
--- a/Example/panAndScroll/index.js
+++ b/Example/panAndScroll/index.js
@@ -13,6 +13,8 @@ import { LoremIpsum } from '../common';
 const windowWidth = Dimensions.get('window').width;
 const circleRadius = 30;
 
+const tap = React.createRef();
+const pan = React.createRef();
 export class TapOrPan extends Component {
   constructor(props) {
     super(props);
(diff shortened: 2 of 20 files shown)


Example/panAndScroll/index.js
```diff
@@ -43,12 +45,13 @@ export class TapOrPan extends Component {
...
       <TapGestureHandler
-        id="tap"
-        waitFor="pan"
+        ref={tap}
+        waitFor={pan}
         onHandlerStateChange={this._onTapHandlerStateChange}
         shouldCancelWhenOutside>
         <PanGestureHandler
-          id="pan"
+          waitFor={this.a}
```
kmagiera: what is `this.a`?

Example/multitap/index.js
```diff
@@ -26,16 +26,17 @@ export class PressBox extends Component {
       Alert.alert('D0able tap, good job!');
     }
   };
+  doubleTapRef = React.createRef();
```
kmagiera: lets move var definition before the methods

Example/panAndScroll/index.js
```diff
@@ -75,7 +78,7 @@ export class TapOrPan extends Component {
 export default class Example extends Component {
   render() {
     return (
-      <ScrollView waitFor={['tap', 'pan']}>
+      <ScrollView waitFor={[tap, pan]}>
```
kmagiera: The whole point of this change is so that we can use `waitFor` and related methods in a "contextual" manner. That is it should be possible to instantiate as many instances of `Example` component as we could wish.

Now it isn't possible because tap and pan are global and they would be reused in the second instance of `<TapOrPan>` which would result in crash.

So instead of doing it you should declare these refs as instance variables of the `Example` component. Then you want to pass them to `<TapOrPan>` for example like that:
```
  tapRef = React.createRef();
  panRef = React.createRef();
  render() {
    return (
      <ScrollView waitFor={[this.tapRef, this.panRef]}>
        <LoremIpsum words={150} />
        <TapOrPan tapRef={this.tapRef} panRef={this.panRef}/>
        <LoremIpsum words={150} />
      </ScrollView>
    );
  }
}
```

Then use `this.props.panRef` and `this.props.tapRef` in the render method of `TapOrPan`

README.md
```diff
@@ -199,15 +199,17 @@ This type of button component should be used with simple icon-only or text-only
 
 ## Controlling gesture handlers interactions
 
-Gesture handler library API allows for defining some basic interaction between handler components. Interactions can be defined by first setting a string identifer for a handler component with the `id` property and then referencing it with `waitFor` or `simultaneousHandlers` props in other handler component.
+Gesture handler library API allows for defining some basic interaction between handler components. Interactions can be defined by first setting a ref object (`React.createRef()`) for a handler component with the `ref` property and then referencing it with `waitFor` or `simultaneousHandlers` props in other handler component.
```
kmagiera: maybe link to https://reactjs.org/docs/refs-and-the-dom.html so that it is clear that it the standard way of setting refs as of React 16.3

GestureHandler.js
```diff
@@ -112,9 +114,15 @@ function transformIntoHandlerTags(handlerIDs) {
   if (!Array.isArray(handlerIDs)) {
     handlerIDs = [handlerIDs];
   }
+
   // converts handler string IDs into their numeric tags
   return handlerIDs
-    .map(handlerID => handlerIDToTag[handlerID] || -1)
+    .map(
+      handlerID =>
+        typeof handlerID === 'string'
```
kmagiera: Can we do sth less strict here like for example:
```
handlerID => handlerIDToTag[handlerID] || (handlerID.current && handlerID.current._handlerTag) || -1
```

Also to avoid using typeof

README.md
```diff
@@ -202,15 +202,19 @@ This type of button component should be used with simple icon-only or text-only
...
+
+You could read about refs in React [docs](https://reactjs.org/docs/refs-and-the-dom.html)
```
kmagiera: could -> can

README.md
```diff
@@ -202,15 +202,19 @@ This type of button component should be used with simple icon-only or text-only
...
+
+*Note:* You may wish to use older version of React which do not support `createRef` (older than 16.03) and then you could pass string name as `id`. Mind that handlers names have to unique in whole app. This solutions is suppose to be deprecated and will be removed soon.
```
kmagiera: Rephrase:
> If you want to use link two gesture handlers (by using `waitFor` or `simultaneousHandlers `) with React older than 16.03, you need to set a globally unique string identifier for the handler with `id` property and then use it in `waitFor` or `simultaneousHandlers` props for the other handler. Note that this way of linking handlers is deprecated and will be removed in future releases

PR #163, opened 2018-05-04
Enable custom styles for DrawerLayout child container

brettdh:
As discussed in [this Stack Overflow question](https://stackoverflow.com/questions/50136741/react-native-drawerlayout-shadow-on-child-appears-on-ios-but-not-on-android), I have a need to style the wrapper component that `DrawerLayout` renders around my child component, and I could not achieve the effect I wanted by simply styling my child component, since the parent `View` was clipping the shadow on Android.

This PR enables setting custom styles on that component. (I noticed that there might be a similar need for the custom styling on the wrapper of the drawer itself, but I don't have that need currently. I'm happy to look into it in this PR if you agree that it's valuable, or we can punt it.)

### DIFF
--- a/DrawerLayout.js
+++ b/DrawerLayout.js
@@ -44,6 +44,7 @@ export type PropType = {
   hideStatusBar?: boolean,
   statusBarAnimation?: 'slide' | 'none' | 'fade',
   overlayColor: string,
+  contentContainerStyle?: any,
 
   // Properties not yet supported
   // onDrawerSlide?: Function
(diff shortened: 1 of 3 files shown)


kmagiera:
Thanks for working on it @brettdh ! The changes look good and will be happy to merge it in. My only suggestion would be around the naming. I'd prefer to stick to the convention used in RN's scroll view and instead of using `childContainerStyle` use `contentContainerStyle` (see it documented here http://facebook.github.io/react-native/docs/scrollview.html#contentcontainerstyle).

Great job!

brettdh:
Thanks for the feedback; renamed and squashed.

PR #178, opened 2018-05-19
Fix oneOfType usages

lnikkila:
Seeing some warnings due to these, I believe the parameter needs to be an array.


lnikkila:
Thanks for the great library!

Issue #58, opened 2017-11-23
Customize color of ripple on Android


brentvatne:
For RectButton I believe this is done with underlayColor, but there doesn't seem to be an option for this on BorderlessButton

gtfargo:
Wanted to note that `underlayColor` does not seem to be working on `<RectButton />` right now.

Just put together a Snack and realized this issue may be caused by setting a `backgroundColor` on the `<RectButton />`, though I'm not sure this is intended.

https://snack.expo.io/Bk2OQYRHG

MyGuySi:
Just wanted to add I'm getting this issue @gtfargo described. I had set the `backgroundColor` but removing that doesn't help, the ripple is still semi-transparent black

PR #183, opened 2018-05-21
Issue #58 - Ripple color button android

osdnk:
## Motivation 
It is response for issue https://github.com/kmagiera/react-native-gesture-handler/issues/58.
Wish to add functionality to select ripple color on Android

## Changes
Colour and props processing in JS.
Add parameter and its handling in Android Core based on https://stackoverflow.com/questions/33061050/how-can-i-modify-ripple-color-when-using-attr-selectableitembackground-as-backg/33085033
Minor stylefixes made automatically


jozan:
I don't know if this PR is going to get merged but I would very much like to have control over ripple color. Can I help to get this PR land? :)

osdnk:
@jozan 
Thank you for this message. We haven't got time to take a concern about this PR, so it has not been even reviewed. If you wish, you could do it and it might be helpful.
I see they're some dummy conflicts which are to be resolved soon

jozan:
I can test this PR after you’ve rebased to current master. I don’t have any knowledge how ripple is supposed to be used natively nor any other Android native knowledge in that regard. :)

osdnk:
I did make some minor change. Well, then I am to ping @kmagiera

vlkosinov:
Guys, let's merge it!

@osdnk
@kmagiera

brentvatne:
this would be valuable to have, it blocks me from using RNGH for the back button in react-navigation currently. cc @kmagiera

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -198,6 +215,11 @@ public void setBorderRadius(ButtonViewGroup view, float borderRadius) {
     view.setBorderRadius(borderRadius);
   }
 
+  @ReactProp(name = "rippleColor")
+  public void setRippleColor(ButtonViewGroup view, int rippleColor) {
+    view.setRippleColor(rippleColor);;
```
kmagiera:
```suggestion
    view.setRippleColor(rippleColor);
```

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -198,6 +215,11 @@ public void setBorderRadius(ButtonViewGroup view, float borderRadius) {
     view.setBorderRadius(borderRadius);
   }
 
+  @ReactProp(name = "rippleColor")
+  public void setRippleColor(ButtonViewGroup view, int rippleColor) {
```
kmagiera: how do we handle the case when ripple color has been set and then reset to null? There is this `mUseCustomRippleColor` variable but we only set it to `true` and never to `false`

kmagiera:
@brentvatne how is that navigation needs ripple customization? I think it should be enough that you can use any background color you like and default semi-transparent grey ripple should do.

brentvatne:
@kmagiera - you don't always want to do that, for example maybe you have a white background but want a blue ripple. currently this is supported in react-navigation (https://github.com/react-navigation/react-navigation-stack/blob/72914657c46963c4337fc8aa1de5a580d7fd5dda/src/views/TouchableItem.js#L46-L48) 
and if it's not supported by rngh and we switch the back button to rngh then we'll regress that functionality.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -105,6 +114,14 @@ private void updateBackground() {
             ((RippleDrawable) selectable).setDrawableByLayerId(android.R.id.mask, mask);
           }
         }
+        if (mUseCustomRippleColor
+                && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP
+                && selectable instanceof RippleDrawable) {
+          int[][] states = new int[][] { new int[] { android.R.attr.state_enabled} };
+          int[] colors = new int[] { mRippleColor };
+          ColorStateList colorStateList = new ColorStateList(states, colors);
+          ((RippleDrawable) selectable).setColor(colorStateList);
```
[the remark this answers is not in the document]
osdnk: Are you sure? How about line 119?

GestureHandler.js
```diff
@@ -563,11 +563,12 @@ class BaseButton extends React.Component {
   };
 
   render() {
-    const { style, ...rest } = this.props;
+    const { style, rippleColor, ...rest } = this.props;
 
     return (
       <RawButton
         style={[{ overflow: 'hidden' }, style]}
+        rippleColor={processColor(rippleColor)}
```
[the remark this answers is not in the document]
osdnk: nothing. It gives 0 or null

GestureHandler.js
```diff
@@ -563,11 +563,13 @@ class BaseButton extends React.Component {
...
+        useRippleColor={!!rippleColor}
```
kmagiera: why do we need this new property?

kmagiera: why can't we just use `rippleColor` property and if we want the default color then it should be `undefined`?

osdnk: Undefined is casted do zero on Android side. I know that maybe it could be achieved with one property, but I suppose it's clearer and easier solution

kmagiera: I think if you change type of rippleColor property in java to `Integer` instead of `int` it should give you `null` in that case:
```
@ReactProp(name = "rippleColor")
public void setRippleColor(ButtonViewGroup view, Integer rippleColor)
```

osdnk: Yes, you're right. I changed it this way.

PR #189, opened 2018-05-22
Add setImmediate to handler's componentDidUpdate

osdnk:
## Motivation
https://github.com/kmagiera/react-native-gesture-handler/issues/182#issuecomment-390638025
Issue was deeply inspected here and 
## Changes
In order to handle proper lifecycle of adding tag of handling decided to add `setImmediate` to CDU.

This PR it connected with https://github.com/kmagiera/react-native-gesture-handler/pull/157

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -239,25 +239,30 @@ function createHandler(handlerName, propTypes = null, config = {}) {
       });
     }
 
-    componentDidUpdate(prevProps, prevState) {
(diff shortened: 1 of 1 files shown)


henrikra:
Why this worked before?

osdnk:
@henrikra 
As long as we didn't tested issue then ComponentDidUpdate was triggered immediately after ComponentDidMount, everything was fine.

PR #209, opened 2018-06-08
Added podspec.

Burnsy:
I have added a pod spec file so users could install though cocoa pods. 
I have been using it so far with no errors.

### DIFF
--- a/react-native-gesture-handler.podspec
+++ b/react-native-gesture-handler.podspec
@@ -1,0 +1,18 @@
+require "json"
+
+json = File.read(File.join(__dir__, "package.json"))
+package = JSON.parse(json).deep_symbolize_keys
+
+Pod::Spec.new do |s|
+  s.name = package[:name]
+  s.version = package[:version]
+  s.license = { type: "MIT" }
+  s.homepage = package[:homepage]
+  s.authors = package[:author]
+  s.summary = package[:description]
+  s.source = { git: package[:repository][:url] }
+  s.source_files = "ios/**/*.{h,m}"
+  s.platform = :ios, "8.0"
+
+  s.dependency "React"
+end


kmagiera:
Thanks @Burnsy for your PR!

As I haven't use podspec based RN dependency for a while now I don't quite understand how that setup should be made. What I know is that we have a podspec located under `ios/` folder: https://github.com/kmagiera/react-native-gesture-handler/blob/master/ios/RNGestureHandler.podspec

Can you advice on if it should be placed there or somewhere else? Can you use that podspec instead of the one you submitted?

Also cc @itchingpixels who updated that other podspec a while ago in #186

Burnsy:
I would recommend these three things that most other libraries do, I overlooked it as I did not see what I was looking for, the current one is fine but here are things that differ between them. 
1) Most libraries normally have them in the base directories. (but having it in the iOS is also fine)
2) I used quite a few pod files and they normally follow the same naming scheem as the library name such as (  react-native-gesture-handler.podspec / react-native-blur.podspec etc) easier to find then also.
3) There is differences in the pod files, mine pulls more information from the packages.json file, but both are fine. 

Here the current pod file and how to install it 
`  pod 'RNGestureHandler', :path => '../../node_modules/react-native-gesture-handler/ios'`

Here the one I put forward that can be removed
`  pod 'react-native-gesture-handler', :path => '../../node_modules/react-native-gesture-handler/'`

Both will work fine, but the things I outlined are the reasons I missed it (1,2) and submitted my own.

kmagiera:
Would placing the file in different location now or changing the pod file name will break builds for people who already rely on the existing podfile?

If so I'd prefer to keep the file at the same location and only add missing entries you mentioned. Then we can clean naming and location when getting closer to some major release

jeanregisser:
@kmagiera placing the podspec at the root of the repo makes it work out of the box with `react-native link` for people using CocoaPods.
i.e. for people already having a `Podfile` it will get added there.
And it it will be business as usual for people not using CocoaPods.

gongdao123:
@kmagiera Surely, we can manually put podspec file into `Podfile`. 
But like @jeanregisser pointed out placing the podspec would be a better idea.

for people already having a `Podfile`, now since there's no podspec file in the root of the repo 
The `react-native link` command would keep link the project to the `Libraries` folder every time it runs. Which will create duplicate with Podfile.

screen shot 2018-11-27 at 5 52 12 pm

kmagiera:
Thanks for explaining @gongdao123 and @jeanregisser 
I'm open to merging this PR. But since the podspec from this PR is different than the one we have under `ios/RNGestureHandler.podspec` can you please verify that the version from this PR works ok for you? Once you confirm I'll merge this and publish new minor release

gongdao123:
@kmagiera I tried the PR and it does not work out of box. 
So I created another one #366

kmagiera:
Thanks @gongdao123 
Closing this in favor of #366

Issue #205, opened 2018-06-04
undefined is not an object (evaluating 'RNGestureHandlerModule.State')


Albert0405:
What i did 
1. npm install --save react-native-gesture-handler
2.  react-native link react-native-gesture-handler
3. Edit my MainActivity
4. Edit my MainApplication
5. import { TapGestureHandler } from 'react-native-gesture-handler';
6. Got the error below

![34484119_1694071200628127_4496503740102082560_n](https://user-images.githubusercontent.com/38030623/40913265-9e74c068-6826-11e8-8bc5-316098311ed3.png)

Albert0405:
```
"dependencies": {
    "@expo/vector-icons": "^6.3.1",
    "axios": "^0.18.0",
    "firebase": "^5.0.3",
    "react": "16.3.1",
    "react-native": "~0.55.2",
    "react-native-gesture-handler": "^1.0.4",
    "react-native-nested-scroll-view": "^6.0.1",
    "react-native-progress": "^3.4.0",
    "react-native-responsive-dimensions": "^1.0.2",
    "react-native-swiper": "^1.5.13",
    "react-native-text": "0.0.7",
    "react-native-vector-icons": "^4.6.0",
    "react-navigation": "^2.0.1",
    "watchman": "^1.0.0"
  }
```

victorporof:
I had a similar issue. I'm using CocoaPods instead of react-native link.

Add this to your fodfile: 
```
pod 'RNGestureHandler', :path => '../node_modules/react-native-gesture-handler/ios'
```

Albert0405:
Hi i solved it now what i did was is to create a new RN project using `react-native init nameOfYourProject` and install the following:

```
npm install --save react-native-gesture-handler
react-native link react-native-gesture-handler
```

and edit my MainActivity and MainApplication:

//MainActivity.java
```
package com.mlguide; // make sure you change it to your project's name, mine is com.mlguide 

import com.facebook.react.ReactActivity;
import com.facebook.react.ReactActivityDelegate;
import com.facebook.react.ReactRootView;
import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;

public class MainActivity extends ReactActivity {

    /**
     * Returns the name of the main component registered from JavaScript.
     * This is used to schedule rendering of the component.
     */
    @Override
    protected String getMainComponentName() {
        return "mlguide"; // this one too
    }

    @Override
[...]

PR #219, opened 2018-06-21
improve android installation instructions

pcooney10:
discussed in #205


kmagiera:
Awesome @pcooney10 ! Makes perfect sense. Thanks a lot!

pcooney10:
No problem! 

The docs still point to this link: https://kmagiera.github.io/react-native-gesture-handler/docs/getting-started.html, which does not seem to be updated with the new `getting-started.md` file.

kmagiera:
Should be up to date now. I forgot to deploy to gh-pages. Thanks for reminder @pcooney10

PR #224, opened 2018-06-22
Enable PanHandler to began in "TouchedBegan" method if it's possible

osdnk:
## Motivation
Issues has been described here: https://github.com/kmagiera/react-native-gesture-handler/issues/223

## Changes
Change some logic of states ion iOS in order to behave like on Android
Add possibility to set state to `BEGAN` in `touchesBegan` method under some conditions on iOS.
Remove `[self setTranslation:CGPointMake(0, 0) inView:self.view];` as it implements default behaviour which does not call for extra handling


osdnk:
Close this issue as decided to make android behave like iOS

Issue #233, opened 2018-06-30
TapGestureHandler position information inconsistent


cwhenderson20:
When using the TapGestureHandler's event on state change to determine the location of the tap, that location relative to the root, denoted by `absoluteX`/`absoluteY`, is accurate when the handler requires only one tap (`numberOfTaps={1}`), but becomes more than occasionally _inconsistent_ when more than one tap is required.

Here is a snack to demonstrate:
https://snack.expo.io/@cwhenderson20/gesture-handler-double-tap-coordinate-bug

Try tapping the small green dot and note the approximate coordinates (approximately `{ x: 198, y: 341 }` on a 4.7" device). Then double-tap and you may notice that those coordinates also approximately double (to `{ x: 396, y: 680 }`). It doesn't happen every time, however, and from what I can tell, it only happens on iOS devices.

Similar odd behavior happens if you make the double-tap handler instead a triple-tap handler. Additionally, even after removing the single-tap handler completely (so the double-tap handler is the only one registered), the inconsistencies still occur.

Happy to provide more info if necessary. Thank you for the great library!

**Edit:**
It looks like this issue began when the `UITapGestureRecognizer` was replaced with `UIPanGestureRecognizer` in PR #143. Version 1.0.0-alpha.43 (the last version before the PR) gives the correct values, while 1.0.0 does not.

cwhenderson20:
@osdnk maybe you'll have some thoughts on this as the author of the PR?

osdnk:
Hi, @cwhenderson20 
I did take a lot on it for a while but IK didn't manage to find out the issue. I didn't have enough time to take a deeper care about it, but I wish to do it asap

osdnk:
@cwhenderson20, 
Did you find any deterministic way to repro it?

cwhenderson20:
@osdnk not yet; I’ve been trying though. It seems random, but I’m sure it’s not. I’ll continue to experiment. Thanks!

osdnk:
I did manage to find solution I suppose and I will handle to do it as soon as possible

cwhenderson20:
Great news! I'll be happy to help test if you need.

osdnk:
@cwhenderson20 
Should be fixed.

PR #240, opened 2018-07-06
Issue #233 - Add touchesMoved invoke to touchesBegan in Tap

osdnk:
## Motivation 
https://github.com/kmagiera/react-native-gesture-handler/issues/233
Wish to fix this issue and manage problem when Tap's invoke is not connected with any move event.
## Repro
The most simply way to reproduce this issue is to use computer mouse on emulator as it gives the possibility to invoke touch down and up event without move. It is also possible on device but seems to be quite difficult as event stream is very sensitive
## Changes
~~Struggled to do it on various way like changing state manually, reseting UIPanGestureRecognizerState manually or even by setting proper translation
However the problem is that we need to control gesture "machine state" and it does not fit into Pan normal event flow.
Therefore TouchesMoved should to be called on began as well. 
As in TapHandler it's not a special case (contrary to PanHandler) the first touch (down event) has to be handled in the same way as move in order to allow handle to change its state not only if there was some move event during recognising
I made also codestyle fix as it didn't fit style of another files.
**So please consider that the only "real" change is line:** https://github.com/kmagiera/react-native-gesture-handler/pull/240/files#diff-5a7a364517bdcc250e56e4b950c8a083R91~~
Revert Tap changes in https://github.com/kmagiera/react-native-gesture-handler/pull/143/files and rewrite `translationInView` to be calculated every time.

### DIFF
--- a/ios/Handlers/RNTapHandler.m
+++ b/ios/Handlers/RNTapHandler.m
@@ -12,6 +12,10 @@
 
 #import <React/RCTConvert.h>
 
+// RNBetterTapGestureRecognizer does not extend UITapGestureRecognizer so that as parameters like
+// maxDelay, maxDuration, minPointers, maxDelta are not configurable in UITapGestureRecognizer
+// and therefore require custom implementation.
+
 @interface RNBetterTapGestureRecognizer : UIGestureRecognizer
 
 @property (nonatomic) NSUInteger numberOfTaps;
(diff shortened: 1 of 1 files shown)


ios/Handlers/RNTapHandler.m
```diff
@@ -108,36 +101,39 @@ - (void)touchesMoved:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
 
     self.state = UIGestureRecognizerStatePossible;
     [self triggerAction];
-    return;
+}
+
+- (CGPoint)translationInView:(UIView *)view {
```
kmagiera: don't pass view as parameter

Issue #223, opened 2018-06-22
PanGestureHandler.onBegan not consistent between iOS and Android


tapz:
In Android PanGestureHandler.onBegan event is sent immediately when the user touches the screen. However, in iOS the event is not sent before the user moves her finger.

Version: 1.0.5

j-piasecki:
We made `PanGestureHandler` on iOS to follow the same state flow as on Android in https://github.com/software-mansion/react-native-gesture-handler/pull/1611. It should now fire `onBegan` when the user touches the screen.

PR #249, opened 2018-07-10
Change activation login in Pan on Android :hand:

osdnk:
## Motivation
The author of #223 observed that the flow of activation PanGH on Android and iOS is different as on Android it gets activated immediately, but on iOS it has to wait for "move"
## Changes
As iOS behaviour seems to be more intuitive because "panning" is referring to some king of "moving" I decided to add extra condition on Android to check whether the event in "Move". If it's only `ACTION_DOWN` or `ACTION_POINTER_DOWN` it won't get active
The problem is if the event  `ACTION_POINTER_DOWN` will happen immediately after `ACTION_DOWN` and will be connected with some movement of first pointer. But this situation seems to be impossible to reproduce in real use case (even on emulator)

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.java
@@ -203,7 +203,9 @@ public class PanGestureHandler extends GestureHandler<PanGestureHandler> {
       mLastEventOffsetY = event.getRawY() - event.getY();
     }
 
-    if (state == STATE_UNDETERMINED && event.getPointerCount() >= mMinPointers) {
+    if (state == STATE_UNDETERMINED
+            && event.getPointerCount() >= mMinPointers
+            && action == MotionEvent.ACTION_MOVE) {
       mStartX = mLastX;
       mStartY = mLastY;
       mOffsetX = 0;


kmagiera:
This may cause orchestrator to move handlers from undetermined -> cancel

This may or may not be ok. Lets try to see how this works on iOS. Specifically to be able to tell how can we make this consistent. Also see #9 where other inconsistencies are described as far as delivering state changes

android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.java
```diff
@@ -203,14 +204,14 @@ protected void onHandle(MotionEvent event) {
       mLastEventOffsetY = event.getRawY() - event.getY();
     }
 
-    if (state == STATE_UNDETERMINED && event.getPointerCount() >= mMinPointers) {
+    if (state == STATE_UNDETERMINED
+            && event.getPointerCount() >= mMinPointers
+            && !mHasProperNumberOfPointers) {
```
kmagiera:
```
boolean actionPointersChanged = 
  action == MotionEvent.ACTION_DOWN 
  || action == MotionEvent.ACTION_UP 
  || action == MotionEvent.ACTION_POINTER_UP 
  || action == MotionEvent.ACTION_POINTER_DOWN;

if (state == STATE_UNDETERMINED 
    && event.getPointerCount() >= mMinPointers 
    && actionPointersChanged) {
  ...
}
```

android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.java
```diff
@@ -240,13 +241,21 @@ protected void onHandle(MotionEvent event) {
     } else if (state == STATE_BEGAN) {
       if (shouldFail()) {
         fail();
-      } else if (shouldActivate()) {
+      }
+    } else
+      if (state == STATE_UNDETERMINED
+              && action == MotionEvent.ACTION_MOVE
+              && mHasProperNumberOfPointers
```
kmagiera: ```event.getPointerCount() >= mMinPointers ```

android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.java
```diff
@@ -255,6 +264,7 @@ protected void onReset() {
       mVelocityTracker.recycle();
       mVelocityTracker = null;
     }
+    mHasProperNumberOfPointers = false;
```
kmagiera: delete

osdnk:
plz merge, @kmagiera

android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.java
```diff
@@ -221,7 +224,7 @@ protected void onHandle(MotionEvent event) {
     if (action == MotionEvent.ACTION_UP) {
       if (state == STATE_ACTIVE) {
         end();
-      } else {
+      } else if (state == STATE_BEGAN) {
```
kmagiera: Are you sure that this handler is going to be removed from active handlers list after getting UP event if it was in UNDETERMINED state before?

Issue #218, opened 2018-06-19
[Android] Buttons are clickable when React Native Inspector is opened


angly-cat:
To date on Android if you open React Native Inspector and try to select a RNGH-button, that button would execute it's click action, making debugging rather inconvenient. On iOS inspector works as expected.

osdnk:
Hi, @angly-cat 
Thank you a lot for this issue, however, I'm not able to fix it without changes in react-native code. Trigering inspectind has no side effects and then I'n not able to hook it. Switching inspector, does not imply any rerender or any callback. What's more, I found that as it's is possible to block touches on Touchables components, it's not possible to block them in our Library.
Am I right, @kmagiera?

osdnk:
I decided to close this issue. I understand that problem is present but as it's really minor we are going to have radar on it later

angly-cat:
Hi, @osdnk! Thanks for responding!

> I understand that problem is present but as it's really minor we are going to have radar on it later

This "clicking through" is very inconvenient when debugging RNGH's buttons' and handlers' styles/hitboxes etc.

> Trigering inspectind has no side effects and then I'n not able to hook it. Switching inspector, does not imply any rerender or any callback.

I've dived into `react-native`'s code and indeed found no ways to hook on inspector in `java` code. However, it's possible in JS-thread (found in [Libraries/ReactNative/AppContainer.js](https://github.com/facebook/react-native/blob/e7391438093cd5dd5033204cdce62e66509e66e1/Libraries/ReactNative/AppContainer.js#L58))

```
require('RCTDeviceEventEmitter').addListener(
    'toggleElementInspector',
    () => { /* React Native Inspector is toggled */ }
);
```

With that, if fixing this can be done in JS without putting too much effort, could you give this a second thought, please?

osdnk:
Ok, but I'm not convinced is it a good idea to use is. Maybe it is good idea to make PR for react native to export some constant with "inspect enabled" state

angly-cat:
> but I'm not convinced is it a good idea to use is

@osdnk I don't think that hooking to inspector toggling event is the best idea too. Therefore I dug into react-native's sources to find out how they "disable" touchables, so we could think of a non-hacky solution.

[...]

PR #252, opened 2018-07-13
issue #218 - Add conditional event handling with inspector

osdnk:
## Motivation
The author of the #218 issue noticed that on Android inspection does not block touches which is a great inconvenience in developing

## Changes
As there's no native or JS hook on enabling "Inpsect", I decided to use 'toggleElementInspector' which seems to be fairly fine.

I suppose it's a matter for discussion, @kmagiera, @angly-cat if it is a good idea to block every touch event depending on inspector's state

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -15,6 +15,7 @@ import {
   StyleSheet,
   FlatList,
   Platform,
+  DeviceEventEmitter,
 } from 'react-native';
 import Touchable from 'react-native/Libraries/Components/Touchable/Touchable';
 
@@ -52,6 +53,17 @@ UIManager.RCTView.directEventTypes = {
     registrationName: 'onGestureHandlerStateChange',
   },
 };
+
+let shouldHandleInspection = false;
+// Toggled inspector block touches events in ordes to allow indpecting
+// As event emitter could be hooked only on toggling an inpector
+// it's impossible to allow touches on disabling "Inspect" without toggling inpector
+if (__DEV__) {
+  DeviceEventEmitter.addListener(
+    'toggleElementInspector',
+    () => (shouldHandleInspection = !shouldHandleInspection)
+  );
+}
 
 const State = RNGestureHandlerModule.State;
 
@@ -182,6 +194,7 @@ function createHandler(handlerName, propTypes = null, config = {}) {
     }
 
     _onGestureHandlerEvent = event => {
+      if (shouldHandleInspection) return;
       if (event.nativeEvent.handlerTag === this._handlerTag) {
         this.props.onGestureEvent && this.props.onGestureEvent(event);
       } else {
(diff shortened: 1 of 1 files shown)


GestureHandler.js
```diff
@@ -53,6 +54,17 @@ UIManager.RCTView.directEventTypes = {
   },
 };
 
+let shouldHandleInspection = false;
+// Toggled inspector block touches events in order to allow inspecting on Android
+// As event emitter could be hooked only on toggling an inspector
+// it's impossible to allow touches on disabling "Inspect" without toggling inspector
+if (__DEV__ && Platform.OS === 'android') {
```
angly-cat: How about mirroring a check for `!global.__RCTProfileIsProfiling` here just as they do [there](https://github.com/facebook/react-native/blob/e7391438093cd5dd5033204cdce62e66509e66e1/Libraries/ReactNative/AppContainer.js#L56) before creating a listener for inspector toggling event?

osdnk: do you think it's a good idea not to "block" GHs if profiler is enabled?

angly-cat: Check the link I provided in previous comment: https://github.com/facebook/react-native/blob/e7391438093cd5dd5033204cdce62e66509e66e1/Libraries/ReactNative/AppContainer.js#L56
In React Native they don't create a listener for `'toggleElementInspector'` if profiler is enabled. This line was introduced in this commit: https://github.com/facebook/react-native/commit/28c07f8698fcc39a4727ddec2c1b909a1b1acf39
It seems that React Native explicitly disallow Inspector to be toggled when profiler is enabled. So it would be sensible to add similar check here, so RNGH wouldn't listen to the event that would never be emitted when profiler is enabled.

osdnk: ok, fine

GestureHandler.js
```diff
@@ -53,6 +54,17 @@ UIManager.RCTView.directEventTypes = {
   },
 };
 
+let shouldHandleInspection = false;
+// Toggled inspector block touches events in order to allow inspecting on Android
+// As event emitter could be hooked only on toggling an inspector
+// it's impossible to allow touches on disabling "Inspect" without toggling inspector
```
kmagiera: can you please rephrase that

GestureHandler.js
```diff
@@ -182,6 +194,9 @@ function createHandler(handlerName, propTypes = null, config = {}) {
     }
 
     _onGestureHandlerEvent = event => {
+      if (shouldHandleInspection) {
```
kmagiera: if we set it like that would that even work if GH is setup with an Animated.event?

osdnk: Not really, but I think it's fine enough

GestureHandler.js
```diff
@@ -196,12 +186,12 @@ function createHandler(
         }
         handlerIDToTag[props.id] = this._handlerTag;
       }
+      this.state = {
+        allowTouches: true,
```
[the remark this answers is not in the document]
osdnk: Yes. You're right. I suppose we should add another check. I'll do it as soon as possible

osdnk:
done

PR #253, opened 2018-07-13
Change states flow 

osdnk:
## Motivation
Gestures flow (especially "beginning" on Android and iOS) are inconsistent. E.g. Pan, Pinch and rotate are "beginning" on the first touch on Android, but "activating" while fulfilling conditions to be activated. On iOS gesture is "beginning" while conditions are fulfilled and immediately is started in the next frame (see apple docs)

## Changes
Having observed states flow on iOS I came to conclusion than there's a small redundancy in gestures state

Following https://developer.apple.com/documentation/uikit/uigesturerecognizer/state/began I came to conclusion than `began` and `changed` could be treated as the same RNState, because there's no need to  favourise the first "active" frame as it is currently on iOS.

On android side every handler is getting "began" immediately and I suppose that there's no need to inform about it via gesture state change. 

Taking it into consideration I decided to treat "began" as initial state for every gesture handler and only send information about "activation" of each as the first event (similarly to iOS native behaviour). 

So I filter every cancellation or failure of undetermined GH on Android (like on iOS). I also changed a bit translation of native -> RN states on iOS.

So now it became useless to use `UNDETERMINED` state on js side. I made a small research and found out that hardly nobody (on Github) were doing it, so it's not very breaking. Using `Undetermined` state on js side is useless now, but I decided to add warning and maybe remove it totally in the next version

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -55,6 +55,14 @@ UIManager.RCTView.directEventTypes = {
 
 const State = RNGestureHandlerModule.State;
 
+// Used code below as it was impossible to do it with restructuration
+Object.defineProperty(State, 'UNDETERMINED', {
+  get() {
+    console.warn('It is deprecated to use state UNDETERMINED');
+    return 0;
+  },
+});
+
 const Directions = RNGestureHandlerModule.Direction;
 
 let handlerTag = 1;
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -56,6 +56,7 @@ UIManager.RCTView.directEventTypes = {
 const State = RNGestureHandlerModule.State;
 
 // Used code below as it was impossible to do it with restructuration
+// FIXME to be removed
 Object.defineProperty(State, 'UNDETERMINED', {
   get() {
     console.warn(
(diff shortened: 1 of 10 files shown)


GestureHandler.js
```diff
@@ -55,6 +55,17 @@ UIManager.RCTView.directEventTypes = {
 
 const State = RNGestureHandlerModule.State;
 
+// Used code below as it was impossible to do it with restructuration
+// FIXME to be removed
+Object.defineProperty(State, 'UNDETERMINED', {
```
kmagiera: can we limit that to only work in DEV mode? In prod users won't see warnings anyways and we can avoid using proxies which are pretty inefficient on Android right now

docs/state.md
```diff
@@ -7,8 +7,7 @@ sidebar_label: Handler State
 As described in ["About Gesture Handlers"](about-handlers.md) section gesture handlers can be treated as ["state machines"](https://en.wikipedia.org/wiki/Finite-state_machine).
 Each handler instance at any given time has an assigned state that can change when new touch events arrive or can be forced by the touch system to change it state in certain circumstances.
 
-There are six possible states for the handler:
- - [UNDETERMINED](#undetermined)
```
kmagiera: ok, but if we remove it what is going to be the "oldState" when we go into "BEGAN"?

osdnk: I didn't wish to inform about this change.

kmagiera:
What we agreed on is to abandon this for the time being as it seem not to be necessary for the "Custom GH" PR

PR #259, opened 2018-07-17
Make chat boxes reanimated

osdnk:
## Motivation
Wish to illustrate  how powerful could be RNGH with reanimated 🎉 

## Changes
Add reanimated libary
Add chat head example
Release menu item

https://twitter.com/i/status/1019185938864443392


kmagiera:
Would be better to pull this example into reanimated repo. We already have dependency on gesture handler there and I prefer to keep gesture handler free of not-yet-standard dependencies

PR #265, opened 2018-07-29
Docs: Add an example on how to detect swipe direction

sonaye:
Hi @kmagiera,

First of all I want to thank you for your efforts working on this amazing addition to the React Native ecosystem, I've been following your progress over the last year.

I've added a simple snippet on how to use the pan handler to detect a swipe direction, I believe that adding this type of small-stripped-down use cases can be very beneficial to understand how to use this tool.

Thanks again, and I hope this work eventually gets included in core.

### DIFF
--- a/docs/handler-pan.md
+++ b/docs/handler-pan.md
@@ -99,7 +99,7 @@ See [set of event attributes from base handler class](handler-common.md#event-da
 
 Translation of the pan gesture along X axis accumulated over the time of the gesture. The value is expressed in the point units.
 
-Here a simple example of how to detect the direction of a user's swipe.
+Here is a simple example of how to detect the direction of a user's swipe.
 
 ```js
 _onHandlerStateChange = event => {
--- a/docs/handler-pan.md
+++ b/docs/handler-pan.md
@@ -103,7 +103,7 @@ Here is a simple example of how to detect the direction of a user's swipe.
 
 ```js
 _onHandlerStateChange = event => {
-  if (e.nativeEvent.oldState === State.ACTIVE) {
+  if (event.nativeEvent.oldState === State.ACTIVE) {
     const direction = event.nativeEvent.translationX < 0 ? 'right' : 'left';
     console.log(direction);
   }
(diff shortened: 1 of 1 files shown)


sonaye:
Can you share an example where your idea is used? I'm not sure I'm following your point. How can I use multiple handlers?

kmagiera:
Same way we do in pinchAndRotate sample where we use rotation and pinch handlers for the same view https://github.com/kmagiera/react-native-gesture-handler/blob/master/Example/scaleAndRotate/index.js#L87

osdnk:
I'm not convinced is it good idea to detect it via this handler. 
Please take a look that in your example swiping right (e.g.50 pts) and the left provides in consequence to trigger both `console.log`s simultaneously.
In fact, would you mind considering using `FlingGestureHandler` which seems to be exactly what you're looking for.

sonaye:
@osdnk Are you referring the example in https://github.com/kmagiera/react-native-gesture-handler/pull/265#issuecomment-410304603? because it's not working and still need some help with that. I'm ok with discarding this PR if it doesn't add a real value btw.

I tried using the `FlingGestureHandler` actually at first, but we still end up having to detect the direction in`onHandlerStateChange` I believe, are you suggesting to use two `FlingGestureHandler`?

sonaye:
Closing this PR, would appreciate any help on doing the detection the right way though for the sake of learning.

osdnk:
Yes, I suppose it's a good idea.
I think I will consider adding 'direction' to event payload because it might be unpleasant to use two handlers instead of one

satya164:
I was trying to detect the direction and ended up with this:

```js
cond(
  greaterThan(
    abs(this.translationX),
    this.swipeDistanceThreshold
  ),
  divide(this.translationX, abs(this.translationX)),
  divide(this.velocityX, abs(this.velocityX))
)
```

Using only `translationX` wasn't enough since the user could move the finger in certain direction, and then fling in the other direction. A `direction` property in the event will definitely be much better.

PR #268, opened 2018-08-07
Allow onHandlerStateChange to be an object

ckknight:
This enables onHandlerStateChange to be passed an Animated.Event which is technically not a function. Support for that case already exists, but was not represented in propTypes.


osdnk:
This issue is not related to RNR.
It's related to RNGH and fixed since 17 seconds 😄

PR #277, opened 2018-08-21
Add extra fields to long press, fling and tap event payload on Android

osdnk:
Related to https://github.com/kmagiera/react-native-gesture-handler/issues/270
## Motivation 
There was a need ot implement `absoluteX`, `absoluteY`, `x` and `y` on Android in js event payload. 
It is already done on iOS.

## Changes
I found it already done for Tap, but it was not documented.

I decided to move it to `GestureHandler.java` as the mechanism of calculating values is the same.
However, we do not provide support for `Rotation`, because it has no sense and is not done on iOS. So I decided no to export these fields to JS in case of rotation GH 

Add docs.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
@@ -220,6 +220,15 @@ public class RNGestureHandlerModule extends ReactContextBaseJavaModule {
         handler.setMaxDist(PixelUtil.toPixelFromDIP(config.getDouble(KEY_LONG_PRESS_MAX_DIST)));
       }
     }
+
+    @Override
+    public void extractEventData(LongPressGestureHandler handler, WritableMap eventData) {
+      super.extractEventData(handler, eventData);
+      eventData.putDouble("x", PixelUtil.toDIPFromPixel(handler.getLastRelativePositionX()));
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
```diff
@@ -220,6 +220,15 @@ public void configure(LongPressGestureHandler handler, ReadableMap config) {
         handler.setMaxDist(PixelUtil.toPixelFromDIP(config.getDouble(KEY_LONG_PRESS_MAX_DIST)));
       }
     }
+
+    @Override
+    public void extractEventData(LongPressGestureHandler handler, WritableMap eventData) {
+      super.extractEventData(handler, eventData);
+      eventData.putDouble("x", PixelUtil.toDIPFromPixel(handler.getLastRelativePositionX()));
```
kmagiera: Should we add it to base class `HandlerFactory` implementation instead of having it here maybe? How do we handle that on iOS?

osdnk: Please see at description, @kmagiera:
"However, we do not provide support for Rotation, because it has no sense and is not done on iOS. So I decided no to export these fields to JS in case of rotation GH"

PR #287, opened 2018-09-10
Filter animated events in config

osdnk:
There's no need to send reanimated and animated nodes in config.
Now the issue of freezing reanimated nodes is solved! 

Updated lint-staged. It's not so important

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -109,7 +109,7 @@ const stateToPropMappings = {
   [State.END]: 'onEnded',
 };
 
-function isEventHandler(param, name) {
+function isConfigParam(param, name) {
   return (
     param !== undefined &&
     typeof param !== 'function' &&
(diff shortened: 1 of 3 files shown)


GestureHandler.js
```diff
@@ -145,6 +145,8 @@ function filterConfig(props, validProps, defaults = {}) {
         if (typeof value !== 'object') {
           value = { top: value, left: value, bottom: value, right: value };
         }
+      } else if (key === 'onHandlerStateChange' || key === 'onGestureEvent') {
```
osdnk: check if adding reanimated `__isNative` in RNR fixes

GestureHandler.js
```diff
@@ -109,11 +109,13 @@ const stateToPropMappings = {
   [State.END]: 'onEnded',
 };
 
-function canUseNativeParam(param) {
+function isEventHandler(param, name) {
```
kmagiera: but this check if it is *not* an event handler. Also we do `param !== undefined` which will filter some other things too. I suggest renaming to `isConfigParam`

PR #292, opened 2018-09-21
Unify PanGestureHandler activation/fail criteria

osdnk:
## Motivation
We now can set a range for when gesture can be activated (have minOffsetY and minDeltaY) but the criteria for cancelling can only be specified as a delta (that is we can only have symmetrical range in both directions using maxDeltaY)

The usecase for that would be when we want to have PanGestureHandler that can activate after pulling down 20 pts but immediately cancel when pulling up. Then if we have another handler waiting for that Pan we have to cancel instead of staying idle while pulling up.

I think the improvement we may try to make is to change the set of parameters on the native side to accept "minOffsetRangeStart" and "minOffsetRangeEnd" for both X and Y and then also have "maxOffsetRangeStart" and "maxOffsetRangeEnd" for both X and Y. Then we can use these to implement "minDelta", "minOffset" and "maxDelta" for compatibility (of course some mix of these props will not be supported)

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -154,13 +154,10 @@ function filterConfig(props, validProps, defaults = {}) {
   return res;
 }
 function backwardCompatibleTransformProps(handlerName, props) {
-  console.warn('XX');
   const res = {};
   if (handlerName === 'PanGestureHandler') {
-    console.warn('XX2');
     // TODO inv
     if (props.minDeltaX) {
-      console.warn('XXX');
       res.minOffsetRangeStartX = -props.minDeltaX;
       res.minOffsetRangeEndX = props.minDeltaX;
     }
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -412,22 +412,22 @@ function backwardCompatibleTransformPanProps(props) {
   if (__DEV__) {
     if (props.minDeltaX && props.activeOffsetX) {
       throw new Error(
-        `It's not supported use minDeltaX with activeOffsetStartX or activeOffsetEndX`
+        `It's not supported use minDeltaX with activeOffsetXStart or activeOffsetXEnd`
       );
     }
     if (props.maxDeltaX && props.failOffsetX) {
       throw new Error(
-        `It's not supported use minDeltaX with activeOffsetStartX or activeOffsetEndX`
+        `It's not supported use minDeltaX with activeOffsetXStart or activeOffsetXEnd`
       );
     }
     if (props.minDeltaY && props.activeOffsetY) {
       throw new Error(
-        `It's not supported use minDeltaX with activeOffsetStartY or activeOffsetEndY`
+        `It's not supported use minDeltaX with activeOffsetYStart or activeOffsetYEnd`
       );
     }
     if (props.maxDeltaY && props.failOffsetY) {
       throw new Error(
-        `It's not supported use minDeltaX with activeOffsetStartY or activeOffsetEndY`
+        `It's not supported use minDeltaX with activeOffsetYStart or activeOffsetYEnd`
       );
     }
     if (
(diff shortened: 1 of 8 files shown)


ios/Handlers/RNPanHandler.m
```diff
@@ -12,153 +12,169 @@
...
-        // We use "minimumNumberOfTouches" property to prevent pan handler from recognizing
-        // the gesture too early before we are sure that all criteria (e.g. minimum distance
-        // etc. are met)
-        super.minimumNumberOfTouches = 20;
-    } else {
-        super.minimumNumberOfTouches = _realMinimumNumberOfTouches;
-    }
-    [super touchesBegan:touches withEvent:event];
+  if (_hasCustomActivationCriteria) {
+    // We use "minimumNumberOfTouches" property to prevent pan handler from recognizing
```
osdnk: Changest intent

GestureHandler.js
```diff
@@ -153,6 +153,63 @@ function filterConfig(props, validProps, defaults = {}) {
   });
   return res;
 }
+function backwardCompatibleTransformProps(handlerName, props) {
+  const res = {};
+  if (handlerName === 'PanGestureHandler') {
```
kmagiera: I'd recommend that we add additional callback to `createHandler` called `transformProps` and then call it in `componentDidMount`:
```
this._config = filterConfig(
      transformProps(this.props),
      this.constructor.propTypes,
      config
);
```

Then for `PanHandler` we can have a transform that translates old props into new style props and for other handlers that callback would be empty (no props transforming).

Another thing is that I'd like to have that code which checks for props compatibility wrapped in `__DEV__` so that it only runs in dev mode and use `console.warning` to report.

Finally, I'd like if we could add "shortcut" props for defining the range and allow `minOffsetRange` to be an array `[start, end]`

android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.java
```diff
@@ -122,20 +139,20 @@ public PanGestureHandler setMinVelocityY(float minVelocityY) {
 
   private boolean shouldActivate() {
     float dx = mLastX - mStartX + mOffsetX;
-    if (mMinDeltaX != MIN_VALUE_IGNORE && Math.abs(dx) >= mMinDeltaX) {
+    if (mMinOffsetRangeStartX != MIN_VALUE_IGNORE && dx <=  mMinOffsetRangeStartX) {
```
kmagiera: I don't think this is quite correct, if the minOffsetRange is (-50, -20) it means I only want to activate when pulled up by 20px but because of that it will activate immediately because initial offset of 0 is greater than -50

kmagiera:
Following our offline discussion about the logic for activating / canceling I think it makes sense to keep it this way but I'd prefer that we rename props to make easier to distinguish (e.g. min vs max is kind of loosely associated with what exactly happens when we hit certain threshold).

I suggest the following:
1) We expose only two new props: `activeOffset` and `failOffset`, both props can take an array of two elements (first lower or equal to 0, second greater or equal to 0) or can take a single number `X` which then gets converted to range `[-X, X]`
2) Under the hood we add four props (`activeOffsetStart`, `activeOffsetEnd`, etc.) – this is important as having it split as separate props will allow us to pass animated values in the future (animated does not support complex props like arrays)
3) Transform logic + params validation should be done as stated before and inside `if (__DEV__)` block

osdnk:
TODO: maybe some example, review docs

@kmagiera, maybe it's high time to split GH into files?

GestureHandler.js
```diff
@@ -400,15 +406,165 @@ const LongPressGestureHandler = createHandler(
   },
   {}
 );
+
+function backwardCompatibleTransformPanProps(props) {
```
kmagiera: Don't like this naming. I suggest that we split this method into two: `transformPanGestureHandlerProps` and `validatePanGestureHandlerProps`

The first one will handle transforming while the second one will be called from inside __DEV__ block and perform all the checks.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
```diff
@@ -251,26 +253,36 @@ public PanGestureHandler create(Context context) {
     public void configure(PanGestureHandler handler, ReadableMap config) {
       super.configure(handler, config);
       boolean hasCustomActivationCriteria = false;
-      if (config.hasKey(KEY_PAN_MIN_DELTA_X)) {
-        handler.setMinDx(PixelUtil.toPixelFromDIP(config.getDouble(KEY_PAN_MIN_DELTA_X)));
+      if(config.hasKey(KEY_PAN_ACTIVE_OFFSET_START_X)) {
+        handler.setactiveOffsetStartX(PixelUtil.toPixelFromDIP(config.getDouble(KEY_PAN_ACTIVE_OFFSET_START_X)));
```
kmagiera: camel case -> `setActiveOffsetStartX` (I'd actually prefer it to be called `setActiveOffsetXStart` because "activeOffsetX" is a name of a range that has "start" and "end")

docs/handler-pan.md
```diff
@@ -47,45 +47,84 @@ See [set of properties inherited from base handler class](handler-common.md#prop
...
+When the given number of fingers is placed on the screen and handler hasn't yet [activated](state.md#active) it will fail recognizing the gesture. Should be a positive integer.
 
 ---
-### `minDeltaX`
+### `activeOffsetX`
 
-Minimum distance expressed in points along X axis the finger (or multiple finger) need to travel (left or right) before the handler [activates](state.md#active). Unlike [`minoffsetx`](#minoffsetx) this parameter accepts only non-negative numbers that represents the distance in point units. If you want for the handler to [activate](state.md#active) for the movement in one particular direction use [`minOffsetX`](#minoffsetx) instead.
+Range along X axis where fingers travels without activation of handler. Moving outside this range implies activation of handler. Range could be given as an array or a single number.
+ If range is set as array, first value must be negative, a the second one positive.
```
kmagiera: as an array

negative -> lower or equal to 0
positive -> higher or equal to 0

osdnk:
I'll also change some examples

PR #296, opened 2018-09-26
Factor in the current drawer translation when animating

brentvatne:
If you close a drawer in the middle of opening it, then it should animate from the position that it was at rather than jumping to the fully opened position then animating to close. Same thing if you open while closing.

### DIFF
--- a/DrawerLayout.js
+++ b/DrawerLayout.js
@@ -92,11 +92,6 @@ export default class DrawerLayout extends Component<PropType, StateType> {
     const touchX = new Animated.Value(0);
     const drawerTranslation = new Animated.Value(0);
 
-    // This is necessary to ensure that the value we get from stopAnimation is
-    // kept up to date. If we don't do this then values on the JS side never get
-    // updated and drawerTranslation value in JS will always be 0.
-    this.drawerTranslationListener = drawerTranslation.addListener(() => {});
-
     this.state = {
       dragX,
       touchX,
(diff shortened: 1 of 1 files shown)


DrawerLayout.js
```diff
@@ -92,6 +92,11 @@ export default class DrawerLayout extends Component<PropType, StateType> {
     const touchX = new Animated.Value(0);
     const drawerTranslation = new Animated.Value(0);
 
+    // This is necessary to ensure that the value we get from stopAnimation is
+    // kept up to date. If we don't do this then values on the JS side never get
+    // updated and drawerTranslation value in JS will always be 0.
+    this.drawerTranslationListener = drawerTranslation.addListener(() => {});
```
brentvatne: 😢

osdnk: 😢

kmagiera: Are you sure we need start value to pass to `_animatedDrawer`? Looks like we use it to call `setValue` so maybe in some cases we can skip calling `setValue` and just let it animate

brentvatne: yeah that would work too

brentvatne:
fixed

DrawerLayout.js
```diff
@@ -254,18 +254,37 @@ export default class DrawerLayout extends Component<PropType, StateType> {
...
+    toValue: number,
+    velocity: number,
+  }) => {
     this.state.dragX.setValue(0);
     this.state.touchX.setValue(
       this.props.drawerPosition === 'left' ? 0 : this.state.containerWidth
     );
-    this.state.drawerTranslation.setValue(fromValue);
+
+    if (typeof fromValue === 'number') {
```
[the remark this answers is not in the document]
brentvatne: probably not. makes more sense to check for undefined

Issue #316, opened 2018-10-18
LongPress gesture cancellation


rt2zz:
On Android, when a long press is activated, and the user then pans away (without lifting the gesture) the LongPressGestureHandler enters a cancelled state.

in iOS the gesture does not cancel and continues to fire gesture events until the gesture is released by lifting the finger.

This comment seems to suggest that the iOS behavior is intended and that android has a bug: https://github.com/kmagiera/react-native-gesture-handler/issues/94#issuecomment-355750006

The bug prevents (afaict) the feature I am working on from being possible in RNGH: activate a chart scrubber upon long press.

jeanregisser:
I'm affected by this bug as well.

Did you find any workaround @rt2zz?

rt2zz:
I have not, I suspect this could be achieved with PanGestureHandler with simultaneous handlers setup, but there were some implementation challenges I ran into and am not sure what a solution would look like.

lvmtam:
I also have the same issue. Seem to be strange.

rt2zz:
found the issue, on android `maxDist` defaults to a small value thus cancelling the gesture after a short pan. Set that to screen width, or whatever makes sense for your component to fix this issue.

Not sure why the iOS behavior does not match?

PR #321, opened 2018-10-22
Send more transition state-related events from iOS

osdnk:
## Motivation 
Cancellation from `UIGestureRecognizerStatePossible` of GH on iOS was previously done silently. Now it's being sent.

## Changes
Split sending-related logic to two methods
Add invoking state changes-related sender from state setter


kmagiera:
How is this related to #316 if that issue states that it is a bug in Android implementation whereas this PR only make changes to iOS?

PR #322, opened 2018-10-22
Fix newProps calculation on update

osdnk:
Inspired by https://github.com/kmagiera/react-native-gesture-handler/issues/311
With with props I forgot to update calculation mechanism on `componentDidUpdate`

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -278,8 +278,8 @@ function createHandler(
 
     _update() {
       const newConfig = filterConfig(
-        this.props,
-        this.constructor.propTypes,
+        transformProps ? transformProps(this.props) : this.props,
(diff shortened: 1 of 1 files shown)


GestureHandler.js
```diff
@@ -382,7 +382,6 @@ const TapGestureHandler = createHandler(
     numberOfTaps: PropTypes.number,
     maxDeltaX: PropTypes.number,
     maxDeltaY: PropTypes.number,
-    minPointers: PropTypes.number,
```
osdnk: Duplication with https://github.com/kmagiera/react-native-gesture-handler/pull/322/files#diff-56d925ec2aada9e1b3b613fe5b782837L387

PR #330, opened 2018-11-06
[TS] Add testID to button props

Zycon42:
All RN views support testID out of the box. It's handled by default view manager implementation.


diegolmello:
@kmagiera Can you bump a version? Thanks :)

PR #345, opened 2018-11-20
Add support fo RN 0.57 (+ fix deprecated configuration)

guix77:


usrbowe:
I would also consider upgrading the default versions of sdk to this (from RN build.gradle)
```
buildToolsVersion = "27.0.3"
minSdkVersion = 16
compileSdkVersion = 27
targetSdkVersion = 26
```

guix77:
Wouldn't that make it less downwards compatible with devices?

usrbowe:
> Wouldn't that make it less downwards compatible with devices?

Oh yes, that's a good point. Let's keep the current, if it doesn't cause any issues.

PR #348, opened 2018-11-22
update Gradle wrapper

dulmandakh:
Starting with RN 0.56, RN projects expose SDK configuration for reuse. This PR will try to make use of or use default values provided in Gradle. Otherwise it'll try to download many things, like Android 25 API and build tools for it.

It works for me 👍


dulmandakh:
Ok, will do it tomorrow.

>

kmagiera:
Looks good! Thanks!

PR #351, opened 2018-11-25
changed `gestureHandlerRootHOC.ios.js` to `gestureHandlerRootHOC.js`

EvanBacon:
This will make `react-native-gesture-handler` _more_ compatible with web.


EvanBacon:
probably safer to just add a new file for versioning.

kmagiera:
I was more asking if packager won't just ignore .android and .ios when .js is present. As that's what imo is the default behavior, but haven't tested it

EvanBacon:
AFAIK priority is ordered as such: `.android.js` -> `.native.js` -> `.js` and this is preserved in the build.

PR #366, opened 2018-12-03
Add Pod Spec [BREAKING]

gongdao123:
The other PR #209 use use different project name than ios/RNGestureHandler.podspec, this might create some conflict.

Also podspec file need to be included with `package.json`, otherwise it will be ignored during `npm install`.

**[BREAKING]** Podspec location has been changed from `./ios/RNGestureHandler.podspec` to `./RNGestureHandler.podspec`. The previous consumers of the podspec should update their podfile to reflect that change

### DIFF
--- a/RNGestureHandler.podspec
+++ b/RNGestureHandler.podspec
@@ -2,7 +2,7 @@ require "json"
 
 Pod::Spec.new do |s|
   # NPM package specification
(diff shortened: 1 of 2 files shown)


kmagiera:
Tested this and appears to be working. This is a breaking change for podspec consumers and will require them to update podspec path after upgrade. I added a comment re that to your PR description and will make sure one is included in release notes too. Thanks!

PR #368, opened 2018-12-03
Set deployment target to 8.0 on iOS

osdnk:
Bump deployment target version to the lowest possible on iOS.
Issue has been described here:
https://github.com/kmagiera/react-native-gesture-handler/issues/349


kmagiera:
I understand that with this we drop support for some earlier versions of React Native. If so it would be useful to know which versions we will no longer support so that this can be added to the docs

henrikra:
In React Native 0.56 the minimum ios version is 9 https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#ios-9-is-now-the-minimum-required-version

Example/ios/GestureHandler.xcodeproj/project.pbxproj
```diff
@@ -270,6 +270,20 @@
 			remoteGlobalIDString = 2D2A28201D9B03D100D4039D;
 			remoteInfo = "RCTAnimation-tvOS";
 		};
+		66DFB4CB21B57F240019EA62 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 146833FF1AC3E56700842450 /* React.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = EBF21BDC1FC498900052F4D5;
+			remoteInfo = jsinspector;
```
kmagiera: what are these new entries added here?

osdnk:
@kmagiera @henrikra 
Realised that I was bumping iOS version for Example, not for library. Sorry 😄

PR #369, opened 2018-12-04
Optimize DrawerLayout by avoiding setState during animation

kmagiera:
This PR adds two optimizations that makes using DrawerLayout implementation much more enjoyable, specifically by minimizing junked caused by the fact we need to do JS roundtip in between the end of a gesture and start of the animation.

Optimizations made:
 - avoid keeping `drawerShown` in state. Instead we keep references to a couple of views that we need to update and call `setNativeProps` directly
 - added `setNativeProps` implementation to gesture handler component so that it can be updated via ref without re-rendering
 - predict next animation position when using native driver after the gesture is release. Previously it was causing an extra frame of deley as even when the animation has started it'd still start from the last position of a drawer and only with the next frame it'd start moving. Now we force it to move a bit before starting the animation as we know that the animation will start delayed.

### DIFF
--- a/DrawerLayout.js
+++ b/DrawerLayout.js
@@ -48,7 +48,6 @@ export type PropType = {
 };
 
 export type StateType = {
-  drawerShown: boolean,
   dragX: any,
   touchX: any,
   drawerTranslation: any,
@@ -80,6 +79,10 @@ export default class DrawerLayout extends Component<PropType, StateType> {
   };
   _openValue: ?Animated.Interpolation;
   _onGestureEvent: ?AnimatedEvent;
+  _accessibilityIsModalView = React.createRef();
+  _pointerEventsView = React.createRef();
+  _panGestureHandler = React.createRef();
+  _drawerShown = false;
 
   constructor(props: PropType, context: any) {
     super(props, context);
@@ -92,7 +95,6 @@ export default class DrawerLayout extends Component<PropType, StateType> {
       dragX,
       touchX,
       drawerTranslation,
-      drawerShown: false,
       containerWidth: 0,
     };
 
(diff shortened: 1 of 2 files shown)


brentvatne:
I tested this on Expo SDK 31 by just patching GestureHandler.js to add setNativeProps and it works nicely, couldn't find any issues. When this patch is released we should bump react-native-gesture-handler to 1.1.0 because adding `setNativeProps` is a new feature that we will be depending on, and I'll need to change the peer dependency in react-navigation to react-native-gesture-handler@^1.1.0.

Also, I applied this diff locally to react-navigation's vendored version of DrawerLayout, where I have fixes RTL support, drawerLockMode support, and a `gestureRef` prop that I have been too lazy to upstream here. Would you mind if I opened a single PR with all of these changes or do you prefer that I open one PR for each feature?

kmagiera:
Feel free to submit your improvements in the most convenient way for you. I could list all the improvements separately in release notes so hope that people who are interested will get to see it regardless.

Also, I tested this change some more today and various phones and also try patching navigation – seems to be working good. So I'm going to merge this in and wait for @brentvatne changes mentioned above before publishing.

As for the version bump I kind of wanted to start 1.1.0 with my [overflow_hittest](https://github.com/kmagiera/react-native-gesture-handler/tree/overflow_hittest) changes merged in. They rely on one method added to RN 57.3 and therefore would make that version incompatible with previous versions of RN. Then I could maintain minor matches under 1.0.X for people who are still below 0.57. But I'm open to discuss other ways of handling that

brentvatne:
@kmagiera - looks good to land this from my point of view, I'll submit a PR afterwards with the changes from react-navigation's fork

PR #372, opened 2018-12-07
Add touchables

osdnk:
Please review with "no whitespaces". I have linted all Obj-C files I touched.

## Motivation
This PR provides a gesture-handler's native button backed implementation of Touchabnles. The goal is to provide a compatibility layer for Touchable components from RN core but make it backed by the gesture-handler components that integrates better with the native touch system.

## Changes
In order to handle it I prepared following polyfills inspired by RN's core implementation. Using `LongPressGestureHandler` and `TapGestureHandler` and few new features and fixes in RNGH I managed to imitate RN's touchables. 

### How it works?
My implementation of Touchables is a states' machine which appears to be be a bit simpler than RN's implementation and has beed documented with comments in code.

### Add examples
Added example for many use cases of Touchables (compared RN's and GN's touchables).

## Differences
`TouchableWithNativeFeedback`s are a bit different here and in RN's core. It's visible on long pressing. I decided not to change it since it would call for a lot of fixes in native node. However my implementation is closer do native behaviour. RN's implementation was manually triggering ripple effect via bridge from JS.


osdnk:
@kmagiera 
If you're fine with these changes, I'll add docs and typings

kmagiera:
Have you considered adding an event prop to the regular gesture event that tells if the event has been emitted inside/outside of the bounds? It seem like that would simplify the code quite a bit

osdnk:
Yes, I've considered, but we need to inform about every passing of bounds

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
```diff
@@ -51,7 +50,7 @@ private static void initPointerProps(int size) {
   private View mView;
   private int mState = STATE_UNDETERMINED;
   private float mX, mY;
-  private boolean mWithinBounds;
+  protected boolean mWithinBounds;
```
kmagiera: This does not seem very consistent. We change visibility of this member only so that we can add public getter to it in a subclass? Why can't we do the same as we do in all the other variables around `mX`, `mY` etc and just have a getter directly in this class?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -60,6 +61,15 @@ public void setBorderRadius(float borderRadius) {
       mNeedBackgroundUpdate = true;
     }
 
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    private RippleDrawable withRipple(RippleDrawable selectable) {
```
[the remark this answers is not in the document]
kmagiera: ok, but can we change the name to be a bit more descriptive? In Java the convention for methods is usually to start with a verb. So, for example, we could have `applyBackgroundForRipple`

Also it seem that in each place you call this method you cast to RippleDrawable. You may want to consider doing this in this method + in addition do the check if `mRippleColor` is null etc. I believe we should check for it in other places too

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java
```diff
@@ -333,6 +333,7 @@ public void extractEventData(PanGestureHandler handler, WritableMap eventData) {
       eventData.putDouble("translationY", PixelUtil.toDIPFromPixel(handler.getTranslationY()));
       eventData.putDouble("velocityX", PixelUtil.toDIPFromPixel(handler.getVelocityX()));
       eventData.putDouble("velocityY", PixelUtil.toDIPFromPixel(handler.getVelocityY()));
+      eventData.putBoolean("pointerInside",handler.getWithinBounds());
```
kmagiera: space

ios/Handlers/RNLongPressHandler.m
```diff
@@ -19,28 +19,25 @@ - (id)initWithGestureHandler:(RNGestureHandler*)gestureHandler;
...
-
-    if (_gestureHandler.shouldCancelWhenOutside) {
-        CGPoint pt = [self locationInView:self.view];
-        if (!CGRectContainsPoint(self.view.bounds, pt)) {
-            self.enabled = NO;
-            self.enabled = YES;
-        }
-    }
+  [super touchesMoved:touches withEvent:event];
+  CGPoint pt = [self locationInView:self.view];
```
kmagiera: why this change? It does not seem to change anything? If it is needed than it could be because of some side effect which in such a case should be properly documented

touchables/TouchableNativeFeedback.js
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
+import { View, Text, StyleSheet } from 'react-native';
+
+/**
+ * UI copied from RN
```
kmagiera: what? how do you copy UI?

touchables/TouchableNativeFeedback.js
```diff
@@ -0,0 +1,39 @@
...
+  static SelectableBackground = () => ({});
+  static SelectableBackgroundBorderless = () => ({});
+  static Ripple = () => ({});
+  static canUseNativeForeground = () => false;
+
+  render() {
+    return (
+      <View style={[styles.container, this.props.style]}>
+        <Text style={styles.info}>
+          TouchableNativeFeedback is not supported on this platform!
```
kmagiera: Are you sure this is how that view renders in React Native on iOS?

touchables/TouchableOpacity.js
```diff
@@ -0,0 +1,62 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * TouchableOpacity base timing animation which has been used in RN's core
+ */
+export default class TouchableOpacity extends TouchableWithoutFeedback {
+  static defaultProps = {
+    ...TouchableWithoutFeedback.defaultProps,
+    // value copier from RN's implementation
```
kmagiera: typo

touchables/TouchableWithoutFeedback.js
```diff
@@ -0,0 +1,346 @@
...
+      accessibilityStates: this.props.accessibilityStates,
+      accessibilityTraits: this.props.accessibilityTraits,
+      nativeID: this.props.nativeID,
+      testID: this.props.testID,
+      onLayout: this.props.onLayout,
+      hitSlop: this.props.hitSlop,
+    };
+
+    return (
+      <PanGestureHandler
```
kmagiera: why do we need both pan and tap handlers? can we maybe render longpress only optionally? 99% of touchables people use don't need longpress and we could maybe only have that when longpress callback is provided as a prop

ios/Handlers/RNPanHandler.m
```diff
@@ -84,9 +86,10 @@ - (void)touchesMoved:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
     self.state = UIGestureRecognizerStateFailed;
     return;
   }
-  if ((self.state == UIGestureRecognizerStatePossible || self.state == UIGestureRecognizerStateChanged) && _gestureHandler.shouldCancelWhenOutside) {
+  if ((self.state == UIGestureRecognizerStatePossible || self.state == UIGestureRecognizerStateChanged)) {
     CGPoint pt = [self locationInView:self.view];
-    if (!CGRectContainsPoint(self.view.bounds, pt)) {
+    _isWithinBounds = [_gestureHandler containsPointInView:pt];
```
kmagiera: can we calculate `isWithinBounds` on demand instead of storing it under _isWithinBounds variable? It also seem that `containsPointInView` method you added would be a good candidate for that. It also seem like it does not need to take `pt` as argument because you always call to `[self locationInView:]` prior to using it

touchables/TouchableWithoutFeedback.js
osdnk: Tap is needed because we cannot handle touch without any movement by pan. Pan is needed because it's a continuous gesture which could send event on each frame.

kmagiera: Can we use GH buttons in that case? (RectButton etc)

osdnk: Not really since event delivered from this gesture handler does not tell anything about accurate position. I think it's not a big cost to use two of them and it's not worth to add extra parameters for `NativeViewGestureHandler` for this specific case.

osdnk:
Following up our Slack conversion, I'm going to change this PR in order not to use inheritance in favour of composition

touchables/TouchableNativeFeedback.js
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
+import { View, Text, StyleSheet } from 'react-native';
+
+/**
+ * This implementation is a copy of react-native behavior.
```
kmagiera: In this case can we import `TouchableNativeFeedback` directly from RN and export it here instead of copying the code?

index.js
```diff
@@ -1,3 +1,4 @@
 export { default as Swipeable } from './Swipeable';
 export { default as DrawerLayout } from './DrawerLayout';
 export * from './GestureHandler';
+export * from './touchables';
```
zocoi: @osdnk this breaks the latest installation as the folder `touchables` is in Examples

kmagiera: Sorry for the trouble @zocoi we noticed that few seconds after publishing 1.0.13 and 1.0.14 is already published on npm and includes missing files

zocoi: @kmagiera will you cut a new release soon? Or can I use an older version?

kmagiera: As I mentioned – 1.0.14 is already out and it fixes the problem

PR #374, opened 2018-12-09
Force Touch  GH

osdnk:
## Motivation
Add possibility of handling force touch gesture. 

## Changes
Add proper handler on iOS and no-op fallback on Android. On iOS for calculating force touch only first finger is concerned which is normal behaviour I observed on iOS. Force is expressed as value greater of equal 0 and lesser or equal 1.
I added two options of customisation.

### `minForce`
`minForce` is a minimal pressure which has to be applied in order to activate handler.

### `maxForce`
`minForce` is a maximal force which could be applied. If greater force applied, handler fails. 

On releasing first finger gesture fails or ends (if has been activated already).

And added simple check method to determine is it possible to use `forceTouch` on current device which exporting RN's constant.

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -4,6 +4,7 @@ import {
   requireNativeComponent,
   Animated,
   NativeModules,
+  Text,
   ScrollView,
   Slider,
   Switch,
@@ -411,18 +412,20 @@ const FlingGestureHandler = createHandler(
 
 const ForceTouchFallback = props => props.children;
 
-const ForceTouchGestureHandler =
-  Platform.OS === 'ios'
-    ? createHandler(
-        'ForceTouchGestureHandler',
-        {
-          minForce: PropTypes.number,
-          maxForce: PropTypes.number,
-          feedbackOnActivation: PropTypes.bool,
-        },
-        {}
-      )
-    : ForceTouchFallback;
+const ForceTouchGestureHandler = NativeModules.PlatformConstants
+  .forceTouchAvailable ? (
+  createHandler(
+    'ForceTouchGestureHandler',
+    {
+      minForce: PropTypes.number,
+      maxForce: PropTypes.number,
+      feedbackOnActivation: PropTypes.bool,
+    },
+    {}
+  )
+) : (
+  <Text>Force Touch is not available on this platform</Text>
+);
 
 ForceTouchGestureHandler.forceTouchAvailable =
   NativeModules.PlatformConstants.forceTouchAvailable || false;
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -418,7 +418,7 @@ const ForceTouchGestureHandler =
         {
           minForce: PropTypes.number,
           maxForce: PropTypes.number,
-          feedbackOnActivation: PropTypes.boolean,
+          feedbackOnActivation: PropTypes.bool,
         },
         {}
       )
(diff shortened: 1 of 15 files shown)


GestureHandler.js
```diff
@@ -409,6 +410,28 @@ const FlingGestureHandler = createHandler(
   {}
 );
 
+const ForceTouchFallback = props => (
+  <Text style={{ color: 'red' }}>
+    Force Touch is not available on this platform
```
kmagiera: I don't think it is a good idea to display text like that. Some people may still only test their stuff on the newest iPhone and never see that until it is too late. Lets keep it render the same way as it used to but add `console.warn` to `componentDidMount` maybe?

GestureHandler.js
```diff
@@ -409,6 +409,31 @@ const FlingGestureHandler = createHandler(
   {}
 );
 
+class ForceTouchFallback extends React.Component {
+  componentDidMount() {
+    console.warn('Force Touch is not available on this platform');
```
kmagiera:
```suggestion
    console.warn('ForceTouchGestureHandler is not available on this platform. Please use ForceTouchGestureHandler.forceTouchAvailable to conditionally render other components that would provide a fallback behavior specific to your usecase ');
```

ios/Handlers/RNForceTouchHandler.m
```diff
@@ -0,0 +1,144 @@
...
+  }
+  [super touchesEnded:touches withEvent:event];
+  if (self.state == UIGestureRecognizerStateBegan || self.state == UIGestureRecognizerStateChanged) {
+    self.state = UIGestureRecognizerStateEnded;
+  } else {
+    self.state = UIGestureRecognizerStateFailed;
+  }
+}
+
+- (void) handleForceWithTouches:(NSSet<UITouch *> *)touches {
```
kmagiera: no space after `(void)`

ios/Handlers/RNForceTouchHandler.m
```diff
@@ -0,0 +1,144 @@
...
+  _force = _firstTouch.force / _firstTouch.maximumPossibleForce;
+}
+
+- (void) reset {
```
kmagiera: same

ios/Handlers/RNForceTouchHandler.m
kmagiera:
```suggestion
- (void)handleForceWithTouches:(NSSet<UITouch *> *)touches {
```

ios/Handlers/RNForceTouchHandler.m
kmagiera:
```suggestion
- (void)reset {
```

ios/RNGestureHandlerEvents.m
```diff
@@ -44,6 +44,24 @@ + (RNGestureHandlerEventExtraData *)forPan:(CGPoint)position
                            @"numberOfPointers": @(numberOfTouches)}];
 }
 
++ (RNGestureHandlerEventExtraData *)forForce:(CGFloat)force
+                                 forPosition:(CGPoint)position
+                        withAbsolutePosition:(CGPoint)absolutePosition
+                         withNumberOfTouches:(NSUInteger)numberOfTouches
+{
+  {
```
kmagiera: why these braces?

osdnk: not sure, works without them as well

PR #381, opened 2018-12-11
Upstream various changes made in the react-navigation fork of DrawerLayout

brentvatne:
- Support for RTL
- Support for `drawerLockMode`
- Add onGestureRef prop - this one I'm not so sure about -- in the React.createRef world, what is the right way to use a ref in two places?

### DIFF
--- a/DrawerLayout.js
+++ b/DrawerLayout.js
@@ -9,11 +9,22 @@
 // that could be found when using the drawer component
 
 import React, { Component } from 'react';
-import { Animated, StyleSheet, View, Keyboard, StatusBar } from 'react-native';
 import invariant from 'invariant';
+import {
+  Animated,
+  StyleSheet,
+  View,
+  Keyboard,
+  StatusBar,
+  I18nManager,
+} from 'react-native';
 import { AnimatedEvent } from 'react-native/Libraries/Animated/src/AnimatedEvent';
 
-import { PanGestureHandler, TapGestureHandler, State } from './GestureHandler';
+import {
+  PanGestureHandler,
+  TapGestureHandler,
+  State,
+} from 'react-native-gesture-handler';
 
 const DRAG_TOSS = 0.05;
 
@@ -25,6 +36,7 @@ export type PropType = {
   children: any,
   drawerBackgroundColor?: string,
   drawerPosition: 'left' | 'right',
+  drawerLockMode?: 'unlocked' | 'locked-closed' | 'locked-open',
   drawerWidth: number,
   keyboardDismissMode?: 'none' | 'on-drag',
   onDrawerClose?: Function,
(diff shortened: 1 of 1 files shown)


DrawerLayout.js
```diff
@@ -463,11 +494,15 @@ export default class DrawerLayout extends Component<PropType, StateType> {
 
     return (
       <PanGestureHandler
-        ref={this._panGestureHandler}
+        ref={this._setPanGestureRef}
         hitSlop={hitSlop}
         minOffsetX={gestureOrientation * minSwipeDistance}
+        maxDeltaY={15}
```
brentvatne: this is also potentially controversial and perhaps should be a prop. the reason for adding it is that we need some criteria for drawer to fail at if we want to wait for drawer to fail before starting another gesture

osdnk: Hi, @brentvatne 
Could you please use https://kmagiera.github.io/react-native-gesture-handler/docs/handler-pan.html#maxdeltax instead of this old API?

brentvatne: do you mean `failOffsetY`? if so, done. also changed `minOffsetX` to `activeOffsetX`

osdnk: yes, ofc

brentvatne:
1. Yeah I'm not sure what the recommended pattern is to support all types of refs right now (string, function, React.createRef). I called it `onGestureRef` to be explicit that it needs to be a function (`onX` is a convention for function callbacks, I also typed it as such). If you want to use a ref created through React.createRef then you would need to provide a callback that sets the current property on that. I don't have time at the moment to investigate how to handle all possible ref types, but I need support for something like this in react-navigation and the perfect API doesn't matter so much to me because it's an implementation detail in my case. I understand if you'd rather not land this particular change if that's not something you want to add to RNGH, I can remove if so.
2. Will do
3. Will do
4. I think nobody has actually tried this API with RTL layout outside of react-navigation before, or they would see that it doesn't work: https://github.com/react-navigation/react-navigation/issues/5279. The behavior of this is the same as DrawerLayoutAndroid default but I can add some docs if you still think that's important. I can add a note in docs though.

Is it possible to push docs changes back into a follow-up PR? I believe @kmagiera wanted to do a release tomorrow and I won't have time to do them before then.

osdnk:
Ok, so I leave decision for @kmagiera.

kmagiera:
@brentvatne Regarding RTL: From react-navigation perspective do you allow users to set drawer props directly (do people are allowed to set `drawerPosition` manually or you have an abstraction on top of that). I wonder, because it needs to be confusing that the available positions are `left` and `right` and so even after setting it to `left` it may appear on the right. If you have some other prop in navigation that hides this detail then the best thing to do would be to use I18n module there and pass proper value of `drawerPosition`. Regardless, I think there is one more place in the code where we check `drawerPosition` prop and if we want I18n support we likely need to update that place too

kmagiera:
I see, makes sense. I'm going to add some comment that will allow me to remember that in the future :)

PR #383, opened 2018-12-12
Add e2e basic tests and travis config

osdnk:
## Motivation
In this state test provided by me does not really check gesture-handler related features. However I see that there's need even for checking if example application is actually building.

## Features
I made some test with mocha and detox and they check some minor things (don't really test gesture handling). I've added building scripts on travis for iOS and Android. I also added running tests on travis on iOS. It was impossible to do it on Android because travis doesn't support HAXM (neither circleCI)


kmagiera:
LGTM 👍

kmagiera:
jk

osdnk:
So here's the point when iOS story is finished. Let's start a new chapter of the adventure

osdnk:
So here's android done. Now merge them

Example/android/build.gradle
```diff
@@ -3,7 +3,7 @@
 buildscript {
     ext {
         buildToolsVersion = "27.0.3"
-        minSdkVersion = 16
```
osdnk: It's just an example, but required by travis

Example/multitap/index.js
```diff
@@ -11,38 +11,55 @@ import {
...
+              ref={this.doubleTapRef}
+              onHandlerStateChange={this._onDoubleTap}
+              numberOfTaps={2}>
+              <View style={styles.box} testID="rectangle" />
+            </TapGestureHandler>
           </TapGestureHandler>
-        </TapGestureHandler>
-      </LongPressGestureHandler>
+        </LongPressGestureHandler>
+        {this.state.lastGH && (
```
osdnk: I didn't manage to catch modal text

osdnk:
Revert example changes

Example/android/gradlew.bat
```diff
@@ -1,90 +1,90 @@
-@if "%DEBUG%" == "" @echo off
```
osdnk: It's a matter of linebreaking. It's not breaking

PR #385, opened 2018-12-14
Handle case of null-parent

osdnk:
## Motivation
This PR is a solution for rarely observed problem of parent-null issue on Android devices. If occurs when we want to extract position od handler when one of its parent has been detached. 

## Why it helps?
I believe that this issue might be related to some race-condition on Android devices when it comes to rendering some complicated structures. I could believe that event sending from `componentWillUnmount` might update GH after gesture event appear in some edge cases.

I cannot prove that it's actually a problem and it's actually solution, but it's not breaking and if it could make our users' apps more reliable, I'd like to have it merged.


android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
```diff
@@ -263,18 +261,7 @@ public void deliverEventToGestureHandlers(MotionEvent event) {
     }
   }
 
-  public GestureHandler getLastActivatedHandler() {
```
osdnk: not used

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
```diff
@@ -249,9 +249,7 @@ public void deliverEventToGestureHandlers(MotionEvent event) {
     // Copy handlers to "prepared handlers" array, because the list of active handlers can change
     // as a result of state updates
     int handlersCount = mGestureHandlersCount;
-    for (int i = 0; i < handlersCount; i++) {
```
osdnk: it's faster

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
```diff
@@ -298,7 +285,16 @@ private void deliverEventToGestureHandler(GestureHandler handler, MotionEvent ev
       return;
     }
     float[] coords = sTempCoords;
-    extractCoordsForView(handler.getView(), event, coords);
+    if (!extractCoordsForView(handler.getView(), event, coords)) {
+      if (!isFinished(handler.getState())) {
+        if (handler.getState() == GestureHandler.STATE_ACTIVE) {
+          handler.end();
```
kmagiera: I don't think we ever want to `end` gesture in such a situation. Especially that we are not even able to reliably reproduce scenario in which this code gets triggered. I think the safest solution would be to follow the same code path that we normally execute when view is detached and call `cancel` on the handler -> https://github.com/kmagiera/react-native-gesture-handler/blob/master/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.java#L65

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
```diff
@@ -321,21 +317,32 @@ private void deliverEventToGestureHandler(GestureHandler handler, MotionEvent ev
...
-  private void extractCoordsForView(View view, MotionEvent event, float[] outputCoords) {
+  /**
+   * extractCoordsForView performs extraction of coords for event in rootView.
+   * In addition it checks whether all of parents for view related to handler are attached.
+   * There might be a rarely observed issue when view has been detached and handler's
+   * state hasn't been change to failed or ended yet. Probably it's a result of some race condition
+   * and stopping delivering for this handler and changing its state to failed of end appear
+   * to be good enough solution.
+   */
+  private boolean extractCoordsForView(View view, MotionEvent event, float[] outputCoords) {
```
kmagiera: I prefer that we don't change the signature of this method. It isn't very clear on when it should return what. We should either change its name if we want it to perform an additional test but I think the best would be to do the following:
1) leave this method untouched
2) add a new method that tests if view is attached under `mWrapperView` (e.g. `isViewAttachedUnderWrapper` which traverses the hierarchy up until it reaches wrapper)
3) use the above method directly in `deliverEventToGestureHandler` even before we test `handler.wantEvents`
4) if view is not attached we immediately call `cancel` on the corresponding handler to match the behavior from https://github.com/kmagiera/react-native-gesture-handler/blob/master/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.java#L65

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
```diff
@@ -321,6 +312,25 @@ private void deliverEventToGestureHandler(GestureHandler handler, MotionEvent ev
...
   }
 
+  /**
+   * isViewAttachedUnderWrapper checks whether all of parents for view related to handler
+   * view are attached. Since there might be an issue rarely observed when view
+   * has been detached and handler's state hasn't been change to canceled, failed or
+   * ended yet. Probably it's a result of some race condition and stopping delivering
+   * for this handler and changing its state to failed of end appear to be good enough solution.
+   */
+  private boolean isViewAttachedUnderWrapper(View view) {
```
kmagiera:
```suggestion
  private boolean isViewAttachedUnderWrapper(@Nullable View view) {
```

kmagiera: you'll also need to add an import for `@Nullable`

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
```diff
@@ -321,6 +312,25 @@ private void deliverEventToGestureHandler(GestureHandler handler, MotionEvent ev
...
+    if (view == mWrapperView) {
+      return true;
+    }
+
+    if (view == null || !(view.getParent() instanceof View)) {
+      return false;
+    }
+
+    return isViewAttachedUnderWrapper((View) view.getParent());
```
kmagiera: you changed simple loop to arraycopy few line earlier and here you do a recursion instead of a simple while loop?

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.java
```diff
@@ -321,6 +314,23 @@ private void deliverEventToGestureHandler(GestureHandler handler, MotionEvent ev
...
 
+  /**
+   * isViewAttachedUnderWrapper checks whether all of parents for view related to handler
+   * view are attached. Since there might be an issue rarely observed when view
+   * has been detached and handler's state hasn't been change to canceled, failed or
+   * ended yet. Probably it's a result of some race condition and stopping delivering
+   * for this handler and changing its state to failed of end appear to be good enough solution.
+   */
+  private boolean isViewAttachedUnderWrapper(@Nullable View view) {
+    while (view != mWrapperView) {
```
kmagiera: `getParent` method is a part of `ViewParent` interface, therefore we don't need casting or instanceof checks:

```
    @Nullable ViewParent parent = view;
    while (parent != null && parent != mWrapperView) {
        parent = parent.getParent();
    }
    return parent == mWrapperView;
```

osdnk: Ok, I'll do it, but what was do reason you've done it like this before?

osdnk: And can't `view` do not fill implement `ViewParent` interface?

PR #395, opened 2018-12-24
Bump RN version to 0.58.0-rc.2 🎄

osdnk:
🎄 Added JSC to linked libs on iOS
🎄 But for JSC, everything works like a charm
🎄 It might be necessary to drop node_modules, xcode's cache and npm's cache as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "lint-staged": "^8.0.0-beta.1",
     "prettier": "^1.13.7",
     "react": "^16.0.0",
-    "react-native": "^0.50.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
     "lint-staged": "^8.0.0-beta.1",
     "prettier": "^1.13.7",
     "react": "^16.0.0",
-    "react-native": "^0.50.1",
+    "react-native": "^0.58.0-rc.2",
```
kmagiera:
```suggestion
     "react-native": "^0.50.1",
```

PR #401, opened 2019-01-06
Only include the ios folder in the podspec

janicduplessis:
I migrated my app to use cocoapods instead of manually linking projects and had a build issue when depending on react-native-gesture-handler from github. What happens is when installing from github it also includes the Example project and the podspec sources pattern also includes the code from it. This causes a duplicate symbol error because the project will contain 2 AppDelegate classes.

To fix this I simply make the sources pattern only include the ios directory.


janicduplessis:
cc @osdnk

PR #405, opened 2019-01-08
Fix typo in GenericTouchable

angly-cat:
`undermined` -> `undetermined`.


osdnk:
Thanks, @angly-cat 
🛡

PR #406, opened 2019-01-09
fix: web compatibility

jaulz:
The motivation for me was to use `react-navigation@^3` and I stumbled across quite a few missing imports. This PR will introduce some shims so it shouldn't be a problem anymore.


osdnk:
Hi, @jaulz 
Thanks for this PR. It looks good. I dom't have string opinion about support of GH for web, but I believe it's nice to make in runnable at least.

1. Can't we do 
```javascript
const Handler = props => children.props
```
Maybe there's no need for something more than only a stub?

2. Then could you add some comments in `GestureHandler.web.js` motivating your decisions?

osdnk:
@jaulz 
Is it WIP? If yes, please add proper label?

jaulz:
@osdnk sorry, I found some issues with the original commit and pushed a fix now 😊As requested I added some comments and it should be in a state where you could please take a look. I tried to touch the existing files at a minimum and also cross checked on native devices if it still works as expected.

GestureHandler.web.js
```diff
@@ -0,0 +1,150 @@
...
+  DrawerLayoutAndroid,
+  WebView,
+  FlatList,
+  TouchableWithoutFeedback,
+  findNodeHandle,
+} from 'react-native';
+import gestureHandlerRootHOC from './gestureHandlerRootHOC';
+
+const State = {
+  UNDETERMINED: 'UNDETERMINED',
```
osdnk: Can we move it out somewhere?

jaulz: What should we move out in detail? State constants?

osdnk: Yes. If you need them here, I wish to have them in separated file in order not to duplicate code.

jaulz: Sure, makes more than sense 😊 Was not sure before since these constants were used in `GestureHandler.js` as well.

osdnk: Oh, well, we intentionally wanted to take them from native site every time. How I think there's no need to do it. You could think about removing extracting from from Java and Objective-C if it's not big deal for you.

jaulz: Hm, the constants are used in the native part as well. Hence it actually makes sense to read them in the JavaScript part from the native side. Should I undo the changes in `GestureHandler.js` for the time being?

AnimatedEvent.web.js
```diff
@@ -0,0 +1,3 @@
+import AnimatedEvent from 'react-native-web/dist/vendor/react-native/Animated/AnimatedEvent';
```
necolas: If `AnimatedEvent` needs to be used directly, you may want to ask that it is added to the `Animated` export from React Native rather than relying on the fragility of reaching into internals.

GestureHandler.web.js
```diff
@@ -0,0 +1,145 @@
+import React from 'react';
+import {
+  ScrollView,
+  Slider,
+  Switch,
+  TextInput,
+  ToolbarAndroid,
+  ViewPagerAndroid,
+  DrawerLayoutAndroid,
+  WebView,
```
necolas: This module is being removed from React Native and isn't included in `react-native-web` either

GestureHandler.web.js
```diff
@@ -0,0 +1,145 @@
...
+        // Attach handler to DOM node again
+        const { enabled } = this.props;
+        if (enabled !== false) {
+          attachNativeHandler(node, this.props);
+        }
+      }
+    }
+
+    setNativeProps() {
+      // No implementation so far but is needed to avoid null calls
```
necolas: You should be able to use `this.container.current.setNativeProps`

GestureHandler.web.js
```diff
@@ -0,0 +1,145 @@
...
+  const { onHandlerStateChange = () => {} } = props;
+  const clickHandler = ({ x }) =>
+    onHandlerStateChange({
+      nativeEvent: {
+        oldState: State.ACTIVE,
+        state: State.UNDETERMINED,
+        x,
+      },
+    });
+  node.addEventListener('click', clickHandler);
```
necolas: Using these DOM events directly means you're not hooked into the responder event system

GestureHandler.web.js
```diff
@@ -0,0 +1,145 @@
...
+const RotationGestureHandler = createHandler('RotationGestureHandler');
+
+// Factory for Button component
+// (at the moment this is a plain TouchableWithoutFeedback)
+function createButton(name) {
+  return class Button extends React.Component {
+    static displayName = name;
+
+    render() {
+      return <TouchableWithoutFeedback {...this.props} />;
```
necolas: If these are meant to be used as buttons in the UI, you should add `accessibilityRole="button"`

GestureHandler.web.js
osdnk: No, it's fine now

AnimatedEvent.web.js
jaulz: Okay, I will check on their repository and raise a PR.

GestureHandler.web.js
necolas: I'm not familiar with the behaviour/implementation of the native `TapGestureHandler`. But if it turns out to be preferable to use the responder events, you could pass the event handlers to `createHandler` and forward them to the component.

miracle2k:
Sorry for the interjection: 

I was trying this PR, but the thing I run into is that my babel transforms the access to `this` in the static property in this line (https://github.com/kmagiera/react-native-gesture-handler/blob/master/touchables/GenericTouchable.js#L56) into 

```GenericTouchable2.propTypes = _objectSpread({}, (void 0).internalPropTypes, (void 0).publicPropTypes);```

(i.e., it replaces this with undefined).

Clearly something is wrong with the way I have setup my babel for the `react-native-web` compile. Can @jaulz or someone else share a working babel config?

**Comment by necolas** [comment]
That just looks like incorrect code. `this` should be the class name instead

**Comment by miracle2k** [comment]
@necolas Funny. If I change those lines to 

```
static propTypes = {
    ... GenericTouchable.internalPropTypes,
    ... GenericTouchable.publicPropTypes,
  };
```

It compiles with webpack, but then `react-native@0.57.8` in the iOS emulator says:

```
Cannot read property 'internalPropTypes' of undefined
```

The exact opposite behaviour.

**Comment by miracle2k** [comment]
I am fairly certain this is related to the Hot Module Reloading transform inserted by metro [here](https://github.com/facebook/metro/blob/v0.48.5/packages/metro/src/reactNativeTransformer.js#L131) by merging into `babelrc` [this partial config](https://github.com/facebook/metro/blob/master/packages/metro-react-native-babel-preset/src/configs/hmr.js).

Transforming the file in question with `babel` and using only `metro-react-native-babel-preset`, the output is sensible: https://gist.github.com/miracle2k/3bc7f1c50080397dbf0d92cfb3101677#file-generictouchable-babel-with-metro-preset-js.

But the actual code generated by `react-native` for the iOS emulator looks like this: https://gist.github.com/miracle2k/3bc7f1c50080397dbf0d92cfb3101677#file-generictouchable-rn-compilation-js
Notice the `_wrapComponent` calls which come from the `react-transform` plugin: `https://github.com/facebook/metro/blob/master/packages/metro-babel7-plugin-react-transform/src/index.js#L420`.

The problem goes away if I manually edit the file linked above (`reactNativeTransformer.js#L131`) to ensure the transform is not added. Note that the current react-native ships with an old metro version, `v0.48.5`, so the problem may or may not be fixed in a newer one.

I can see that @osdnk added this code quite recently. Maybe as a workaround for this issue we could rewrite that part and pull the re-used proptypes into the global module scope.

**Comment by osdnk** [comment]
I will merge it if needed.

File: GestureHandler.web.js (hunk 6 above)
Comment (jaulz) [reply GestureHandler.web.js]: Thanks for the hint!

File: GestureHandler.web.js (hunk 3 above)
Comment (jaulz) [reply GestureHandler.web.js]: Which one in detail? Currently, I see that all of these components are exported though some of them as `UnimplementedView` which is fine I guess.

Comment (SimenB) [reply GestureHandler.web.js]: https://github.com/react-native-community/discussions-and-proposals/pull/3

Comment (jaulz) [reply GestureHandler.web.js]: Thanks for the info!

File: GestureHandler.web.js (hunk 4 above)
Comment (jaulz) [reply GestureHandler.web.js]: I intentionally left this out since I think we do not need to pass on props for stub components. This would just generate warnings since most of the props are custom.

**Comment by jaulz** [comment]
@osdnk @brentvatne it would be great if could you have another look. I assume we could even implement a proper `PanGestureHandler` with the help of `PanResponder` from `react-native-web` but this might be something for the future.

File: DrawerLayout.js [hunk 7]
```diff
@@ -507,7 +507,7 @@ export default class DrawerLayout extends Component<PropType, StateType> {
 }
 
 const styles = StyleSheet.create({
-  drawer: { flex: 0 },
+  drawer: {},
```
osdnk: why?

jaulz: With `flex: 0` the `Drawer` will not be visible on web and though I removed it it was working the same on native. Did you notice any difference or why did you choose `flex: 0`?

GestureHandler.web.js
```diff
@@ -0,0 +1,132 @@
...
+import gestureHandlerRootHOC from './gestureHandlerRootHOC';
+import Directions from './Directions';
+import State from './State';
+
+// Factory for stub Handler components
+function createStubHandler(name) {
+  return class NativeViewGestureHandler extends React.Component {
+    static displayName = name;
+
+    container = React.createRef();
```
osdnk: if it used anywhere?

GestureHandler.web.js
```diff
@@ -0,0 +1,132 @@
...
+      // If for any reason we need to pass on native props we can fallback to this:
+      // this.container.current.setNativeProps(...args)
+    }
+
+    render() {
+      const { children, ...rest } = this.props;
+
+      // We don't want to create another layer, so instead we clone it only but keep the reference
+      const child = React.Children.only(children);
+      return React.cloneElement(child, {
```
osdnk: https://reactjs.org/docs/forwarding-refs.html

Is it actually needed to clone component for obtaining ref?

GestureHandler.web.js
```diff
@@ -0,0 +1,132 @@
...
+class TapGestureHandler extends React.Component {
+  setNativeProps() {}
+
+  render() {
+    const { children, enabled, onHandlerStateChange, style } = this.props;
+
+    return (
+      <TouchableWithoutFeedback
+        style={style}
+        onPress={({ x }) => {
```
osdnk: how about y?

osdnk: Hm, can't we make it more complex? I believe it's fully possible to achieve more similar behaviour to native `TapGestureHandler` by by combing `onPressIt`, `onPressOut` and `pressRetentionOffset`. Is it possible on web?

jaulz: Yep, is included now.

GestureHandler.web.js
jaulz: This was something left over from the initial version, sorry.

GestureHandler.web.js
jaulz: Yeah, I think in the future we should try to respect the gesture props even more. `pressRetentionOffset` means something else if I understand correctly and cannot be mapped to any of the props. As a start I implemented support for `maxDurationMs` prop but I think additional props should be part of future PRs. What do you think?

GestureHandler.web.js
jaulz: This was something left over from the initial version, sorry.

AnimatedEvent.web.js
EvanBacon: `AnimatedEvent` is just used as a type. This shouldn't prohibit any functionality.

alidcastano:
@osdnk @jaulz anything remaining in this PR?

jaulz:
@alidcastano I am waiting for @osdnk feedback on this but I would love to see this being merged since it would solve a big pain point for a lot of `react-native-web` users right now. Once we have it we should consider the actual implementations of the gesture handlers.

GestureHandler.web.js
osdnk: how about:
```
render() {
	return this.children
}
```

GestureHandler.web.js
```diff
@@ -0,0 +1,135 @@
...
+        },
+      });
+    }
+  };
+
+  render() {
+    const { children, maxDurationMs, style } = this.props;
+
+    return (
+      <TouchableWithoutFeedback
```
osdnk: And can't we use `onPressIn` and `onPressIn` for achieving more native-like states' flow?

osdnk:
Also, please provide some guide how to test it. Maybe try to extend our `Example` app with web support and add some script for simple start?

jaulz:
Sure, I will tackle the example next week. For the time being I enhanced the `TapGestureHandler` to respect most of the props except for `minPointers`. Actually the `touches` attribute of the event (https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/touches) should return the number of pointers but for some reason it does return an empty array both on desktop and on mobile devices and at maximum an array with a length of 1 if pressing takes longer than a few milliseconds.

osdnk:
Looks so good!
I think that in the future we're going to split handlers into separated files and then we could simple do `TapHandler.web.js`. But now, it's perfect as is! 😍

jaulz:
@osdnk finally I could make the example work... you can run it via `yarn run web`. Since the last commit I did a major rewrite because I thought rewriting the whole `GestureHandler.js` would not make sense and now I just outsourced the parts that needed a different implementation on web and native. I also removed the `*.web` suffix so we can offer the same for any other non iOS or Android platform.

GestureHandlerModule.js
```diff
@@ -0,0 +1,191 @@
...
+const handlers = {
+  NativeViewGestureHandler: class NativeViewGestureHandler extends BaseGestureHandler {
+    render() {
+      const { children } = this.props;
+
+      return children;
+    }
+  },
+
+  TapGestureHandler: class TapGestureHandler extends BaseGestureHandler {
```
osdnk: I'm not really into using inheritance in that way. I think it's not a common practice to use it when in comes to react component. It's not very readable then and it's not very popular for these cases. Would you mind rather using a composition here in similar way we're doing it in `GestureHandler.js`?

jaulz: Just changed it 😊

jaulz:
I'm not 100% sure about the suffixes but e.g., the use of `createRef` in DrawerLayout already limits us to `react@^16.3` so I assume it already limits us to newer React Native versions. Is there any mapping between React Native and React versions?

Anyway I renamed the files and we could always change it again 😊

EvanBacon:
@osdnk 
>  I mean that I am not convinced if using .native suffix for native platforms and no suffix for web will always fallback properly for old version of RN.

The `.native` suffix has been around since at least April 2018.

jaulz:
@EvanBacon yep, it's like this now

GestureHandler.js
```diff
@@ -15,46 +13,19 @@ import {
...
-    registrationName: 'onGestureHandlerStateChange',
-  },
-};
-
-const State = RNGestureHandlerModule.State;
-
-const Directions = RNGestureHandlerModule.Direction;
+import Directions from './Directions';
+import State from './State';
+import PlatformConstants from './PlatformConstants.web';
```
Jekiwijaya: it is intendeed to import `PlatformConstants.web` ? Same with the `GestureHandlerButton` and `GestureHandlerModule`

jaulz: Oops, good catch. Seems that VS code automatically changed the imports when I renamed the files. Will change it on Monday latest.

GestureHandler.js
```diff
@@ -15,46 +13,19 @@ import {
...
   Platform,
+  processColor,
+  Touchable,
 } from 'react-native';
-import processColor from 'react-native/Libraries/StyleSheet/processColor';
-import Touchable from 'react-native/Libraries/Components/Touchable/Touchable';
 import deepEqual from 'fbjs/lib/areEqual';
 import PropTypes from 'prop-types';
 
+import GestureHandlerButton from './GestureHandlerButton.web';
```
Jekiwijaya:
```suggestion
import GestureHandlerButton from './GestureHandlerButton';
```

GestureHandler.js
```diff
@@ -15,46 +13,19 @@ import {
...
+import GestureHandlerModule from './GestureHandlerModule.web';
```
Jekiwijaya:
```suggestion
import GestureHandlerModule from './GestureHandlerModule';
```

GestureHandler.js
Jekiwijaya:
```suggestion
import PlatformConstants from './PlatformConstants';
```

Jekiwijaya:
@jaulz the android example currently is crashed. Here is the fix, https://github.com/jaulz/react-native-gesture-handler/pull/1

Jekiwijaya:
Much appreciate if this could go in, my projects using react-native-web and react-navigation, which this pull request could make my projects upgrade from react-navigation v2 to react-navigation v3 in web platform.

Anyway, thanks for great library! 👍 👍 👍

Example/android/build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
     ext {
         buildToolsVersion = "27.0.3"
         minSdkVersion = 16
-        compileSdkVersion = 27
+        compileSdkVersion = 28
```
osdnk: I don't really know why it is required. In other PRs everything is passing without this change 🤔

Jekiwijaya: 1. It's because the version react-native of Example is changes, from `0.58.0-rc.2` to `"0.58.0-rc.2` https://github.com/kmagiera/react-native-gesture-handler/pull/406/files#diff-77190fabae122eec07dd4211d79a25f7R20
2. And because of that, the Example taking latest react-native version, which is `0.58.3` the latest release (6 days ago from now).
3. In 0.58.0, there's already an update from android support 0.27 to 0.28 https://github.com/facebook/react-native/commit/fe6f6cd46f077b96733f67b6e3d6769cdfb5ee21#diff-4a7ba4beeeab016621279eed94a6270bL315
4. Because the android support 0.28 is being used, we need to change the compileSdkVersion to 28.

osdnk: ok, thanks!

zamponotiropita:
I think `package.json` has to be adapted to include the new files (`Directions.js`, `State.js`, ...), currently they aren't fetched by npm

jaulz:
@zamponotiropita thanks for the hint

GestureHandler.js
```diff
@@ -673,7 +643,7 @@ function createNativeWrapper(Component, config = {}) {
             typeof source[methodName] === 'function' &&
             this[methodName] === undefined
           ) {
-            this[methodName] = source[methodName].bind(node);
```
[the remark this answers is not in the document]
jaulz: Yeah, I saw the same warning. I just reset my changes to the original state so we can merge your fix on top.

zamponotiropita:
@jaulz Still not packaging correctly on my side, for some reason `GestureHanlderModule*.js` are ommited

Tested with following dependency in package.json: "react-native-gesture-handler": "jaulz/react-native-gesture-handler#fix/web-compatibility",

jaulz:
@zamponotiropita can you check if the `package.json` is up to date? Maybe it's still cached.

zamponotiropita:
@jaulz Tested after doing `rm -r node_modules package-lock.json && npm i`

jaulz:
@zamponotiropita please try again now.

GestureHandler.js
```diff
@@ -15,46 +13,19 @@ import {
...
-  onGestureHandlerEvent: { registrationName: 'onGestureHandlerEvent' },
-  onGestureHandlerStateChange: {
-    registrationName: 'onGestureHandlerStateChange',
-  },
-};
-
-const State = RNGestureHandlerModule.State;
-
-const Directions = RNGestureHandlerModule.Direction;
+import Directions from './Directions';
```
kmagiera: We used to take directions list from `RNGestureHandlerModule` can we have a separate implementation for `Directions.web.js` to keep it the same way on mobile platforms?

GestureHandler.js
```diff
@@ -352,7 +322,7 @@ function createHandler(
       }
 
       const child = React.Children.only(this.props.children);
-      let children = child.props.children;
+      let children = GestureHandlerModule.getChildren(this.props);
```
kmagiera: This line seem to be changing behavior on mobile platforms. It returns a single child on mobile as far as I can see. Have you tested this on mobile with overlay debugger ON?

GestureHandlerModule.web.js
```diff
@@ -0,0 +1,187 @@
...
+  handleHandlerStateChange(props, event, State.BEGAN, State.ACTIVE);
+}
+
+function handleBegan(props, event) {
+  handleHandlerStateChange(props, event, State.UNDETERMINED, State.BEGAN);
+}
+
+class UnimplementedGestureHandler extends React.Component {
+  render() {
+    return this.props.children;
```
kmagiera: do we want to add a warning if an unsupported GH is being rendered on web?

Example/common.js
```diff
@@ -35,3 +33,13 @@ Praesent ac risus nec augue commodo semper eu eget quam. Donec aliquam sodales c
 
 Fusce interdum nisi a risus aliquet, non dictum metus cursus. Praesent imperdiet sapien orci, quis sodales metus aliquet id. Aliquam convallis pharetra erat. Fusce gravida diam ut tellus elementum sodales. Fusce varius congue neque, quis laoreet sapien blandit vestibulum. Donec congue libero sapien, nec varius risus viverra ut. Quisque eu maximus magna. Phasellus tortor nisi, tincidunt vitae dignissim nec, interdum vel mi. Ut accumsan urna finibus posuere mattis.
 `;
+
+export const Alert = {
```
kmagiera: We only use `Alert` to notify that some action has been triggered. I checked around our example app codebase and we never use non-standard options like buttons and alert types. I'd recommend that instead of adding this to commons we just use `alert` method that is globally available in RN on web and mobile

Swipeable.web.js
```diff
@@ -0,0 +1,4 @@
+// At the moment no swipe is supported
+export default function Swipeable({ children }) {
+  return children;
```
kmagiera: add `console.warn` here please

DrawerLayout.js
kmagiera: Let's remove `styles.drawer` than altogether, what's the point of having an empty style?

GestureHandler.js
```diff
@@ -368,8 +338,9 @@ function createHandler(
           })
         );
       }
+
       return React.cloneElement(
-        child,
+        GestureHandlerModule.render(handlerName, this.props),
```
kmagiera: on web we seem to have handler specific components that we render here. So in my option it does not make sense to render and immediately clone it. So maybe lets only use clone on mobile where we want to clone the underlying child component?

Another aspect of that is it appears to me like we inadvertently try to reuse logic of the component returned from `createHandler` while it seem the whole logic there is only so that we call `GestureHandlerModule` methods which are no-ops on web. So in my opinion we should diverge this implementation at an earlier stage.

GestureHandlerModule.web.js
jaulz: Yep, it's implemented now.

Example/common.js
jaulz: Sure, but I still kept it in `common.js` in case we want to differentiate the behavior in the future.

Swipeable.web.js
jaulz: Yep, it's implemented now.

GestureHandler.js
jaulz: I just restructured it once again and introduced a `GestureHandlerRenderer.js` which should improve this. e.g., on web we do not need to render the debug views explicitly.

GestureHandler.js
jaulz: Yep, it's implemented now.

jaulz:
@kmagiera thanks for your feedback! I just pushed some changes which should improve the overall structure. Could you please have another look?

zamponotiropita:
@jaulz super funkt danke

DrawerLayout.js
jaulz: Removed in https://github.com/kmagiera/react-native-gesture-handler/pull/406/commits/23af89f77c0620cf3b20372d8df0636172793f7a

Example/common.js
```diff
@@ -35,3 +33,7 @@ Praesent ac risus nec augue commodo semper eu eget quam. Donec aliquam sodales c
 
 Fusce interdum nisi a risus aliquet, non dictum metus cursus. Praesent imperdiet sapien orci, quis sodales metus aliquet id. Aliquam convallis pharetra erat. Fusce gravida diam ut tellus elementum sodales. Fusce varius congue neque, quis laoreet sapien blandit vestibulum. Donec congue libero sapien, nec varius risus viverra ut. Quisque eu maximus magna. Phasellus tortor nisi, tincidunt vitae dignissim nec, interdum vel mi. Ut accumsan urna finibus posuere mattis.
 `;
+
+export const alert = message => {
+  window.alert(message);
```
Jekiwijaya: Is the native apps use window.alert too?

I think it's better just to use global `alert`.

jaulz: Okay, it's in now.

kmagiera:
Thanks @jaulz for responding to my comments – this looks way better now!

I might've not been clear enough with one of my suggestion earlier and would like to run this idea through you once more. So when it comes to the code split between web and mobile I was more thinking if we could have separate implementations of `createHandler` method ([line 166](https://github.com/kmagiera/react-native-gesture-handler/pull/406/files#diff-56d925ec2aada9e1b3b613fe5b782837L166)). In my opinion this way we could avoid introducing unnecessary abstractions for "renderer" and "mobule" and have web version use your implementation while mobile version use native modules. We would also avoid having so many web-related check in `createHandler`.

I am not completely sure if that suggestion wouldn't complicate some other things, so would appreciate if you could share what you think about it before you invest your time in making the change

jaulz:
@kmagiera I understand what your point is but the reason why I chose this way is because we can reuse all the things are common (e.g. prop types, resolving refs etc.) without a major rewrite of the `GestureHandler`. Also, going forward the web related checks might not only be applicable for web only but also for other platforms - maybe other platforms require `attachGestureHandler` as well? In that case the current set up will help us to have consistent `GestureHandlers` across multiple platforms.

kmagiera:
I see. However in my opinion we don't need this whole ref hackery if handlers are mapped to the actual components. So I am not sure if there is any value in reusing these.

What are "web related checks" you mentioned? To be honest I am more concerned about someone accidentally breaking web compatibility because it may not be clear for them in what circumstances module methods are not present.

jaulz:
To be honest I'm not 100% sure if the ref hacks are needed on web (e.g., are there cases when `ref.current` is null?) but we might need the ref hacks for other platforms and their native modules in case there is demand in the future  (e.g. the ones listed here: https://facebook.github.io/react-native/docs/out-of-tree-platforms). 

With "web related checks" I meant the checks that are not handled via different file name suffixes which are these at the moment:
https://github.com/kmagiera/react-native-gesture-handler/pull/406/files#diff-56d925ec2aada9e1b3b613fe5b782837R206

https://github.com/kmagiera/react-native-gesture-handler/pull/406/files#diff-56d925ec2aada9e1b3b613fe5b782837R220

https://github.com/kmagiera/react-native-gesture-handler/pull/406/files#diff-56d925ec2aada9e1b3b613fe5b782837R230

https://github.com/kmagiera/react-native-gesture-handler/pull/406/files#diff-56d925ec2aada9e1b3b613fe5b782837R238

Since I do not have a preferred option and am open for both: shall I come up with another commit with `createHandler.js` and `createHandler.web.js` and incorporate the current `GestureHandlerModule` and `GestureHandlerRenderer` in those?

kmagiera:
I think from a maintainer perspective it would be easier for us to accept it if we have fewer "diversion points". So if you think this wouldn't take too much time I'd be glad if you could change it the way you proposed with `createHandler.js` and `createHandler.web.js`

axhamre:
May I ask what's the status here? 

Web supprt would be such a milestone.

jaulz:
@kmagiera could you please check the latest version?

iamnader:
excited for this.  thanks for all the work here folks

createHandler.web.js
```diff
@@ -0,0 +1,203 @@
...
+    _refHandler = node => {
+      this._viewNode = node;
+    };
+
+    setNativeProps = (...args) => {
+      this._viewNode.setNativeProps(...args);
+    };
+
+    render() {
+      let Handler = handlers[handlerName];
```
kmagiera: nit: I think this check could be moved outside of the render method and directly to `createHandler`

jaulz:
@kmagiera do you have any timeline for a new release? Thanks for merging by the way 😊

kmagiera:
will try to send out new release tomorrow

alidcastano:
just tried using release with this PR for web and ran into errors. what does we need to configure to get it running - looks like `index.web` needs to be resolved over `index`, anything else? 

it'd be great we we could just have a simple alias (like we do react-native to react-native-web)

jaulz:
@alidcastano if you add the `.web.js` suffix to your Webpack configuration (in `resolve.extensions`) it should work immediately.

createHandler.js
```diff
@@ -0,0 +1,328 @@
...
+      RNGestureHandlerModule.attachGestureHandler(this._handlerTag, newViewTag);
+    };
+
+    _updateGestureHandler = newConfig => {
+      this._config = newConfig;
+
+      RNGestureHandlerModule.updateGestureHandler(this._handlerTag, newConfig);
+    };
+
+    _dropGestureHandler = () => {
```
osdnk: Why is it not used?

jaulz: What do you mean?

osdnk: `_dropGestureHandler` is not used anywhere. Isn't it a bug?

jaulz: Yep, good catch! Sorry..

osdnk: np, thanks!

zamponotiropita:
After 3 months off keyboard and a new laptop I stumbled across the following bugs in both chrome and safari (the second one is major):
* I have to set `cardStyle: { flex: 1 }` in the options of `createStackNavigator`, otherwise the cards (individual screens of the stack navigator) will have height `0`.
* When pushing a new screen into the navigation stack, the previous screen will become blank and occupy half of the top of the screen instead of disappearing completely. The bottom half of the screen will contain the pushed screen.

Testing on android I don't have those issues.

I don't know whether these issues are related to react-native-gesture-handler, react-navigation or react-native-web. I would appreciate any hints.

I used the most recent versions for all packages (react-native v0.59.9, react-navigation v3.0.9 and react-native-web v0.11.4)

zamponotiropita:
I opened an issue for encountered bugs here:
https://github.com/kmagiera/react-native-gesture-handler/issues/632

Manishalexin:
When trying to import from RNGH in react native web project, it throws with this error,
"Support for the experimental syntax 'exportDefaultFrom' isn't currently enabled", adding babel plugins for these doesn't help.

Minimum reproducible demo,
https://codesandbox.io/s/polished-paper-tdmyp

brentvatne:
@Manishalexin - you should use babel-preset-expo and the react-native-web setup that you get when you init a project with `expo-cli` - `npm i -g expo-cli && expo init`

Manishalexin:
Thank you @brentvatne it's working with babel-preset-expo in babel config and I have to use `expo start`. React scripts & react-scripts-rewired still fails. So, now the storybook fails as it still uses react scripts. I used the setup here, `https://github.com/expo/examples/tree/master/with-storybook` but it doesn't work as I believe it's only for the expo projects. Any suggestions on how to make it work?

I am using a bare workflow, with react native cli.

jpaas:
Also running into this on a `create-react-app` project using RNGH 1.6.1
```
SyntaxError: .../node_modules/react-native-gesture-handler/DrawerLayout.js: Support for the experimental syntax 'flow' isn't currently enabled (30:8):

  28 | const SETTLING = 'Settling';
  29 |
> 30 | export type PropType = {
     |        ^
  31 |   children: any,
  32 |   drawerBackgroundColor?: string,
  33 |   drawerPosition: 'left' | 'right',
```

jpaas:
Why does that js file even contain typescript?

kopax-polyconseil:
Hi, I am trying to add react-native-web support on an existing react native ios android application, 

I still have error such as this one : 

```
 FAIL  src/features/search/pages/__tests__/SearchFilter.test.tsx
  ● Test suite failed to run

    TypeError: (0 , _reactNative.requireNativeComponent) is not a function

      1 | import { t } from '@lingui/macro'
      2 | import React, { useState } from 'react'
    > 3 | import { TouchableOpacity } from 'react-native-gesture-handler'
        | ^
      4 | import styled from 'styled-components/native'
      5 | 
      6 | import { DateFilter } from 'features/search/atoms/Buttons'

      at Object.<anonymous> (node_modules/react-native-gesture-handler/GestureHandlerButton.js:3:32)
      at Object.<anonymous> (node_modules/react-native-gesture-handler/GestureButtons.js:6:1)
      at Object.<anonymous> (node_modules/react-native-gesture-handler/GestureHandler.js:10:1)
      at Object.<anonymous> (node_modules/react-native-gesture-handler/Swipeable.js:10:1)
      at Object.<anonymous> (node_modules/react-native-gesture-handler/index.js:1:1)
      at Object.<anonymous> (src/features/search/sections/OfferDate.tsx:3:1)
      at Object.<anonymous> (src/features/search/sections/index.ts:8:1)
      at Object.<anonymous> (src/features/search/pages/SearchFilter.tsx:10:1)
      at Object.<anonymous> (src/features/search/pages/__tests__/SearchFilter.test.tsx:12:1)
```

Looking at your discussion, I assume it should work out of the box with react-native-gesture-handler v1.6.0, why am I still getting those requireNative? Thanks a lot for this awsome lib!

PR #409, opened 2019-01-10
Exclude the "Example" folder in the podspec

matt-oakes:
This needs to be done because the `s.source_files  = "ios/**/*.{h,m}"` was also picking up files inside `Example/ios/*`. Inside this folder, there is a file called `AppDelegate` which was causing duplicate symbol errors when using this library through cocoapods.

This is fixed by explicitly excluding the entire `Example` folder.

### DIFF
--- a/RNGestureHandler.podspec
+++ b/RNGestureHandler.podspec
@@ -4,15 +4,16 @@ Pod::Spec.new do |s|
   # NPM package specification
   package = JSON.parse(File.read(File.join(File.dirname(__FILE__), "package.json")))
 
-  s.name         = "RNGestureHandler"
-  s.version      = package["version"]
-  s.summary      = package["description"]
-  s.homepage     = "https://github.com/kmagiera/react-native-gesture-handler"
-  s.license      = "MIT"
-  s.author       = { package["author"]["name"] => package["author"]["email"] }
-  s.platform     = :ios, "7.0"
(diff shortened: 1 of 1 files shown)


matt-oakes:
Thanks for the review @osdnk! 

The `=` signs were previously lined up so I lined them up again once the length changed. It looked a bit messy without the additional spaces in the other lines.

The Example folder was the only one which caused any issues.

osdnk:
Hm, do you know why? Because when I was doing e2e tests I just copied Example app and they're configured in the same way. Di you use master branch build for checking?

matt-oakes:
I realised what was causing this issue.

I was including the podspec file using:

`pod 'RNGestureHandler', :podspec => '../node_modules/react-native-gesture-handler'`

rather than:

`pod 'RNGestureHandler', :path => '../node_modules/react-native-gesture-handler'`

Using `podspec` rather than `path` causes Cocoapods to checkout the project from git, which means the files that are in the `.npmignore` file are actually included, which is then causing this issue.

I should have been using `path`, so really this isn't an issue which needs fixing, however, you might want to do it anyway just to make sure it works for people who mess up like I did.

You should also ensure that `e2e` is included in the `.nomignore` file to avoid that causing any issues.

I'll close this and leave it up to you if you want to make the changes or not.

Thanks for the help!

PR #423, opened 2019-01-24
[Android][Regression] Fix gestures not being delivered to RN

### LINKED ISSUE #422: [1.0.14] App stucked
Hi,
I don't have a lot of informations, but updating from 1.0.12 to 1.0.14 made my app unresponsive to touch, only the drawer works.

Just to inform.

angly-cat:
Version 1.0.14 introduced a regression: on Android gestures made outside of RNGH elements can't make it to RN elements.

The regression comes from https://github.com/kmagiera/react-native-gesture-handler/commit/3def2e1d72f8556dd368cb85570f9a4ab00b0fec.

Apparently, `handler.cancel()` line is to blame. My fix works, but I'm not sure if it's correct. Is it ok to just `return;` here without canceling a handler, just like in similar `if`s below?

Fixes #422.


kmagiera:
Thanks @angly-cat for working on this. Apparently this is not the right fix as it completely invalidates the changes introduced in 3def2e1d72f8556dd368cb85570f9a4ab00b0fec

I have a proper fix for this and will land it very soon.

PR #427, opened 2019-01-25
Fix problem with cancelling RootViewHandler

### LINKED ISSUE #422: [1.0.14] App stucked

kmagiera:
PR #385 introduced a regression on Android that'd prevent regular RN touch events from propagating. That PR was ought to address crashes in which we were still passing touch events to handlers even so the corresponding views has been detached. For this purpose `isViewAttachedUnderWrapper` has been added that was meant to detect if a given view is attached under root view wrapper. This method however did not work if the view provided as an argument was the root view. As a result handlers (root view handler) attached to the wrapper view were getting immediately cancelled.

Fixes #422


kmagiera:
cc @angly-cat

kmagiera:
Thanks for testing. I'd want to send one more unrelated patch and then I will push out an update to npm

scarlac:
@kmagiera thank you for a quick turnaround on this. I was scratching my head quite a bit.

Issue #403, opened 2019-01-07
[Swipeable] Overflow hidden limits usability


tedcurrent:
A Swipeable component not full width clips on swipe. This is due to
```
container: {
    overflow: 'hidden'
  },
```

Can we have a style prop for managing the `Animated.View` that wraps the children?

If this breaks something, I'm open for suggestions. Full width is not really an option.

roydigerhund:
I'm currently having the same issue. A card that isn't fullwidth and get's cropped on swipe due to the `overflow: 'hidden'` style settings. 

A `containerStyle` would help, or omit `overflow: 'hidden'` completely, I haven't experienced issues without this style.

osdnk:
🎉

PR #437, opened 2019-01-30
Add containerStyle Prop

roydigerhund:
PR for the issue #403 

This will add `containerStyle` as a property. This is for example useful to override the `overflow: 'hidden'` style.


osdnk:
Hi, @roydigerhund 
Thank you!
Please add docs and ts definitions :)

PR #447, opened 2019-02-04
Add "SectionList" Support

kamilzamerski:
Hi, in our app we need to add support for SectionList. It can be useful also for others ;)

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -14,6 +14,7 @@ import {
   WebView,
   StyleSheet,
   FlatList,
+  SectionList,
   Platform,
 } from 'react-native';
 import processColor from 'react-native/Libraries/StyleSheet/processColor';
@@ -704,6 +705,12 @@ function createNativeWrapper(Component, config = {}) {
 const WrappedScrollView = createNativeWrapper(ScrollView, {
   disallowInterruption: true,
 });
+
+const WrappedSectionList = createNativeWrapper(SectionList, {
+  disallowInterruption: true,
+  shouldCancelWhenOutside: false,
+});
+
 const WrappedSlider = createNativeWrapper(Slider, {
   shouldCancelWhenOutside: false,
   shouldActivateOnStart: true,
@@ -931,4 +938,5 @@ export {
   FlatListWithGHScroll as FlatList,
   gestureHandlerRootHOC,
   Directions,
+  WrappedSectionList as SectionList
 };


osdnk:
Hm, I don't think it's very needed. I mean that there are a lot of components which might be wrapped as well and we want to wrap only most important ones. Feel free to wrap `SectionList`, but I am not convinced if it should be inside core of RNGH.

osdnk:
However, thanks a ton for your interest in RNGH and PRs are always very welcome! 😄

stigi:
@osdnk I wonder if you would be open to PRs that would add `SectionList` to RNGH in the meantime?

I consider its importance on pair with `FlatList` and would love to get out of the box support for it. The implementation should be straight forward and follow the [FlatList example](https://github.com/software-mansion/react-native-gesture-handler/blob/f242d82d7cf0820cf0d9ab717f9c6695c78650f2/GestureComponents.js#L45-L57).

dani-mp:
Yes, please! 🙌

izakfilmalter:
For those wanting a SectionList:
```tsx
import {
  ForwardedRef,
  forwardRef,
  PropsWithChildren,
  ReactElement,
  RefAttributes,
} from 'react'
import {
  SectionList as RNSectionList,
  SectionListProps as RNSectionListProps,
} from 'react-native'
import {
  ScrollView,
  NativeViewGestureHandlerProps,
} from 'react-native-gesture-handler'

const commonProps = [
  'id',
  'enabled',
  'minPointers',
  'shouldCancelWhenOutside',
  'hitSlop',
] as const

const componentInteractionProps = ['waitFor', 'simultaneousHandlers'] as const

const baseGestureHandlerProps = [
  ...commonProps,
  ...componentInteractionProps,
  'onBegan',
  'onFailed',
  'onCancelled',
  'onActivated',
  'onEnded',
  'onGestureEvent',
  'onHandlerStateChange',
] as const

const nativeViewGestureHandlerProps = [
  'shouldActivateOnStart',
  'disallowInterruption',
] as const

const nativeViewProps = [
  ...baseGestureHandlerProps,
  ...nativeViewGestureHandlerProps,
] as const

export const SectionList = forwardRef((props, ref) => {
  const sectionListProps = {}
  const scrollViewProps = {}
  for (const [propName, value] of Object.entries(props)) {
    // https://github.com/microsoft/TypeScript/issues/26255
    if ((nativeViewProps as readonly string[]).includes(propName)) {
      // @ts-ignore - this function cannot have generic type so we have to ignore this error
      scrollViewProps[propName] = value
    } else {
      // @ts-ignore - this function cannot have generic type so we have to ignore this error
      sectionListProps[propName] = value
    }
  }
  return (
    // @ts-ignore - this function cannot have generic type so we have to ignore this error
    <RNSectionList
      ref={ref}
      {...sectionListProps}
      renderScrollComponent={(scrollProps) => (
        <ScrollView {...{ ...scrollProps, ...scrollViewProps }} />
      )}
    />
  )
}) as <ItemT = unknown>(
  props: PropsWithChildren<
    RNSectionListProps<ItemT> &
      RefAttributes<SectionList<ItemT>> &
      NativeViewGestureHandlerProps
  >,
  ref: ForwardedRef<SectionList<ItemT>>,
) => ReactElement | null

// eslint-disable-next-line @typescript-eslint/no-redeclare
export type SectionList<ItemT = unknown> = typeof SectionList &
  RNSectionList<ItemT>
```

PR #452, opened 2019-02-05
Export exclusiveTouch property on iOS and make touchables exclusive on Android

osdnk:
and make it default behavior in touchables

## Changes
Export `excluding` from iOS and make it default. 

Change activation logic in Android buttons. Now it becomes active on the very first touch. Previously in was activating by native handlers which are waiting some time from first move action. That was the reason I made touchables active on beginning instead of activating and it was leading to unwanted effect -- excluding other touched was not performed correctly and some blinkings were present.
Now Android buttons are excluding each other since all excluding logic is related to activation which is performed on first touch.

Clean up `GenericTouchable` code.

Lint

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -748,6 +748,7 @@ const RawButton = createNativeWrapper(
 class BaseButton extends React.Component {
   static propTypes = {
     ...RawButton.propTypes,
+    exclusiveTouch: PropTypes.bool,
     onPress: PropTypes.func,
     onActiveStateChange: PropTypes.func,
   };
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -748,9 +748,13 @@ const RawButton = createNativeWrapper(
 class BaseButton extends React.Component {
   static propTypes = {
     ...RawButton.propTypes,
-    exclusiveTouch: PropTypes.bool,
+    exclusive: PropTypes.bool,
     onPress: PropTypes.func,
     onActiveStateChange: PropTypes.func,
+  };
+
+  static defaultProps = {
+    exclusive: true,
   };
 
   constructor(props) {
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -753,10 +753,6 @@ class BaseButton extends React.Component {
     onActiveStateChange: PropTypes.func,
   };
 
-  static defaultProps = {
-    exclusive: true,
-  };
-
   constructor(props) {
     super(props);
     this._lastActive = false;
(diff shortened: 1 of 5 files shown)


GestureHandler.js
```diff
@@ -748,10 +748,15 @@ const RawButton = createNativeWrapper(
 class BaseButton extends React.Component {
   static propTypes = {
     ...RawButton.propTypes,
+    exclusive: PropTypes.bool,
```
kmagiera: why is it not in `RawButton` then?

ios/RNGestureHandlerModule.m
```diff
@@ -27,6 +27,11 @@ @implementation RNGestureHandlerButtonManager
 
 RCT_EXPORT_VIEW_PROPERTY(enabled, BOOL)
 
+RCT_CUSTOM_VIEW_PROPERTY(exclusive, BOOL, RNGestureHandlerButton)
+{
+  [view setExclusiveTouch:[RCTConvert BOOL: json]];
```
kmagiera: space after colon

touchables/GenericTouchable.js
```diff
@@ -123,16 +120,10 @@ export default class GenericTouchable extends Component {
     clearTimeout(this.pressOutTimeout);
     if (this.props.delayPressOut) {
       this.pressOutTimeout = setTimeout(() => {
-        if (this.STATE === TOUCHABLE_STATE.UNDETERMINED) {
```
osdnk: It was actually useless

GestureHandler.js
kmagiera: I think if someone uses RawButton then the defaults behavior is going to be different. Maybe we should defined the defaults on the native side in this case. That is when we create new native view instance it should have `exclusiveTouch` enabled by default and we don't need to have a default value specified in JS

kmagiera:
Is the PR title still relevant? Does the JS changes you made bring compatibility to Android? Or is there going to be a follow up with Android changes?

GestureHandler.js
osdnk: Well, that's fine. I changed it

osdnk:
Changes on Android side has been done for make behavior consistent on iOS and Android as well. On iOS buttons are being activated immediately which appears to be an expected behavior. On Android it was waiting a while and that was the reason I was activating animation on began instead of activation.

GestureHandler.js
kmagiera: So I understand this should no longer be needed here (bc `RawButton.propTypes` already contains `exclusive`)

osdnk:
Close in favor of https://github.com/kmagiera/react-native-gesture-handler/pull/486

PR #458, opened 2019-02-08
Add handleSwipeableLeftRelease/handleSwipeableRightRelease Props

scottmas:
I was trying to implement a swipeable row that on release animates back to the closed position. This is useful for instance on swipe to toggle functionality (For example, simply swiping right on the Gmail app causes the email to be archived)

To support this, I added two props: `handleSwipeableLeftRelease` and `handleSwipeableRightRelease`. They come with the "handle" prefix because the return value is merged into the animation options. In my specific use case I'm just returning `{ toValue: 0 }` to prevent the opening  animation, but its flexible enough to other use cases.

### DIFF
--- a/Swipeable.js
+++ b/Swipeable.js
@@ -35,9 +35,12 @@ export type PropType = {
   onSwipeableRightWillOpen?: Function,
   onSwipeableWillOpen?: Function,
   onSwipeableWillClose?: Function,
+  handleSwipeableLeftRelease?: Function,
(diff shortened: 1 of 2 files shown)


scottmas:
The more I think about this PR, it's runs counter to the uses of the swipeable component. There's got to be a better way. Closing.

torigeorge:
This is what I need. Did you get this to work?

PR #478, opened 2019-02-27
Memoize tap GH state for sending proper value after releasing finger

osdnk:
## Motivation 
Issue happens on iOS. https://github.com/kmagiera/react-native-gesture-handler/issues/476. 

After releasing finger in tap GH which is waiting for second GH and failure of second GH, values are being set zero and then zeros are send to JS instead on corrects values.
 
## Changes
After reseting values are memoized and placed in nativeEvent. If this kind of situation is detected (no finger on screen) we fallback to memoized valued values instead of evaluating them.

### DIFF
--- a/ios/Handlers/RNTapHandler.m
+++ b/ios/Handlers/RNTapHandler.m
@@ -12,8 +12,8 @@
 
 #import <React/RCTConvert.h>
 
-// RNBetterTapGestureRecognizer extends UIGestureRecognizer instead of UITapGestureRecognizer 
-// because the latter does not allow for parameters like maxDelay, maxDuration, minPointers, 
+// RNBetterTapGestureRecognizer extends UIGestureRecognizer instead of UITapGestureRecognizer
+// because the latter does not allow for parameters like maxDelay, maxDuration, minPointers,
 // maxDelta to be configured. Using our custom implementation of tap recognizer we are able
 // to support these.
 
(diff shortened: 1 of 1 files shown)


kmagiera (changes requested):
I'd prefer if we always use memoized value for the event and memoize it in a place where we are certain the values are right. I'm not certain that the conditions you added in `eventExtraData` are going to work at all times (e.g. could there be a case that numberOfTouches is zero but the code where we memoize hasn't been triggered?)

farfromrefug:
@kmagiera i think i fixed this the way you wanted.
I actually use your Native code for a Nativescript module.
Here is the commit if that is of any interest to you
https://github.com/Akylas/nativescript-gesturehandler/commit/6c3ed76e958660ea83364f1e3e9e6fcef8e7a43b

kmagiera:
This looks more like how I'd imagine. Thanks for sending @farfromrefug. Willing to send a PR with this change maybe?

farfromrefug:
@kmagiera sorry i dont use your code nor react-native

jkadamczyk:
@kmagiera @jakub-gonet I can apply changes that @farfromrefug proposed.
Is it still relevant?

m-bert:
I'm closing this PR as the issue no longer persists on v2 and v3. Let me know if this should be reopened.

PR #486, opened 2019-03-01
Add exclusive touch prop on Android

osdnk:
## Motivation
The main issue which is the source of this problem is an inconsistent behavior of touchables on android and iOS. The main case was a feedback of touchables (e.g. fading in `TouchableOpacity`) which was not exclusive (it means that it could also occurs on on multiple buttons) even that exclusive-related logic was already done in RNGH (in was impossible to make two buttons `active` in the same time)

## Changes
I decided that buttons by default should not only be forbidden to activate simultaneously but also they shouldn't even begin immediately (it's a behavior on iOS)! It was an important point because touchables' feedback was hooked to beginning of gesture, not activation. I can't simple change it because on Android buttons do not become active immediately.
I moved catching responder to a bit earlier phase and connect it with `begin()` in `NativeViewGestureHandler` which might appears to be a bit dirty, but it was necessary. Catching responder (mutex for exclusive touch) is not handled by `setResponder` method.
Then it became extremely easy to make exclusiveness configurable. We could simply "not catch" responder it there's `exclusive` prop set to `false` which follows iOS behavior.

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/NativeViewGestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/NativeViewGestureHandler.java
@@ -5,6 +5,8 @@ import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 
+import com.swmansion.gesturehandler.react.RNGestureHandlerButtonViewManager;
+
 public class NativeViewGestureHandler extends GestureHandler<NativeViewGestureHandler> {
 
   private boolean mShouldActivateOnStart;
@@ -86,7 +88,14 @@ public class NativeViewGestureHandler extends GestureHandler<NativeViewGestureHa
         view.onTouchEvent(event);
         activate();
       } else if (state != STATE_BEGAN) {
-        begin();
+        // setting flag for exclusive touch for buttons
+        if (view instanceof RNGestureHandlerButtonViewManager.ButtonViewGroup) {
+          if (((RNGestureHandlerButtonViewManager.ButtonViewGroup) view).setResponder()) {
+            begin();
+          }
+        } else {
+          begin();
+        }
       }
     } else if (state == STATE_ACTIVE) {
       view.onTouchEvent(event);
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
@@ -21,10 +21,10 @@ import com.facebook.react.uimanager.annotations.ReactProp;
 public class RNGestureHandlerButtonViewManager extends
         ViewGroupManager<RNGestureHandlerButtonViewManager.ButtonViewGroup> {
 
-  static class ButtonViewGroup extends ViewGroup {
+  public static class ButtonViewGroup extends ViewGroup {
 
     static TypedValue sResolveOutValue = new TypedValue();
-    static ButtonViewGroup sResponder;
+    public static ButtonViewGroup sResponder;
 
     int mBackgroundColor = Color.TRANSPARENT;
     // Using object because of handling null representing no value set.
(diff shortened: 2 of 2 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -225,6 +236,11 @@ public void setRippleColor(ButtonViewGroup view, Integer rippleColor) {
     view.setRippleColor(rippleColor);
   }
 
+  @ReactProp(name = "exclusive")
+  public void setExclusive(ButtonViewGroup view, Boolean  exclusive) {
```
kmagiera: can we just use `boolean` instead of `Boolean` (you can also set default value via `@ReactProp` in case it is missing)

android/lib/src/main/java/com/swmansion/gesturehandler/NativeViewGestureHandler.java
```diff
@@ -86,7 +88,14 @@ protected void onHandle(MotionEvent event) {
         view.onTouchEvent(event);
         activate();
       } else if (state != STATE_BEGAN) {
-        begin();
+        // setting flag for an exclusive touch for buttons
+        if (view instanceof RNGestureHandlerButtonViewManager.ButtonViewGroup) {
```
kmagiera: not sure I like adding this check here. I'd prefer to try one of the following ways:
1) Can we use `shouldRecognizeSimultaneously` abstraction to implement this behavior?
2) Can we add it directly to `NativeViewGestureHandler` or maybe even to all types of gesture handlers?

osdnk: 1. not sure, since `shouldRecognizeSimultaneously` allows for staying in began state and I don't want to change this logic.

2. I'm not convinced how could it look like without any `instanceof RNGestureHandlerButtonViewManager.ButtonViewGroup`. I moved it to another function for a bit cleaner code.

kmagiera:
Also if the above is going to be consistent with the behavior on iOS?

osdnk:
Updated PR information

micheleb:
also checking it to thank @osdnk for the PR, and for an FYI: if I have button B inside a larger button C, and I click on B, this PR correctly makes C's handler ignore the click, and only B is triggered. However, if after the first click I click on B again, C's handler is triggered instead. This obviously happens only if clicking on B does not trigger a re-render of the screen.

omerman:
Any news on this PR?

braincore:
Can we get this merged into master? It still rebases cleanly with master AFAICT and I'm using it successfully.

LcTwisk:
@braincore @micheleb @osdnk Seems to be working fine with some small modifications: https://github.com/LcTwisk/react-native-gesture-handler/tree/fix-exclusive-touch-on-android-2
Screen Recording 2020-08-26 at 21 30 25
Can someone verify?

jkadamczyk:
I resolved conflicts, from what I got out of the conversation it is more less ready to go
CC @kmagiera @jakub-gonet

AireshBhat:
Just want to say thank you for all the efforts put in for this project. This issue seems to be a deal breaker for me as of now though, anyway merging of this pull request can be fast-tracked? Thank you again.

yfunk:
The builds are failing, because @jkadamczyk seems to have accidentally removed a closing bracket on Line 266 in `RNGestureHandlerButtonViewManager.java`. 
With that fixed i also noticed the issue @micheleb mentioned (Emulator, API29, RN0.63.2), the solution suggested by @LcTwisk however does **not** work for me, both touchables are pressed.

yfunk:
Since with React Native Gesture Handler 1.9.0 every touchable now has a tap sound this PR would be even more important to merge. Nested touchables on Android will now not only execute the `onPress` action of their parents but also play sounds for them, which results in a weird "multi-clicking" sound. I attempted to find a fix for the [issue that seems to be holding this PR back](https://github.com/software-mansion/react-native-gesture-handler/pull/486#issuecomment-563310225), but I'm not familiar enough with how RNGH works. As mentioned in my last comment the builds are failing because of a missing closing bracket.

Could anyone who is currently working on `react-native-gesture-handler` or contributed to this PR please have another look at this?
@osdnk @jakub-gonet @kmagiera @jkadamczyk

jakub-gonet:
I need to test it locally, but I'm not sure if buttons should be emitting sounds when nested.

PR #487, opened 2019-03-01
Remove peerDependencies for web

brentvatne:
react-native-web shouldn't be a peerDep, and I'm not sure why react-art is there. I think we could also remove react and react-native and just put them as `'*'` and specify more clear version requirements in the README / installation instructions, because it all just gets lots in the peerDependencies warning mess that react-native is in.

screen shot 2019-03-01 at 12 15 01 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,10 +59,7 @@
   },
   "peerDependencies": {
     "react": ">= 16.3.2",
-    "react-art": ">= 16.3.2",
-    "react-dom": ">= 16.3.2",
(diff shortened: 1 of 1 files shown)


kmagiera:
Can you advise @jaulz ?

Also shouldn't we update yarn.lock here too?

jaulz:
@brentvatne is of course right... I don't think that we need to update the peer dependencies since we are changing the `peerDependencies` only.

PR #493, opened 2019-03-04
Add mocks for RNGH

osdnk:
## Motivation
Need to provide some mocks for RNGH in order to make it workable with jest.

## Changes
1. Remove need of mocking `PlatformConstants`. Thanks @tsapeta! Now I check whether `PlatformConstants` exists before usage.
2. Add mocks of RNGH module. 
3. Also abstract RNGH module for separated file for easier mocking.


__mocks__/RNGestureHandlerModule.js
```diff
@@ -0,0 +1,13 @@
+export default {
```
dsznajder: I've also find out that you may need to mock those: 
```js
jest.mock('NativeModules', () => ({
  UIManager: {
    RCTView: () => {},
  },
  PlatformConstants: {
    forceTouchAvailable: false,
  },
}))
```

I'm not sure but it may help others :)

dsznajder: Oh, nevermind. You've changed implementation of `GestureHandler`

osdnk: Certainly I do not need `PlatformConstants`. See the changes and you'll figure out why :).
Why do you think I do need `RCTView`?

dsznajder: Oh, sorry. This seems to be obsolete. I've had this one in past and thought it's still necessary.

osdnk:
I'll share project later. I don't want to have it done it Example because of really strange config, which makes it more difficult and less universal.

osdnk:
https://github.com/mehranabi/rn-gh-error/blob/master/__tests__/App.js
It's quite useful

dsznajder:
@osdnk 	As you see [there](https://github.com/mehranabi/rn-gh-error/blob/7501841caba8d4b8dc7b60595ad0b62088c2015c/__tests__/App.js#L14) it has been necessary to pass the tests. That's why I mentioned this one :)

osdnk:
@dsznajder 
Yes, but https://youtu.be/HG7I4oniOyA?t=115

I see that people are doing it, but I don't want to make it matter of faith. I suppose it was important it previous version because of this:
https://github.com/kmagiera/react-native-gesture-handler/commit/35c57a0f4930210c3b152e33e814e97ec1e23d92#diff-56d925ec2aada9e1b3b613fe5b782837

But since this commit I see it does not have any impact.

dsznajder:
I can confirm that also. It does not make any difference with `1.1.0` and `1.0.17` so I'm pretty sure that your mock is 100% correct ;)

dsznajder:
@osdnk I find out that with 1.1.0 and RN 0.59.1 I got an error: 
`TypeError: Cannot read property 'setJSResponder' of undefined`
caused by [this line](https://github.com/kmagiera/react-native-gesture-handler/blob/70451b949e5072413b4bfd48a13f2ab3f0aa84c1/createHandler.js#L13)

It seems to be necessary to mock `UIManager` with an empty object or sth.

package.json
```diff
@@ -33,6 +34,8 @@
     "PlatformConstants.web.js",
     "react-native-gesture-handler.d.ts",
     "README.md",
+    "RNGestureHandlerModule.js",
+    "jestConfig.js",
```
FireyFly: This should be `jestSetup.js`, I think? Which probably explains why the `jestSetup.js` appears to be missing from the 1.2.0 release.

PR #503, opened 2019-03-06
ignore exclusive touch on tvOS

kasinskas:
exclusiveTouch is not supported on tvOs


osdnk:
I find it quite similar 😄 
https://github.com/kmagiera/react-native-gesture-handler/pull/505

PR #505, opened 2019-03-06
Except exclusiveTouch from TV build

osdnk:
Fixes https://github.com/kmagiera/react-native-gesture-handler/issues/502

This method is not available on TV.

### DIFF
--- a/ios/RNGestureHandlerButton.m
+++ b/ios/RNGestureHandlerButton.m
@@ -38,7 +38,9 @@
   self = [super init];
   if (self) {
     _hitTestEdgeInsets = UIEdgeInsetsZero;
+#if !TARGET_OS_TV
     [self setExclusiveTouch:YES];
+#endif
   }
   return self;
 }
--- a/ios/RNGestureHandlerModule.m
+++ b/ios/RNGestureHandlerModule.m
@@ -29,7 +29,9 @@ RCT_EXPORT_VIEW_PROPERTY(enabled, BOOL)
 
 RCT_CUSTOM_VIEW_PROPERTY(exclusive, BOOL, RNGestureHandlerButton)
 {
+#if !TARGET_OS_TV
   [view setExclusiveTouch: json == nil ? YES : [RCTConvert BOOL: json]];
+#endif
 }
 
 RCT_CUSTOM_VIEW_PROPERTY(hitSlop, UIEdgeInsets, RNGestureHandlerButton)


lukebars:
When deployed to the simulator I get an error "Native module cannot be null" which can be dismissed. Any ideas?

osdnk:
TV simulator on iPhone?

kasinskas:
TV

osdnk:
Hm. I've just made simple RN app and compiled it on TV with my changes and see no problem. Could you provide some steps to repro?

osdnk:
Closing in favor of https://github.com/kmagiera/react-native-gesture-handler/pull/503#pullrequestreview-211494627

lukebars:
> Hm. I've just made simple RN app and compiled it on TV with my changes and see no problem. Could you provide some steps to repro?

Literally just this stack navigator in app.js and a dummy home page. Or two dummy pages with a link, to give it some actual purpose.
```
import { createAppContainer, createStackNavigator } from 'react-navigation';

import Home from './home';

const AppNavigator = createStackNavigator({
    Home: {
        screen: Home,
    },
});

export default createAppContainer(AppNavigator);
```

osdnk:
@lukebars 
Do you observe it with #503 ?

lukebars:
@osdnk yeah it's same, the #503 was just a more elegant solution since the exact code that was recently added for the exclusive touch was commented out, but the output was still the same.

osdnk:
hm, strange. I really made an app and it worked 😢. Could you prepare some minimal example? Maybe it's just some config

lukebars:
@osdnk did you use createStackNavigator and CreateAppContainer? When I define a navigator that provides an error. But it allows me to dismiss it ( Esc ) and everything works just fine and that's really strange.

osdnk:
Well, I haven't use react-navigation at all. Maybe try adding `import react-native-gesture-handler` on the top  of some main file?

lukebars:
Turns out react-navigation was causing this. Had to add podspec RCTLinkingIOS and it's gone. 
[https://github.com/react-navigation/react-navigation/issues/460#issuecomment-282078380]

lukebars:
Is it possible to get a new release? Kinda sucks to wrap the exclusive touch with if's every time 🦉

godwinap:
> When deployed to the simulator I get an error "Native module cannot be null" which can be dismissed. Any ideas?

Me too. I'm seeing null is not an object (evaluating 'RNGestureHandlerModule.Direction')

osdnk:
@godwinap 
I think it's different issue

lukebars:
@godwinap
@osdnk is right. That's completely different. Mine was because of react-navigation on Apple TV, yours is RNGestureHandler related.

godwinap:
> @godwinap
> @osdnk is right. That's completely different. Mine was because of react-navigation on Apple TV, yours is RNGestureHandler related.

@osdnk , I'm not sure if it's the same issue or not. But I'm trying to do the same thing as @lukebars. My app uses react-navigation, which run fine on Android TV but not on Apple TV. When I checked further I found that `const { RNGestureHandlerModule } = NativeModules;` from file `node_modules/react-native-gesture-handler/Directions.js` `RNGestureHandlerModule` is giving `null`.

seajean:
still got this problem

lukebars:
@seajean what error are you getting?

PR #506, opened 2019-03-06
Remove ViewPager and Slider in favor of exporting createNativeWrapper

### LINKED ISSUE #482: ViewPagerAndroid and Slider moved from RN 0.59 to external packages
Warning: ViewPagerAndroid has been extracted from react-native core and will be removed in a future release. It can now be installed and imported from '@react-native-community/viewpager' instead of 'react-native'. See https://github.com/react-native-community/react-native-viewpager

Warning: Slider has been extracted from react-native core and will be removed in a future release. It can now be installed and imported from '@react-native-community/slider' instead of 'react-native'. See https://github.com/react-native-community/react-native-slider

karanjthakkar:
Fixes #482 

In RN 0.59 `ViewPager` and `Slider` are going to be dropped and therefore they call for being dropped from RNGH core. However, since there're use cases of these components, decided not to drop support for them at all, but to expose function for creating generic components with pure native ones.

### DIFF
--- a/Example/android/app/build.gradle
+++ b/Example/android/app/build.gradle
@@ -126,6 +126,7 @@ android {
 }
 
 dependencies {
+    implementation project(':@react-native-community_slider')
     compile project(':react-native-vector-icons')
     compile project(':react-native-gesture-handler')
     compile fileTree(dir: "libs", include: ["*.jar"])
--- a/Example/android/app/build.gradle
+++ b/Example/android/app/build.gradle
@@ -126,6 +126,7 @@ android {
 }
 
 dependencies {
+    implementation project(':@react-native-community_viewpager')
     implementation project(':@react-native-community_slider')
     compile project(':react-native-vector-icons')
     compile project(':react-native-gesture-handler')
--- a/Example/android/app/src/main/java/com/swmansion/gesturehandler/react/example/MainApplication.java
+++ b/Example/android/app/src/main/java/com/swmansion/gesturehandler/react/example/MainApplication.java
@@ -3,6 +3,7 @@ package com.swmansion.gesturehandler.react.example;
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
+import com.reactnativecommunity.slider.ReactSliderPackage;
 import com.oblador.vectoricons.VectorIconsPackage;
 import com.facebook.react.ReactNativeHost;
 import com.facebook.react.ReactPackage;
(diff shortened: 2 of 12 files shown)


osdnk:
Hello, @karanjthakkar 
I'm not into adding this dependencies. How about exporting function for creating components?

karanjthakkar:
@osdnk Do you mean removing ViewPagerAndroid and Slider support from RNGH and allowing consumers to wrap them on their own?

osdnk:
Yes, I don't think if it's good to add peers if we actually do not require them (I know it's done already but I see no good reason to add it here as well)

karanjthakkar:
That's fine. So if I understand correctly:

1. instead of using the third party packages, you want to remove Slider and ViewPagerAndroid from the codebase
2. and we want to let consumers create their own wrappers by exposing `createNativeWrapper` externally

Is that right?

osdnk:
yes. I know it's a breaking change, but I'm afraid that it's not big deal since it's the only issue related and it's not very popular. 
I'll include it in release notes.

karanjthakkar:
Sounds good. Would you like me to do the necessary changes in this PR?

osdnk:
I could do it as well. However, it's your honour to be so reactive and thanks to you we will manage to prevent users' crashes in the nearest future! 😃

osdnk:
@karanjthakkar 
Could you add some docs for this illustrating how to it with these examples (ViewPager and Slider)?

karanjthakkar:
1. Where do you think would be the right place to add docs for `createNativeWrapper`? 
2. There is still some usage of `ViewPagerAndroid` and `Slider` in the `Examples`. How do you recommend I remove it? Or do you want to keep it there?

osdnk:
1. I think that `docs/handlers.md` is right place (under list of exported components)
2. I'd like to keep them alive. It would be very nice example of using this wrapper.

karanjthakkar:
> I think that docs/handlers.md is right place (under list of exported components)

I'll check it out.

> I'd like to keep them alive. It would be very nice example of using this wrapper.

That would mean I need to add the third party packages in the Examples project right? And then use createNativeWrapper to create the wrapped version of it and use that in place of the original one?

osdnk:
Yup

osdnk:
Ooops, sorry

karanjthakkar:
No problem. Thanks for the tips. I'll take a look soon.

osdnk:
Is it ready, @karanjthakkar?
May I merge?

karanjthakkar:
I'll ping once I have tested the Examples that I changed. Give me some time :)

karanjthakkar:
@osdnk I've verified that the examples work as expected. Feel free to merge once the CI is green. Thanks for helping me out 🙂

ujwal-setlur:
Any idea when a new release with this change will be pushed out?

PR #515, opened 2019-03-11
RTL support

ShaMan123:
motivation: #319 
~~further work: [implement a helper function](https://github.com/kmagiera/react-native-gesture-handler/issues/319#issuecomment-471767752)~~


Swipeable.js
```diff
@@ -351,10 +351,10 @@ const styles = StyleSheet.create({
...
-    flexDirection: 'row',
+    flexDirection: I18nManager.isRTL? 'row-reverse': 'row',
   },
   rightActions: {
     ...StyleSheet.absoluteFillObject,
-    flexDirection: 'row-reverse',
+    flexDirection: I18nManager.isRTL ? 'row' : 'row-reverse',
   },
-});
+});
```
adamczyk777: Line break at the end of file is missing :/

adamczyk777:
@osdnk Would you mind share your opinion about this? It certainly fixes the issue of not being able to swipe the row. The only thing I see is that your implementation of an Apple-styled swipeable row does not work in the same way as without RTL enabled. My guess is it is not related to the library but just wanted you to take a look at this one as you were the author.

Swipeable.js
ShaMan123: I can't find it as not fixed. Could it be because I fixed it in my fork after sending the PR?

ShaMan123: OK I think I managed to sync the PR

ShaMan123:
I will give a look. 
For now I'm not able to build the example app on android....
Did you read my [suggestion](https://github.com/kmagiera/react-native-gesture-handler/issues/319#issuecomment-471767752). After looking again at the file I have an idea to merge the suggestion easily.

ShaMan123:
I updated the Example app. 
Everything works as it should for both layouts.
Because `rtl` layout needs `flexDirection: row-reverse` there's more dynamic styling to be applied higher up the component tree.
I think that in some use cases additional interpolation will be needed to be applied on `progress`/`dragX` passed to `renderRightActions` to achieve the desired behavior. 
I tested both.
I think that styling is the answer in most use cases.

It might be good to document this because it might be frustrating handling right actions with `rtl` layout without peeking under the hood. (i.e. if someone needs to calculate interpolation they need to take into account that some measurements are relative to the right while most logic is relative to the left, meaning they might need to compensate for the width of the view in question or something alike for the calculation to be correct)

Hope I was clear

ShaMan123:
I will submit another PR for androidX build for Example app

Example/swipeable/AppleStyleSwipeableRow.js
```diff
@@ -44,13 +44,13 @@ export default class AppleStyleSwipeableRow extends Component {
...
-  renderRightActions = progress => (
-    <View style={{ width: 192, flexDirection: 'row' }}>
-      {this.renderRightAction('More', '#C8C7CD', 192, progress)}
-      {this.renderRightAction('Flag', '#ffab00', 128, progress)}
-      {this.renderRightAction('More', '#dd2c00', 64, progress)}
-    </View>
-  );
+    renderRightActions = progress => (
+        <View style={{ width: 192, flexDirection: I18nManager.isRTL?'row-reverse':'row' }}>
+            {this.renderRightAction('More', '#C8C7CD', 192, progress)}
```
osdnk: Could you fix whitespaces?

ShaMan123: My IDE goes mad because of whitespace diffs - dome my best.

ShonP:
this is wrong!! you actually check if the device is an rtl device but what you should check is the inherited View and know what direction you are in https://facebook.github.io/react-native/docs/layout-props#direction

ShaMan123:
AFAIK `I18nManager` is the standard for `rtl` layout logic.
Anyways this PR fixes calculations that shouldn't change (I think) due to `direction` prop.
Did you run the example app? and read the docs I've edited?

ShaMan123:
Any intention on merging this?

vovkind:
Guys, any progress here, please? 
We really love this plugin, and we must have this fix.

adamczyk777:
If you want to use the version with a fix, you can simply provide GitHub link in package.json I guess. It was long time ago, I have limited time now, I will review it once again I guess. Not sure when though.

ShaMan123:
I've been using it since, seems to be fixed

ShaMan123:
It is a very minor change to js files

vovkind:
> If you want to use the version with a fix, you can simply provide GitHub link in package.json I guess. It was long time ago, I have limited time now, I will review it once again I guess. Not sure when though.

Yep. It is what I did later...It is working.

@ShaMan123 thank you!

adamczyk777:
Dudes I know I am not a full time maintainer here as @osdnk was and I don't want to break anything, as I won't have time to fix it.

osdnk:
Please remove native changes and all white spaces changes and I’ll review it then

ShaMan123:
done

ShaMan123:
@vovkind 
I tagged you to let you know I updated my fork from base so it should be good for RN0.61+.
Build passes.

Maintainers,
This repo is excellent.
Please make it RTL compatible, review this PR and accept it.

calintamas:
Any plans on merging this?

osdnk:
Sorry about this delay!

Example/package.json
```diff
@@ -33,6 +33,7 @@
     "@babel/polyfill": "^7.2.5",
     "@babel/preset-env": "^7.3.1",
     "@babel/runtime": "^7.4.4",
+    "@react-native-community/cli": "^1.9.4",
```
Irislm, line 36: Hi, i found if use this package, `pod install` will give an error `Invalid `Podfile` file: A JSON text must at least contain two octets!.` Because there is no config in @react-native-community/cli-1.9.4 ! Please check!

PR #522, opened 2019-03-15
Added try/catch for dispatchTouchEvent

cvlehe:
Fix for this issue: https://github.com/kmagiera/react-native-gesture-handler/issues/519 - Crash when using multiple fingers to interact with certain components. 

Fixed by adding try/catch for ```IllegalArgumentException```  when trying to call ```dispatchTouchEvent```.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEnabledRootView.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEnabledRootView.java
@@ -37,6 +37,7 @@ public class RNGestureHandlerEnabledRootView extends ReactRootView {
       }
     }catch (IllegalArgumentException e) {
       e.printStackTrace();
+      return false;
     }
     return false;
   }
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEnabledRootView.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEnabledRootView.java
@@ -29,8 +29,8 @@ public class RNGestureHandlerEnabledRootView extends ReactRootView {
   @Override
   public boolean dispatchTouchEvent(MotionEvent ev) {
     try {
-      if (mGestureRootHelper != null) {
-        mGestureRootHelper.dispatchTouchEvent(ev)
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEnabledRootView.java
```diff
@@ -28,10 +28,18 @@ public void requestDisallowInterceptTouchEvent(boolean disallowIntercept) {
 
   @Override
   public boolean dispatchTouchEvent(MotionEvent ev) {
-    if (mGestureRootHelper != null && mGestureRootHelper.dispatchTouchEvent(ev)) {
-      return true;
+    try {
+      if (mGestureRootHelper != null && mGestureRootHelper.dispatchTouchEvent(ev)) {
+        return true;
+      }
+      if (super.dispatchTouchEvent(ev)) {
```
osdnk: why not `return super.dispatchTouchEvent(ev);`

osdnk:
I will happily merge it as long as I could reproduce it and will figure out if an error is not related to our codebase.

cvlehe:
> I will happily merge it as long as I could reproduce it and will figure out if an error is not related to our codebase.

To reproduce I made rapid taps on my Android device on a view that implements `PanGestureHandler`.

osdnk:
So is it possible to break something from `Example` as well in this way?

cvlehe:
> So is it possible to break something from `Example` as well in this way?

Yes, in any view utilizing the `PanGestureHandler`.

cvlehe:
I'm able to reproduce consistently by rapidly tapping on a view with a `PanGestureHandler` with multiple fingers. After doing that for a few seconds, the app will crash. We're seeing a large number of events for this issue in our crash log.

osdnk:
Could you tell me sth more? What's your phone? 
I'm using OnePlus A6003 / API 28

https://drive.google.com/open?id=1ubENepOgoyHGE6svzuzLS1-Ip4v-oALg

cvlehe:
Seems to be resolved by in `1.1.0`. Closing this PR

abudhia-oyo:
@cvlehe @osdnk 

Is there any reason why the issue would still exists in v1.7.0 #1205

PR #525, opened 2019-03-18
Fixes Detox running with RNGH(possibly other libraries)

ddzirt:
This PR comes from this issue:
https://github.com/wix/Detox/issues/1207

Removes false positives and false positives out of this check:
https://github.com/kmagiera/react-native-gesture-handler/blob/c9f7c46b42966c0ec9c7f97df3d5a119ce789ae6/createHandler.js#L42

and allows for 3d partly libraries to add their gestures on top(if they pass null or custom functions that can possibly break this check)

### DIFF
--- a/createHandler.js
+++ b/createHandler.js
@@ -36,8 +36,8 @@ let handlerTag = 1;
 const handlerIDToTag = {};
 
 function isConfigParam(param, name) {
-  // param !== Object(param) returns false if param is a function and 
-  // returns true if param is null
+  // param !== Object(param) returns false if `param` is a function
+  // or an object and returns true if `param` is null
   return (
     param !== undefined &&
     (param !== Object(param) || !('__isNative' in param)) &&
--- a/createHandler.js
+++ b/createHandler.js
@@ -35,17 +35,12 @@ UIManager.genericDirectEventTypes = {
 let handlerTag = 1;
 const handlerIDToTag = {};
 
-// This check ensures that we do not pass
-// false positives(like null) or false negatives(functions)
-// that can possibly come from other libraries
-function isObject(obj) {
-  return obj === Object(obj);
-}
-
 function isConfigParam(param, name) {
+  // param !== Object(param) returns false if param is a function and 
+  // returns true if param is null
   return (
-    param &&
-    (!isObject(param) || !('__isNative' in param)) &&
+    param !== undefined &&
+    (param !== Object(param) || !('__isNative' in param)) &&
     name !== 'onHandlerStateChange' &&
     name !== 'onGestureEvent'
   );
--- a/createHandler.js
+++ b/createHandler.js
@@ -35,11 +35,17 @@ UIManager.genericDirectEventTypes = {
 let handlerTag = 1;
 const handlerIDToTag = {};
 
+// This check ensures that we do not pass
+// false positives(like null) or false negatives(functions)
+// that can possibly come from other libraries
+function isObject(obj) {
+  return obj === Object(obj);
+}
+
 function isConfigParam(param, name) {
   return (
-    param !== undefined &&
-    typeof param !== 'function' &&
-    (typeof param !== 'object' || !('__isNative' in param)) &&
+    param &&
+    (!isObject(param) || !('__isNative' in param)) &&
     name !== 'onHandlerStateChange' &&
     name !== 'onGestureEvent'
   );


ddzirt:
> The code looks good, but I'm afraid that it's a little bit unclear and will lead to mistakes in the future.
> 
>     1. Can we inline it? I see no point for making it a separated function.
> 
>     2. Can you write a little bit more low-level comment like `This check gives true if it's function or object but false if it's null` etc?

@osdnk we can replace function with additional check within your previous solution to look like this:
```javascript
// This will do the same as my solution, but the check becomes a bit strange and hard 
// to understand reading it first time, imo
return (
    param && 
    typeof param !== 'function' &&
    (typeof param !== 'object' || !('__isNative' in param)) &&
    name !== 'onHandlerStateChange' &&
    name !== 'onGestureEvent'
  );
```

or inline this to something like this:
```javascript
// This, to me at least, looks cleaner and easier to understand
return (
    param &&
    (param !== Object(param) || !('__isNative' in param)) &&
    name !== 'onHandlerStateChange' &&
    name !== 'onGestureEvent'
);
```

Regarding comment, I feel like it is completely not needed if we inline the check removing function in progress.

I can update this later today, depending on what feels better to leave as final solution.

osdnk:
The second solution looks fine. However, I would really appreciate at least a minor comment. It's not obvious that `(param !== Object(param)` gives `false` for functions and `true` to null

ddzirt:
> The second solution looks fine. However, I would really appreciate at least a minor comment. It's not obvious that `(param !== Object(param)` gives `false` for functions and `true` to null

So something like:
```javascript
// param !== Object(param) returns false if param is a function and 
// returns true if param is null
return (
    param &&
    (param !== Object(param) || !('__isNative' in param)) &&
    name !== 'onHandlerStateChange' &&
    name !== 'onGestureEvent'
);
```

osdnk:
Great!

ddzirt:
> Ok, one more thing.
> If param is `0` then `isConfigParam` return `false`. It should not be like.
> 
> My idea is
> 
> ```js
> function isConfigParam(param, name) {
>   return (
>     param !== undefined &&
>     typeof param !== 'function' &&
>     (typeof param !== 'object' || (param !== null && !('__isNative' in param))) &&
>     name !== 'onHandlerStateChange' &&
>     name !== 'onGestureEvent'
>   );
> }
> ```
> 
> I also consider it a temporary solution and I want to handle `null`somehow for fallback to default values

So, you mean, that if within an array of valid props that we have in `filterConfig` function, some can pass 0 as a valid value, check has to return true. OK

And if value passed as `null` has to be recognized as a fallback to a default value(what will be default value in this case)? This means that check has to pass properly(returning true or false) and allow `filterConfig` perform it's own logic. OK

osdnk:
Don’t care about this default values. It’s a plan for a future :)?;

ddzirt:
> Don’t care about this default values. It’s a plan for a future :)

Wow, is there any road map about changes/new features?

osdnk:
@ddzirt 
Not now. 
I'm afraid that now (and for few next weeks) we are very limited when it comes to our time resources so I'm relying on community rather than me and @kmagiera. So now I see it difficult to give any promises what (and when) will be finished. 

If you want to help us, I'm thinking about implementing "native fallback to default values". I mean that currently values are set in a native constructor. And if they're set to `null` after creating (by changing props) they're probably ignoring. So I believe it's not difficult to do some primitive mechanism for it.

osdnk:
Also, I'm merging it, but I cannot make a release now. 
Ping @kmagiera

ddzirt:
> @ddzirt
> Not now.
> I'm afraid that now (and for few next weeks) we are very limited when it comes to our time resources so I'm relying on community rather than me and @kmagiera. So now I see it difficult to give any promises what (and when) will be finished.
> 
> If you want to help us, I'm thinking about implementing "native fallback to default values". I mean that currently values are set in a native constructor. And if they're set to `null` after creating (by changing props) they're probably ignoring. So I believe it's not difficult to do some primitive mechanism for it.

What I was asking about was merely a small list of features that will be implemented in future or distant future, so that if I have time I could help out a bit.

"native fallback to default values", sounds interesting, but I don't really know were to look for that so, it will take me some time to investigate first, so I'll look into that and hopefully add a new PR

osdnk:
I think I'll add it soon. I thought about it about ear ago but I was a time when there was no so active help from the community so it would be just for my convenience. 
Now I see it very reasonable.

yogeshthanvi:
After adding this tap actions stopped working we are not able to navigate from one screen to another

ddzirt:
@yogeshthanvi what is your setup(react native, react navigation), I'd like to have a look at it

yogeshthanvi:
Mine project config 
"react-native": "0.59.0",
"react-navigation": "3.0.8",

micabe:
Same here:
Mine project config
"react-native": "0.59.2",
"react-navigation": "3.5.1",

ddzirt:
Issue regarding Animations was not connected to RNGH, https://github.com/wix/Detox/issues/1234 fixed any issues Detox had with React Navigation.

PR #527, opened 2019-03-18
Add SectionList to GHScroll

team55:


osdnk:
Hi, @team55 
Thanks for this contribution, but I'm afraid I don't want to have it merge. It's not a very common (comparing with e.g. ScrollView) component and I see no point in adding every single component users require in RNGH core. 

Also, it's possible to achieve this functionality without changes in RNGH core.

MikeMajara:
> Hi, @team55
> Thanks for this contribution, but I'm afraid I don't want to have it merge. It's not a very common (comparing with e.g. ScrollView) component and I see no point in adding every single component users require in RNGH core.
> 
> Also, it's possible to achieve this functionality without changes in RNGH core.

Could you provide a simple way to do this? I fail to see any downside to provide this out of the box.

Issue #358, opened 2018-11-28
Use Animated.event with Button


ckknight:
I'd like to use `Animated.event` with the handlers of the Buttons, without having to rely on using `TapGestureHandler` or similar.

Given the following code:

```jsx
<RawButton onGestureEvent={Animated.event([{ nativeEvent: { state: new Animated.Value(-1) } }])}>
  <View style={{width: 50, height: 50, backgroundColor: 'red'}} />
<RawButton>
```

This error occurs when tapping:

Expected `onGestureHandlerStateChange` listener to be a function, instead got a value of `object` type.

This is the case whether using `react-native`'s `Animated` or using `react-native-reanimated`.

osdnk:
Hi, @ckknight 
Thanks for this issue. However I didn't manage to repro using newest version of RNGH

brentvatne:
@osdnk - I was able to reproduce this on iOS with the latest version of RNGH and Reanimated

```js
import * as React from 'react';
import { Text, View, StyleSheet } from 'react-native';
import { RawButton, State } from 'react-native-gesture-handler';
import Animated from 'react-native-reanimated';
const { event, eq, cond, Value } = Animated;

export default class App extends React.Component {
  constructor(props) {
    super(props);

    const state = new Value();

    // This does not work
    //
    this._onGestureEvent = event([
      {
        nativeEvent: { state },
      },
    ]);

    // This works!
    //
    // this._onGestureEvent = ({ nativeEvent }) => {
    //   console.log(nativeEvent.state);
    //   state.setValue(nativeEvent.state)
    // };

    this._opacity = cond(eq(state, State.ACTIVE), 0.1, 1);
  }

  render() {
    return (
      <View style={styles.container}>
        <RawButton onHandlerStateChange={this._onGestureEvent}>
          <Animated.View style={[styles.button, { opacity: this._opacity }]}>
            <Text style={styles.buttonText}>I am a button!</Text>
          </Animated.View>
        </RawButton>
      </View>
    );
  }
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    backgroundColor: '#ecf0f1',
    padding: 8,
  },
  button: {
    paddingVertical: 15,
    paddingHorizontal: 20,
    backgroundColor: '#000',
    borderRadius: 3,
  },
  buttonText: {
    color: '#fff',
  },
});
```

osdnk:
[...]

PR #531, opened 2019-03-20
Export PureNativeButton

osdnk:
## Motivation
@ckknight 
I'd like to use `Animated.event` with the handlers of the Buttons, without having to rely on using `TapGestureHandler` or similar.

Given the following code:

```jsx
<RawButton onGestureEvent={Animated.event([{ nativeEvent: { state: new Animated.Value(-1) } }])}>
  <View style={{width: 50, height: 50, backgroundColor: 'red'}} />
<RawButton>
```

This error occurs when tapping:

Expected `onGestureHandlerStateChange` listener to be a function, instead got a value of `object` type.

This is the case whether using `react-native`'s `Animated` or using `react-native-reanimated`.

## Changes
Exporting PuteNativeButton allows to preparing Animated component on users' own. `createNativeWrapper` makes a wrapper for this component and only if it's an animated component, we could use the code above. 

So now it will be possible to do it like:

```jsx
import { createNativeWrapper, PureNativeButton } from 'react-native-gesture-handler';
const { event, Value, createAnimatedComponent } = Animated;
const AnimatedRawButton = createNativeWrapper(
  createAnimatedComponent(PureNativeButton),
  {
    shouldActivateOnStart: false,
  }
);
export default class App extends React.Component {
  constructor(props) {
    super(props);
    const state = new Value();
    this._onGestureEvent = event([
      {
        nativeEvent: { state },
      },
    ]);
  }
  render() {
    return (
      <AnimatedRawButton onHandlerStateChange={this._onGestureEvent}/>
    );
  }
}
```

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -624,5 +624,6 @@ export {
   /* Other */
   FlatListWithGHScroll as FlatList,
   gestureHandlerRootHOC,
+  GestureHandlerButton as PureNativeButton,
   Directions,
 };
--- a/docs/component-buttons.md
+++ b/docs/component-buttons.md
@@ -109,7 +109,7 @@ It should be used if you wish to handle non-crucial actions and supportive behav
 
 ### `PureNativeButton`
 
-Use `PureNativeButton` for accessing native Component used for build more complex buttons listed above.
+Use a `PureNativeButton` for accessing the native Component used for build more complex buttons listed above.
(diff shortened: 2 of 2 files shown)


osdnk:
Fixes  #358

brentvatne:
looks great! 🆗

osdnk:
Can you approve @brentvatne?

PR #533, opened 2019-03-20
Change style prop-type to any

scottmas:
The style prop can actually be pretty much anything: an object, a number, a falsey value, or an array containing any mix of the preceding types. We could write up a complex type check for that, but it seems overkill. And prop-types are being removed from react-native core, so I'm not sure we even need these prop checks. 

If you feel strongly about a complex type check, let me know and I'll see what I can do.


osdnk:
Thanks!

PR #547, opened 2019-04-01
[Pinch] Only update focal point when multiple fingers present

### LINKED ISSUE #546: Pinch focalX/focalY in state change callback don't match Animated values
I am working with `PinchGestureHandler` and using `focalX`/`focalY` variables to calculate `translateX`/`translateY` values. Everything works fine during the gesture.

When the gesture finishes, I have set up `curX` and `curY` variables, to which I offload the current values of `translateX`/`translateY` (derived from the focal coordinates). However, I have noticed that the `focalX` and `focalY` results reported to `PinchGestureHandler`'s `onHandlerStateChange` do not match the values reported to the `Animated.Value`s passed in through `onGestureEvent`.

~~On iOS this issue does not occur - the values match up.~~ On Android however, because the values do not match up I end up with a jerky effect when the gesture resolves.

I was able to compare the values by calling `__getValue()` on the `Animated.Value`s in `onHandlerStateChange`. ~~On iOS all the values line up~~, but on Android only the scale values match - the `focalX` and `focalY` values do not.

Is this a known issue? Any ideas as to what might be causing it?

Ashoat:
Fixes #546. Android's [ScaleGestureDetector](https://developer.android.com/reference/android/view/ScaleGestureDetector.html#getFocusX()) says "the focal point is between each of the pointers forming the gesture". However, that means that at the end of the gesture, when the first finger is removed from the screen, the focal point will suddenly be redefined as the location of the second finger, rather than being between the two fingers.

This PR prevents the focal point from being updated when only one finger is present.

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.java
@@ -14,6 +14,9 @@ public class PinchGestureHandler extends GestureHandler<PinchGestureHandler> {
   private float mStartingSpan;
   private float mSpanSlop;
 
+  private float focalX = Float.NaN;
+  private float focalY = Float.NaN;
+
   private ScaleGestureDetector.OnScaleGestureListener mGestureListener =
           new ScaleGestureDetector.OnScaleGestureListener() {
 
@@ -51,19 +54,14 @@ public class PinchGestureHandler extends GestureHandler<PinchGestureHandler> {
 
   @Override
   protected void onHandle(MotionEvent event) {
-    if (getState() == STATE_UNDETERMINED) {
+    int state = getState();
+    if (state == STATE_UNDETERMINED) {
       Context context = getView().getContext();
       mLastVelocity = 0f;
       mLastScaleFactor = 1f;
       mScaleGestureDetector = new ScaleGestureDetector(context, mGestureListener);
       ViewConfiguration configuration = ViewConfiguration.get(context);
       mSpanSlop = configuration.getScaledTouchSlop();
-
-      begin();
-    }
-
-    if (mScaleGestureDetector != null) {
-      mScaleGestureDetector.onTouchEvent(event);
     }
 
     int activePointers = event.getPointerCount();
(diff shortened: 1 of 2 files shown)


Ashoat:
Turns out iOS actually had the same issue, I just didn't notice until I tested on a real device instead of a simulator. On a real device one of the fingers gets removed first, and at that point the focal coordinates jerk to an incorrect value.

I've updated this PR to fix the issue on both platforms.

osdnk:
Hi, @Ashoat 
I want @adamczyk777 to take look on it 😄

adamczyk777:
Hey @Ashoat. As far as I can see it is a natural behavior of native gesture recognizers on both Android and iOS to update the focal point to the position of the pointer, when only one is present on the screen, by that I mean these rapid changes that you described in an issue before. The thing is, shouldn't we consider this library a wrapper on native gesture recognizers, and so on the source of truth and leave it as it is? In the events, there is a property of how many pointers are on the screen so it is most likely possible to make it suitable for your use case. What I am afraid of is that when someone has an edge case and actually needs the focal point where it actually is, then there is no way that person is able to get it without idk another handler for Pan that is not functioning just to take the location of that single finger. Let me know what you think because I might be wrong 😃

Ashoat:
You're right that it's the natural behavior of the native gesture recognizers, and this library is just reflecting that.

When I first used `PinchGestureHandler` I tried to model my use on the example in the [docs](https://kmagiera.github.io/react-native-gesture-handler/docs/handler-pinch.html). The way that example works is that the pinch scale gets "offloaded" from the native event value to the local value in the `onHandlerStateChange` callback.

That sort of strategy will lead to experiencing this sort of bug. As it stands, the native pinch focal values just can't be trusted when the state isn't `ACTIVE`. On the other hand, the native pinch scale values can be trusted. By implicitly trusting the pinch value in `onHandlerStateChange`, the example will probably lead to a lot of folks assuming they can trust the focal values too.

I've since worked around it by modeling my code more on the [`ImagePreview`](https://github.com/kmagiera/react-native-reanimated/tree/master/Example/imageViewer) example in `react-native-reanimated`. The idea is do the "offload" continuously rather than all at the end. For more complicated calculations (like the focal point -> translation calculations), this involves doing a bit of calculus, but it's entirely doable, and in my case led to more elegant code.

I'm not entirely sure if this sort of "continuous" approach is possible without `react-native-reanimated`. But I imagine most people who are using `react-native-gesture-handler` are probably also using `react-native-reanimated`.

Ashoat:
> What I am afraid of is that when someone has an edge case and actually needs the focal point where it actually is

I frankly don't think this will happen. The concept of a "pinch focal point" isn't really defined when a pinch isn't occurring. At that point you're just tracking where the finger on screen is, which would be handled by `PanGestureHandler`.

That said, this PR admittedly introduces some complexity. My view is that either the examples should be changed to avoid the perception that animated value can be trusted when the state isn't `ACTIVE`, or the animated values should all be "trustable" regardless of state. But that's just my 2¢. With that said, feel free to close this PR if you'd rather not merge :)

adamczyk777:
@Ashoat Thank you so much for such a detailed response. We will think it through shortly.

adamczyk777:
Hi @Ashoat. After spending some time thinking about this case we decided to close the issue as we want to keep the native behavior of gesture responders. Apart from that, we are aware of some inconsistencies between Android and iOS in pinch gesture handler, which I hope to address in the near future. Anyway, thank you so much for your contribution it is highly appreciated 🏅

PR #559, opened 2019-04-12
Add to and improve docs

naftalibeder:


docs/handlers.md
```diff
@@ -54,13 +56,13 @@ Library exports a `State` object that provides a number of constants used to exp
...
 
- - `shouldActivateOnStart`
+ - `shouldActivateOnStart` (**Android only**)
  - `disallowInterruption`
 
 #### `LongPressGestureHandler` extra properties
 
  - `minDurationMs`
- - `maxDist`
+ - `maxDist` (**on Android, may need to set to a very large number, e.g. 5000, to prevent immediate cancellation**)
```
osdnk: is it only Android? 😮

docs/handlers.md
```diff
@@ -54,13 +56,13 @@ Library exports a `State` object that provides a number of constants used to exp
 
 #### `NativeViewGestureHandler` extra properties
 
- - `shouldActivateOnStart`
+ - `shouldActivateOnStart` (**Android only**)
```
osdnk: If yes, I think it's rather a bug in the code, not in docs 😃

docs/handlers.md
naftalibeder: As far as I could tell, this only happens on Android. I was not experiencing it on iOS - the default `maxDist` value worked fine.

docs/handlers.md
naftalibeder: I sort of agree - it's not technically a bug in the code. Rather, the property in the Android codebase is simply not linked up to anything. So it's more like it's unimplemented.

docs/handlers.md
osdnk: I made a try with maxDist=150 and got expected result. Maybe you have it wrapped with ScrollView?
https://streamable.com/2bj0r

jakub-gonet:
@naftalibeder, what's the status of this PR?

naftalibeder:
@jakub-gonet This can probably just be closed. I opened this a while ago, and a lot has changed in the documentation since then. Looking through the current docs, I still do feel like `NativeViewGestureHandler` doesn't get the prominent explanation that it deserves, given its utility, but at least it's described - in the previous version, the link to its page was simply broken.

jakub-gonet:
Got it, thanks for the contribution anyway, I appreciate it.

PR #561, opened 2019-04-13
Fix TouchableNativeFeedback background propTypes

gmaclennan:
Currently the propType for background is `string`. It should be `object`.


osdnk:
:shipit:

Issue #468, opened 2019-02-16
Buttons and accessibility on Android


satya164:
I don't seem to be able to interact with gesture handler buttons using the accessibility tools on Android.

When you enable TalkBack on Android, there are 2 issues with the gesture handler's button:
- The `accessibilityRole` doesn't seem to be applied. talkback doesn't announce it as a button
- It's not possible to interact with the button via double tap

In this snack, I added a button with `TouchableNativeFeedback` where these two work as I expect, and button from `GestureHandler` where they don't work:

https://snack.expo.io/@satya164/gesture-handler-button-accessibility

satya164:
Fix:

```js
<RectButton
  onPress={() => alert('Clicked RectButton')}>
  <View accessible accessibilityRole="button">
    <Text>RectButton</Text>
  </View>
</RectButton>
```

Thanks @osdnk

osdnk:
Thanks @adamczyk777

kkafar:
Closing as following the docs should resolve the problem

PR #565, opened 2019-04-15
Add a note about buttons accessibility to docs

adamczyk777:
Just a note on how to make buttons accessible for now.

Related to issue #468


docs/component-buttons.md
```diff
@@ -16,6 +16,24 @@ Currently Gesture handler library exposes three components that render native to
...
+    </RectButton>);
+  // Accessible:
+  const AccessibleButton = () => (
+    <RectButton onPress={this._onPress}>
+      <View accessible>
+        <Text>Bar</Text>
+      </View>
+    </RectButton>);
+```
+If your children are already wrapped in some kind of `View` just add `accessible` prop to that View.
```
osdnk: I think it's not needed

satya164:
Maybe this should be built-in? The button is a common component, so accessibility should be baked in. Also, many people don't bother to add accessibility related props.

osdnk:
@satya164 
Maybe, but I wanna omit wrapping and cloning.

adamczyk777:
Updated the docs notice. It affects both iOS and Android.

PR #566, opened 2019-04-16
Add warning when accessibility is not explicitly set

jkadamczyk:
# Motivation
In order to make Buttons from React Native Gesture handler accessible one is required to decorate the button's child with `accessible` prop. I wanted to make people aware that their buttons are not accessible to people with disabilities.

# Changes
I added `console.warn` that gets triggered when you don't explicitly pass accessible prop to children of a component wrapped by `NativeViewGestureHandler`. It may be `RawButton`, `BaseButton` or `RectButton`, but also a completely new component that does not exist yet. Checking for accessibility is opt-in. You have to pass `checkAccessibility: true` in config object for `createNativeWrapper` function.

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -363,7 +363,7 @@ function createNativeWrapper(Component, config = {}) {
         });
         if (!accessibilitySetProperly) {
           console.warn(
-            'Content of BaseButton is not accesible. Consider adding accesible prop. If you dont wan’t to make it accesible, get rid of this warning by setting accesible prop explicitly to false.'
+            "Wrapped component does not support accessibility features. Consider adding accessible prop. If you don't want to make it accessible, get rid of this warning by setting accessible prop explicitly to false."
           );
         }
       }
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -347,11 +347,31 @@ const NATIVE_WRAPPER_PROPS_FILTER = {
   onGestureHandlerStateChange: PropTypes.func,
 };
 
-function createNativeWrapper(Component, config = {}) {
+function createNativeWrapper(
+  Component,
+  config = {},
+  checkAccessibility = false
+) {
   class ComponentWrapper extends React.Component {
     static propTypes = {
       ...Component.propTypes,
     };
+
+    componentDidMount() {
+      if (checkAccessibility) {
+        let accessibilitySetProperly = false;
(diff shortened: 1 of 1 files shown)


GestureHandler.js
```diff
@@ -444,6 +444,20 @@ class BaseButton extends React.Component {
     this._lastActive = false;
   }
 
+  componentDidMount() {
```
osdnk: Imho we should do it only with ‘RawButton’. I see no need for displaying this warning for every component which can be wrapped with ‘NativeViewGestureHandler’

GestureHandler.js
```diff
@@ -444,6 +444,20 @@ class BaseButton extends React.Component {
...
+    let accessibilitySetProperly = false;
+    React.Children.forEach(this.props.children, child => {
+      if (child.props.accessible !== undefined) {
+        accessibilitySetProperly = true;
+      }
+    });
+    if (!accessibilitySetProperly) {
+      console.warn(
+        'Content of BaseButton is not accesible. Consider adding accesible prop. If you dont wan’t to make it accesible, get rid of this warning by setting accesible prop explicitly to false.'
```
osdnk:
```suggestion
        'Content of BaseButton is not accessible. Consider adding accessible prop. If you don't want to make it accessible, get rid of this warning by setting accessible prop explicitly to false.'
```

GestureHandler.js
```diff
@@ -347,12 +347,32 @@ const NATIVE_WRAPPER_PROPS_FILTER = {
   onGestureHandlerStateChange: PropTypes.func,
 };
 
-function createNativeWrapper(Component, config = {}) {
+function createNativeWrapper(
+  Component,
+  config = {},
+  checkAccessibility = false
```
osdnk: Can we move it to `config`?

jakub-gonet:
Linked #565

jakub-gonet:
RN uses `accessible={true}` for all touchables so I think we should rather mark all handlers, buttons, and swipeables as accessible than issue a warning if it's missing. This way developers won't have to remember about accessibility when using RNGH and it removes some boilerplate when writing components.

jkadamczyk:
Closing this one, that's definitely not the optimal solution. Guess it was implemented as a feel-good solution so that we address the issue in any way. It should be implemented like @jakub-gonet describes it.

PR #570, opened 2019-04-17
* check for null in isConfigParam

mizerlou:
in `isConfigParam` if param is `null` then the `'__isNative' in param` check blows up.


mizerlou:
fixes #571

mizerlou:
yeah, if i `yarn add https://github.com/kmagiera/react-native-gesture-handler.git#master`, then my problem goes away.  sorry for the noise.

osdnk:
Whop whop 
Thanks, @mizerlou 
Today we were faster 🥳

Issue #569, opened 2019-04-17
hitSlop with right and width properties are not respected on Android


satya164:
When a `hitSlop` of value `{ left: 0, width: 48 }` is supplied, the result is expected, 48dpi on the left activates a gesture. But if a value like `{ right: 0, width: 48 }`, it only works on iOS, i.e. 48dpi on the right activates the gesture on iOS, but on Android, it behaves like it's not specified.

The `DrawerLayout` component in the repo suffers from this bug when `drawePosition` is `right`.

For better understanding, here is a repro: https://snack.expo.io/@satya164/gesture-handler-hitslop-bug

cc @osdnk

adamczyk777:
Hi @satya164 I will be addressing this issue in the near future I guess 😄

adamczyk777:
The PR addressing this issue has been merged, allow me to close this issue 😌

PR #572, opened 2019-04-18
FIX hitSlop right and left properties not being respected on android

adamczyk777:
Relates to #569 

Incorrect values were taken into consideration in function which determines whether the pointer is within bounds. We were using hitSlop padding values set through props instead of calculated absolute bounds (left, right, bottom, top).


adamczyk777:
Sorry for the slight delay 😃

PR #585, opened 2019-04-26
Hooks hooks hooks

osdnk:
is it dope? 🤔


osdnk:
Does someone have some initial thoughts? 
@kmagiera @adamczyk777

osdnk:
Try: https://reactjs.org/docs/hooks-reference.html#usememo

PR #586, opened 2019-04-26
fix: drawer opens when dragging from the middle in RTL

alikazemkhanloo:
When in RTL mode dragging from the center of the screen would open the drawer.
I guess this only happens in RTL. This PR fixes that.

### DIFF
--- a/DrawerLayout.js
+++ b/DrawerLayout.js
@@ -293,7 +293,7 @@ export default class DrawerLayout extends Component<PropType, StateType> {
     // activate when gesture happens not further than SLOP away from the edge
     const hitSlop = fromLeft
       ? { left: 0, width: showing ? undefined : edgeWidth }
-      : { right: 0, width: showing ? undefined : edgeWidth };
+      : { right: I18nManager.isRTL ? this.state.containerWidth - edgeWidth : 0, width: showing ? undefined : edgeWidth };
     this._panGestureHandler.current &&
       this._panGestureHandler.current.setNativeProps({
         hitSlop,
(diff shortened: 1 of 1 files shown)


adamczyk777:
Hi, could you provide a repro, or tell me if it is present in Example app? I tried every DrawerLayout in RNGH Example app with RTL and I can't experience the issue of drawer opening when dragging from the center of the screen 😞

alikazemkhanloo:
Hi,
I'm using react-navigation3.
I created [this snack](https://snack.expo.io/@alikazemkhanloo/cmVwcm) and realized this happens only on Android, and not just RTL.
should this fix apply to the Android platform instead of RTL?

seriiix:
> Hi,
> I'm using react-navigation3.
> I created [this snack](https://snack.expo.io/@alikazemkhanloo/cmVwcm) and realized this happens only on Android, and not just RTL.
> should this fix apply to the Android platform instead of RTL?

Thank you! I was going to create a snack. It seems you are always one step ahead of me 😄. I also noticed that when you make drawer RTL, dragging from left-to-right opens drawer. This issue is also present in the [snack](https://snack.expo.io/@alikazemkhanloo/cmVwcm) you made.

alikazemkhanloo:
here is my situation:
The snack I made has this issue ( on the web simulator and in my phone's expo).
But my own app is working fine without this PR.
All versions are the same.
:|

jakub-gonet:
Hi, what's the status of this? If you want this merged please provide repro in the description which ideally uses only RNGH.

jakub-gonet:
Closing due to no response, if you want to continue working on this please reply.

alikazemkhanloo:
Hi @jakub-gonet, Sorry for not responding in-time.
This issue is not present anymore in my project.

jakub-gonet:
All right, thanks for the contribution anyway.

PR #589, opened 2019-04-27
Migrate Android Support Libraries To AndroidX

aarongrider:
This is a new PR following: https://github.com/kmagiera/react-native-gesture-handler/pull/579

Currently I am using 2 forked dependencies to get the Android tests to pass. We will probably want to wait till our dependencies take the AndroidX upgrade and install directly from npm before merging this branch.

https://github.com/react-native-community/react-native-slider/pull/56
https://github.com/react-native-community/react-native-viewpager/pull/22

### DIFF
--- a/Example/android/build.gradle
+++ b/Example/android/build.gradle
@@ -6,7 +6,6 @@ buildscript {
         minSdkVersion = 16
         compileSdkVersion = 28
         targetSdkVersion = 28
-        supportLibVersion = "1.0.2"
     }
     repositories {
         google()
--- a/Example/android/build.gradle
+++ b/Example/android/build.gradle
@@ -2,18 +2,18 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "27.0.3"
+        buildToolsVersion = "28.0.3"
         minSdkVersion = 16
         compileSdkVersion = 28
-        targetSdkVersion = 26
-        supportLibVersion = "27.1.1"
+        targetSdkVersion = 28
+        supportLibVersion = "1.0.2"
     }
     repositories {
         google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.4'
+        classpath 'com.android.tools.build:gradle:3.2.1'
     }
 }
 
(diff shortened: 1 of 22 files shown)


osdnk:
Supercool. Is there anything we should be afraid of merging it? Are we going drop support for some versions with it?

aarongrider:
As far as I understand AndroidX libraries are not compatible with non AndroidX projects, so all projects that consume this library must also be migrated to AndroidX. 

I wanted to leave it to the maintainers to decide if this is something they would want to do. Unfortunately, the reverse is also true, any Android projects that have migrated to AndroidX, cannot use this library without these changes.

Do we have any idea how other RN libraries are handling this situation?

aarongrider:
It looks like Flutter has depreciated support library support in favor of AndroidX and has included instructions on how to upgrade your project, along with a list of the last plugin versions to support the legacy support libraries: https://flutter.dev/docs/development/packages-and-plugins/androidx-compatibility

This might be a good approach?

aarongrider:
There is also a React Native AndroidX migration plan here:
https://github.com/facebook/react-native/issues/23112

osdnk:
@aarongrider Please apply changes in your PR to Slider

aarongrider:
Should we just wait for RN 60?

Example/android/build.gradle
```diff
@@ -2,18 +2,18 @@
...
 buildscript {
     ext {
-        buildToolsVersion = "27.0.3"
+        buildToolsVersion = "28.0.3"
         minSdkVersion = 16
         compileSdkVersion = 28
-        targetSdkVersion = 26
-        supportLibVersion = "27.1.1"
+        targetSdkVersion = 28
+        supportLibVersion = "1.0.2"
```
rozPierog: There is no use for this in AndroidX. _Almost_ every AndroidX library has different versioning not like in Support Library where everything was chained together.
```suggestion

```

kmagiera:
Hey, I'd love to finally get that androidX migration in, however this PR has too many conflicts I am not sure how to resolve. Do you mind updating it?

se-bastiaan:
Of all the AndroidX PRs this seems to be the most complete one. It updates the example app and all dependencies and keeps the CI intact.

Instead of using forked versions of the viewpager and slider for the example app I would suggest using https://github.com/mikehardy/jetifier instead though and update the react-native version to 0.60+.

osdnk:
we already merged another PR switching to AndroidX

PR #600, opened 2019-05-09
Fix TouchableHighlight initial 'underlayColor' value

paolospag:
This PR fixes unexpected initial underlayColor using RNGH `TouchableHighlight`.


osdnk:
Thanks!

PR #603, opened 2019-05-12
Don't fail PanGestureHandler on tap

osdnk:
@james-macindoe
It's interesting. could you tell sth more?


james-macindoe:
Yeah so this is a fix for https://github.com/osdnk/react-native-reanimated-bottom-sheet/issues/15

If there's a Touchable inside a PanGestureHandler, when you press on the Touchable, the PanGestureHandler will begin() but not activate() (I think because the dy of the pan is too small). The PanGestureHandler then receives the ACTION_UP, which causes it to fail() at the line this PR changes. I'm assuming this is a valid use case that it shouldn't fail() on.

PR #616, opened 2019-05-29
[Typescript] Adds argument to FlatList

localnet:
FlatList: Adds argument ItemT, as on react-native original definition


osdnk:
Thanks!

PR #617, opened 2019-05-29
Lazy import RN components

kmagiera:
This PR changes the way we export gesture-handler-wrapped components. Instead of importing them from RN we rely on lazy-load mechanism (that is also utilize in RN core) to only pull in core components that we need. On top of that I added memoization mechanism such that for each component only one wrapper component is created.

This change makes GH more future proof as some core components (e.g. AndroidToolbar) are being removed from react-native. This will allow consumers to keep using these so long they rely on RN version that have them included (full backward compatibility). And also allow for consumers relying on new RN versions to wrap components from react-native-community github in case they want to continue using these.

### DIFF
--- a/GestureHandler.js
+++ b/GestureHandler.js
@@ -1,15 +1,10 @@
 import React from 'react';
 import {
   Animated,
-  ScrollView,
-  Switch,
-  TextInput,
-  ToolbarAndroid,
-  DrawerLayoutAndroid,
   StyleSheet,
-  FlatList,
   Platform,
   processColor,
+  default as ReactNative,
 } from 'react-native';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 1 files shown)


EvanBacon:
This breaks tree-shaking on web, and with strict configs like expo, it breaks altogether when you use import/export and module.exports.

GestureHandler.js
```diff
@@ -1,15 +1,10 @@
...
-  ScrollView,
-  Switch,
-  TextInput,
-  ToolbarAndroid,
-  DrawerLayoutAndroid,
   StyleSheet,
-  FlatList,
   Platform,
   processColor,
+  default as ReactNative,
```
EvanBacon: This will include everything in React Native which means things like react-art which aren't used by anyone will get bundled.

PR #655, opened 2019-06-21
Use UNSAFE_componentWillUpdate in Swipable.js

janicduplessis:
To silence the react warning, ideally we'd migrate to new React apis.

### DIFF
--- a/Swipeable.js
+++ b/Swipeable.js
@@ -88,7 +88,7 @@ export default class Swipeable extends Component<PropType, StateType> {
     );
   }
 
-  componentWillUpdate(props: PropType, state: StateType) {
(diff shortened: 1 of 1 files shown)


osdnk:
Thank you, @janicduplessis!
Can't we use `compenentDidUpdate`? 😃

janicduplessis:
@osdnk Sadly no since `_updateAnimatedEvent` sets a bunch of instance vars that are used in the subsequent render. If moved to `compenentDidUpdate` the wrong values would be used.

I cool way to fix it would be to use hooks. Otherwise we'd need `getDerivedStateFromProps` but that would be a lot more involved since we'd need to save prev props in state to diff.

PR #658, opened 2019-06-22
Remove Math.sign polyfill

janicduplessis:
RN only supports iOS 9+ now so we can remove this according to the comment.


waters4200 (commented):
4AEE18F83AFDEB23

PR #679, opened 2019-07-10
Disable the background accessibility when the drawer opens and overlaps (Android)

mrjoema:
Previously, the screen / components behind the drawer are still focusable because we didn't add the "importantForAccessibility" (only Android) to disable the accessibility service depending if the drawer opens


mrjoema:
https://facebook.github.io/react-native/docs/accessibility 
Search "importantForAccessibility" in the link above for more details

PR #694, opened 2019-07-18
[Typescript] Adds rippleColor to RectButtonProperties

rszalski:
This fixes the following TS error:

```
Type '{ children: Element; style: any; underlayColor: string; rippleColor: string; onPress: () => void; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<RectButton> & Readonly<{ children?: ReactNode; }> & Readonly<RectButtonProperties>'.
  Property 'rippleColor' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<RectButton> & Readonly<{ children?: ReactNode; }> & Readonly<RectButtonProperties>'.ts(2322)
```

### DIFF
--- a/react-native-gesture-handler.d.ts
+++ b/react-native-gesture-handler.d.ts
@@ -362,6 +362,7 @@ declare module 'react-native-gesture-handler' {
     onPress?: (pointerInside: boolean) => void;
     onActiveStateChange?: (active: boolean) => void;
     style?: StyleProp<ViewStyle>;
+    rippleColor?: string;
   }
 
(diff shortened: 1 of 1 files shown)


Andarius:
Shouldn't it be added to `BaseButton` instead ?  https://kmagiera.github.io/react-native-gesture-handler/docs/component-buttons.html#basebutton

rszalski:
@Andarius You are correct, thank you. I somehow looked at the wrong button in the docs 😅 

I'll re-test and fix this.

PR #708, opened 2019-07-26
Update RotationHandler docs

### LINKED ISSUE #706: Documentation for RotationGestureHandler says you can access focalX and focalY when really its anchorX and anchorY
~~Hello, I've been attempting to write an app which has an image that you can zoom in on, rotate and pan. Adjusting zooming to be around the focal point was fairly easy, however, while writing the logic for adjusting the image during rotation (to allow rotating around the focal point) I noticed that the focalX and focalY values returned by the rotation handler are always 0. To confirm this i printed the values via a call() method in the handlers. for rotation, I always saw 0 yet for zooming i saw correct values.~~

Turns out the documentation is just wrong for the RotationGestureHandler, after poking around the code, to access the focal points for rotation you access anchorX and anchorY rather than focalX and focalY. Will make a pr to fix this

damiansf:
Docs incorrectly had focalX and focalY as properties of a rotation handler. They are called anchorX and anchorY in the actual handler

fixes #706

### DIFF
--- a/docs/handler-rotation.md
+++ b/docs/handler-rotation.md
@@ -27,11 +27,11 @@ The rotation of the gesture in radians.
 Velocity of the pan gesture the current moment. The value is expressed in point units per second.
 
 ---
-### `focalX`
+### `anchorX`
 Position expressed in points along X axis of center anchor point of gesture 
 
 ---
-### `focalY`
+### `anchorY`
 Position expressed in points along Y axis of center anchor point of gesture  
 
 ## Example


wcandillon:
I can only push for this to be merged.

damiansf:
@osdnk @jakub-gonet 

just thought id tag you two since I've seen you merge PR's in the past. Not a high priority change but it's been sitting around for a while, hoping to get it merged.

damiansf:
@kmagiera

jakub-gonet:
Sorry for the late response, will check it out in the moment.

jakub-gonet:
Yup, that's correct, many thanks for the PR 🎉

PR #734, opened 2019-08-21
use AndroidX

dulmandakh:
This PR makes Gesture Handler compatible with AndroidX, by using Nullable annotation and Pools from AndroidX. But also removes **buildToolsVersion** as it's no longer required by Android Gradle Plugin since version 3.1.0, see https://developer.android.com/studio/releases/gradle-plugin#3-1-0, and bumped compileSdkVersion, targetSdkVersion to 28.

Upgraded **Example** app, and had to bump Android Gradle Version, thus Gradle wrapper version. Also had to run **jetify** because some dependencies are not AndroidX ready.


dulmandakh:
Android CI is green 👍

dulmandakh:
@kmagiera CI is green, and good to be merged 😎

kmagiera:
Awesome job @dulmandakh – thanks so much for your help!

bananer:
Hi, I'm wondering why `android/lib/build.gradle` still points to `com.android.support:appcompat-v7` depdency. Should this not have been changed to `androidx.appcompat:appcompat`? It might be jetifier fixing it for builds automatically...

https://github.com/software-mansion/react-native-gesture-handler/blob/06c44568e46e2bde4aacf443dcbc1aa161b00e47/android/lib/build.gradle#L27

PR #752, opened 2019-09-06
Swipeable ios default thresholds

shufflingB:
Hi,

Small(ish) backward compatible changes that adds the ability to have the Swipeable component animate and trigger default action menu items. 

It's been added in order to enable the component to achieve a similar default action menu item animation and triggering behaviour to that seen in many of ios list based apps e.g. mail where on LHS  default triggers the toggling of Read/Unread marking and RHS default Deletes the email. 

The principal changes are to Swipeable and are chiefly around triggering the default action on drag event release and in the determining their respective exposed open action menu widths. 

Of the two, the approach used for triggering of the default is straighforward. While the rational and design behind how the snapped open menu widths are determined for those action menus with default action triggering is a bit more involved, but is as follows:

1. When fully extended action menus with default actions can animate beyond the snapped open action width across the entire row.
1. This stops the previously used onrender approach for determining the snapped open menu width from being useful (returns row width).
1. Working around the problem by having the user specify the snapped open menu width allows the code changes to be localised and keeps things simple  

I've added docs and tweaked the example app to show it in operation.

Thanks for great component, docs and examples.

Jon

PS: First pull request, be gentle  ;-)

### DIFF
--- a/Example/package.json
+++ b/Example/package.json
@@ -20,6 +20,7 @@
     "react-art": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-native": "^0.60.0",
+    "react-native-responsive-screen": "^1.2.2",
     "react-native-vector-icons": "5.0.0",
     "react-native-web": "^0.10.0-alpha.3",
     "react-navigation": "^3.1"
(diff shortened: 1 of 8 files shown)


osdnk:
Are you sure it's correct? Scope of your description looks like a small fix and your PR has +4,114 −2,117

jakub-gonet:
I'm closing it due to no response. If you want it to be merged please respond.

PR #785, opened 2019-10-07
Updated drawer container style type

zachariahtimothy:
Drawer container style typings to not match code [here](https://github.com/kmagiera/react-native-gesture-handler/blob/master/DrawerLayout.js#L51). Since TS Ignore cannot be used within JSX thought I would fix.


zachariahtimothy:
Can this get checked and merged soon? I have an upcoming demo and there is a @ts-ignore on the screen because this definition is wrong.

PR #792, opened 2019-10-18
Fix for #320, "Unsupported top level event type "onGestureHandlerEvent""

### LINKED ISSUE #320: Unsupported top level event type "onGestureHandlerStateChange" dispatched
I am getting crash when I touch the PanGestureHandler.

```
return (
			<PanGestureHandler
				onGestureEvent={this.onGestureEvent}
			>
				<Animated.View style={[style, panStyle]} {...rest}>
					{children}
				</Animated.View>
			</PanGestureHandler>
		);
```

package.json
```
    "react-native-gesture-handler": "1.0.8",
    "react": "^16.5.0",
    "react-native": "^0.57.2",
```

android versions:
```
  compileSdkVersion 27
    buildToolsVersion '27.0.3'

    defaultConfig {
        minSdkVersion 16
        targetSdkVersion 27
  }

    implementation "com.android.support:appcompat-v7:27.1.1"

```

crash log
```
Unsupported top level event type "onGestureHandlerStateChange" dispatched
extractEvents
    ReactNativeRenderer-dev.js:2471:6
extractEvents
    ReactNativeRenderer-dev.js:1016:8
runExtractedEventsInBatch
    ReactNativeRenderer-dev.js:1070:4
<unknown>
    ReactNativeRenderer-dev.js:2713:6
batchedUpdates$1
    ReactNativeRenderer-dev.js:15311:14
batchedUpdates
    ReactNativeRenderer-dev.js:2616:31
_receiveRootNodeIDEvent
    ReactNativeRenderer-dev.js:2711:17
receiveEvent
    ReactNativeRenderer-dev.js:2731:26
__callFunction
    MessageQueue.js:349:47
<unknown>
    MessageQueue.js:106:26
__guard
    MessageQueue.js:297:10
callFunctionReturnFlushedQueue
    MessageQueue.js:105:17
```

related issue: https://github.com/expo/expo/issues/2067

paddlefish:
hook up one NativeEventEmitter to whitelist our event types

This fixes #320

### DIFF
--- a/createHandler.js
+++ b/createHandler.js
@@ -2,6 +2,7 @@ import React from 'react';
 import {
   findNodeHandle as findNodeHandleRN,
   NativeModules,
+  NativeEventEmitter,
   Touchable,
   Platform,
 } from 'react-native';
@@ -15,6 +16,20 @@ function findNodeHandle(node) {
 }
 
 const { UIManager = {} } = NativeModules;
+
+// We need to hook up one listener so that react-native will call
+// into our module and bless our events declared there in supportedEvents
+// of RNGestureHandlerModule.m. Otherwise when we start sending
+// events later, we might crash (seemingly only in release builds) with
+// an error like:
+//   Unsupported top level event type "onGestureHandlerEvent"
+// Fixes: https://github.com/kmagiera/react-native-gesture-handler/issues/320
+const unusedEmitter = new NativeEventEmitter(NativeModules.RNGestureHandlerModule)
+// It's not enough to create an emitter -- you must also add a listener because
+// react-native does this initialization lazily
+const subscription = unusedEmitter.addListener("onGestureHandlerEvent",function(next){console.warn(next)})
+// Now that our events are blessed by the system we can let go of the listener
+subscription.remove()
 
 const customGHEventsConfig = {
   onGestureHandlerEvent: { registrationName: 'onGestureHandlerEvent' },
--- a/createHandler.js
+++ b/createHandler.js
@@ -27,7 +27,7 @@ const { UIManager = {} } = NativeModules;
 const unusedEmitter = new NativeEventEmitter(NativeModules.RNGestureHandlerModule)
 // It's not enough to create an emitter -- you must also add a listener because
 // react-native does this initialization lazily
-const subscription = unusedEmitter.addListener("onGestureHandlerEvent",function(next){console.warn(next)})
+const subscription = unusedEmitter.addListener("onGestureHandlerEvent",function(next){})
 // Now that our events are blessed by the system we can let go of the listener
 subscription.remove()
 


solidfox:
I'm also seeing the same behavior for `onGestureHandlerStateChange`. Do we need to add a dummy listener for that as well?

mikehardy:
Hello everyone :-), honest question: has anyone done a pure test on this without the workaround of importing react-native-gesture-handler early, and can reliably crash, then apply this fix and never crash? Just curious if this has been tested rigorously.

If not, I could try that, and hopefully it could at least be provisionally merged until there was a good solution, as I'd like to remove the workaround in my own project and it would help new users I'm sure.

@paddlefish I assume this nailed the problem for you? @kmagiera @solidfox have you tried it for onGestureHandlerStateChange?

tomtargosz:
@mikehardy I gave this a shot by using this fix locally on react-native 0.61.3 and gesture handler 1.5.0 but replacing `onGestureHandlerEvent` with `onGestureHandlerStateChange` without the import workaround and I'm still still getting the same error. 😞

solidfox:
@mikehardy I also tried adding listeners for both events and that still saw the problem.

paddlefish:
I think there is a deeper problem for sure. I thought maybe this workaround would be a band aid... but perhaps just like real band aids they sometimes come off :-/

I did notice that the core react-native code has special case handling for gesture handler events.   Maybe there is a newer better way events are supposed to be sent that would mean that this workaround isn’t needed AND that react-native doesn’t need to have special cases either.

omnikron:
We are currently upgrading to React Native 0.61.3 and as part of that have upgraded react-native-gesture-handler 1.5.0 to fix https://github.com/facebook/react-native/issues/26626#issuecomment-547920759.  

I only started seeing this _'Unsupported top level event type "onGestureHandlerEvent"'_ crash **after** upgrading from react-native-gesture-handler 1.4.1 to 1.5.0.  

I can't offer any more repro steps than have already been mentioned in the various tickets but maybe this helps to narrow it down...?

kmagiera:
Hey @paddlefish – Thanks for working on this. 

Have you maybe tried your workaround when running on iOS in release mode? I think currently the main issue is with the release setup as otherwise the gesture handler events are whitelisted in RN core.

kmagiera:
I checked this PR and it appears not to be fixing the issue in release mode. Can you double check or confirm whether you tried release mode on iOS?

kmagiera:
I think I managed to find yet another workaround and submitted https://github.com/kmagiera/react-native-gesture-handler/pull/845

paddlefish:
> Hey @paddlefish – Thanks for working on this.
> 
> Have you maybe tried your workaround when running on iOS in release mode? I think currently the main issue is with the release setup as otherwise the gesture handler events are whitelisted in RN core.

I discovered it only in release builds (TestFlight). After my patch my TestFlight builds have been working. :shrug:

slorber:
Hey, for what it's worth, I can confirm I only see this issue in iOS release mode, and see it in an Expo ejected 36 app despite using 1.5.3 JS, while it's supposed to be fixed in 1.5.1 in https://github.com/software-mansion/react-native-gesture-handler/pull/845/files

I guess Expo might use 1.5.0 native so upgrading just the JS is not enough (but at least does not crash anyway)

kmagiera:
@slorber for versioning gesture-handler libs we use this scheme where patch numbers are only JS updates. This is specifically to make it possible for expo users to receive some updates that does not require native code. It also means that 1.5.1 is compatible with 1.5.0 native code. Also note that the fix from #845 did not include any native changes

slorber:
Ah I see, I thought I had seen some native in the PR when I looked.

Was already on 1.5.3 (with Expo SDK36), and still had the problem, and that was fixed by importing RNGH at the entry point in the app. My app is perhaps particular in that it's like uber driver + uber customer in one app, so I require("./uberDriverApp") only after knowing the user type to avoid initializing js unnecessary, that may be a reason.

mnzaki:
Thanks @omnikron! 
> Edit: very anecdotally, but moving `import "react-native-gesture-handler";` to the top of my `index.ios.js` and `index.android.js`files seems to have fixed it.

Can confirm, adding `import 'react-native-gesture-handler'` somewhere near the beginning of the bundle resolved the issue for me

RN 0.59
react-native-gesture-handler 1.3.0

jkadamczyk:
I'll close this PR as from what I understand, RNGH events are whitelisted in RN core.
If you still experience any issues of the kind described in this PR, we can reopen it or work on a new solution.

PR #795, opened 2019-10-21
Added mocks for pan and scroll gesture components

jdmg94:
Basically `react-native-tab-view` uses this components and without them testing is not possible

### DIFF
--- a/__mocks__/RNGestureHandlerModule.js
+++ b/__mocks__/RNGestureHandlerModule.js
@@ -1,4 +1,8 @@
+import { View } from 'react-native'
+
 export default {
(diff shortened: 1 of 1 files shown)


__mocks__/RNGestureHandlerModule.js
```diff
@@ -1,4 +1,8 @@
+import { View } from 'react-native'
+
 export default {
+  ScrollVew: View,
```
osdnk: Please use ScrollView from RN here

jdmg94: done

PR #796, opened 2019-10-21
Stop calling cancelTouches

osdnk:
`cancelTouches` beneath only calls  `cancel` of `touchHandler` of `contentView`. I just call it directly keeping the same logic.

### DIFF
--- a/ios/RNGestureHandlerManager.m
+++ b/ios/RNGestureHandlerManager.m
@@ -159,7 +159,9 @@
 
     UIView *parent = rootContentView.superview;
     if ([parent isKindOfClass:[RCTRootView class]]) {
-        [(RCTRootView*)parent cancelTouches];
+      // Previously we were using cancelTouches method from RCTRootView.
+      // Now it's deprecated and we call cancel directly on touchHandler of contentView
+      [(RCTTouchHandler *)[[parent valueForKey:@"contentView"] valueForKey:@"touchHandler"] cancel];
     }
 }
 


osdnk:
https://github.com/facebook/react-native/blob/aee88b6843cea63d6aa0b5879ad6ef9da4701846/React/Base/RCTRootView.m#L402

ios/RNGestureHandlerManager.m
```diff
@@ -159,7 +159,9 @@ - (void)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
 
     UIView *parent = rootContentView.superview;
     if ([parent isKindOfClass:[RCTRootView class]]) {
-        [(RCTRootView*)parent cancelTouches];
+      // Previously we were using cancelTouches method from RCTRootView.
+      // Now it's deprecated and we call cancel directly on touchHandler of contentView
+      [(RCTTouchHandler *)[[parent valueForKey:@"contentView"] valueForKey:@"touchHandler"] cancel];
```
kmagiera: I don't like that we use `valueForKey` here. IMO it is more likely that this code will break comparing to the use of deprecated method.

If we want to access `_contentView` can we maybe instead of looking for `RCTRootView` parent look for `RCTRootContentView`? I understand that _contentView is a direct subview of the rootview and that should work

kmagiera:
Check this one here https://github.com/kmagiera/react-native-screens/pull/191

brentvatne:
the approach from https://github.com/kmagiera/react-native-screens/pull/191 looks reasonable, we should get this ported over to rngh and released asap. we'll need to include it in the next expo release

osdnk:
Fixed on master

PR #799, opened 2019-10-22
[WEB] Fix x/y for PanGestureHandler onGestureEvent

awinograd:
https://kmagiera.github.io/react-native-gesture-handler/docs/handler-pan.html#x

> X/Y coordinate of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the view attached to the handler. Expressed in point units.

Not sure what the performance implications are of calling `getBoundingClientRect()` in terms of layout thrashing. It appears we already call it once in the event handler anyway (https://github.com/kmagiera/react-native-gesture-handler/blob/5713701f4359dc11f4fb7ad3e4d78100b49730ce/web/GestureHandler.js#L102) so I would guess it doesn't add extra overhead

### DIFF
--- a/web/DraggingGestureHandler.js
+++ b/web/DraggingGestureHandler.js
@@ -6,6 +6,7 @@ class DraggingGestureHandler extends GestureHandler {
   }
 
   transformNativeEvent({ deltaX, deltaY, velocityX, velocityY, center: { x, y } }) {
+    const rect = this.view.getBoundingClientRect(); 
     return {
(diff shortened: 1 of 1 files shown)


awinograd:
Anything I can do / provide to move this forward? cc @osdnk

osdnk:
I want @EvanBacon to review this :)

awinograd:
Thanks for the review!

PR #844, opened 2019-11-21
Remove excessive logging

mo22:
Hi,

as far as I can tell these log messages (initialise, tear down) are predictable and do not aid in debugging.

I'd like to remove these to not clutter the console.

I've removed the teardown warning also, as this happens frequently when using react-native's reload mechanism and as far as I know native modules are never dealloced in production builds.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.java
@@ -1,7 +1,6 @@
 package com.swmansion.gesturehandler.react;
 
 import android.os.SystemClock;
-import android.util.Log;
 import android.view.MotionEvent;
 import android.view.ViewGroup;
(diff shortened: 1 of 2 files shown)


osdnk:
I think I'd rather keep them here. Sometimes it's good to know if RNGH was correctly initialized

mo22:
> I think I'd rather keep them here. Sometimes it's good to know if RNGH was correctly initialized

How about a verbose setting that can be initialised from your AppDelegate / Activity before react-native is started? That could also include more detailed logging about what gestures in what state were detected on which view etc.

That way it would be a really helpful tool to hunt down bugs but wouldn't clutter the console.

osdnk:
> How about a verbose setting that can be initialised from your AppDelegate / Activity before react-native is started?

Sounds cool

jakub-gonet:
Hi @mo22, any update on this?

mo22:
No not really - I'll code this MR

PR #853, opened 2019-11-25
Update the readme for running the examples on ios

brobertsUPS:
Thanks for the project!

I had to do some extra steps to get the Example to run locally in iOS so I updated the readme and pod lockfile.


osdnk:
Thanks!

Issue #607, opened 2019-05-14
<Touchable*> styling is not compatible with React Native Touchables


radex:
RNGH Touchables have a different behavior from RN Touchables when styling — making them not drop-in compatible.

Take for example:

this is the behavior of RN TouchableOpacity (and of RNGH `BaseButton`):

<img width="503" alt="Captura de pantalla 2019-05-14 11 02 45" src="https://user-images.githubusercontent.com/183747/57686487-f8bf2280-7639-11e9-8fe8-fcb8cbf39c1a.png">

this is the behavior of RNGH touchables:

<img width="527" alt="Captura de pantalla 2019-05-14 11 03 01" src="https://user-images.githubusercontent.com/183747/57686489-f957b900-7639-11e9-8ca2-6f266ee66dcd.png">

the yellow box is a touchable with `flex: 1`.

The problem is that all RNGH Touchables render two wrapped views:

https://github.com/kmagiera/react-native-gesture-handler/blob/baf2ba72ea4fca22a11b507cffb4ee94f0126480/touchables/GenericTouchable.js#L258-L266

The styles are applied to the child view, so applying `flex: 1` doesn't do what one would expect.

I _think_ the only way to avoid this incompatibility is to render just one native view. This should also be a little more performant.

I'm willing to help with implementation (at least on JS+iOS side), but I would need some basic pointers as to how to make `RawButton` accept `View` props…

vinnyHero:
any update on this ? @kmagiera @brentvatne

njdancer:
@radex #854 is not exactly what you’ve proposed but does allow for a workaround.

In my use case, supplying '''containerStyle={{ flex: 1 }}''' to an RNGH Touchable along with my existing '''style''' prop, mirrors the native behaviour.

mjarraya:
If not fixed, this needs at minimum to be mentioned in the docs

lukebars:
Is there any progress on this one? Looks like crucial part of gesture handler's future, to be as consistent to react-native styling as possible.

j-piasecki:
`containerStyle` prop has been added to Gesture Handler's Touchables quite a while ago in https://github.com/software-mansion/react-native-gesture-handler/pull/854. I'll close the issue because using that prop will solve the problem of not being able to style the outer view, and there is a lot of open issues regarding styling of Touchables and buttons coming from the same root cause.

PR #854, opened 2019-11-26
Add option to style touchable container

njdancer:
RNGH Touchable components usually have a child view that is wrapped with BaseButton. Styles supplied to the Touchables are applied directly to the child view and in some circumstances a need arises to be able to style the container as well as the child. In my specific use case I'm running into the same issue as #607. That issue proposes removing the child view entirely which is probably preferable however this patch enables a fairly simple workaround by allowing additional styling on the container.

Regarding the type information, I wasn't aware of a way to only have this apply when the view is actually being wrapped, however I believe that the additional styling will just be ignored in those cases so I don't see any issues with this.

### DIFF
--- a/react-native-gesture-handler.d.ts
+++ b/react-native-gesture-handler.d.ts
@@ -403,21 +403,25 @@ declare module 'react-native-gesture-handler' {
     BorderlessButtonProperties
   > {}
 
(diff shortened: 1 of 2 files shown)


osdnk:
Thanks!

PR #860, opened 2019-12-03
Add missing LongPressGestureHandlerGestureEvent; fixes #723

### LINKED ISSUE #723: Typescript definitions file missing LongPressGestureHandlerGestureEvent
According to [LongPressGestureHandlerProperties ](https://github.com/kmagiera/react-native-gesture-handler/blob/master/react-native-gesture-handler.d.ts#L281). onGestureEvent has an GestureHandlerGestureEvent type. But `long press onGestureEvent` should has a `absoluteX` and `absoluteY` field. LongPressGestureHandlerGestureEvent should be implemented

lafiosca:
This is a simple addition/fix to the existing typings and does not touch any of the logic.

### DIFF
--- a/react-native-gesture-handler.d.ts
+++ b/react-native-gesture-handler.d.ts
@@ -129,6 +129,12 @@ declare module 'react-native-gesture-handler' {
     absoluteY: number;
   }
 
+  export interface LongPressGestureHandlerGestureEvent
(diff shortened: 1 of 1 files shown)


lafiosca:
I see the CI check has failed, but it doesn't appear to be related to the changes. I assume that #858 is about fixing this.

PR #872, opened 2019-12-09
Adding accessibilityLabel to NativeViewGestureHandlerProperties

BlairWoods:
Adding "accessibilityLabel" to buttons so that so they can be picked up by test automation frameworks such as WebdriverIO and Appium. More specifically, updating the RawButtonProperties interface to include the "accessibilityLabel" property:

export interface RawButtonProperties
extends NativeViewGestureHandlerProperties {
exclusive?: boolean;
testID?: string;
accessibilityLabel?: string;
}

This provides a fix to the issue: https://github.com/kmagiera/react-native-gesture-handler/issues/865

### DIFF
--- a/react-native-gesture-handler.d.ts
+++ b/react-native-gesture-handler.d.ts
@@ -372,6 +372,7 @@ declare module 'react-native-gesture-handler' {
     extends NativeViewGestureHandlerProperties {
     exclusive?: boolean;
     testID?: string;
+    accessibilityLabel?: string;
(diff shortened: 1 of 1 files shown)


kmagiera:
Thanks for adding this. I'm going to merge this one but I think long term it'd be better to migrate to rely on some more generic way of listing props that are coming from RN's View component. In general button exposes all the props View does, so this list is currently way to short

PR #896, opened 2019-12-30
Replaced HammerJS with EGJS/HammerJS

codymurphyjones:
I replaced all the imports for egjs and hammer.

I was running into an issue where my ssr wouldnt build properly for production.  The issue existed within hammerjs for the gesture handler, I had used the module swap for the testing process but next/build didnt work.

I figured it wasnt logical for it not to be in this repo and did the change myself.

Feel free to correct me if im wrong and we are choosing to not migrate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
   },
   "homepage": "https://github.com/software-mansion/react-native-gesture-handler#readme",
   "dependencies": {
-    "@egjs/hammerjs": "^2.0.8",
+    "hammerjs": "https://github.com/naver/hammer.js.git",
     "hoist-non-react-statics": "^2.3.1",
     "invariant": "^2.2.4",
     "prop-types": "^15.7.2"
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -63,7 +63,8 @@
   },
   "homepage": "https://github.com/software-mansion/react-native-gesture-handler#readme",
   "dependencies": {
-    "hammerjs": "^2.0.8",
+    "@egjs/hammerjs": "^2.0.17",
+    "hammerjs": "https://github.com/naver/hammer.js.git",
```
EvanBacon: does egjs require hammerjs to be installed as well? the docs seem to imply that only `@egjs/hammerjs` is required https://github.com/naver/hammer.js/#yarn

codymurphyjones: I do not believe hammerjs is still needed, for this actually. Just the replacement

PR #925, opened 2020-01-17
Export a component for gesture handler root view

satya164:
Currently the user always needs to add code to wrap the root view in their MainActivity.java on Android, or in case of libraries such as wix's native navigation, wrap each screen in `gestureHandlerRootHOC`.

This works well for apps, but for libraries which use gesture handler, it often increases the setup steps for the user. In case of React Navigation and other libraries I work on, we have found that people always miss this configuration step and open issues about their gestures not working. This wrapper will let library authors avoid this extra step for the user by wrapping their library code in `GestureHandlerRootView`, and will only need auto-linking to work.

**Overview**

- New `GestureHandlerRootView` component for libraries to use, takes care of nesting multiple root views with context API so only 1 of them is rendered and rest are replaced with `View`
- Change `gestureHandlerRootHOC` to use the `GestureHandlerRootView` component which also fixes a platform inconsistency
- Fix the native `RNGestureHandlerRootView` (Android) to make sure that nesting it inside `RNGestureHandlerEnabledRootView` doesn't crash it, which can happen if the user already configured their `Mainactivity.java`.

**Test plan**

Tested by wrapping one of the examples in the example app in `GestureHandlerRootView` and whole app in `gestureHandlerRootHOC`, both with and without the setup code in `MainActivity.java`

**Questions**

Should we remove the `MainActivity.java` step from the docs and only use the component/hoc? Arguably it's easier to setup than `MainActivity.java` because you don't have to touch native code.


osdnk:
Can you test it and "undraft"?

osdnk:
I'm supportive of these changes

docs/getting-started.md
```diff
@@ -157,6 +157,22 @@ You can check out [this example project](https://github.com/henrikra/nativeNavig
 
 Remember that you need to wrap each screen that you use in your app with `gestureHandlerRootHOC` as with native navigation libraries each screen maps to a separate root view. It will not be enough to wrap the main screen only.
 
+### For library authors
+
+If you're using gesture handler in your component library, you may want to wrap your library's code in the `GestureHandlerRootView` component. This will avoid extra configuration in `MainActivity.java` for the user.
```
kmagiera: We may want to mention here that if you use props like `shouldCancelWhenOutside`, `simultaneousHandlers` or `waitFor` in the handlers they need to be mounted under a single root component. So it is important that GH root view is as close to the actual root view as possible.

GestureHandlerRootView.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+import { View } from 'react-native';
+
+export default View;
```
kmagiera: can we just render children in this case? No need to introduce another view layer by default when root view isn't necessary

docs/getting-started.md
satya164: Thanks! Will update

GestureHandlerRootView.js
satya164: I think it won't be correct to not render a view. For example, if I'm using `GestureHandlerRootView` in my library/app, I could pass some styles such as `backgroundColor`, `flexDirection` etc. If we don't return a view here, those styles will only be applied on Android.

In practice, if I'm able to pass styles here directly and make it work on both platforms, I wouldn't have to add an extra view to do it.

satya164:
Thanks for the review. I'll test it this week and let you know.

kmagiera:
Also you be important to test this with expo. In expo we had this kind of special setup where the native GH root view was instantiated but not "initialized" to use gesture handler unless at leas one GH component is used.

satya164:
@kmagiera yes it does with MainAcitivity configured. I have tested it in the example app.

kmagiera:
Ah, I found the relevant code for the expo setup: https://github.com/software-mansion/react-native-gesture-handler/blob/b2f2e0d69d9fce7092ea94304d8feb80c8061302/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java#L604

kmagiera:
We may need to make sure that we call `RNGestureHandlerEnabledRootView#initialize` despite root container being in the hierarchy. Otherwise the container won't get "enabled" because it sees that the root view is from gesture-handler, but it may still not be initialized

satya164:
> Also you be important to test this with expo. In expo we had this kind of special setup where the native GH root view was instantiated but not "initialized" to use gesture handler unless at leas one GH component is used.

Right now I'm exporting a dummy `View` for `GestureHandlerRootView` with `android.expo.js` similar to iOS, so nothing should change for Expo.

PR #937, opened 2020-01-30
Make RNGH workable inside modals on Android

osdnk:
## Motivation
Modals are currently not working with RNGH, because handlers need to be located under RootView in the native hierarchy. 

## Changes

It's fixed it by wrapping content of modals with `gestureHandlerHOC`. However, it called for a few native changes. 
In all places we were using RootView ref, I made it more generic to accept any `ViewGroup`. In places where methods of `RootView` were accessed I added checks and castings. 

Also, modal wrapper (`ReactModalHostView.DialogRootViewGroup`) is not accessible from another package so I added the file to the package (`com.facebook.react.views.modal`) exporting important functionalities.

Added docs.

## Test
I verified it iterating through example app and checking if every example (including these with RN's `PanResponer`) are working correctly.

### DIFF
--- a/android/src/main/java/com/facebook/react/views/modal/RNGHModalUtils.java
+++ b/android/src/main/java/com/facebook/react/views/modal/RNGHModalUtils.java
@@ -1,0 +1,17 @@
+package com.facebook.react.views.modal;
+
+
+import android.view.MotionEvent;
+import android.view.ViewGroup;
+import android.view.ViewParent;
+
+public class RNGHModalUtils {
+    static public void dialogRootViewGroupOnChildStartedNativeGesture(ViewGroup modal, MotionEvent androidEvent) {
+        ((ReactModalHostView.DialogRootViewGroup) modal).onChildStartedNativeGesture(androidEvent);
+    }
+
+    static public boolean isDialogRootViewGroup(ViewParent modal) {
+        return modal instanceof ReactModalHostView.DialogRootViewGroup;
+    }
+
+}
--- a/android/src/main/java/com/facebook/react/views/modal/RNGHModalUtils.java
+++ b/android/src/main/java/com/facebook/react/views/modal/RNGHModalUtils.java
@@ -11,11 +11,11 @@ import android.view.ViewParent;
  */
 
 public class RNGHModalUtils {
-  static public void dialogRootViewGroupOnChildStartedNativeGesture(ViewGroup modal, MotionEvent androidEvent) {
+  public static void dialogRootViewGroupOnChildStartedNativeGesture(ViewGroup modal, MotionEvent androidEvent) {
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/facebook/react/views/modal/RNGHModalUtils.java
```diff
@@ -0,0 +1,21 @@
...
+import android.view.ViewParent;
+
+/**
+ * For handling gestures inside RNGH we need to have access to some methods of
+ * `ReactModalHostView.DialogRootViewGroup`. This class is not available outside
+ * package so this file exports important features.
+ */
+
+public class RNGHModalUtils {
+  static public void dialogRootViewGroupOnChildStartedNativeGesture(ViewGroup modal, MotionEvent androidEvent) {
```
adamczyk777: Could you please change `static public` to `public static`? It's more common to see the latter in Java.

android/src/main/java/com/facebook/react/views/modal/RNGHModalUtils.java
```diff
@@ -0,0 +1,21 @@
...
+    ((ReactModalHostView.DialogRootViewGroup) modal).onChildStartedNativeGesture(androidEvent);
+  }
+
+  static public boolean isDialogRootViewGroup(ViewParent modal) {
```
adamczyk777: Same as above

docs/getting-started.md
```diff
@@ -157,6 +157,31 @@ You can check out [this example project](https://github.com/henrikra/nativeNavig
...
+
+export default function Example() {
+  return (
+    <Modal animationType="slide" transparent={false}>
+      <ExampleWithHoc />
+    </Modal>
+  );
+}
+``` 
+
```
adamczyk777: Just a question, would it be possible to export a modal from RNGH that already handles gestures as it should on Android?

osdnk: Good point!

osdnk: ... but let's do it in follow-up

grahammendick:
@osdnk You don't need to use the inaccessible `DialogRootViewGroup` class.  You can use the public `RootView` interface instead. Both `ReactRootView` and `DialogRootViewGroup` implement it and it contains the method `onChildStartedNativeGesture` that you're after.  It would simplify the code and mean that you don't need to use the `com.facebook.react.views.modal` package.

docs/getting-started.md
Risbot: I try to use example, but it does not work :(

khacbac:
not work if Modal inside react-native-stack

jerearaujo03:
> not work if Modal inside react-native-stack

Same here with `react-navigation-5`

lexguy:
> > not work if Modal inside react-native-stack
> 
> Same here with `react-navigation-5`

same here width @react-navigtaion/stack 5.14.2 , too...

PR #951, opened 2020-02-06
Remove extra hammerjs code

EvanBacon:
Remove code missed in this PR https://github.com/software-mansion/react-native-gesture-handler/pull/896

### DIFF
--- a/.yarnclean
+++ b/.yarnclean
@@ -1,1 +1,0 @@
-hammerjs/tests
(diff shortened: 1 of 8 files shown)


samjt:
Could this please be merged and released?

Installation of react-navigation, for example, fails because https://github.com/naver/hammer.js.git cannot be resolved properly.

`error Couldn't find any versions for "hammerjs" that matches "naver-hammer.js-master.tar.gz-art-external"`

The only current solution is to roll back to a version that doesn't try to use a git repo, and it's a duplicate dependency anyway!

PR #953, opened 2020-02-08
[Web] Fix TapGestureHandler x/y coordinates being undefined

### LINKED ISSUE #943: [web] TapGestureHandler x/y coordinates are undefined
This is probably caused by `transformNativeEvent` inside of `DiscreteGestureHandler`. It currently has the following signature, but `x` and `y` are undefined.

```
transformNativeEvent({ x, y }) {
  return {
    absoluteX: x,
    absoluteY: y,
    x,
    y,
  };
}
```

Making it similar to `DraggingGestureHandler` fixes the problem. 

```
transformNativeEvent({ center: { x, y }}) {
```

Not sure about the implications though. Happy to submit a PR if this seems like the right solution.

latviancoder:
Resolves #943

### DIFF
--- a/web/DiscreteGestureHandler.js
+++ b/web/DiscreteGestureHandler.js
@@ -26,12 +26,14 @@ class DiscreteGestureHandler extends GestureHandler {
     );
   }
 
-  transformNativeEvent({ x, y }) {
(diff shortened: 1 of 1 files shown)


MrLoh:
@kmagiera could you make a new release?

PR #945, opened 2020-02-04
Fix GestureHandlerRootView type

tobiasbu:
The exported `GestureHandlerRootView` component in type definitions, is throwing the following error. Occurs at line [457](https://github.com/software-mansion/react-native-gesture-handler/blob/d9c7553c8d7112dcff9bd498975a5f4b0a1fa9bc/react-native-gesture-handler.d.ts#L457).
```
Cannot find name 'ReactComponentType'. ts(2304)
```
TypeScript projects can break due to an incorrect type annotation. This PR fixes that small detail.
Bug was introduced in 1.5.4 version.

PR #954, opened 2020-02-09
Update types definition to avoid using non-existing type

elyalvarado:
The types definition was using ReactComponentType instead of React.ComponentType


jwinn:
@osdnk @kmagiera I cannot recommend this fix enough, as, basically, any typescript project, using `>= 1.5.4`, is broken. The current "fix" is to either manually update the type definition file, or revert version to `1.5.3`. Please advise...

tobiasbu:
Hi :)
Duplicated: #945

kmagiera:
Thanks for this but we already merged two other PRs making the same change :)

PR #973, opened 2020-02-16
Fix typo in exported root view type declaration

james-k-ryan-iii:
It looks like this should be `React.ComponentType` instead of `ReactComponentType` (`tsc` is erroring on the latter). I'm not too familiar with React typings though, so apologies in advance if I'm off base!


james-k-ryan-iii:
Closing in favor of #945

PR #990, opened 2020-02-27
Fixes #930

### LINKED ISSUE #930: onHandlerStateChange of TapGestureHandler does not return coordinates on iOS
When we have 2 handlers - tap and double-tap like in [multitap example](https://github.com/software-mansion/react-native-gesture-handler/blob/master/Example/multitap/index.js), when a single tap `onHandlerStateChange` handler is being called, it returns event with null coordinates like:
```
nativeEvent: {
  "absoluteX": 0,
  "absoluteY": 0,
  "handlerTag": 3,
  "numberOfPointers": 0,
  "oldState": 4,
  "state": 5,
  "target": 163,
  "x": 0,
  "y": 0,
}
```

Bug occurs on iOS only, Android works fine and returns right `x`, `y`, etc. values.
It also occurs only when we have 2 `TapGestureHandler`s and one fo them wait for other one.

hhunaid:
This PR fixes issue #930. Having both single and double-tap would not report correct coordinates for single tap gestures.

Problem:
RNBetterTapGestureRecognizer doesn't return correct coordinates after the gesture has completed. i.e. `state == UIGestureRecognizerStateEnded`

Fix:
Simply keep the coordinates from the previous state of the gesture and pass those instead.

You can see the difference here.

Problematic: https://streamable.com/oqftv
Fixed: https://streamable.com/znhm3

### DIFF
--- a/ios/Handlers/RNTapHandler.m
+++ b/ios/Handlers/RNTapHandler.m
@@ -166,7 +166,9 @@
 
 @end
 
-@implementation RNTapGestureHandler
+@implementation RNTapGestureHandler {
+    RNGestureHandlerEventExtraData * _lastData;
+}
 
 - (instancetype)initWithTag:(NSNumber *)tag
 {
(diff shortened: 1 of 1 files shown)


todorone:
@kmagiera @osdnk Can you guys please review/merge?

kmagiera:
Makes sense, thank you for working on this!

blainekasten:
@kmagiera any chance ya'll will do a new release soon? Blocked on this. Thanks for your hard work! 💜

PR #999, opened 2020-03-06
DrawerLayout: add `onDrawerSlide()`

rorygarand:
Related to: https://github.com/software-mansion/react-native-gesture-handler/issues/469

Adds the option to listen for changes on dragX as a user is sliding open a drawer.


DrawerLayout.js
```diff
@@ -198,9 +196,22 @@ export default class DrawerLayout extends Component<PropType, StateType> {
       }
     );
 
+    const gestureOptions = {
+      useNativeDriver: props.useNativeAnimations
+    };
+
+    if( this.props.onDrawerSlide ) {
```
jakub-gonet, line 203: Formatting.

```suggestion
    if(this.props.onDrawerSlide) {
```

DrawerLayout.js
```diff
@@ -198,9 +196,22 @@ export default class DrawerLayout extends Component<PropType, StateType> {
...
+        const position = translationX / this.state.containerWidth;
+
+        this.props.onDrawerSlide(position);
+      };
+    }
+
     this._onGestureEvent = Animated.event(
       [{ nativeEvent: { translationX: dragXValue, x: touchXValue } }],
-      { useNativeDriver: props.useNativeAnimations }
+      gestureOptions
```
jakub-gonet, line 214: I'm not sure extracting this object is more readable than older code.

jakub-gonet (approved):
LGTM, could you resolve the conflicts so we can merge?

DrawerLayout.js
rorygarand: How would you prefer to conditionally add the listener function?

DrawerLayout.js
rorygarand: ok.

rorygarand:
Resolved the merge conflicts

DrawerLayout.js
jakub-gonet: Oh, I missed this assignment. 
This could be written with spread and `&&` operator, but i don't think it's more readable than the current solution.

jakub-gonet:
Thanks for your contribution, much appreciated.

PR #1037, opened 2020-04-16
fix(web): enable text selection in GestureHandler

jaulz:
Currently, you cannot select any text in the browser when it's wrapped with a `GestureHandler`. In the worst case you cannot select anything because the `GestureHandler` wraps the whole page (e.g. the Drawer navigation from `react-navigation`). This PR enables text selection within all `GestureHandlers`.

### DIFF
--- a/web/GestureHandler.js
+++ b/web/GestureHandler.js
@@ -216,7 +216,11 @@ class GestureHandler {
     this.ref = ref;
 
     this.view = findNodeHandle(ref);
-    this.hammer = new Hammer.Manager(this.view);
+    this.hammer = new Hammer.Manager(this.view, {
+      cssProps: {
+        userSelect: 'auto',
+      },
+    });
 
     this.oldState = State.UNDETERMINED;
     this.previousState = State.UNDETERMINED;


RWOverdijk:
This works for me, why was this closed?

jaulz:
It makes a lot of gestures unusable which I noticed only after I raised the PR.

satya164:
I think it makes sense to not disable text selection if gestures are disabled.

RWOverdijk:
@satya164 Agreed. It's tough to make the distinction between swiping and selecting though. Maybe there's a way to see if it's touch or cursor? Or define regions that should respond to touch events?

For now I'm on this patch branch, with gestured disabled. But that's not going to be a very solid solution for long I think.

xn3cr0nx:
I agree with @RWOverdijk, this cannot be a solution. Anyway, I think it's really important to enable selectable text. In may case no text is selectable, this is annoying. I hope we can figure out a solution.

mattmischuk:
I’m also having this issue, @RWOverdijk you seem to be 12 days ahead of our team on a similar initiative, if you’ve found a solution for this please feel free to share here or elsewhere.

RWOverdijk:
@mattmischuk Honestly, I'm using the fork this PR is made from and I disabled gestures on nav for web. It's not pretty but so far it works.

satya164:
Latest versions of React Navigation Stack and Drawer don't use gesture handler on Web

xn3cr0nx:
@satya164 I had to manually set gestureEnabled on the web to avoid a bug due to gestures. What you're saying is what the doc is saying, but it doesn't seems to be the truth

satya164:
You can check the source code and verify yourself.

RWOverdijk:
@xn3cr0nx React-navigation creates new releases about 293847 times a day, so chances are you just need to update again. I dropped react-navigation for web so I can't verify for you.

cglacet:
Can someone explain its solution to this issue?

PR #1043, opened 2020-04-20
fixed #1042 'waitFor' for sibling pan handlers

### LINKED ISSUE #1042: iOS: waitFor doesn't work when handlers belong to different view hierarchies
The following code works properly on Android - until first pan gesture is finished, the second can't be started, but on iOS both handlers can be active simultaneously:

```js
import React, {Component} from 'react';
import {StyleSheet, View} from 'react-native';

import {
  ScrollView,
  State,
  PanGestureHandler,

export class PressBox extends Component {
  panFirst = React.createRef();
  panSecond = React.createRef();
  _onPanFirst = event => {
    if (event.nativeEvent.state === State.ACTIVE) {
      console.log('pan first');
    }
  };
  _onPanSecond = event => {
    if (event.nativeEvent.state === State.ACTIVE) {
      console.log('pan second');
    }
  };
  render() {
    return (
      <>
        <PanGestureHandler
          ref={this.panFirst}
          waitFor={this.panSecond}
          onHandlerStateChange={this._onPanFirst}
          numberOfTaps={1}>
          <View style={[styles.box, {backgroundColor: 'green'}]} />
        </PanGestureHandler>
        <PanGestureHandler
          ref={this.panSecond}
          waitFor={this.panFirst}
          onHandlerStateChange={this._onPanSecond}
          numberOfTaps={1}>
          <View style={[styles.box, {backgroundColor: 'blue'}]} />
        </PanGestureHandler>
      </>
    );
  }
}

  render() {
    return (
      <ScrollView style={styles.scrollView}>
        <PressBox />
      </ScrollView>
    );
  }
}

  scrollView: {
    flex: 1,
  },
  box: {
    width: 150,
    height: 150,
    alignSelf: 'center',
    backgroundColor: 'plum',
    margin: 10,
    zIndex: 200,
  },
});
```

It looks like `RNGestureHandler` methods `gestureRecognizer:shouldBeRequiredToFailByGestureRecognizer:` and `gestureRecognizer:shouldRequireFailureOfGestureRecognizer:` are not called when handlers doesn't wrap each other.

iyegoroff:


osdnk:
It's modifying a bit behavior of UIKit but basically looks good. Can you add some more description and comparison to Android?

ios/RNGestureHandler.m
```diff
@@ -58,6 +58,7 @@ CGRect RNGHHitSlopInsetRect(CGRect rect, RNGHHitSlop hitSlop) {
     return rect;
 }
 
+static NSHashTable<RNGestureHandler *> *allGestureHandlers;
```
osdnk, line 61: just wondering. Can't we do `[NSHashTable weakObjectsHashTable]` here?

iyegoroff: It won't compile because of "Initializer element is not a compile-time constant" error.

iyegoroff:
>Can you add some more description and comparison to Android?

I have described the problem here #1042 - in order to wait for each other on iOS handlers should wrap each other:
```js
// this works on iOS and Android
<PanGestureHandler waitFor={inner} ref={outer}>
  <PanGestureHandler waitFor={outer} ref={inner}>
    {...}
  </PanGestureHandler>
</PanGestureHandler>
```

But when dependent handlers aren't members of the same hierarchy - `waitFor` will not work on iOS:

```js
// works only on Android
<PanGestureHandler waitFor={second} ref={first} />
<PanGestureHandler waitFor={first} ref={second} />
```

PR #1044, opened 2020-04-20
New prop 'waitForGroup'

iyegoroff:
When dealing with gesture handlers as FlatList item views it is quite hard to prevent simultaneous activation of several handlers by using `waitFor` and arrays of refs. `waitForGroup` prop is intended to simplify this task - handlers that have `waitForGroup` prop set to the same string can't be active simultaneously.

This PR also includes all changes from #1043

### DIFF
--- a/createHandler.js
+++ b/createHandler.js
@@ -77,6 +77,9 @@ function filterConfig(props, validProps, defaults = {}) {
       let value = props[key];
       if (key === 'simultaneousHandlers' || key === 'waitFor') {
         value = transformIntoHandlerTags(props[key]);
+      } else if (key === 'waitForGroup') {
+        const group = props[key];
+        value = typeof group === 'string' ? group : group?.current?._groupTag;
       } else if (key === 'hitSlop') {
         if (typeof value !== 'object') {
           value = { top: value, left: value, bottom: value, right: value };
@@ -114,7 +117,11 @@ function hasUnresolvedRefs(props) {
     }
     return refs.some(r => r && r.current === null);
   };
-  return extract(props['simultaneousHandlers']) || extract(props['waitFor']);
+  return (
+    extract(props['simultaneousHandlers']) ||
+    extract(props['waitFor']) ||
+    extract(props['waitForGroup'])
+  );
 }
 
 const stateToPropMappings = {
--- a/docs/handler-common.md
+++ b/docs/handler-common.md
@@ -46,7 +46,7 @@ Accepts a react ref object or an array of refs to other handler components (refs
 ---
 ### `waitForGroup`
 
-Accepts a string or a react ref to `WaitForGroup` element. Handlers that have `waitForGroup` prop set to the same object can't be active simultaneously. This prop is intended to simplify prevention of simultaneous activation for handlers located inside `FlatList` item views. Value for this prop can also be obtained from `useWaitForGroup` hook - [usage example](../Example/swipeableAndWaitForGroup/index.js).
+Accepts a string or a react ref to `WaitForGroup` element. Handlers that have `waitForGroup` prop set to the same object can't be active simultaneously. This prop is intended to simplify prevention of simultaneous activation for handlers located inside `FlatList` item views. Value for this prop can also be obtained from `useWaitForGroup` hook - [usage example](https://github.com/software-mansion/react-native-gesture-handler/blob/master/Example/swipeableAndWaitForGroup/index.js).
 
 ---
 ### `hitSlop`
--- a/docs/handler-common.md
+++ b/docs/handler-common.md
@@ -44,6 +44,11 @@ Accepts a react ref object or an array of refs to other handler components (refs
 Accepts a react ref object or an array of refs to other handler components (refs should be created using [`React.createRef()`](https://reactjs.org/docs/refs-and-the-dom.html)). When set the handler will not [activate](state.md#active) as long as the handlers provided by their refs are in [began](state.md#began) state. Read more in the [cross handler interaction](interactions.md#awaiting-other-handlers) section.
 
 ---
+### `waitForGroup`
+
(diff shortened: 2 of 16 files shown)


osdnk:
Hi, thanks, looks good.
I thinking if we can use some better name e.g. `activationGroup`, cause not I'm not sure what's actually the group component need to wait for. But I'm not having a strong in this opinion.

Also, can we also support refs (like in `waitFor` etc.)? Cause now e.g. if I want to render two FlatList I need to somehow care about unique keys. With `useRef` it will be simpler.

iyegoroff:
Hi, @osdnk. 
>cause not I'm not sure what's actually the group component need to wait for

I chose `waitForGroup` name for that prop because it enables behavior similar to `waitFor` and is related to a group of entities. So if handlers have the same `waitForGroup` they have to wait for each other.

>Also, can we also support refs (like in waitFor etc.)? Cause now e.g. if I want to render two FlatList I need to somehow care about unique keys.

I don't understand how this could be useful - items in these lists could have `waitForGroup` props set to different strings:

```js
class List extends React.Component {
    render() {
      return <FlatList {...this.props} renderItem={this.renderItem} />
    }

    renderItem(info) {
      return (
        <PanGestureHandler waitForGroup={this.props.group} />
      )
    }
}

const firstList = <List group={'First'} />
const secondList = <List group={'Second'} />
```
Here, for example, two handlers from `firstList` can't be active simultaneously, but one from `firstList` and one from `secondList` can.

osdnk:
> I chose waitForGroup name for that prop because it enables behavior similar to waitFor and is related to a group of entities. So if handlers have the same waitForGroup they have to wait for each othe

Ok, works for me

> I don't understand how this could be useful - items in these lists could have waitForGroup props set to different strings:

Take a look from w bit wider perspective. What you're actually asking users about is to make groups' names globally unique. I don't say it's very difficult, but it may lead to some unexpected issues (eg. in the complicated app where none of the developers has end to end knowledge about the whole app or when some library will start to use your feature). I don't want to make people need to care about namespace. 

My suggestion is to support `refs` additionally to your current solution.
E.g.

```jsx
function renderItem(info) {
  return (
    <PanGestureHandler waitForGroup={info.waitFor} />
  )
}

function List(props) {
  const waitGroup = useRef()
  return <FlatList {...props} renderItem={item => renderItem({ ...item, waitGroup })} />
}

```

Then I can use 

```
<List/>
```
in any place of the app not caring about any namespace


Also, please add an example.

iyegoroff:
Ok, this makes sense.

osdnk:
actually `useRef` might be misleading. You can consider adding `useGroup` hook just for convenience.

docs/handler-common.md
```diff
@@ -43,6 +43,11 @@ Accepts a react ref object or an array of refs to other handler components (refs
 
 Accepts a react ref object or an array of refs to other handler components (refs should be created using [`React.createRef()`](https://reactjs.org/docs/refs-and-the-dom.html)). When set the handler will not [activate](state.md#active) as long as the handlers provided by their refs are in [began](state.md#began) state. Read more in the [cross handler interaction](interactions.md#awaiting-other-handlers) section.
 
+---
+### `waitForGroup`
+
+Accepts a string or a react ref to `WaitForGroup` element. Handlers that have `waitForGroup` prop set to the same object can't be active simultaneously. This prop is intended to simplify prevention of simultaneous activation for handlers located inside `FlatList` item views. Value for this prop can also be obtained from `useWaitForGroup` hook - [usage example](../Example/swipeableAndWaitForGroup/index.js).
+
```
osdnk, line 50: most likely this link won't work on gh pages

iyegoroff: done

iyegoroff:
@osdnk , are there any issues keeping this PR from merging?

osdnk:
I'm no longer managing this repo, I'm sorry, cc @jakub-gonet

iyegoroff:
Hi @jakub-gonet, could you please give some feedback on this PR?

createHandler.js
```diff
@@ -77,6 +77,9 @@ function filterConfig(props, validProps, defaults = {}) {
       let value = props[key];
       if (key === 'simultaneousHandlers' || key === 'waitFor') {
         value = transformIntoHandlerTags(props[key]);
+      } else if (key === 'waitForGroup') {
+        const group = props[key];
+        value = typeof group === 'string' ? group : group?.current?._groupTag;
```
jakub-gonet, line 82: Should we add 'waitForGroup' key to the object when `value === undefined`?

ios/RNGestureHandler.m
```diff
@@ -309,10 +317,13 @@ - (BOOL)containsPointInView
 
 - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
 {
-    if ([_handlersToWaitFor count]) {
+    if ([_handlersToWaitFor count] || _waitForGroup != nil) {
         for (RNGestureHandler *handler in [allGestureHandlers allObjects]) {
             if (handler != nil
                 && (handler.state == RNGestureHandlerStateActive || handler->_recognizer.state == UIGestureRecognizerStateBegan)) {
+                if (_waitForGroup != nil && [_waitForGroup isEqualToString:handler->_waitForGroup]) {
```
jakub-gonet, line 324: You used `_waitForGroup != nil && [_waitForGroup isEqualToString:handler->_waitForGroup]` three times, maybe it's a good idea to extract it to some helper function?

jakub-gonet (commented):
I've left some comments.

jakub-gonet:
@iyegoroff 🏓

jakub-gonet:
@iyegoroff, do you plan on working on this?

jakub-gonet:
This wasn't active for some time so I'm closing it. If you plan on working on this PR please reply so we can reopen this one.

PR #1053, opened 2020-04-29
Fix typo

mrmuhammadali:


jakub-gonet:
Thanks for your contribution, your change is included in #1018

PR #1055, opened 2020-04-29
Fix typings for enums

iyegoroff:
This PR fixes invalid enum-based typings for `State` and `Directions`. Something like `State[State.END]` should return `'END'` instead of `undefined` because Typescript enum generates code like this - https://www.typescriptlang.org/play/index.html#code/KYOwrgtgBAIglgJ2AYwC5wPYgM5QN4BQUxxASgJIDiAEgCpQC8UAjADRElQAyAogGL0mAJnacoAVQAKjKABZRnGAHkA6gDkZADg4BfIA - which is different from actual `State` and `Directions` objects

### DIFF
--- a/react-native-gesture-handler.d.ts
+++ b/react-native-gesture-handler.d.ts
@@ -22,21 +22,25 @@ declare module 'react-native-gesture-handler' {
 
   /* GESTURE HANDLER STATE */
 
-  export enum Directions {
-    RIGHT = 1,
-    LEFT = 2,
-    UP = 4,
-    DOWN = 8,
+  export const Directions: {
+    readonly RIGHT: 1
+    readonly LEFT: 2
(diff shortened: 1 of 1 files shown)


jakub-gonet (approved):
Thanks!

mrousavy:
Nice! Just a question; wouldn't the same be achieveable using `as const` at the end of the enum?

jakub-gonet:
Using an enum like that results in the syntax error. `as const` assertion only adds `readonly` to every field, so this PR could be improved in that manner.

We have to use `typeof State[keyof typeof State]` to be able to use numbers and enums constants interchangeably.

mrousavy:
I see. I'm not very experienced with enums, I've been using union types everywhere. 🙈 thanks for clarifying!

PR #1098, opened 2020-06-22
fix LongPressGestureHandler web discrepancy on move events

### LINKED ISSUE #1097: LongPressGestureHandler discrepancy between web and mobile implementations
LongPressGestureHandler on mobile (ios/android) is implemented as a continuous gesture, i.e. we can drag it around after long-press activated. However, on the web, it is implemented as a `PressGestureHandler`, which is a `DiscreteGestureHandler`. Judging from [onRawEvent](https://github.com/software-mansion/react-native-gesture-handler/blob/f9fe5d82c47da8b1370eaa8b1e3a00a0c31400c5/web/PressGestureHandler.js#L105): it seems to only fire `INPUT_END` event when the gesture is released, i.e. it ignores all gesture movements.

Not sure if this behavior discrepancy is intended, figure I should ask before attempting the fix.

connectdotz:
This PR fixes the following issues for the web implementation

1. The LongPressGuestureHandler on mobile (ios/android) emits move events after being activated but the web implementation does not. Modified the `PressGestureHandler.onRawEvent` to pass along the move event.
1. in `createHander` static class proprType is not supported in `react-native-web`, fixed in `createHander`

fix #1097

### DIFF
--- a/createHandler.js
+++ b/createHandler.js
@@ -236,7 +236,10 @@ export default function createHandler(
       this._createGestureHandler(
         filterConfig(
           transformProps ? transformProps(this.props) : this.props,
-          { ...this.constructor.propTypes, ...customNativeProps },
+          {
+            ...(this.constructor.propTypes || propTypes),
+            ...customNativeProps,
+          },
           config
         )
       );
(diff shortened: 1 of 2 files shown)


noldidrita:
I can confirm this works.

EvanBacon (approved):
ok

Issue #935, opened 2020-01-28
Wrong fire onSwipeableWillOpen


dmytro-kupriianov:
Method onSwipeableWillOpen fired only after open panel, but not when start movement. How fix this moment?

MoOx:
Having the same issue. I would have expected that this event is called on start. How this differ from onSwipeableOpen then?

MoOx:
The idea in #1101 could be a solution if changing `onSwipeableWillOpen` is not an option (eg: undesired breaking change).

jakub-gonet:
#990 should make this possible now.

MoOx:
Not really as discussed in #1102. #999 is for Drawer only as we are here discussing about Swipeable.

MoOx:
Btw, what is the difference between onSwipeableWillOpen and onSwipeableOpen. If there is a good reason, I think new lifecycle method are necessary. If there is no good reasons (= not really a bug but still is?) it should be changed, but would be a breaking change. Otherwise, we could add new lifecycle (onSwipeStart+End) as onSwipeableWillOpen might be legit for programmatic opening as you mentioned in https://github.com/software-mansion/react-native-gesture-handler/pull/1102#issuecomment-733060255.
I am open for a documented PR if we agree on something :)

jakub-gonet:
> Btw, what is the difference between onSwipeableWillOpen and onSwipeableOpen.

I'd want to know it myself. We can check it in the code but that's not the point – having will/did and plain lifecycle methods is confusing to the developers using this component. 

Currently, we have the following description in docs:

> ## onSwipeableOpen
> method that is called when action panel gets open (either right or left).

> ## onSwipeableWillOpen
> method that is called when action panel starts animating on open (either right or left).

I imagine this is confusing to the new developers using RNGH. `onSwipeableWillOpen` gets called before opening but at the start of the animation? `onSwipeableOpen` gets called when opened swipeable or in the same time when `will` variant? What about people who use TS as their documentation?

> Otherwise, we could add new lifecycle (onSwipeStart+End) as onSwipeableWillOpen might be legit for programmatic opening as you mentioned in #1102 (comment).

For me, there's little sense to create another lifecycle method; we added will/did, left/right prefix and it effectively makes the number of lifecycle methods eightfold. Adding onSwipeStart/End makes situation worse.

Ideally, it'd be only one/two callbacks:

- `onSwipeableOpen` taking a function with two args: `(direction, animating)`
[...]

PR #1102, opened 2020-06-25
add hooks for swipe start/end

### LINKED ISSUE #1101: add hook for `isDragging` to Swipeable
right now its possible to tell when drag ends (`onSwipeableWillClose`) but there is no way to attach functionality to when dragging starts as far as I can tell. `onSwipeableWillOpen` is only called when opened programmatically, not when the user starts to drag the component

chrisdrackett:
closes #1101

### DIFF
--- a/Swipeable.js
+++ b/Swipeable.js
@@ -19,6 +19,8 @@ export type PropType = {
   overshootLeft?: boolean,
   overshootRight?: boolean,
   overshootFriction: number,
+  onSwipeStart?: Function,
+  onSwipeEnd?: Function,
   onSwipeableLeftOpen?: Function,
   onSwipeableRightOpen?: Function,
   onSwipeableOpen?: Function,
@@ -160,7 +162,11 @@ export default class Swipeable extends Component<PropType, StateType> {
   };
 
   _onHandlerStateChange = ({ nativeEvent }) => {
+    if (nativeEvent.state === State.ACTIVE) {
+      this.props.onSwipeStart && this.props.onSwipeStart()
+    }
     if (nativeEvent.oldState === State.ACTIVE) {
+      this.props.onSwipeEnd && this.props.onSwipeEnd()
       this._handleRelease(nativeEvent);
     }
   };


MoOx:
Would be nice to agree on something as this is a trivial change that is pending for a while. Need any help somehow?

pribeh:
Bump. Tested here and works fine.

jakub-gonet:
I'd say this can be now achieved by #999 so I'll close this one.

Thanks for your contribution!

MoOx:
@jakub-gonet wait noooo. #999 only works for  Drawer while this one is for Swipeable component.

jakub-gonet:
Ohhhhh.

MoOx:
@jakub-gonet what about merging it now? 😅

jakub-gonet:
I think we should allow `onSwipeableWillOpen` to be fired on touch dragging instead of only when opening programmatically if RN does this (I haven't checked that yet). For me adding yet another lifecycle method isn't ideal.

> @jakub-gonet what about merging it now? 😅

This PR lacks a docs and types update so can't right now, but I'd want to discuss ^ first.

jakub-gonet:
Tagging for visibility: @chrisdrackett, @MoOx 

Per [my comment in #935](https://github.com/software-mansion/react-native-gesture-handler/issues/935#issuecomment-733090839) I think this PR should be closed or changed to `onDrag` or a similar method. I'm open for discussion though.

(We're in process of merging PRs, so I'll close this one if there will be no activity for a week – nothing permanent, it can be always reopened if you want to work on this one)

chrisdrackett:
I don't have a strong option on what these props are called myself.

We moved to just rolling our own swappable in July as there wasn't activity on this issue so I don't really have skin in the game for this PR anymore.

jakub-gonet:
All right, thanks for responding. I'll close this one for now – this is a small change codewise but requires us to support two new API methods for Swipeable. It can be reopened in the future after we decide what we should do with Swipeable/Drawer APIs.

Thanks for your contribution anyway, really appreciated.

PR #1104, opened 2020-06-29
Make Touchable* work with Detox 17

todorone:
All `Touchable*` are broken when trying to run e2e tests using recently released Detox 17. It's because `testID` props are attached to `Animated.View` component instead of `BaseButton`. This PR fix it making e2e tests using Detox working again.

### DIFF
--- a/touchables/GenericTouchable.js
+++ b/touchables/GenericTouchable.js
@@ -251,7 +251,6 @@ export default class GenericTouchable extends Component {
       accessibilityStates: this.props.accessibilityStates,
       accessibilityTraits: this.props.accessibilityTraits,
       nativeID: this.props.nativeID,
-      testID: this.props.testID,
       onLayout: this.props.onLayout,
       hitSlop: this.props.hitSlop,
     };
@@ -266,6 +265,7 @@ export default class GenericTouchable extends Component {
         hitSlop={this.props.hitSlop}
         shouldActivateOnStart={this.props.shouldActivateOnStart}
         disallowInterruption={this.props.disallowInterruption}
+        testID={this.props.testID}
         {...this.props.extraButtonProps}>
         <Animated.View {...coreProps} style={this.props.style}>
           {this.props.children}


todorone:
@jakub-gonet Is it possible to merge this PR?

jakub-gonet:
Sure, I don't see any blockers.

Next release should be out in the next few weeks.

Guuz:
@jakub-gonet Any chance we can do this sooner than "a few weeks"? This prevents users from updating to RN 0.63 and Detox 0.17 (which are related).

jakub-gonet:
If that's a blocker I can prepare a release tomorrow.

jakub-gonet:
1.7.0 has been released
https://github.com/software-mansion/react-native-gesture-handler/releases/tag/1.7.0

Guuz:
@todorone With 1.7.0 i'm still getting this issue, did it work for you?

> Test Failed: View “<RCTTextView: 0x7fd56c7310a0>” is not hittable at point “{"x":58.5,"y":12.166664123535156}”: Another view “<RNGestureHandlerButton: 0x7fd56e21add0>” is hittable at window point “{"x":187.5,"y":210.16664886474609}”

todorone:
@Guuz Hm, it fixed the issue for me...

Guuz:
@todorone yeah. i'm digging deeper into this. You probably do `by.id()`, right? we use `by.text("foo")` which still gives this error... 🤔  because it probably finds the text node, which is not "tappable".

Guuz:
Also see the discussion here with Detox: https://github.com/wix/Detox/issues/2156
Using TouchableHighlight from RN (not RNGH) works. So there is a difference with how these things interact which each other. Not sure where the bug is and how to fix it :( But you cant use Detox `by.text()` with RNGH buttons.

LeoNatan:
As I mentioned in the issue, there is likely some pointer events set for the text view, or its parent, which is making it decline touch events. Detox 17 and above ensures that elements can be interacted with before actually performing the interaction.

Issue #1100, opened 2020-06-23
Natural swipe behaviour with touchpads


robertherber:
I'm looking for a natural behaviour on touchpads. Using the PanResponder requires pressing down on the touchpad and dragging while pressing - whereas I feel the natural behaviour would be swiping with two fingers. This is the case for both on the web implementation as well as natively on the iPad touchpad.

I'm building an implementation on top of ScrollView now - it's not fully optimal but the only way I've found to get a natural swipe on touchpads (and works for both iPad with touchpad and the web). The best would be if PanResponder would support this pattern out of the box with a configuration flag.

jakub-gonet:
Implemented in #1128.

robertherber:
> Implemented in #1128.

Yeah, for iPad! ;) Web would still be an open feature request! :)

jakub-gonet:
Oh, right. Reopening :>

j-piasecki:
This looks like a duplicate of https://github.com/software-mansion/react-native-gesture-handler/issues/800.

robertherber:
@j-piasecki Looks very much so.. 🤦‍♂️😂

PR #1128, opened 2020-07-15
Allow two finger touchpad scrolling on iPadOS

robertherber:
## Description

Adding support for touchpad in PanGestureHandler by setting `allowedScrollTypesMask` to `UIScrollTypeMaskAll` for  all devices with iOS 13.4+. Important for iPads with Touchpads and should work with Mac Catalyst through the same API as well. As described in #1100 - this covers the ipadOS part of that issue.

## Test plan

Run on an iPad with touchpad. It's now possible to two-finger swipe to interact with the PanGestureHandler.

<details>
<summary>Show code</summary>

```jsx
import React from 'react';
import { AppRegistry, SafeAreaView, Animated } from 'react-native';
import { PanGestureHandler, State } from 'react-native-gesture-handler';

const Example = () => {
  const translateX = new Animated.Value(0);
  const translateY = new Animated.Value(0);
  const lastOffset = { x: 0, y: 0 };

  const onPanGestureEvent = Animated.event([
    {
      nativeEvent: {
        translationX: translateX,
        translationY: translateY,
      },
    },
  ]);

  const onHandlerStateChange = ({
    nativeEvent: { translationX, translationY, oldState },
  }) => {
    if (oldState === State.ACTIVE) {
      // compensate a start of pan
      lastOffset.x += translationX;
      lastOffset.y += translationY;
      translateX.setOffset(lastOffset.x);
      translateY.setOffset(lastOffset.y);

      translateX.setValue(0);
      translateY.setValue(0);
    }
  };

  return (
    <SafeAreaView style={styles.container}>
      <PanGestureHandler
        onGestureEvent={onPanGestureEvent}
        onHandlerStateChange={onHandlerStateChange}>
        <Animated.View
          style={[
            styles.rectangle,
            {
              transform: [{ translateX }, { translateY }],
            },
          ]}
        />
      </PanGestureHandler>
    </SafeAreaView>
  );
};

const styles = {
  container: { flex: 1, justifyContent: 'center', alignItems: 'center' },
  rectangle: {
    backgroundColor: 'pink',
    width: 100,
    height: 100,
  },
};

AppRegistry.registerComponent('Example', () => Example);
```

</details>


jakub-gonet:
This breaks CI.

robertherber:
> This breaks CI.

I'm guessing it's because the XCode used in the CI is 10.2 (current is 11.7) and this flag requires iOS 13 (which should required XCode 11).

jakub-gonet:
I finally bumped CI and if checks will pass I'll merge this one.

jakub-gonet:
One question though - should we do that by default?

robertherber:
> One question though - should we do that by default?

I think in most cases this would be the expected default behaviour but there might be some edge case I'm not aware of.

DrawerLayout.d.ts
```diff
@@ -36,6 +36,7 @@ declare module 'react-native-gesture-handler/DrawerLayout' {
     statusBarAnimation?: StatusBarAnimation;
     overlayColor?: string;
     contentContainerStyle?: StyleProp<ViewStyle>;
+    iosEnableTouchpadTwoFingerGesture?: boolean;
```
jakub-gonet, line 39: I'm not sure if we should prefix this property with `ios`. For example in Buttons we have [`activeOpacity`](https://docs.swmansion.com/react-native-gesture-handler/docs/component-buttons#activeopacity-ios-only) which is ios only or `borderless` which is android specific.

DrawerLayout.js
```diff
@@ -50,6 +50,7 @@ export type PropType = {
   drawerContainerStyle?: any,
   contentContainerStyle?: any,
   onGestureRef?: Function,
+  iosEnableTouchpadTwoFingerGesture?: Boolean,
```
jakub-gonet, line 53: I think this should be `boolean`

ios/Handlers/RNPanHandler.m
```diff
@@ -206,6 +206,16 @@ - (void)configure:(NSDictionary *)config
   APPLY_FLOAT_PROP(failOffsetYStart);
   APPLY_FLOAT_PROP(failOffsetYEnd);
   
+  if (@available(iOS 13.4, *)) {
+    id prop = config[@"iosEnableTouchpadTwoFingerGesture"];
+    if(prop != nil){
+      bool iosEnableTouchpadTwoFingerGesture = [RCTConvert BOOL:prop];
```
jakub-gonet, line 212: I think we can convert in the same line with `config[(...)]` if `[RCTConvert BOOL:nil]` returns `nil`.

jakub-gonet (commented):
Left some inline comments.

DrawerLayout.d.ts
robertherber: Changed it to `enableTrackpadTwoFingerGesture`

ios/Handlers/RNPanHandler.m
robertherber: Updated and works

jakub-gonet:
I pasted some test code into a description. I don't have access to the touchpad connected to the iPad, you please verify that it works with this PR?

robertherber:
@jakub-gonet I've verified that the test code works - and of course the examples. Ready to merge?

jakub-gonet (approved):
LGMT, thanks for the patch.

PR #1144, opened 2020-07-22
Fix installation and documentation links

kaueDM:
## Description

Fixed two broken links.

btw when clicking in `Basics > Getting Started` [here](https://docs.swmansion.com/react-native-gesture-handler/docs/), it redirects to a 404 page.


jakub-gonet:
Thanks!

We are aware of this 404, it seems that our docusaurus config is a bit misconfigured, but I wasn't able to narrow it down.

PR #1160, opened 2020-08-07
Fix RNGestureHandlerManager to use explicit imports

PeteTheHeat:
## Description

I work on RN Core at Facebook, and am encountering a unique problem. I'm trying to refactor `RCTEventDispatcher`, but cannot because of the implicit dependency on `RCTEventDispatcher` in `RNGestureHandlerManager`. We use rn-gesture-handler internally, and builds for it fail after the refactor. 

The fix is to explicitly depend on `RCTEventDispatcher` instead of relying on the `RCTUIManager -> RCTViewManager -> RCTEventDispatcher` import chain. 

## Test plan

I unfortunately did not test this. I can't get react-native-gesture-handler building locally. I assume that there are CI builds that would fail on a simple compilation error.

### DIFF
--- a/ios/RNGestureHandlerManager.h
+++ b/ios/RNGestureHandlerManager.h
@@ -1,7 +1,9 @@
 #import <Foundation/Foundation.h>
 
 #import <React/RCTBridgeModule.h>
-#import <React/RCTUIManager.h>
+
+@class RCTUIManager;
+@class RCTEventDispatcher;
 
 @interface RNGestureHandlerManager : NSObject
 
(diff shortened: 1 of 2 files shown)


jakub-gonet (approved):
It's interesting that the RN uses RNGH internally, is it open-sourced?

PeteTheHeat:
Sorry, I should have been more clear.

**Facebook** uses RNGH internally in a few apps. In order to land my refactor to RN core, those apps must build.

jkadamczyk (approved):

jkadamczyk:
I think we can merge that. Builds are passing on CI.
CC @jakub-gonet

jkadamczyk:
Thanks for your contribution @PeteTheHeat

PR #1172, opened 2020-08-22
Add sound to Touchables on Android

### LINKED ISSUE #1146: TouchableNativeFeedback is missing touch sound
On Android, touching a button in a native app will result (if user configured) in a sound to be played. This is an indicator to the user that the button has been pressed.
When using the `TouchableNativeFeedback` element of React native, this sound is played (see: [Touchable.js](https://github.com/facebook/react-native/blob/caf010914c5b3124d76562e624dd4936c96f049c/Libraries/Components/Touchable/Touchable.js#L876) ). 

However, when using the `TouchableNativeFeedback` of this repository, this sound is not played. Making it feel less like a native experience.

When looking further into the code, it seems that React native is using [this](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/sound/SoundManagerModule.java) code to actually play the sound.

jakub-gonet:
## Description

Fixes #1146.

Seems like even `TouchableWithoutFeedback` plays sound in RN implementation, so we will stick to that.

Extracted from changes:

Buttons in RN are wrapped in NativeViewGestureHandler which manages `onTouchEvent` call after activation of the handler. Problem is, in order to verify that the underlying button implementation is interested in receiving touches we have to call onTouchEvent and check if the button is pressed.
This leads to invoking onTouchEvent twice which isn't idempotent in View - it calls OnClickListener and plays a sound effect if OnClickListener was set.

To mitigate this behavior we use `mLastEventTime` variable to check that we already handled the event in `onInterceptTouchEvent`. We assume here that different events will have different event times. We can't use ids of the event objects because they can be recycled and reused.

## Test plan

Tested on example code:
```jsx
import React from 'react';
import {
  Text,
  View,
  TouchableNativeFeedback,
  TouchableOpacity,
  TouchableWithoutFeedback,
} from 'react-native';

import {
  TouchableNativeFeedback as GHTouchableNativeFeedback,
  TouchableOpacity as GHTouchableOpacity,
  TouchableWithoutFeedback as GHTouchableWithoutFeedback,

export default () => {
  return (
    <View
      style={{
        flex: 1,
        justifyContent: 'center',
        alignItems: 'center',
        alignContent: 'space-around',
      }}>
      <TouchableOpacity onPress={() => console.log('pressed TouchableOpacity')}>
        <View
          style={{
            width: 200,
            height: 50,
            backgroundColor: 'lightyellow',
          }}>
          <Text>TouchableOpacity</Text>
        </View>
      </TouchableOpacity>
      <TouchableWithoutFeedback
        onPress={() => console.log('pressed TouchableWithoutFeedback')}>
        <View
          style={{
            width: 200,
            height: 50,
            backgroundColor: 'lightyellow',
          }}>
          <Text>TouchableWithoutFeedback</Text>
        </View>
      </TouchableWithoutFeedback>
      <TouchableNativeFeedback
        onPress={() => console.log('pressed TouchableNativeFeedback')}>
        <View
          style={{
            width: 200,
            height: 50,
            backgroundColor: 'lightyellow',
          }}>
          <Text>TouchableNativeFeedback</Text>
        </View>
      </TouchableNativeFeedback>
      <GHTouchableOpacity
        onPress={() => console.log('pressed RNGH TouchableOpacity')}>
        <View
          style={{
            width: 200,
            height: 50,
            backgroundColor: 'lightblue',
          }}>
          <Text>RNGH TouchableOpacity</Text>
        </View>
      </GHTouchableOpacity>
      <GHTouchableWithoutFeedback
        onPress={() => console.log('pressed RNGH TouchableWithoutFeedback')}>
        <View
          style={{
            width: 200,
            height: 50,
            backgroundColor: 'lightblue',
          }}>
          <Text>RNGH TouchableWithoutFeedback</Text>
        </View>
      </GHTouchableWithoutFeedback>
      <GHTouchableNativeFeedback
        onPress={() => console.log('pressed RNGH TouchableNativeFeedback')}>
        <View
          style={{
            width: 200,
            height: 50,
            backgroundColor: 'lightblue',
          }}>
          <Text>RNGH TouchableNativeFeedback</Text>
        </View>
      </GHTouchableNativeFeedback>
    </View>
  );
};
```

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
@@ -1,5 +1,6 @@
 package com.swmansion.gesturehandler.react;
 
+import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
 import android.content.Context;
 import android.content.res.ColorStateList;
@@ -11,6 +12,7 @@ import android.graphics.drawable.RippleDrawable;
 import android.os.Build;
 import android.util.TypedValue;
 import android.view.MotionEvent;
+import android.view.View;
 import android.view.ViewGroup;
 
 import com.facebook.react.bridge.SoftAssertions;
@@ -27,6 +29,10 @@ public class RNGestureHandlerButtonViewManager extends
 
     static TypedValue sResolveOutValue = new TypedValue();
     static ButtonViewGroup sResponder;
+    static OnClickListener sClickListener = new OnClickListener() {
+      @Override
+      public void onClick(View v) { }
+    };
 
     int mBackgroundColor = Color.TRANSPARENT;
     // Using object because of handling null representing no value set.
@@ -36,6 +42,7 @@ public class RNGestureHandlerButtonViewManager extends
     boolean mUseBorderless = false;
     float mBorderRadius = 0;
     boolean mNeedBackgroundUpdate = false;
+    long mLastEventTime = 0;
     public static final String SELECTABLE_ITEM_BACKGROUND = "selectableItemBackground";
     public static final String SELECTABLE_ITEM_BACKGROUND_BORDERLESS = "selectableItemBackgroundBorderless";
 
@@ -43,6 +50,8 @@ public class RNGestureHandlerButtonViewManager extends
     public ButtonViewGroup(Context context) {
       super(context);
 
+      // we attach empty click listener to trigger tap sounds (see View#performClick())
+      setOnClickListener(sClickListener);
       setClickable(true);
       setFocusable(true);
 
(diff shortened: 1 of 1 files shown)


jkadamczyk (approved):
Looks good to me

jakub-gonet:
While writing this PR I noticed that even after moving outside of Touchable it still runs onPress and plays a sound, it's something we want to fix in the future.

jakub-gonet:
The previous implementation used Sound Manager from RN, which was failing detox tests due to turbo module inclusion.

I prepared a new implementation that uses native tap sound triggered by [presence of `OnClickListener`](https://android.googlesource.com/platform/frameworks/base/+/a175a5b/core/java/android/view/View.java#3524). 

This revealed that we call `onTouchEvent` multiple times on the same event, so we block subsequent calls by comparing the last event's time with the current event time. This is written under the assumption that every event will have a different event time. This approach may have unexpected consequences in the future if this isn't true (e.g. device sends events in batches with the same event time), causing some events to be skipped.

Guuz:
Using the "presence of OnClickListener" as you linked above sounds like a much better strategy, also for RN itself... might be worth it to suggest it there too!

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -27,6 +29,10 @@
 
     static TypedValue sResolveOutValue = new TypedValue();
     static ButtonViewGroup sResponder;
+    static OnClickListener sClickListener = new OnClickListener() {
```
jakub-gonet, line 32: Change it to "sDummyListener"

kmagiera (approved):

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java
```diff
@@ -93,12 +102,38 @@ public boolean onInterceptTouchEvent(MotionEvent ev) {
...
+     * the event in {@link #onInterceptTouchEvent(MotionEvent)}. We assume here that different events
+     * will have different event times.
+     *
+     * Reference:
+     * {@link com.swmansion.gesturehandler.NativeViewGestureHandler#onHandle(MotionEvent)} */
+    @SuppressLint("ClickableViewAccessibility")
+    @Override
+    public boolean onTouchEvent(MotionEvent event) {
+      long eventTime = event.getEventTime();
+      if(mLastEventTime != eventTime || mLastEventTime == 0){
```
jakub-gonet, line 130: Formatting.

jakub-gonet:
For the record: the new implementation fixed tap sound when moving outside of the touchable.

PR #1173, opened 2020-08-22
Fix thread safety of `RNGestureHandlerModule` methods

### LINKED ISSUE #1169: void com.swmansion.gesturehandler.GestureHandlerOrchestrator.onHandlerStateChange(com.swmansion.gesturehandler.GestureHandler, int, int) on a null object reference
```
java.lang.NullPointerException Attempt to invoke virtual method 'void com.swmansion.gesturehandler.GestureHandlerOrchestrator.onHandlerStateChange(com.swmansion.gesturehandler.GestureHandler, int, int)' on a null object reference 
    GestureHandler.java:342 com.swmansion.gesturehandler.GestureHandler.moveToState
    GestureHandler.java:436 com.swmansion.gesturehandler.GestureHandler.cancel
    RNGestureHandlerRegistry.java:69 com.swmansion.gesturehandler.react.RNGestureHandlerRegistry.detachHandler
    RNGestureHandlerRegistry.java:76 com.swmansion.gesturehandler.react.RNGestureHandlerRegistry.dropHandler
    RNGestureHandlerModule.java:512 com.swmansion.gesturehandler.react.RNGestureHandlerModule.dropGestureHandler
    Method.java:-2 java.lang.reflect.Method.invoke
    JavaMethodWrapper.java:372 com.facebook.react.bridge.JavaMethodWrapper.invoke
    JavaModuleWrapper.java:151 com.facebook.react.bridge.JavaModuleWrapper.invoke
    NativeRunnable.java:-2 com.facebook.react.bridge.queue.NativeRunnable.run
    Handler.java:873 android.os.Handler.handleCallback
    Handler.java:99 android.os.Handler.dispatchMessage
    MessageQueueThreadHandler.java:27 com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage
    Looper.java:215 android.os.Looper.loop
    MessageQueueThreadImpl.java:226 com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run
    Thread.java:798 java.lang.Thread.run
```

Not sure when that happens since we received that in production but I think the stacktrace should be self-explanatory.

### Expected behavior
No crashes

### Actual behavior
Crash

## Package versions

- React: 15.6.2
- React Native: 0.63.2
- React Native Gesture Handler: 1.7.0
(using wix react-native-navigation 6.9.1)

jakub-gonet:
## Description

Fixes #1169.

Another instance of changing the handler state outside the UI thread. In this case, calling `handler.cancel()` which calls `mOrchestrator.onHandlerStateChange()` interfered with `scheduleFinishedHandlersCleanup()` method which calls `handler.reset()` and sets `mOrchestrator` to `null`. 

This may happen when threads interleave in the way described below:

1. `RNGestureHandlerModule.dropGestureHandler` is called (directly or in `attachGestureHandler()`; it's called on native modules thread)
2. RNGH dispatches first `onStateChange` event on UI thread
3. after dispatching event `mHandlingChangeSemaphore` is zero and call to `scheduleFinishedHandlersCleanup()` causes `cleanupFinishedHandlers()` to reset handlers and clear `mOrchestrator`
4. Native modules thread continues with execution and tries to call `moveToState` which crashes the app

This may show when `RNGestureHandlerModule.attachGestureHandler()` or `RNGestureHandlerModule.dropGestureHandler()` is rapidly called from JS (e.g. when frequently unmounting or updating handler component).

After checking the call hierarchy of `GestureHandler.moveToState()` it seems like this is the last occurrence of a native-module-calling-UI-only-method problem.

## Example

Fast clicking on the TapGestureHandler crashes with error "Expected to run on UI thread", introduced in #1171. After applying this patch, the error disappears.

```jsx
import { TapGestureHandler } from 'react-native-gesture-handler';
export default () => {
  const [i, setI] = useState(0);
  const updateState = () => setI(i + 1);
  return (
    <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
      <TapGestureHandler onHandlerStateChange={updateState} key={i}>
        <View style={{ height: 100, width: 100, backgroundColor: 'pink' }} />
      </TapGestureHandler>
    </View>
  );
};
```

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.java
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.java
@@ -3,6 +3,7 @@ package com.swmansion.gesturehandler.react;
 import android.util.SparseArray;
 import android.view.View;
 
+import com.facebook.react.bridge.UiThreadUtil;
 import com.swmansion.gesturehandler.GestureHandler;
 import com.swmansion.gesturehandler.GestureHandlerRegistry;
 
@@ -50,7 +51,7 @@ public class RNGestureHandlerRegistry implements GestureHandlerRegistry {
     }
   }
 
-  private synchronized void detachHandler(GestureHandler handler) {
+  private synchronized void detachHandler(final GestureHandler handler) {
     Integer attachedToView = mAttachedTo.get(handler.getTag());
     if (attachedToView != null) {
       mAttachedTo.remove(handler.getTag());
@@ -66,7 +67,12 @@ public class RNGestureHandlerRegistry implements GestureHandlerRegistry {
       // Handler is in "prepared" state which means it is registered in the orchestrator and can
       // receive touch events. This means that before we remove it from the registry we need to
       // "cancel" it so that orchestrator does no longer keep a reference to it.
-      handler.cancel();
+      UiThreadUtil.runOnUiThread(new Runnable() {
+        @Override
+        public void run() {
+          handler.cancel();
+        }
+      });
     }
   }
 


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.java
```diff
@@ -66,7 +67,12 @@ private synchronized void detachHandler(GestureHandler handler) {
       // Handler is in "prepared" state which means it is registered in the orchestrator and can
       // receive touch events. This means that before we remove it from the registry we need to
       // "cancel" it so that orchestrator does no longer keep a reference to it.
-      handler.cancel();
+      UiThreadUtil.runOnUiThread(new Runnable() {
```
jakub-gonet, line 70: Should I extract this Runnable to the member variable?

kmagiera (approved):
👍

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.java
kmagiera: If it is created frequently then it may make sense. I don't think it is though, because it only happens when you change enabled property while there is an ongoing gesture

jakub-gonet: This is called every time when you update/unmount GH which has associated view, so it may be worth doing.

jakub-gonet: Actually, we need to capture `handler` variable and AFAIK there's no simple solution to reuse one Runnable to do that.

PR #1187, opened 2020-09-08
Lazy imports/exports

radex:
## Description

Hi. I've been profiling the launch time of my app, and optimizing the critical path to launch the app. I found RNGH to take ~25ms to launch (to import RNGH), but very few of its exports are actually required to launch my app.

And so I applied the technique react-native uses (explicit export with getters that lazily require sub-modules) to make the reexports fully lazy. Is it ugly? Yes. Is it worth it? Sure.

After this change, RNGH takes ~5ms to import.

## Test plan

I ran automated tests in this repo + ran manual tests on [Nozbe Teams](https://nozbe.com) to check for regressions.


jkadamczyk:
Hi @radex! Thanks for the performance tests and a pull request. 
I will try to get this in sooner than later.
In the meantime would you mind resolving the conflicts?

I will also go to reanimated and work on a PR there as you posted these for both libraries.

Have a good day!

jakub-gonet:
I think we could also use [computed getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#Syntax) to automatically generate getters based on file names.

jakub-gonet:
🏓 @radex

jakub-gonet:
I'm closing this one due to inactivity. If you plan on working on this, please reply.

PR #1202, opened 2020-09-30
Fix GestureHandler isWithinBounds top hitslop

eeynard:
## Description

This PR fixes a bug in `isWithinBounds` making GestureHandler hitSlop `top` not usable.

## Test Plan

Create a new view with a PanGestureHandler and shrink the hitSlop to only the top part of the view. Events should not be sent when gesture are made outside de red area.

**Before** : Events were fired in the white and red area
**After** : Events are fired only on the red area as expected

```ts
import React from 'react';

import { PanGestureHandler } from 'react-native-gesture-handler';

const HEADER_HEIGHT = 200;

const BottomSheet = () => (
  <PanGestureHandler
    onGestureEvent={() => console.log('Gesture handled')}
    hitSlop={{ top: 0, height: HEADER_HEIGHT }}>
      <View style={styles.activeZone} />
    </View>
  </PanGestureHandler>
);

export default BottomSheet;

  container: {
    flex: 1,
    backgroundColor: 'white',
  },
  activeZone: {
    backgroundColor: 'red',
    height: HEADER_HEIGHT,
  },
});

```

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
@@ -409,7 +409,7 @@ public class GestureHandler<T extends GestureHandler> {
         left -= padLeft;
       }
       if (hitSlopSet(padTop)) {
-        top -= padBottom;
+        top -= padTop;
       }
       if (hitSlopSet(padRight)) {
         right += padRight;
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.java
@@ -428,9 +428,9 @@ public class GestureHandler<T extends GestureHandler> {
         }
       }
       if (hitSlopSet(height)) {
-        if (!hitSlopSet(top)) {
+        if (!hitSlopSet(padTop)) {
           top = bottom - height;
-        } else if (!hitSlopSet(bottom)) {
+        } else if (!hitSlopSet(padBottom)) {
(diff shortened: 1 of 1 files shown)


jakub-gonet:
Thanks for the PR.

Could you please post some code you tested this change on?

eeynard:
@jakub-gonet I updated the description :)

eeynard:
@jakub-gonet Is there any chances this could be added in the next release ? I am really looking forward this bug fix. Thanks !

jakub-gonet:
Yeah, probably. I'll dedicate more time to the gesture handler in the next week, so I'll test and merge it. You can use patch-package until then.

jkadamczyk:
Hey @eeynard I copied and pasted your example to expo managed app. Even though it was using the release version of gesture handler it worked like you described in the expected behaviour without your changes.
Would you mind updating the example or recreating it in expo or sharing a project that makes it possible to easily reproduce it?
Thanks for the PR, and have a good day!

eeynard:
Hey @jkadamczyk, you're right, my example does not seems to be relevant. I made another one on snack, here is the link https://snack.expo.io/TGAF2FyWn. Hopefully this time you can reproduce the problem.

I'll edit the PR description as well. Thank you !

Note: On the snack example the counter increments on every events received by RN-gesture-handler. You can see that it does not do anything on the red zone like it should but events are sent when you play with the bottom of the screen.

jkadamczyk:
Thank you!!!
I'll take a look tomorrow!

jkadamczyk:
@eeynard Tested it works as it should.
Merci beaucoup!

PR #1211, opened 2020-10-13
Adjust Platform constants import

### LINKED ISSUE #1210: PlatformConstants no longer a member of NativeModules (breaks ForceTouchGestureHandler)
## Description

ForceTouchGestureHandler was not working on my iPhone XS Max (claiming it was unavailable on the platform which is untrue) and after digging into it I can see that in PlatformConstants.js, `PlatformConstants` is no longer a member of react-native's `NativeModules` export. Because we're querying `PlatformConstants.forceTouchAvailable` in Gestures.js, this results in `<ForceTouchFallback />` being rendered rather than the actual gesture handler.

I wasn't easily able to see which RN version broke this, but since Gestures.js is the only file that imports PlatformConstants.js, I think we could just `import { Platform } from 'react-native'` here and query `Platform.constants.forceTouchAvailable` rather than conditionally import `NativeModules` if on RN v0.60.x, and from `Platform` if on newer RN versions. That being said, I imagine for future development it's nice to have a centralized PlatformConstants file. I just don't know how the software mansion team would approach this issue while maintaining backwards compatibility.

## Steps To Reproduce
Rather than upgrading the entire react-native-gesture-library to support the latest react-native v0.63.3, I just used the reanimated v2 playground repo.
1. `git clone https://github.com/software-mansion-labs/reanimated-2-playground`
2. npm install, pod install
3. Paste the example code below.
4. To fix, replace `import PlatformConstants from './PlatformConstants';` with `import { Platform } from 'react-native';` in node_modules/react-native-gesture-handler/Gestures.js, and change both instances of `PlatformConstants && PlatformConstants.forceTouchAvailable` with `Platform.constants.forceTouchAvailable`.

### Expected behavior
ForceTouchGestureHandler works as expected on force touch-capable devices running latest react-native versions.

### Actual behavior
`<ForceTouchFallback />` is rendered on devices that do in fact support a force touch handler.

Here is a force touch gesture handler example using reanimated v2, though the issue would still be present in reanimated v1:
```
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  useAnimatedGestureHandler,
import React from 'react';
import { ForceTouchGestureHandler } from 'react-native-gesture-handler';

export default function AnimatedStyleUpdateExample(props) {
  const force = useSharedValue(0);

  const animatedBoxStyles = useAnimatedStyle(() => {
    return {
      transform: [
        { scale: 1 + force.value }
      ]
    };
  });

  const onForceTouchGestureEvent = useAnimatedGestureHandler({
    onStart: (evt) => {
      force.value = evt.force
    },
    onActive: (evt) => {
      force.value = evt.force
    },
    onFinish: () => {
      force.value = 0
    }
  })

  return (
      <ForceTouchGestureHandler
        minForce={0}
        feedbackOnActivation
        onGestureEvent={onForceTouchGestureEvent}>
        <Animated.View style={[styles.box, animatedBoxStyles]} />
      </ForceTouchGestureHandler>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center'
  },
  box: {
    width: 100,
    height: 100,
    backgroundColor: 'red'
  }
})
```

## Package versions
```
"react": "16.13.1",
"react-native": "0.63.1",
"react-native-gesture-handler": "^1.6.1",
"react-native-reanimated": "2.0.0-alpha.7",
```

drewandre:
## Description

Fixes #1210 by importing platform constants from react-native's `Platform` API instead of `NativeModules.PlatformConstants`. ForceTouchGestureHandler is the only handler that requires this file, so this change only affects this handler. It is also backwards-compatible.

I'm a bit confused why PlatformConstants would ever be used over Platform. They have a slightly different data structure, but as far as I can tell the constants are the same.

## Test plan

I tested this on the [gesture-handler example app](https://github.com/software-mansion/react-native-gesture-handler/tree/master/Example) which is running RN v0.61.2, as well as the [reanimated v2 playground repo](https://github.com/software-mansion-labs/reanimated-2-playground) which runs RN v0.63.1. I did not test on web, but PlatformConstants is not a file required by any web-specific files. There is a PlatformConstants.web.js file, but the getter defined in this file is the same data structure as exporting `Platform.constants` in PlatformConstants.js for mobile.


PlatformConstants.js
```diff
@@ -1,3 +1,3 @@
-import { NativeModules } from 'react-native';
+import { Platform } from 'react-native';
```
jkadamczyk, line 1:
```suggestion
import { NativeModules, Platform } from 'react-native';

```

PlatformConstants.js
```diff
@@ -1,3 +1,3 @@
...
 
-export default NativeModules.PlatformConstants;
+export default Platform.constants;
```
jkadamczyk, line 3:
```suggestion
export default NativeModules?.PlatformConstants ?? Platform.constants;

```

jkadamczyk (changes requested):
Hey @drewandre 
I added a suggestion on how it can be implemented in case someone is using an older version and `PlatformConstants` is still available.
@jakub-gonet Please let me know if I'm wrong with the solution in any way 😄

jkadamczyk:
@drewandre Thanks for the fixes, I'll try to look into this today and get it merged. Have a good day!

jkadamczyk:
E2E tests are kind of not working I have to investigate why 🤷

jakub-gonet:
Updated branch with master. Sorry for introducing a bunch of unrelated lint changes.

jakub-gonet:
CI is failing due to missing `Platform.constants` types.
I submitted https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51146 so we should update RN types after it's merged and merge this one.

jakub-gonet (approved):
Thanks!

PR #1217, opened 2020-10-22
Use weak reference for storing rootViews in iOS

### LINKED ISSUE #1207: RNGestureHandlerManager's _rootViews causes memory leak
## Description

In `RNGestureHandlerManager.m`, `_rootViews` is a `NSMutableSet` which will retain the elements.

```
NSMutableSet<UIView*> *_rootViews;
```

In fact, the element in _rootViews is `RCTRootContentView`.

When a `RCTRootView` is created, the `RCTRootContentView` will be pushed into the `_rootViews`.

In my case, `RCTBridge` is cached after remove a `RCTRootView`.  `RCTRootView` will dealloc, but `RCTRootContentView` will not dealloc.

Then I allocate a new `RCTRootView` which is inited by my cached `RCTBridge`, a new `RCTRootContentView` will be allocated and it will also be retained by `RNGestureHandlerManager`. At present, the `_rootViews` has two `RCTRootContentView` element, the old and the new. 

I think that the `_rootViews` should not strong retain the element.

In my opinion, the solution is that using `NSHashTable` instead of `NSMutableSet`

```
NSHashTable<UIView *> *_rootViews;
_rootViews = [NSHashTable hashTableWithOptions:NSPointerFunctionsWeakMemory];
```

## Steps To Reproduce

1.

### Expected behavior

When a `RCTRootView` is removed, during the `RCTBridge` is cached, the `RCTRootContentView` should be dealloced

### Actual behavior

When a `RCTRootView` is removed, during the `RCTBridge` is cached, the `RCTRootContentView` did not be dealloced

## Package versions

- React: 16.11.0
- React Native: 0.62.2
- React Native Gesture Handler: 1.7.0

Norcy:
## Description

Please refer to [issue-1207](https://github.com/software-mansion/react-native-gesture-handler/issues/1207)

Fixes #1207

## Test plan

1. Create a RCTRootView, and cache the RCTBridge
2. Remove the RCTRootView, check the memory or the dealloc of RCTRootContentView
3. After my fix, the RCTRootContentView's dealloc will called.

### DIFF
--- a/ios/RNGestureHandlerManager.m
+++ b/ios/RNGestureHandlerManager.m
@@ -40,7 +40,7 @@
 {
     RNGestureHandlerRegistry *_registry;
     RCTUIManager *_uiManager;
-    NSMutableSet<UIView*> *_rootViews;
+    NSHashTable<UIView *> *_rootViews;
     RCTEventDispatcher *_eventDispatcher;
 }
 
@@ -51,7 +51,7 @@
         _uiManager = uiManager;
         _eventDispatcher = eventDispatcher;
         _registry = [RNGestureHandlerRegistry new];
-        _rootViews = [NSMutableSet new];
+        _rootViews = [NSHashTable hashTableWithOptions:NSPointerFunctionsWeakMemory];
     }
     return self;
 }


jakub-gonet (approved):
Looks good to me, thanks!

PR #1220, opened 2020-10-25
Fix destructuring `ref.props` on the web

### LINKED ISSUE #1164: TypeError: Cannot destructure property 'onGestureHandlerStateChange' of 'this.ref.props' as it is undefined.
This is a bug in react native web. I want to detect double tap using *TapGestureHandler*. It works perfectly fine in Android and iOS but in Web, I get this error:
```
TypeError: Cannot destructure property 'onGestureHandlerStateChange' of 'this.ref.props' as it is undefined.
TapGestureHandler.GestureHandler.sendEvent
node_modules/react-native-gesture-handler/web/GestureHandler.js:151
  148 | 
  149 | sendEvent = nativeEvent => {
  150 |   const {
> 151 |     onGestureHandlerStateChange: onHandlerStateChange,
      | ^  152 |     onGestureHandlerEvent: onGestureEvent,
  153 |   } = this.ref.props;
  154 | 
```
## Versions
react-native: 0.62.2
react-native-web: 0.13.6
react-native-gesture-handler: 1.7.0

## Sample code to generate issue
```typescript
// Tapper.tsx
import React from 'react';
import {
  State,
  TapGestureHandler,
  TapGestureHandlerStateChangeEvent,

type Props = {
  onSingleTap?: () => void;
  onDoubleTap?: () => void;
  children: React.ReactNode;
};

const Component: React.FC<Props> = ({ children, onDoubleTap, onSingleTap }) => {
  const doubleTapRef = React.useRef<TapGestureHandler>();
  const handleSingleHandlerStateChange = React.useCallback(
    (event: TapGestureHandlerStateChangeEvent) => {
      if (event.nativeEvent.state === State.ACTIVE) {
        onSingleTap?.();
      }
    },
    [onSingleTap]
  );

  const handleDoubleHandlerStateChange = React.useCallback(
    (event: TapGestureHandlerStateChangeEvent) => {
      if (event.nativeEvent.state === State.ACTIVE) {
        onDoubleTap?.();
      }
    },
    [onDoubleTap]
  );

  return (
    <TapGestureHandler
      onHandlerStateChange={handleSingleHandlerStateChange}
      waitFor={doubleTapRef}
    >
      <TapGestureHandler
        // @ts-ignore
        ref={doubleTapRef}
        onHandlerStateChange={handleDoubleHandlerStateChange}
        numberOfTaps={2}
      >
        {children}
      </TapGestureHandler>
    </TapGestureHandler>
  );
};

Component.displayName = 'Tapper';

export default Component;
```

Just use this component with any component in web.

hosseinmd:
## Description

Fixes #1164.

It seems there is no way to get props from ref after react-native-web migrate to functional component, so I passed props from Handler component to gestureHandler class. Now that works well.

## Test plan

```jsx
import React from "react";
import { Text, View } from "react-native";
import {
  PanGestureHandler,
  TapGestureHandler,
} from "react-native-gesture-handler";

function App() {
  return (
    <>
      <Text>Log PanGestureHandler </Text>
      <View
        style={{
          backgroundColor: "yellow",
          overflow: "hidden",
        }}
      >
        <PanGestureHandler
          onGestureEvent={(gestureEvent) => {
            console.log({ gestureEvent });
          }}
        >
          <TapGestureHandler
            onHandlerStateChange={(stateEvent) => {
              console.log({ stateEvent });
            }}
          >
            <View
              style={{
                backgroundColor: "blue",
                height: 100,
                width: 200,
                margin: 20,
                alignItems: "center",
                justifyContent: "center",
              }}
            ></View>
          </TapGestureHandler>
        </PanGestureHandler>
      </View>
    </>
  );
}

export default App;
```

### DIFF
--- a/createHandler.js
+++ b/createHandler.js
@@ -189,46 +189,36 @@ export default function createHandler(
 
     _createGestureHandler = newConfig => {
       this._config = newConfig;
+      this._events = this._getEventProps();
 
       RNGestureHandlerModule.createGestureHandler(
         handlerName,
         this._handlerTag,
-        newConfig
+        newConfig,
+        Platform.OS === 'web' ? this._events : undefined
       );
     };
 
     _attachGestureHandler = newViewTag => {
       this._viewTag = newViewTag;
+      this._events = this._getEventProps();
 
-      if (Platform.OS === 'web') {
-        RNGestureHandlerModule.attachGestureHandler(
-          this._handlerTag,
-          newViewTag,
-          this.props
-        );
-      } else {
-        RNGestureHandlerModule.attachGestureHandler(
-          this._handlerTag,
-          newViewTag
-        );
-      }
+      RNGestureHandlerModule.attachGestureHandler(
+        this._handlerTag,
+        newViewTag,
+        Platform.OS === 'web' ? this._events : undefined
+      );
     };
 
     _updateGestureHandler = newConfig => {
+      this._events = this._getEventProps();
       this._config = newConfig;
 
-      if (Platform.OS === 'web') {
-        RNGestureHandlerModule.updateGestureHandler(
-          this._handlerTag,
-          newConfig,
-          this.props
-        );
-      } else {
-        RNGestureHandlerModule.updateGestureHandler(
-          this._handlerTag,
-          newConfig
-        );
-      }
+      RNGestureHandlerModule.updateGestureHandler(
+        this._handlerTag,
+        newConfig,
+        Platform.OS === 'web' ? this._events : undefined
+      );
     };
 
     componentWillUnmount() {
@@ -282,9 +272,24 @@ export default function createHandler(
         { ...(this.constructor.propTypes || propTypes), ...customNativeProps },
         config
       );
-      if (!deepEqual(this._config, newConfig)) {
+      if (!deepEqual(this._config, newConfig) || this._isEventsPropsUpdated()) {
         this._updateGestureHandler(newConfig);
       }
+    }
+
+    _getEventProps() {
+      const { onHandlerStateChange, onGestureEvent } = this.props;
+      return { onHandlerStateChange, onGestureEvent };
+    }
+
+    _isEventsPropsUpdated() {
+      if (Platform.OS !== 'web') {
+        return false;
+      }
+
+      const _events = this._getEventProps();
+
+      return !deepEqual(this._events, _events);
     }
 
     setNativeProps(updates) {
--- a/createHandler.js
+++ b/createHandler.js
@@ -141,6 +141,8 @@ export default function createHandler(
       super(props);
       this._handlerTag = handlerTag++;
       this._config = {};
+      this._propsRef = React.createRef(props);
+
       if (props.id) {
         if (handlerIDToTag[props.id] !== undefined) {
           throw new Error(`Handler with ID "${props.id}" already registered`);
@@ -189,36 +191,28 @@ export default function createHandler(
 
     _createGestureHandler = newConfig => {
       this._config = newConfig;
-      this._events = this._getEventProps();
 
       RNGestureHandlerModule.createGestureHandler(
         handlerName,
         this._handlerTag,
-        newConfig,
-        Platform.OS === 'web' ? this._events : undefined
+        newConfig
       );
     };
 
     _attachGestureHandler = newViewTag => {
       this._viewTag = newViewTag;
-      this._events = this._getEventProps();
 
       RNGestureHandlerModule.attachGestureHandler(
         this._handlerTag,
         newViewTag,
-        Platform.OS === 'web' ? this._events : undefined
+        Platform.OS === 'web' ? this._propsRef : undefined
       );
     };
 
     _updateGestureHandler = newConfig => {
-      this._events = this._getEventProps();
       this._config = newConfig;
 
-      RNGestureHandlerModule.updateGestureHandler(
-        this._handlerTag,
-        newConfig,
-        Platform.OS === 'web' ? this._events : undefined
-      );
+      RNGestureHandlerModule.updateGestureHandler(this._handlerTag, newConfig);
     };
 
     componentWillUnmount() {
@@ -255,10 +249,14 @@ export default function createHandler(
           config
         )
       );
+      this._propsRef.current = this.props;
+
       this._attachGestureHandler(findNodeHandle(this._viewNode));
     }
 
     componentDidUpdate() {
+      this._propsRef.current = this.props;
+
       const viewTag = findNodeHandle(this._viewNode);
       if (this._viewTag !== viewTag) {
         this._attachGestureHandler(viewTag);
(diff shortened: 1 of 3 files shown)


connectdotz:
hey @hosseinmd thanks for submitting this PR, I tried it and found a problem: in `createHandler.js` the handler's props are set upon `componentDidMount()` but was never updated in `componentDidUpdate()` when the props changed, for example, a new `onGestureEvent`...

hosseinmd:
> hey @hosseinmd thanks for submitting this PR, I tried it and found a problem: in `createHandler.js` the handler's props are set upon `componentDidMount()` but was never updated in `componentDidUpdate()` when the props changed, for example, a new `onGestureEvent`...

Done

calumjames:
Could this be released as a hotfix please? It stops gesture support on the latest stable version of React Native for Web. Thanks.

connectdotz:
@hosseinmd still not quite working... looks like the `_update` method `if (!deepEqual(this._config, newConfig))` will ignore some props like `onGestureEvent`... see `filterConfig()`

hosseinmd:
@connectdotz Because of deepEqual checking, I just gave it the needed events in props, not the whole props.

connectdotz:
@hosseinmd looks like we are heading down toward a more and more involved solutioin... what do you think if we pass the `this.props` in `createHandler._attachGestureHandler` as a ref instead? and in the `componentDidUpdate()` we just update the ref always `this._propsRef.current = this.props`. Then we don't need to touch any other methods to keep the change scope minimal...?

hosseinmd:
Yes, but something is not right, _attachGestureHandler running just when viewTag change
```js
      const viewTag = findNodeHandle(this._viewNode);
      if (this._viewTag !== viewTag) {
        this._attachGestureHandler(viewTag);
      }
```

connectdotz:
you don't need to call `_attachGestureHandler` again in `componentDidUpdate()`, just update the ref directly should work:
```
componentDidUpdate() {
      const viewTag = findNodeHandle(this._viewNode);
      if (this._viewTag !== viewTag) {
        this._attachGestureHandler(viewTag);
      }
      this._propsRef.current = this.props;
      this._update();
    }
```

hosseinmd:
All right, I understand that now.

createHandler.js
```diff
@@ -200,7 +202,11 @@ export default function createHandler(
     _attachGestureHandler = newViewTag => {
       this._viewTag = newViewTag;
 
-      RNGestureHandlerModule.attachGestureHandler(this._handlerTag, newViewTag);
+      RNGestureHandlerModule.attachGestureHandler(
+        this._handlerTag,
+        newViewTag,
+        Platform.OS === 'web' ? this._propsRef : undefined
+      );
```
connectdotz, line 209: the ios ci build failed, I wonder if you do need to separate the API calls by the platform like in your earlier commit...

connectdotz (commented):
great, I think it looks a lot cleaner now. let's fix the ci build failure...

hosseinmd:
Off topic, examples have old versions of modules, perhaps should be updated to new versions.

connectdotz (approved):
Thanks for fixing it, it looks good now 👍 . I tested it in my use cases and can verify that it worked. Hopefully, this can be released soon as many people are blocked by this issue...

chadhobson:
Thanks to both of you. I can also confirm this fixes the issue.

Here's a patch-package to get folks back on track until this gets merged.

[react-native-gesture-handler+1.8.0.patch.zip](https://github.com/software-mansion/react-native-gesture-handler/files/5465600/react-native-gesture-handler%2B1.8.0.patch.zip)

sekonia:
I ran into the same problems... thank you @hosseinmd and @connectdotz for fixing this and @chadhobson for the patch. Looking forward to a release update. 🎉

web/GestureHandler.js
```diff
@@ -206,13 +203,14 @@ class GestureHandler {
...
 
-  setView(ref) {
+  setView(ref, propsRef) {
     if (ref == null) {
       this.destroy();
       this.view = null;
       return;
     }
 
+    this._propsRef = propsRef;
```
jakub-gonet, line 213: I think this shouldn't be underscored here to be consistent with other properties.

web/GestureHandler.js
```diff
@@ -147,10 +147,7 @@ class GestureHandler {
   }
 
   sendEvent = nativeEvent => {
-    const {
-      onGestureHandlerStateChange: onHandlerStateChange,
-      onGestureHandlerEvent: onGestureEvent,
-    } = this.ref.props;
+    const { onHandlerStateChange, onGestureEvent } = this._propsRef.current;
```
jakub-gonet, line 150: Why did you remove destructuring here? `onGestureHandlerStateChange` and `onHandlerStateChange` (note *Gesture* part) aren't interchangeable.

```
 const { onGestureHandlerStateChange: onHandlerStateChange,	
         onGestureHandlerEvent:       onGestureEvent, } = (...)
```

jakub-gonet (commented):
This PR lack a proper test plan. Could you write some small repro example on which we can test it?

I left some comments inline.

hosseinmd:
> This PR lack a proper test plan. Could you write some small repro example on which we can test it?
> 
> I left some comments inline.

Thanks for your review, for test this we should update the example of this repo to the new version of react-native-web.

web/GestureHandler.js
hosseinmd: Because with take them directly from props and they defined with `onHandlerStateChange, onGestureEvent` in props.

jakub-gonet:
I think this is its own issue, our example projects are terribly outdated. I'd be content with a link to the repo using rn for web & rngh with some small code example.

web/GestureHandler.js
hosseinmd: Excuse me, I don't understand, please explain what's your idea I will do it.

hosseinmd:
> I think this is its own issue, our example projects are terribly outdated. I'd be content with a link to the repo using rn for web & rngh with some small code example.

Ok, I will do it as soon as posible.

web/GestureHandler.js
jakub-gonet: I'm just saying that the rest of the properties don't have `_` prefix, so I think this should be named `propsRef` instead of `_propsRef`

web/GestureHandler.js
jakub-gonet: I'm afraid I don't understand, could you elaborate?

My concern is, we set those two properties in [`React.cloneElement`](https://github.com/software-mansion/react-native-gesture-handler/blob/7564162fe6bf4ddb04f2ce18f8dfd4dfe905c98c/createHandler.js#L347-L348) (this is used to nest handlers). I may be wrong, but it won't ignore those?

Have you tested your PR against nested handlers? If it works then we're okay.

hosseinmd:
@jakub-gonet Please test this repo https://github.com/hosseinmd/test-gesture-handler

web/GestureHandler.js
hosseinmd: You are totally right I understand that not, because this reanimated useAnimatedGestureHandler don't work with this PR.

hosseinmd:
I have done what you say. I added two examples to this repo https://github.com/hosseinmd/test-gesture-handler, but something is wrong yet because useAnimatedGestureHandler doesn't work.

jakub-gonet:
Thanks, I'll try to look into it in the next few days.

jakub-gonet:
Ok, I checked it and seems like the error you were experiencing was fixed in [react-native-reanimated#1402](https://github.com/software-mansion/react-native-reanimated/pull/1402) (yet unreleased). I applied this patch locally and seems like "Log PanGestureHandler" and "Bouncing" example work. The swipeable list doesn't but it also isn't crashing with `ref.props` error.

I also checked nested handlers and it works alright. I'll update the PR description with the example you wrote and I think we're good to go.

jakub-gonet (approved):

hosseinmd:
I tested it in my project and it works. Except for useAnimatedGestureHandler, I think it's not related to this error, I will work on it in another pr. please merge this. Thank you.

jakub-gonet:
Sure, thanks for your contribution!

PR #1275, opened 2020-12-14
Adjust overshootFriction behavior

kolking:
## Description

I found the `overshootFriction` behavior slightly confusing:
- First problem: the overshoot friction starts to work only when the corresponding panel's `overshoot[Left | Right]` props is set to `false` which makes no sense because `overshootRight = false` literally means "disable overshoot for the right action panel".
```js
// overshootFriction won't be applied for the right panel,
// it will be applied for the left (empty) panel instead:
<Swipeable overshootFriction={3} renderRightActions={renderRightActions}>

// And to make it work for the right panel you have to:
<Swipeable overshootFriction={3} renderRightActions={renderRightActions} overshootRight={false}>
```
- Second problem: when `overshootFriction` is greater than `1`, you can't disable overshoot for the panels. For example, you want only the right action panel, and you want to disable overshoot for the left panel (because there are no actions):
```js
// It still can be pulled to the right, despite the fact there is no
// renderLeftActions and you even explicitly set overshootLeft = false
<Swipeable overshootFriction={3} renderRightActions={renderRightActions} overshootLeft={false}>
```

This PR fixes the above issues. The `overshootFriction` will be applied only for the action panels where the overshoot behavior is allowed. Overshoot behavior can be explicitly disabled by setting `overshoot[Left | Right] = false` for the appropriate panel.


jakub-gonet (approved):
I'd like to only include the second change and move the first one to the separate PR since it's breaking. Could you resolve conflicts as well?

PR #1273, opened 2020-12-12
Rewrite Examples to TS

jakub-gonet:
### Session 1: Example Blues

Fixes #1017.

This is the first part of moving into Typescript. Using TS types in Examples will allow us to test breaking changes to TS types API in the future PRs.

#### Note

During the rewrite, we spotted that RNGH `TouchableNativeFeedback` type in `react-native-gesture-handler.d.ts` does not expose static fields of RN's implementation. This is something that should be fixed in follow-up PRs.

## Changes

- Removed unused bits in examples
- Regenerated Example app with latest RN
- Added Eslint & Prettier config file in the root directory
- Removed stray files from examples/
- Updated Travis config to include newer Android build tools & iOS scheme name change
- Moved Example code into src/
- Removed Example tests

## Test plan

I ran a `diff` on two `npm pack`ed version of the library and only changes that were present were related to reformatting.

Manually investigating TS errors & running example app.

## TODO

- [x] Rewrite every example to TS
- [x] Make sure this change won't break packaging the library
- [ ] Test every example on Android
- [x] Test every example on iOS
- [ ] ~~Test every example on Web~~ We don't have setup ready, this can be verified in the follow-up PRs.

PR #1276, opened 2020-12-15
Fix typo in example alert output

cmlarsen:


jakub-gonet:
This is included in #1273.

Thanks!

PR #1280, opened 2020-12-17
Migrate source files to Typescript

jakub-gonet:
### Session 2: Stray Typescript Strut

This PR renames files in RNGH & adds some TS config to prepare sources for the next PR that will add proper types for them.

## Changes

Most of the files were renamed to use .ts(x) extension and given Eslint and TS ignores to ensure that errors won't prevent RNGH from building. They should be gradually removed as we add types.

## `Package.json` changes

`dist/` directory has been included in `package.json` `files` property to ensure we're using transpiled sources. This created a problem with importing submodules (Swipeable, DrawerLayout) because setting the entry point doesn't change root directory for resolution. This makes distributing those much harder because we have to keep them as js files OR make some sort of hack to ensure importing them via `import Swipeable from 'react-native-gesture-handler/Swipeable';` will still work. 

The solution is to provide `exports` property to map `dist` structure to root directory and shadow it. 

However, this works on Node 14+. In order to preserve backwards compatibility, `Swipeable.js` and `DrawerLayout.js` are exporting corresponding modules from `dist/` directory. Alternative solutions to this problem would be to copy built js files to root dir or compile TS files into their location (removing dist/ directory). IMO the current way of doing this is the least invasive way.

See https://github.com/microsoft/TypeScript/issues/8305 and https://github.com/nodejs/node/issues/14970#issuecomment-571887546 for more context.

## Test plan

I manually packed RNGH and used a local file in the example project to test it. I tested Swipeable and draggable examples to test that RNGH still works correctly.

<details>
<summary>Diff between bundles of master and this branch (314bb40) versions of RNGH</summary>

```diff
1486c1486
< },2,[1,3,481,482,484,14,479,485,487,488,490,491,492],"node_modules/react-native-gesture-handler/index.js");
---
> },2,[1,3,481,482,484,14,479,485,487,488,490,491,492],"node_modules/react-native-gesture-handler/dist/index.js");
1505c1505
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/handlers/createNativeWrapper.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/handlers/createNativeWrapper.js";
1546,1547c1546,1547
<           lineNumber: 61,
<           columnNumber: 7
---
>           lineNumber: 54,
>           columnNumber: 17
1554c1554
<           lineNumber: 64,
---
>           lineNumber: 55,
1564c1564
< },3,[1,4,6,7,8,14],"node_modules/react-native-gesture-handler/src/handlers/createNativeWrapper.js");
---
> },3,[1,4,6,7,8,14],"node_modules/react-native-gesture-handler/dist/src/handlers/createNativeWrapper.js");
3758c3758
< },14,[1,7,15,21,480],"node_modules/react-native-gesture-handler/src/handlers/NativeViewGestureHandler.js");
---
> },14,[1,7,15,21,480],"node_modules/react-native-gesture-handler/dist/src/handlers/NativeViewGestureHandler.js");
4690a4691,4692
>   var _interopRequireWildcard = _$$_REQUIRE(_dependencyMap[1], "@babel/runtime/helpers/interopRequireWildcard");
> 
4696c4698
<   var _classCallCheck2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[1], "@babel/runtime/helpers/classCallCheck"));
---
>   var _classCallCheck2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[2], "@babel/runtime/helpers/classCallCheck"));
4698c4700
<   var _createClass2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[2], "@babel/runtime/helpers/createClass"));
---
>   var _createClass2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[3], "@babel/runtime/helpers/createClass"));
4700c4702
<   var _inherits2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[3], "@babel/runtime/helpers/inherits"));
---
>   var _inherits2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[4], "@babel/runtime/helpers/inherits"));
4702c4704
<   var _possibleConstructorReturn2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[4], "@babel/runtime/helpers/possibleConstructorReturn"));
---
>   var _possibleConstructorReturn2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[5], "@babel/runtime/helpers/possibleConstructorReturn"));
4704c4706
<   var _getPrototypeOf2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[5], "@babel/runtime/helpers/getPrototypeOf"));
---
>   var _getPrototypeOf2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[6], "@babel/runtime/helpers/getPrototypeOf"));
4706c4708
<   var _defineProperty2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[6], "@babel/runtime/helpers/defineProperty"));
---
>   var _defineProperty2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[7], "@babel/runtime/helpers/defineProperty"));
4708c4710
<   var _react = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[7], "react"));
---
>   var _react = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[8], "react"));
4710c4712
<   var _reactNative = _$$_REQUIRE(_dependencyMap[8], "react-native");
---
>   var _reactNative = _interopRequireWildcard(_$$_REQUIRE(_dependencyMap[9], "react-native"));
4712c4714
<   var _areEqual = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[9], "fbjs/lib/areEqual"));
---
>   var _areEqual = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[10], "fbjs/lib/areEqual"));
4714c4716
<   var _RNGestureHandlerModule = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[10], "../RNGestureHandlerModule"));
---
>   var _RNGestureHandlerModule = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[11], "../RNGestureHandlerModule"));
4716c4718
<   var _State = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[11], "../State"));
---
>   var _State = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[12], "../State"));
5037c5039
<           if (_reactNative.Touchable.TOUCH_TARGET_DEBUG && child.type && (child.type === 'RNGestureHandlerButton' || child.type.name === 'View' || child.type.displayName === 'View')) {
---
>           if (_reactNative.default.Touchable.TOUCH_TARGET_DEBUG && child.type && (child.type === 'RNGestureHandlerButton' || child.type.name === 'View' || child.type.displayName === 'View')) {
5039c5041
<             grandChildren.push(_reactNative.Touchable.renderDebugView({
---
>             grandChildren.push(_reactNative.default.Touchable.renderDebugView({
5058c5060
< },21,[1,22,23,24,26,28,7,8,29,477,478,479],"node_modules/react-native-gesture-handler/src/handlers/createHandler.js");
---
> },21,[1,4,22,23,24,26,28,7,8,29,477,478,479],"node_modules/react-native-gesture-handler/dist/src/handlers/createHandler.js");
99193c99195
< },478,[29],"node_modules/react-native-gesture-handler/src/RNGestureHandlerModule.js");
---
> },478,[29],"node_modules/react-native-gesture-handler/dist/src/RNGestureHandlerModule.js");
99220c99222
< },479,[],"node_modules/react-native-gesture-handler/src/State.js");
---
> },479,[],"node_modules/react-native-gesture-handler/dist/src/State.js");
99258c99260
< },480,[1,15],"node_modules/react-native-gesture-handler/src/GestureHandlerPropTypes.js");
---
> },480,[1,15],"node_modules/react-native-gesture-handler/dist/src/GestureHandlerPropTypes.js");
99271c99273
< },481,[1,478],"node_modules/react-native-gesture-handler/src/Directions.js");
---
> },481,[1,478],"node_modules/react-native-gesture-handler/dist/src/Directions.js");
99290c99292
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/gestureHandlerRootHOC.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/gestureHandlerRootHOC.js";
99301,99302c99303,99304
<           lineNumber: 12,
<           columnNumber: 7
---
>           lineNumber: 10,
>           columnNumber: 17
99308c99310
<           lineNumber: 13,
---
>           lineNumber: 11,
99324c99326
< },482,[1,6,8,29,483,484],"node_modules/react-native-gesture-handler/src/gestureHandlerRootHOC.js");
---
> },482,[1,6,8,29,483,484],"node_modules/react-native-gesture-handler/dist/src/gestureHandlerRootHOC.js");
99424,99425d99425
<   var _interopRequireDefault = _$$_REQUIRE(_dependencyMap[0], "@babel/runtime/helpers/interopRequireDefault");
< 
99431,99433c99431
<   var _react = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[1], "react"));
< 
<   var _reactNative = _$$_REQUIRE(_dependencyMap[2], "react-native");
---
>   var _reactNative = _$$_REQUIRE(_dependencyMap[0], "react-native");
99437c99435
< },484,[1,8,29],"node_modules/react-native-gesture-handler/src/GestureHandlerRootView.js");
---
> },484,[29],"node_modules/react-native-gesture-handler/dist/src/GestureHandlerRootView.js");
99480c99478
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/components/GestureButtons.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/components/GestureButtons.js";
99555,99556c99553,99554
<             lineNumber: 66,
<             columnNumber: 7
---
>             lineNumber: 48,
>             columnNumber: 17
99621,99622c99619,99620
<             lineNumber: 115,
<             columnNumber: 7
---
>             lineNumber: 80,
>             columnNumber: 17
99637c99635
<             lineNumber: 119,
---
>             lineNumber: 81,
99691,99692c99689,99690
<             lineNumber: 167,
<             columnNumber: 7
---
>             lineNumber: 115,
>             columnNumber: 17
99708c99706
< },485,[1,7,6,70,22,23,24,26,28,15,8,29,3,486,479],"node_modules/react-native-gesture-handler/src/components/GestureButtons.js");
---
> },485,[1,7,6,70,22,23,24,26,28,15,8,29,3,486,479],"node_modules/react-native-gesture-handler/dist/src/components/GestureButtons.js");
99720c99718
< },486,[29],"node_modules/react-native-gesture-handler/src/components/GestureHandlerButton.js");
---
> },486,[29],"node_modules/react-native-gesture-handler/dist/src/components/GestureHandlerButton.js");
99732c99730
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/components/GestureComponents.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/components/GestureComponents.js";
99792,99793c99790,99791
<                   lineNumber: 53,
<                   columnNumber: 13
---
>                   lineNumber: 47,
>                   columnNumber: 151
99800,99801c99798,99799
<               lineNumber: 49,
<               columnNumber: 9
---
>               lineNumber: 47,
>               columnNumber: 67
99811c99809
< },487,[1,6,8,29,3],"node_modules/react-native-gesture-handler/src/components/GestureComponents.js");
---
> },487,[1,6,8,29,3],"node_modules/react-native-gesture-handler/dist/src/components/GestureComponents.js");
100075c100073
< },488,[1,22,23,24,26,28,7,15,8,21,480,489],"node_modules/react-native-gesture-handler/src/handlers/Gestures.js");
---
> },488,[1,22,23,24,26,28,7,15,8,21,480,489],"node_modules/react-native-gesture-handler/dist/src/handlers/Gestures.js");
100086c100084
< },489,[29],"node_modules/react-native-gesture-handler/src/PlatformConstants.js");
---
> },489,[29],"node_modules/react-native-gesture-handler/dist/src/PlatformConstants.js");
100119c100117
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/components/Swipeable.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/components/Swipeable.js";
100136c100134
<     function Swipeable(_props) {
---
>     function Swipeable(props) {
100140c100138
<       _this = _super.call(this, _props);
---
>       _this = _super.call(this, props);
100344,100345c100342
<       var _dragX = new _reactNative.Animated.Value(0);
< 
---
>       var dragX = new _reactNative.Animated.Value(0);
100347c100344
<         dragX: _dragX,
---
>         dragX: dragX,
100355c100352
<       _this._updateAnimatedEvent(_props, _this.state);
---
>       _this._updateAnimatedEvent(props, _this.state);
100359c100356
<           translationX: _dragX
---
>           translationX: dragX
100362c100359
<         useNativeDriver: _props.useNativeAnimations
---
>         useNativeDriver: props.useNativeAnimations
100394,100395c100391,100392
<             lineNumber: 284,
<             columnNumber: 7
---
>             lineNumber: 208,
>             columnNumber: 44
100407c100404
<             lineNumber: 290,
---
>             lineNumber: 213,
100421,100422c100418,100419
<             lineNumber: 299,
<             columnNumber: 7
---
>             lineNumber: 215,
>             columnNumber: 46
100434c100431
<             lineNumber: 305,
---
>             lineNumber: 220,
100447,100448c100444,100445
<             lineNumber: 314,
<             columnNumber: 7
---
>             lineNumber: 222,
>             columnNumber: 17
100456c100453
<             lineNumber: 319,
---
>             lineNumber: 223,
100465c100462
<             lineNumber: 324,
---
>             lineNumber: 226,
100478c100475
<             lineNumber: 327,
---
>             lineNumber: 227,
100505c100502
< },490,[1,4,6,7,22,23,24,26,28,8,29,488,479],"node_modules/react-native-gesture-handler/src/components/Swipeable.js");
---
> },490,[1,4,6,7,22,23,24,26,28,8,29,488,479],"node_modules/react-native-gesture-handler/dist/src/components/Swipeable.js");
100540c100537
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/components/DrawerLayout.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/components/DrawerLayout.js";
100560c100557
<     function DrawerLayout(_props, context) {
---
>     function DrawerLayout(props, context) {
100564c100561
<       _this = _super.call(this, _props, context);
---
>       _this = _super.call(this, props, context);
100810,100811c100807,100808
<             lineNumber: 402,
<             columnNumber: 7
---
>             lineNumber: 249,
>             columnNumber: 21
100820c100817
<             lineNumber: 403,
---
>             lineNumber: 250,
100882,100883c100879,100880
<             lineNumber: 466,
<             columnNumber: 7
---
>             lineNumber: 293,
>             columnNumber: 21
100891c100888
<             lineNumber: 467,
---
>             lineNumber: 294,
100902c100899
<             lineNumber: 483,
---
>             lineNumber: 306,
100910c100907
<             lineNumber: 488,
---
>             lineNumber: 307,
100921,100926c100918,100920
<       var _dragX = new _reactNative.Animated.Value(0);
< 
<       var _touchX = new _reactNative.Animated.Value(0);
< 
<       var _drawerTranslation = new _reactNative.Animated.Value(0);
< 
---
>       var dragX = new _reactNative.Animated.Value(0);
>       var touchX = new _reactNative.Animated.Value(0);
>       var drawerTranslation = new _reactNative.Animated.Value(0);
100928,100930c100922,100924
<         dragX: _dragX,
<         touchX: _touchX,
<         drawerTranslation: _drawerTranslation,
---
>         dragX: dragX,
>         touchX: touchX,
>         drawerTranslation: drawerTranslation,
100934c100928
<       _this._updateAnimatedEvent(_props, _this.state);
---
>       _this._updateAnimatedEvent(props, _this.state);
100975,100976c100969,100970
<             lineNumber: 525,
<             columnNumber: 7
---
>             lineNumber: 349,
>             columnNumber: 17
101019c101013
< },491,[1,4,7,22,23,27,24,26,28,8,30,29,488,479],"node_modules/react-native-gesture-handler/src/components/DrawerLayout.js");
---
> },491,[1,4,7,22,23,27,24,26,28,8,30,29,488,479],"node_modules/react-native-gesture-handler/dist/src/components/DrawerLayout.js");
101058c101052
< },492,[1,493,494,496,497],"node_modules/react-native-gesture-handler/src/components/touchables/index.js");
---
> },492,[1,493,494,496,497],"node_modules/react-native-gesture-handler/dist/src/components/touchables/index.js");
101069c101063
< },493,[29],"node_modules/react-native-gesture-handler/src/components/touchables/TouchableNativeFeedback.js");
---
> },493,[29],"node_modules/react-native-gesture-handler/dist/src/components/touchables/TouchableNativeFeedback.js");
101085c101079
<       _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/components/touchables/TouchableWithoutFeedback.js";
---
>       _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/components/touchables/TouchableWithoutFeedback.js";
101094,101095c101088,101089
<         lineNumber: 5,
<         columnNumber: 3
---
>         lineNumber: 4,
>         columnNumber: 68
101104c101098
< },494,[1,6,8,495],"node_modules/react-native-gesture-handler/src/components/touchables/TouchableWithoutFeedback.js");
---
> },494,[1,6,8,495],"node_modules/react-native-gesture-handler/dist/src/components/touchables/TouchableWithoutFeedback.js");
101139c101133
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/components/touchables/GenericTouchable.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/components/touchables/GenericTouchable.js";
101193,101198c101187
< 
<       for (var _len = arguments.length, args = new Array(_len), _key = 0; _key < _len; _key++) {
<         args[_key] = arguments[_key];
<       }
< 
<       _this = _super.call.apply(_super, [this].concat(args));
---
>       _this = _super.apply(this, arguments);
101393,101394c101382,101383
<             lineNumber: 260,
<             columnNumber: 7
---
>             lineNumber: 228,
>             columnNumber: 17
101401c101390
<             lineNumber: 271,
---
>             lineNumber: 229,
101420c101409
< },495,[1,4,7,6,22,23,24,26,28,8,29,479,485,15],"node_modules/react-native-gesture-handler/src/components/touchables/GenericTouchable.js");
---
> },495,[1,4,7,6,22,23,24,26,28,8,29,479,485,15],"node_modules/react-native-gesture-handler/dist/src/components/touchables/GenericTouchable.js");
101455c101444
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/components/touchables/TouchableOpacity.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/components/touchables/TouchableOpacity.js";
101474,101479c101463
< 
<       for (var _len = arguments.length, args = new Array(_len), _key = 0; _key < _len; _key++) {
<         args[_key] = arguments[_key];
<       }
< 
<       _this = _super.call.apply(_super, [this].concat(args));
---
>       _this = _super.apply(this, arguments);
101523,101524c101507,101508
<             lineNumber: 52,
<             columnNumber: 7
---
>             lineNumber: 40,
>             columnNumber: 17
101530c101514
<             lineNumber: 61,
---
>             lineNumber: 46,
101547c101531
< },496,[1,4,7,6,70,22,23,24,26,28,29,495,8,15],"node_modules/react-native-gesture-handler/src/components/touchables/TouchableOpacity.js");
---
> },496,[1,4,7,6,70,22,23,24,26,28,29,495,8,15],"node_modules/react-native-gesture-handler/dist/src/components/touchables/TouchableOpacity.js");
101582c101566
<   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/src/components/touchables/TouchableHighlight.js";
---
>   var _jsxFileName = "/Users/jgonet/Projects/work/RN/sandbox/v0_63/node_modules/react-native-gesture-handler/dist/src/components/touchables/TouchableHighlight.js";
101656,101657c101640,101641
<               lineNumber: 66,
<               columnNumber: 14
---
>               lineNumber: 56,
>               columnNumber: 20
101682,101683c101666,101667
<             lineNumber: 90,
<             columnNumber: 7
---
>             lineNumber: 66,
>             columnNumber: 17
101704c101688
< },497,[1,4,7,6,70,22,23,24,26,28,8,495,29,15],"node_modules/react-native-gesture-handler/src/components/touchables/TouchableHighlight.js");
---
> },497,[1,4,7,6,70,22,23,24,26,28,8,495,29,15],"node_modules/react-native-gesture-handler/dist/src/components/touchables/TouchableHighlight.js");
```
</details>

As you may see, most differences are caused by changing line and column numbers, using files from `dist/`, and renaming `_variables` to `variables`. TS version doesn't include argument array copying, but I don't think this is dangerous.


DrawerLayout.js
```diff
@@ -1,2 +1,15 @@
...
+via scoped import like "import Swipeable from 'react-native-gesture-handler/Swipeable';".
+
+Even if you set main file to be in the dist/ directory it doesn't mean that scoped file resolution
+will be performed in that director. Actually node will search in <root>/Swipeable (which doesn't exist with .js extension)
+
+This behavior can be mitigated via "exports" key in package.json but it's supported since Node 14 which is quite new.
+
+Please remove this file when we stop supporting node versions lower than 14.
+
+*/
```
jkadamczyk, line 13: Can it be done in a better way if we drop backward compatibility in this migration? Maybe it is worth posting a breaking change for the sake of less tech debt?

Swipeable.js
```diff
@@ -1,2 +1,15 @@
...
+
+Even if you set main file to be in the dist/ directory it doesn't mean that scoped file resolution
+will be performed in that director. Actually node will search in <root>/Swipeable (which doesn't exist with .js extension)
+
+This behavior can be mitigated via "exports" key in package.json but it's supported since Node 14 which is quite new.
+
+Please remove this file when we stop supporting node versions lower than 14.
+
+*/
+import Swipeable from './dist/Swipeable';
```
jkadamczyk, line 14: Same as above, can we avoid it?

src/components/DrawerLayout.tsx
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable eslint-comments/no-unlimited-disable */
+/* eslint-disable */
+// @ts-nocheck TODO(TS) provide types
 // @flow
```
WoLewicki, line 4: Why is flow used here?

src/components/Swipeable.tsx
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable eslint-comments/no-unlimited-disable */
+/* eslint-disable */
+// @ts-nocheck TODO(TS) provide types
 // @flow
```
WoLewicki, line 4: And here

tsconfig.json
```diff
@@ -26,5 +22,14 @@
     "noUnusedLocals": true,
     "outDir": "./dist"
   },
-  "include": ["**/*.ts", "**/*.tsx", "**/*.d.ts", "examples/Example/index.js"]
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx",
+    "src/**/*.d.ts",
```
WoLewicki, line 28: To be removed too.

DrawerLayout.js
jakub-gonet: We can't use scoped imports with TS, so possible solutions are:

- drop node <14 compat
- copy all files from `dist/` after transpiling to the root directory
- maybe try using symlinks? Not sure if this will work, IIRC metro doesn't support symlinks
- move Swipeable and Drawer as exports, dropping `react-native-gesture-handler/Swipeable` import

src/components/DrawerLayout.tsx
jakub-gonet: Swipeable and Drawer were typed using TS but with `// @flow` comment added. Not sure why, this will be removed after we migrate types.

tsconfig.json
jakub-gonet: This will be removed in the next PR.

WoLewicki: Just commenting for convenience along with other comments of removal.

DrawerLayout.js
jkadamczyk: The last one sounds like a fine solution, at least to me. But I might be missing something important there as well.

jakub-gonet: I think we should leave it for now and if the next PR breaks types, then we can do some more breaking changes.

WoLewicki: Can't we do something like `react-native-screens/native-stack`? It is in the `native-stack` folder in `src` with its own `index.tsx` file. Or am I missing something too?

jakub-gonet: Using `name.js` or `name/index.js` is equivalent. I'll look at screens config, maybe we can figure something out.

.eslintrc.json
```diff
@@ -8,14 +8,16 @@
     "react": { "version": "16" },
     "import/resolver": {
       "babel-module": {}
-    }
+    },
+    "import/core-modules": ["react-native-gesture-handler/Swipeable"]
```
jakub-gonet, line 12: This is to make eslint happy, TS can deal with that already.

package.json
```diff
@@ -3,13 +3,21 @@
...
-    "lint-check": "yarn prettier --check src/ examples/Example/src"
+    "ts-check": "yarn tsc --noEmit",
+    "lint-check": "yarn eslint src/ && yarn prettier --check src/"
   },
-  "main": "index.js",
+  "exports": {
+    ".": "./dist/index.js",
+    "./": "./dist/",
+    "./package.json": "./package.json",
+    "./RNGestureHandler.podspec": "./RNGestureHandler.podspec"
```
jakub-gonet, line 18: I'm not sure if any other files should be included here, let me know.

I'll test if this is needed after today's changes.

package.json
```diff
@@ -19,10 +27,13 @@
     "android/lib/src/main/java/",
     "ios/",
     "src/",
-    "index.js",
-    "react-native-gesture-handler.d.ts",
-    "DrawerLayout.d.ts",
+    "dist/",
+    "!dist/examples",
+    "index.ts",
+    "Swipeable.js",
```
jakub-gonet, line 33: To remove.

package.json
```diff
@@ -19,10 +27,13 @@
...
     "Swipeable.d.ts",
+    "DrawerLayout.js",
```
jakub-gonet, line 35: To remove.

tsconfig.json
```diff
@@ -1,10 +1,11 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
+    // TODO(TS) Remove after we migrate to TS entirely
     "paths": {
       "react-native-gesture-handler": ["./react-native-gesture-handler.d.ts"],
-      "react-native-gesture-handler/Swipeable": ["./Swipeable.d.ts"],
-      "react-native-gesture-handler/DrawerLayout": ["./DrawerLayout.d.ts"]
+      "react-native-gesture-handler/Swipeable": ["Swipeable.ts"],
```
jakub-gonet, line 7: Add `./`.

jakub-gonet: Fixed in 85ca6ed.

package.json
jakub-gonet: Fixed in f25dcf4.

package.json
jakub-gonet: Fixed in f25dcf4.

package.json
jakub-gonet: Removed in 314bb40.

Swipeable.js
jakub-gonet: Removed in 314bb40.

DrawerLayout.js
jakub-gonet: Fixed in 314bb40.

I managed to work around this by creating `{DrawerLayout, Swipeable}/package.json` and specifying the entry point there. It doesn't really differ from using `Swipeable.js` linking to `dist/` dir as it was done before but I find it a little clearer. It also allowed removing `exports` section in the `package.json`.

jakub-gonet:
I'm pretty sure this is ready so I'm going to merge it after CI passes.

PR #1290, opened 2020-12-27
Clamp swipe animation in examples

mhkeller:
## Description

Currently, when using this example and you overshoot swiping left, it drags the buttons to the left and adds white space in the space between each of them. This clamps the animation so it doesn't move the buttons with the overshoot. All other behavior remains the same as before.


jakub-gonet (approved):
Thanks!

PR #1291, opened 2020-12-28
Add TS types to the files

jakub-gonet:
### Session 3: Honky Tonk Rewrite

Last part of the TS rewrite process, in this PR we transfer types from declaration files to their implementations and add missing types for other functions and components.

Change lists are probably not exhaustive but we tried to write most of the changes.

### TS

- Removed declaration files for Swipeable and DrawerLayout components.
- Added types compatibility layer for the new types (`gestureHandlerTypesCompat.ts`).
- Added generation of declarations and declarations map when TS is transpiled.

#### Internal

- Added types for all files used in RN.
- Added basic types for the web version.
- Mark web base classes as abstract (DiscreteGH, DraggingGH, etc).

### Source code

- Added `"browser"` target and updated targets for source and types for Swipeable and DrawerLayout components `package.json`s.
- Removed babel config for the main module, moved to examples.
- Changed `"react-native"` entry point to `index.ts` (this ensures that clients using RN will load source instead of TS transpiled files).
- Changed `"types"` to `"dist/index.d.ts"`.

#### Internal

- Changed most underscore prefixed function to non-prefixed, private functions.
- Removed not used functions and arguments.
- Replaced `x && x()` with `x.?()` in various places.
- Removed prop types and replace them with lists of strings when necessary.

#### Internal

- [Changed some TS-specific Eslint rules to "warn" mode.](https://github.com/software-mansion/react-native-gesture-handler/pull/1291/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fR22-R26)
- Removed `no-use-before-define` rule.
- Example's Eslint config extends root config.
- Added support for the web version of Examples.
- `State` and `NativeViewGestureHandler` are now named export internally.
- Renamed `Gestures` to `gestureHandlers `.
- Added `lint` npm script.

## TODO

- [ ] Fill diff between master / ts-rewrite and this PR
- [x] Update links referencing files in the docs
- [ ] List all deprecations and breaking changes
- [x] List all notable changes
- [ ] Review TODOs, `eslint-ignore`s, `@ts-ignore`s, and `@ts-nocheck`s

## Missing types

- [x] `src/components/GestureComponents.web.tsx`
- [x] `src/web/DiscreteGestureHandler.ts`
- [x] `src/web/DraggingGestureHandler.ts`
- [x] `src/web/FlingGestureHandler.ts`
- [x] `src/web/GestureHandler.ts`
- [x] `src/web/IndiscreteGestureHandler.ts`
- [x] `src/web/LongPressGestureHandler.ts`
- [x] `src/web/NativeViewGestureHandler.ts`
- [x] `src/web/PanGestureHandler.ts`
- [x] `src/web/PinchGestureHandler.ts`
- [x] `src/web/PressGestureHandler.ts`
- [x] `src/web/RotationGestureHandler.ts`
- [x] `src/web/TapGestureHandler.ts`

## Further improvements mentioned in review comments

- https://github.com/software-mansion/react-native-gesture-handler/pull/1291#discussion_r549979788

## Test plan

This will be tested by manually checking examples.

Future PRs can be checked by comparing bundle output (create two packages from master and this branch using `npm pack`, install one and then second later, create two bundles via `yarn react-native bundle --entry-file=index.js --bundle-output=rngh.bundle --reset-cache --minify=false`, and compare results with `diff rngh.bundle rngh.ts.bundle`). This PR is too big (18k diff) so we didn't do that.


package.json
```diff
@@ -12,22 +13,20 @@
     "lint-check": "yarn eslint src/ && yarn prettier --check src/"
   },
   "main": "dist/index.js",
-  "types": "react-native-gesture-handler.d.ts",
+  "types": "dist/index.d.ts",
```
WoLewicki, line 16: Should this field be still present in `package.json`? It is not present in `RNScreens`.

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.9.0",
   "description": "Experimental implementation of a new declarative API for gesture handling in react-native",
   "scripts": {
+    "lint": "eslint --ext '.js,.ts,.tsx' src/ Swipeable.ts DrawerLayout.ts",
```
WoLewicki, line 6: Last 2 files are not present anymore.

WoLewicki (commented):
There are also `types` fields present in `package.json` of `Swipeable` and `DrawerLayout`, which should be removed. Am I correct?

package.json
```diff
@@ -12,22 +13,20 @@
...
-    "!dist/examples",
+    "index.tsx",
+    "src/",
     "index.ts",
     "Swipeable/",
-    "Swipeable.d.ts",
     "DrawerLayout/",
-    "DrawerLayout.d.ts",
     "README.md",
     "jestSetup.js",
```
WoLewicki, line 31: There is no such file in the main folder.

package.json
jakub-gonet: AFAIK we should link to the entry file types, we can test it after PR is ready.

package.json
jakub-gonet: Good catch!

package.json
jakub-gonet: This is explained in [Publishing section](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) of TS docs.

> Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.

package.json
jakub-gonet: Fixed in e743947.

src/components/GestureButtons.tsx
```diff
@@ -1,65 +1,108 @@
...
+} from '../handlers/NativeViewGestureHandler';
+
+/* BUTTONS PROPERTIES */
+
+interface RawButtonProperties extends NativeViewGestureHandlerProperties {
+  exclusive?: boolean;
+  testID?: string;
+  accessibilityLabel?: string;
+  // TODO: apply proper value
+  rippleColor?: any; // it was present in BaseButtonProperties before but is used here in code
```
jakub-gonet, line 31: TODO.

src/components/GestureButtons.tsx
```diff
@@ -1,65 +1,108 @@
...
+import {
+  NativeViewGestureHandlerPayload,
+  NativeViewGestureHandlerProperties,
+} from '../handlers/NativeViewGestureHandler';
+
+/* BUTTONS PROPERTIES */
+
+interface RawButtonProperties extends NativeViewGestureHandlerProperties {
+  exclusive?: boolean;
+  testID?: string;
```
jakub-gonet, line 28: Why we add this here?

src/components/GestureButtons.tsx
```diff
@@ -1,65 +1,108 @@
...
+  accessibilityLabel?: string;
```
jakub-gonet, line 29: There are more a11y props to consider, see `AccessibilityProps` interface in RN types.

src/components/GestureButtons.tsx
```diff
@@ -1,65 +1,108 @@
...
-  };
+export interface RectButtonProperties extends BaseButtonProperties {
+  underlayColor?: string;
+  activeOpacity?: number;
+}
+
+export interface BorderlessButtonProperties extends BaseButtonProperties {
+  borderless?: boolean;
+  activeOpacity?: number;
+}
```
jakub-gonet, line 48: `activeOpacity?: number;` is repeated in both interfaces, can we move it to `BaseButtonProperties` or not really?

src/components/GestureButtons.tsx
```diff
@@ -1,65 +1,108 @@
...
+import {
+  GestureEvent,
+  HandlerStateChangeEvent,
+} from '../handlers/gestureHandlers';
+import {
+  NativeViewGestureHandlerPayload,
+  NativeViewGestureHandlerProperties,
+} from '../handlers/NativeViewGestureHandler';
+
+/* BUTTONS PROPERTIES */
```
jakub-gonet, line 24: Unnecessary comment.

src/components/GestureButtons.tsx
```diff
@@ -1,65 +1,108 @@
...
+
+export class BaseButton extends React.Component<BaseButtonProperties> {
+  constructor(props: BaseButtonProperties) {
     super(props);
-    this._lastActive = false;
+    this.lastActive = false;
   }
 
-  _handleEvent = ({ nativeEvent }) => {
+  private lastActive: boolean;
```
jakub-gonet, line 64: Nitpick; we should move it before c-tor.

src/components/GestureButtons.tsx
```diff
@@ -1,65 +1,108 @@
...
-  _onGestureEvent = (e) => {
-    this.props.onGestureEvent && this.props.onGestureEvent(e);
-    this._handleEvent(e);
+  private onGestureEvent = (
+    e: GestureEvent<NativeViewGestureHandlerPayload>
+  ) => {
+    this.props.onGestureEvent?.(e);
+    this.handleEvent(
+      e as HandlerStateChangeEvent<NativeViewGestureHandlerPayload>
+    ); // TODO: maybe it is not correct
```
jakub-gonet, line 105: Type cast + TODO.

src/components/GestureButtons.tsx
```diff
@@ -88,25 +131,25 @@ const btnStyles = StyleSheet.create({
...
 
-  constructor(props) {
+  constructor(props: RectButtonProperties) {
     super(props);
-    this._opacity = new Animated.Value(0);
+    this.opacity = new Animated.Value(0);
   }
 
-  _onActiveStateChange = (active) => {
+  private opacity: Animated.Value;
```
jakub-gonet, line 145: Nitpick; should be placed before c-tor.

src/components/GestureButtons.tsx
```diff
@@ -88,25 +131,25 @@ const btnStyles = StyleSheet.create({
...
+
+  private onActiveStateChange = (active: boolean) => {
     if (Platform.OS !== 'android') {
-      this._opacity.setValue(active ? this.props.activeOpacity : 0);
+      this.opacity.setValue(active ? this.props.activeOpacity! : 0);
```
jakub-gonet, line 149: Not null type assertion.

src/components/GestureButtons.tsx
```diff
@@ -139,28 +182,25 @@ export class RectButton extends React.Component {
...
+    this.opacity = new Animated.Value(1);
   }
 
-  _onActiveStateChange = (active) => {
+  private opacity: Animated.Value;
+
+  private onActiveStateChange = (active: boolean) => {
     if (Platform.OS !== 'android') {
-      this._opacity.setValue(active ? this.props.activeOpacity : 1);
+      this.opacity.setValue(active ? this.props.activeOpacity! : 1);
```
jakub-gonet, line 200: Not null type assertion.

src/components/GestureButtons.tsx
```diff
@@ -139,28 +182,25 @@ export class RectButton extends React.Component {
...
 
-  constructor(props) {
+  constructor(props: BorderlessButtonProperties) {
     super(props);
-    this._opacity = new Animated.Value(1);
+    this.opacity = new Animated.Value(1);
   }
 
-  _onActiveStateChange = (active) => {
+  private opacity: Animated.Value;
```
jakub-gonet, line 196: Nitpick; should be placed before c-tor.

src/web/utils.ts
```diff
@@ -1,18 +1,18 @@
...
+export const TEST_MIN_IF_NOT_NAN = (value: number, limit: number): boolean =>
   !isnan(limit) &&
   ((limit < 0 && value <= limit) || (limit >= 0 && value >= limit));
 export const VEC_LEN_SQ = ({ x = 0, y = 0 } = {}) => x * x + y * y;
-export const TEST_MAX_IF_NOT_NAN = (value, max) =>
+export const TEST_MAX_IF_NOT_NAN = (value: number, max: number) =>
   !isnan(max) && ((max < 0 && value < max) || (max >= 0 && value > max));
 
-export function fireAfterInterval(method, interval) {
+export function fireAfterInterval(method: () => any, interval?: number) {
```
jakub-gonet, line 15: `any`

src/web/PanGestureHandler.ts
```diff
@@ -11,6 +11,7 @@ import DraggingGestureHandler from './DraggingGestureHandler';
 import { isValidNumber, isnan, TEST_MIN_IF_NOT_NAN, VEC_LEN_SQ } from './utils';
 import State from '../State';
 
+import { GestureHandlerProperties } from '../types';
```
jakub-gonet, line 14: Not used. Also, `../types` doesn't exist anymore.

src/web/NodeManager.ts
```diff
@@ -1,24 +1,28 @@
-/* eslint-disable eslint-comments/no-unlimited-disable */
-/* eslint-disable */
-// @ts-nocheck TODO(TS) provide types
-let gestures = {};
+import { ValueOf } from '../typeUtils';
+import { Gestures } from '../RNGestureHandlerModule.web';
 
-export function getHandler(tag) {
+const gestures: Record<number, any> = {};
```
jakub-gonet, line 4: `any`

src/web/Errors.ts
```diff
@@ -1,8 +1,5 @@
-/* eslint-disable eslint-comments/no-unlimited-disable */
-/* eslint-disable */
-// @ts-nocheck TODO(TS) provide types
 export class GesturePropError extends Error {
-  constructor(name, value, expectedType) {
+  constructor(name: string, value: any, expectedType: string) {
```
jakub-gonet, line 2: `any`

src/web/DiscreteGestureHandler.ts
```diff
@@ -47,12 +47,12 @@ class DiscreteGestureHandler extends GestureHandler {
...
       minPointers,
       maxPointers,
-      maxDist,
+      _maxDist,
       maxDeltaX,
       maxDeltaY,
       maxDistSq,
       shouldCancelWhenOutside,
-    },
+    }: any,
```
jakub-gonet, line 55: `any`.

src/web/DiscreteGestureHandler.ts
```diff
@@ -47,12 +47,12 @@ class DiscreteGestureHandler extends GestureHandler {
     {
       minPointers,
       maxPointers,
-      maxDist,
+      _maxDist,
```
jakub-gonet, line 50: Why?

src/web/DiscreteGestureHandler.ts
```diff
@@ -32,7 +32,7 @@ class DiscreteGestureHandler extends GestureHandler {
     );
   }
 
-  transformNativeEvent({ center: { x, y } }) {
+  transformNativeEvent({ center: { x, y } }: any) {
```
jakub-gonet, line 35: `any`.

src/web/DiscreteGestureHandler.ts
```diff
@@ -14,8 +14,8 @@ class DiscreteGestureHandler extends GestureHandler {
   }
 
   shouldFailUnderCustomCriteria(
-    { x, y, deltaX, deltaY },
-    { maxDeltaX, maxDeltaY, maxDistSq, shouldCancelWhenOutside }
+    { x, y, deltaX, deltaY }: any,
+    { maxDeltaX, maxDeltaY, maxDistSq, shouldCancelWhenOutside }: any
```
jakub-gonet, line 18: `any`.

src/components/touchables/TouchableWithoutFeedback.tsx
```diff
@@ -1,13 +1,12 @@
-// @ts-nocheck TODO(TS) provide types
 import React from 'react';
-import GenericTouchable from './GenericTouchable';
+import GenericTouchable, { GenericTouchableProps } from './GenericTouchable';
 
-const TouchableWithoutFeedback = React.forwardRef((props, ref) => (
-  <GenericTouchable ref={ref} {...props} />
-));
+// TODO: it might give wrong type to users
```
jakub-gonet, line 4: TODO.

src/components/touchables/TouchableOpacity.tsx
```diff
@@ -57,7 +58,7 @@ export default class TouchableOpacity extends Component {
         style={[
           style,
           {
-            opacity: this.opacity,
+            opacity: (this.opacity as unknown) as number, // TODO: fix this
```
jakub-gonet, line 61: TODO.

src/components/touchables/TouchableOpacity.tsx
```diff
@@ -38,9 +39,9 @@ export default class TouchableOpacity extends Component {
     }).start();
   };
 
-  onStateChange = (from, to) => {
+  onStateChange = (_: number, to: number) => {
     if (to === TOUCHABLE_STATE.BEGAN) {
-      this.setOpacityTo(this.props.activeOpacity, 0);
+      this.setOpacityTo(this.props.activeOpacity!, 0);
```
jakub-gonet, line 44: Not null type assertion.

src/components/touchables/TouchableOpacity.tsx
```diff
@@ -38,9 +39,9 @@ export default class TouchableOpacity extends Component {
     }).start();
   };
 
-  onStateChange = (from, to) => {
+  onStateChange = (_: number, to: number) => {
```
jakub-gonet, line 42: I think we should leave original names – `_from`.

src/RNGestureHandlerModule.web.ts
```diff
@@ -39,16 +41,20 @@ export default {
...
-  attachGestureHandler(handlerTag, newView, propsRef) {
+  attachGestureHandler(
+    handlerTag: number,
+    newView: number,
+    propsRef: React.RefObject<unknown>
+  ) {
     NodeManager.getHandler(handlerTag).setView(newView, propsRef);
   },
-  updateGestureHandler(handlerTag, newConfig) {
+  updateGestureHandler<T>(handlerTag: number, newConfig: T) {
```
jakub-gonet, line 51: Unrestricted generic type.

src/RNGestureHandlerModule.web.ts
```diff
@@ -24,13 +21,18 @@ const Gestures = {
...
   handleClearJSResponder() {
     console.warn('handleClearJSResponder: ');
   },
-  createGestureHandler(handlerName, handlerTag, config) {
+  createGestureHandler<T>(
+    handlerName: keyof typeof Gestures,
+    handlerTag: number,
+    config: T
+  ) {
+    //TODO(TS) extends config
```
jakub-gonet, line 35: TODO & unrestricted generic type.

src/RNGestureHandlerModule.web.ts
```diff
@@ -24,13 +21,18 @@ const Gestures = {
 
 export default {
   Direction,
-  handleSetJSResponder(tag, blockNativeResponder) {
+  handleSetJSResponder(tag: number, blockNativeResponder: boolean) {
     console.warn('handleSetJSResponder: ', tag, blockNativeResponder);
   },
   handleClearJSResponder() {
     console.warn('handleClearJSResponder: ');
   },
```
jakub-gonet, line 29: Why do we have those warnings here?

src/handlers/gestureHandlerTypesCompat.ts
```diff
@@ -0,0 +1,43 @@
...
+export type PanGestureHandlerStateChangeEvent = HandlerStateChangeEvent<PanGestureHandlerEventPayload>;
+
+export type PinchGestureHandlerGestureEvent = GestureEvent<PinchGestureHandlerEventPayload>;
+export type PinchGestureHandlerStateChangeEvent = HandlerStateChangeEvent<PinchGestureHandlerEventPayload>;
+
+export type RotationGestureHandlerGestureEvent = GestureEvent<RotationGestureHandlerEventPayload>;
+export type RotationGestureHandlerStateChangeEvent = HandlerStateChangeEvent<RotationGestureHandlerEventPayload>;
+
+export type FlingGestureHandlerGestureEvent = GestureEvent<FlingGestureHandlerEventPayload>;
+export type FlingGestureHandlerStateChangeEvent = HandlerStateChangeEvent<FlingGestureHandlerEventPayload>;
```
jakub-gonet, line 43: Should we add a deprecation warning here?

jakub-gonet (commented):
Partial review.

1. I'm wondering if we could replace relative imports with `src/` imports, I think this would be a little cleaner.
2. We use prefixed types for React, but use RN types without prefix, can we settle on one convention?

src/GestureHandlerRootView.android.tsx
```diff
@@ -1,24 +1,18 @@
...
 
 const GestureHandlerRootViewContext = React.createContext(false);
 
-export default function GestureHandlerRootView({ children, ...rest }) {
+export default function GestureHandlerRootView({
+  children,
+  ...rest
+}: {
+  children: React.ReactNode;
+}) {
```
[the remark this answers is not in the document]
jakub-gonet: Fixed in f23e0b4.

src/GestureHandlerRootView.ts
```diff
@@ -1,7 +1,4 @@
-/* eslint-disable eslint-comments/no-unlimited-disable */
-/* eslint-disable */
-// @ts-nocheck TODO(TS) provide types
 import React from 'react';
-import { View } from 'react-native';
+import { View, ViewProps } from 'react-native';
 
-export default View;
+export default View as React.ComponentType<ViewProps>;
```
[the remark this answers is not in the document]
jakub-gonet: Fixed in e860f6b.

src/RNGestureHandlerModule.ts
```diff
@@ -2,4 +2,25 @@ import { NativeModules } from 'react-native';
...
+}>;
+
+export type RNGestureHandlerModuleProps = {
+  Direction: Directions;
+  handleSetJSResponder: (tag: number, blockNativeResponder: boolean) => void;
+  handleClearJSResponder: () => void;
+  createGestureHandler: <T>(
+    handlerName: string,
+    handlerTag: number,
+    config: T //TODO extends config
```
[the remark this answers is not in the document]
jakub-gonet: Fixed in 75ed0cc.

src/RNGestureHandlerModule.ts
```diff
@@ -2,4 +2,25 @@ import { NativeModules } from 'react-native';
...
+  ) => void;
+  attachGestureHandler: (handlerTag: number, newView: number) => void;
+  updateGestureHandler: <T>(handlerTag: number, newConfig: T) => void;
+  dropGestureHandler: (handlerTag: number) => void;
+};
```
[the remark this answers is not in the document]
jakub-gonet: I checked native implementation and they do.

src/components/touchables/GenericTouchable.tsx
```diff
@@ -250,20 +255,22 @@ export default class GenericTouchable extends Component {
...
+      interpolate: undefined,
     };
 
     return (
       <BaseButton
         style={this.props.containerStyle}
         onHandlerStateChange={
-          this.props.disabled ? null : this.onHandlerStateChange
+          // TODO: not sure if it can be undefined instead of null
+          this.props.disabled ? undefined : this.onHandlerStateChange
```
jakub-gonet, line 273: Why this change was introduced?

src/components/touchables/GenericTouchable.tsx
```diff
@@ -250,20 +255,22 @@ export default class GenericTouchable extends Component {
...
       accessibilityRole: this.props.accessibilityRole,
-      accessibilityStates: this.props.accessibilityStates,
-      accessibilityTraits: this.props.accessibilityTraits,
+      // TODO: check if changed to no 's' correctly, also removed 2 props that are no longer available: `accessibilityComponentType` and `accessibilityTraits`,
+      // would be good to check if it is ok for sure, see: https://github.com/facebook/react-native/issues/24016
+      accessibilityState: this.props.accessibilityState,
       nativeID: this.props.nativeID,
       onLayout: this.props.onLayout,
       hitSlop: this.props.hitSlop,
+      interpolate: undefined,
```
jakub-gonet, line 265: Why?

src/components/touchables/GenericTouchable.tsx
```diff
@@ -214,14 +218,15 @@ export default class GenericTouchable extends Component {
...
+        this.props.onPress?.();
       }
     }
   };
 
   onLongPressDetected = () => {
     this.longPressDetected = true;
-    this.props.onLongPress();
+    // checked for in the caller of `onLongPressDetected`, but better to check twice
+    this.props.onLongPress?.();
```
jakub-gonet, line 229: I'm not sure about that one.

src/components/touchables/GenericTouchable.tsx
```diff
@@ -238,7 +243,7 @@ export default class GenericTouchable extends Component {
 
   onMoveOut() {
     // long press should no longer be detected
-    clearTimeout(this.longPressTimeout);
+    clearTimeout(this.longPressTimeout!);
```
jakub-gonet, line 246: Not null assert.

src/components/touchables/TouchableHighlight.tsx
```diff
@@ -61,21 +69,23 @@ export default class TouchableHighlight extends Component {
...
 
   renderChildren() {
     if (!this.props.children) {
       return <View />;
     }
 
-    const child = React.Children.only(this.props.children);
+    const child = React.Children.only(
+      this.props.children
+    ) as React.ReactElement<ViewProps>; // TODO: not sure if OK but fixes error
```
jakub-gonet, line 82: Type assertion.

src/components/touchables/TouchableHighlight.tsx
```diff
@@ -61,21 +69,23 @@ export default class TouchableHighlight extends Component {
...
     return React.cloneElement(child, {
       style: StyleSheet.compose(child.props.style, this.state.extraChildStyle),
     });
   }
 
-  onStateChange = (from, to) => {
+  onStateChange = (_: number, to: number) => {
```
jakub-gonet, line 88:
```suggestion
  onStateChange = (_from: number, to: number) => {
```

src/handlers/createNativeWrapper.tsx
```diff
@@ -13,65 +14,60 @@ import NativeViewGestureHandler from './NativeViewGestureHandler';
...
+>(
+  Component: React.ComponentType<P>,
+  config: Readonly<Record<string, unknown>> = {}
+) {
+  const ComponentWrapper = React.forwardRef<React.ComponentType<P>, P>(
+    (props, ref) => {
+      // filter out props that should be passed to gesture handler wrapper
+      const gestureHandlerProps = Object.keys(props).reduce(
+        (res, key) => {
+          // TS being overly protective with it's types, see https://github.com/microsoft/TypeScript/issues/26255#issuecomment-458013731 fpr more info
```
jakub-gonet, line 33: "for"

jakub-gonet (commented):
Another partial review, 47/53

src/web/PanGestureHandler.ts
WoLewicki: Probably my mistake, can be removed if is not used.

src/web/GestureHandler.ts
```diff
@@ -479,7 +502,7 @@ function ensureConfig(config) {
   return props;
 }
 
-function asArray(value) {
+function asArray<T extends unknown>(value: T): T[] {
```
WoLewicki, line 505: Shouldn't this be in `utils.ts`?

src/web/GestureHandler.ts
```diff
@@ -57,15 +86,11 @@ abstract class GestureHandler {
     return { success: true };
   }
 
-  parseNativeEvent(nativeEvent) {
```
WoLewicki, line 58: Why is it removed?

src/web/GestureHandler.ts
```diff
@@ -159,14 +186,10 @@ abstract class GestureHandler {
 
     const event = this.transformEventData(nativeEvent);
 
-    invokeNullableMethod('onGestureEvent', onGestureHandlerEvent, event);
+    invokeNullableMethod(onGestureHandlerEvent, event);
```
WoLewicki, line 189: Why is this method changed?

WoLewicki (commented):
Good job!

src/web/GestureHandler.ts
jakub-gonet: Probably, the web is still not finished.

src/web/GestureHandler.ts
jakub-gonet: Not used, see the header of 2154801.

src/components/touchables/GenericTouchable.tsx
```diff
@@ -14,61 +24,51 @@ import PropTypes from 'prop-types';
...
  */
-export const TOUCHABLE_STATE = {
+export const TOUCHABLE_STATE: Record<string, TouchableStateValuesType> = {
   UNDETERMINED: 0,
   BEGAN: 1,
   MOVED_OUTSIDE: 2,
 };
 
-const PublicPropTypes = {
+type TouchableStateValuesType = 0 | 1 | 2;
```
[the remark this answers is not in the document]
WoLewicki: If it works, I am ok with it.

src/web/GestureHandler.ts
jakub-gonet: Not used, see 95fac51.

src/RNGestureHandlerModule.ts
```diff
@@ -2,4 +2,25 @@ import { NativeModules } from 'react-native';
...
+
+export default RNGestureHandlerModule as RNGestureHandlerModuleProps;
```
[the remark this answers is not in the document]
WoLewicki: Following https://reactnative.dev/docs/native-modules-ios#better-native-module-export, it looks like there is no better way to do it.

src/RNGestureHandlerModule.web.ts
```diff
@@ -11,7 +8,7 @@ import PinchGestureHandler from './web/PinchGestureHandler';
 import RotationGestureHandler from './web/RotationGestureHandler';
 import TapGestureHandler from './web/TapGestureHandler';
 
-const Gestures = {
+export const Gestures = {
```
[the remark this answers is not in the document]
WoLewicki: It is used in `NodeManager.ts`, so I guess it needs to?

src/components/GestureButtons.tsx
WoLewicki: It should be in `BaseButtonProperties`, as it is used in `BaseButton` in `GenericTouchable.tsx`. I will move it then. Is it what you meant?

src/components/GestureButtons.tsx
WoLewicki: It is taken from the former `rngh.d.ts` file. I believe it is not needed there since it is used in `GenericTouchable.tsx`, where other a11y props are provided and used.

src/components/GestureButtons.tsx
WoLewicki: I think not really, based on the code, these are `props` intended for these components only.

src/components/GestureButtons.tsx
WoLewicki: `this.handleEvent` is called both in `onGestureEvent` and `onHandlerStateChange`, which take different types of native events, so one of the calls must cast the type unfortunately. Maybe it is a mistake in implementation. `handleEvent` uses `oldState`, which is available only in `HandlerStateChangeEvent<NativeViewGestureHandlerPayload>`, so I cast the former into it. From testing it in example, `oldState` is never present in `onGestureEvent`'s native event. It would be good to check how to correctly resolve this.

src/components/GestureButtons.tsx
WoLewicki: It is inevitable in class component with `defaultProps`, which take care of always providing the value.

src/components/GestureButtons.tsx
WoLewicki: It is inevitable in class component with `defaultProps`, which take care of always providing the value.

src/web/Errors.ts
WoLewicki: We need types `web/FlingGestureHandler` for me to know what type can it have, `unknown` is not a good option since it is later used in a comment.

src/web/DiscreteGestureHandler.ts
WoLewicki: Because it is not used.

src/web/DiscreteGestureHandler.ts
WoLewicki: These web files are now marked with a `no-check`, so I will just leave these `any` to be removed when we type it.

src/components/touchables/TouchableWithoutFeedback.tsx
WoLewicki: Checked and it shows all the options, so it is ok.

src/components/touchables/TouchableOpacity.tsx
WoLewicki: `this.opacity` is an `Animated.Value`, so I don't know how to correctly do it. Maybe the implementation is incorrect?

src/components/touchables/TouchableOpacity.tsx
WoLewicki: It is inevitable in class component with defaultProps, which take care of always providing the value.

src/handlers/gestureHandlerTypesCompat.ts
WoLewicki: It would be nice to point people to use the newer API. Deprecation is easily visible, so it seems like a good option. It requires only changing the name of the event, yes?

src/State.ts
```diff
@@ -1,22 +1,14 @@
...
-      return keys[i];
-    }
-  }
-};
+} as const;
 
-export default State;
+// It is needed for backwards compability, since it can be used as a type and a value
+// eslint-disable-next-line no-redeclare
+export type State = typeof State[keyof typeof State];
```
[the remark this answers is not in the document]
WoLewicki: It is taken from `react-native-gesture-handler.d.ts`, so it can be used by users. `typeof State[keyof typeof State];` returns `type State = 0 | 1 | 2 | 3 | 4 | 5`, whereas `typeof State` returns 
```ts
type State = {
    readonly UNDETERMINED: 0;
    readonly FAILED: 1;
    readonly BEGAN: 2;
    readonly CANCELLED: 3;
    readonly ACTIVE: 4;
    readonly END: 5;
}
```

src/components/touchables/GenericTouchable.tsx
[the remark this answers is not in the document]
WoLewicki: I can change it to: 
```ts
export const TOUCHABLE_STATE = {
  UNDETERMINED: 0,
  BEGAN: 1,
  MOVED_OUTSIDE: 2,
} as const;

type TouchableStateType = typeof TOUCHABLE_STATE[keyof typeof TOUCHABLE_STATE];
```

It will then follow the `State` implementation.

src/components/touchables/GenericTouchable.tsx
```diff
@@ -14,61 +24,51 @@ import PropTypes from 'prop-types';
...
-  onLongPress: PropTypes.func,
-  nativeID: PropTypes.string,
-  testID: PropTypes.string,
-  delayPressIn: PropTypes.number,
-  delayPressOut: PropTypes.number,
-  delayLongPress: PropTypes.number,
-  shouldActivateOnStart: PropTypes.bool,
-  disallowInterruption: PropTypes.bool,
-};
+  // e.g. onBlur and onFocus as well as deprecated props. - TODO: this comment may be unuseful in this moment
```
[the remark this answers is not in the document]
WoLewicki: I am not sure what should we do about it.

src/components/touchables/TouchableHighlight.tsx
WoLewicki: `children` is `React.ReactNode` by default. Maybe we should type `children` explicitly to be `React.ReactElement<ViewProps>` or maybe something else?

src/components/touchables/GenericTouchable.tsx
WoLewicki: Because `onHandlerStateChange` does not accept null. I think `undefined` is safe since there is a check in code of `BaseButton` (`this.props.onHandlerStateChange?.(e)`) that will treat `null` and `undefined` the same.

src/components/touchables/GenericTouchable.tsx
WoLewicki: What do you mean? I think it was wrong that there was no check before and TS does not allow it without `?.` too.

src/components/touchables/GenericTouchable.tsx
```diff
@@ -123,7 +123,7 @@ export default class GenericTouchable extends Component {
 
   // handleGoToUndetermined transits to UNDETERMINED state with proper delay
   handleGoToUndetermined() {
-    clearTimeout(this.pressOutTimeout);
+    clearTimeout(this.pressOutTimeout!); // TODO: maybe it can be undefined
```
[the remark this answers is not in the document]
WoLewicki: I don't know how to type these timeouts properly. The current version is the only working one I could make.

src/RNGestureHandlerModule.web.ts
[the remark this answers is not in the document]
jakub-gonet: You're right.

src/components/GestureButtons.tsx
jakub-gonet: I mean, this is for tests and I'm not sure why we should support this prop. AFAIK this is quite common, if we should apply it to every exported component then maybe we should add some type that adds it?

src/components/GestureButtons.tsx
jakub-gonet: But Touchables aren't buttons, right? So we should somehow support every other a11y prop in Buttons.

src/web/Errors.ts
jakub-gonet: it's `number`, I left some comments for myself so don't worry about that.

src/components/touchables/TouchableOpacity.tsx
jakub-gonet: Hard to say, I'll take a look at it after we finish web.

src/handlers/gestureHandlerTypesCompat.ts
jakub-gonet: I'm not sure what you mean. If migration, then yes, users would only need to substitute existing types with new declarations from here.

src/State.ts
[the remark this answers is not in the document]
jakub-gonet: I see.
I also tested other variants and they don't work so let's stay with existing types.

src/components/touchables/GenericTouchable.tsx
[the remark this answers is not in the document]
jakub-gonet: Definitely looks more maintainable IMO.

src/components/touchables/GenericTouchable.tsx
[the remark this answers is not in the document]
jakub-gonet: I can take a look at that.

src/components/touchables/GenericTouchable.tsx
jakub-gonet: Sorry, I made really ambiguous comments that day 😄 

I think I referred to the comment, We should rewrite it in the cleaner manner, so TS can infer if it can be undefined or not.

WoLewicki: Not sure if it is possible, this function could be used somewhere else too, and there could be no check. I think it is safe to leave as it is now.

src/web/GestureHandler.ts
jakub-gonet: Actually, not really, this is used as a helper only there.

jakub-gonet:
I resolved most of `any`/TODO/typecast comments, there are too many to keep track of everything.

I think we should now check if every changed type is backward compatible and fix the rest incrementally. RNGH code (especially web) is too complex to type some things from untyped code directly.

src/RNGestureHandlerModule.ts
```diff
@@ -2,4 +2,28 @@ import { NativeModules } from 'react-native';
 
 const { RNGestureHandlerModule } = NativeModules;
 
-export default RNGestureHandlerModule;
+type Directions = Readonly<{
+  RIGHT: 1;
+  LEFT: 2;
+  UP: 4;
+  DOWN: 8;
+}>;
```
jkadamczyk, line 10: Can't it be an ENUM?

src/RNGestureHandlerModule.web.ts
```diff
@@ -39,16 +41,20 @@ export default {
...
-  attachGestureHandler(handlerTag, newView, propsRef) {
+  attachGestureHandler(
+    handlerTag: number,
+    newView: number,
+    propsRef: React.RefObject<unknown>
+  ) {
     NodeManager.getHandler(handlerTag).setView(newView, propsRef);
   },
-  updateGestureHandler(handlerTag, newConfig) {
+  updateGestureHandler(handlerTag: number, newConfig: any) {
```
jkadamczyk, line 51: Will it work?
```suggestion
  updateGestureHandler(handlerTag: number, newConfig: object) {
```

src/State.ts
```diff
@@ -1,22 +1,14 @@
-/* eslint-disable eslint-comments/no-unlimited-disable */
-/* eslint-disable */
-// @ts-nocheck TODO(TS) provide types
-const State = {
+// TODO use State from RNModule
+
+export const State = {
```
jkadamczyk, line 3: Wouldn't an enum work for this case?

src/RNGestureHandlerModule.ts
jakub-gonet: IIRC enum semantics is different from an object. I'd like to keep possibly incompatible changes outside of this PR and maybe fix them in the follow-up.

src/State.ts
jakub-gonet: IIRC enum semantics is different from an object. I'd like to keep possibly incompatible changes outside of this PR and maybe fix them in the follow-up.

src/RNGestureHandlerModule.web.ts
jakub-gonet: Yeah, but config is more constrained than `object`, `Record<string, unknown>` would be better but let's keep `any` until we'll be sure which properties can go into `newConfig` and which cannot.

src/components/GestureComponents.tsx
```diff
@@ -1,63 +1,153 @@
...
+    );
+    // we use literal object since TS gives error when using RN's `positions`
+    // @ts-ignore FIXME(TS) maybe this should be removed?
+    DrawerLayoutAndroid.positions = { Left: 'left', Right: 'right' };
     return DrawerLayoutAndroid;
   },
-  get FlatList() {
-    if (!MEMOIZED.FlatList) {
+
+  // @ts-ignore xddddd
```
jkadamczyk, line 70: 🤔

src/components/touchables/GenericTouchable.tsx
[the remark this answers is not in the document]
jkadamczyk: If it seems not useful for you maybe we should delete it? If someone will spot there's something missing, git history for that file is up for exploration. For others, I think it might be just noise. Let me know what you think

src/components/touchables/GenericTouchable.tsx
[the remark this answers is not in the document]
jkadamczyk: Maybe you can call it only when there is need to do so.
```suggestion
    this.pressOutTimeout && clearTimeout(this.pressOutTimeout);
```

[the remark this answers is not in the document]
jkadamczyk: Not sure if that syntax will be enough. Maybe if statement is necessary

src/components/GestureComponents.tsx
jakub-gonet: This was added in WIP commit, will be fixed 😄

src/components/GestureComponents.tsx
```diff
@@ -10,6 +10,7 @@ import {
   DrawerLayoutAndroidProps as RNDrawerLayoutAndroidProps,
   FlatList as RNFlatList,
   FlatListProps as RNFlatListProps,
+  FlatListProps,
```
WoLewicki, line 13: Why import it 2 times?

src/components/touchables/GenericTouchable.tsx
jakub-gonet: Unresolving for visibility.

src/components/GestureComponents.tsx
```diff
@@ -1,63 +1,82 @@
...
+      {...props}
+      renderScrollComponent={(scrollProps) => <ScrollView {...scrollProps} />}
+    />
+  )
+);
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type FlatList<ItemT> = React.ComponentType<
+  RNFlatListProps<ItemT> &
+    NativeViewGestureHandlerProperties &
+    React.RefAttributes<any>
```
WoLewicki, line 67: What does it do?

src/components/GestureComponents.tsx
```diff
@@ -1,63 +1,82 @@
...
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type ScrollView = typeof ScrollView & {
+  scrollTo(
+    y?: number | { x?: number; y?: number; animated?: boolean },
+    x?: number,
+    animated?: boolean
+  ): void;
+  scrollToEnd(options?: { animated: boolean }): void;
+};
+export const Switch = createNativeWrapper<RNSwitchProps>(RNSwitch, {
```
WoLewicki, line 35: Wouldn't adding blank lines between components be good?

src/components/GestureComponents.web.tsx
```diff
@@ -25,12 +23,15 @@ export const TextInput = createNativeWrapper(RNTextInput);
 export const DrawerLayoutAndroid = createNativeWrapper(RNDrawerLayoutAndroid, {
   disallowInterruption: true,
 });
+// @ts-ignore -- TODO(TS) to investigate if it's needed
```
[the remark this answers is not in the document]
WoLewicki: Oh, it is web, so I am not sure.

tsconfig.json
```diff
@@ -17,15 +11,20 @@
...
     "noUnusedLocals": true,
     "outDir": "./dist"
   },
-  "include": ["*.ts", "src/**/*.ts", "src/**/*.tsx", "**/*.d.ts"]
+  "include": [
+    "index.ts",
+    "src/**/*.ts",
+    "src/**/*.tsx",
+    "src/**/*.js",
+    "src/components/GestureComponents.js"
```
WoLewicki, line 28: Are these two still necessary?

WoLewicki (commented):
Left some comments, looks good overall!

tsconfig.json
jakub-gonet: Nope, good catch!

jakub-gonet: Fixed in 24c54dd.

src/components/GestureComponents.tsx
jakub-gonet: This allows passing refs, they're excluded from props by default.

src/GestureHandlerRootView.android.tsx
```diff
@@ -1,24 +1,17 @@
...
 const GestureHandlerRootViewNative = requireNativeComponent(
-  'GestureHandlerRootView',
-  iface
+  'GestureHandlerRootView'
 );
 
 const GestureHandlerRootViewContext = React.createContext(false);
 
-export default function GestureHandlerRootView({ children, ...rest }) {
+type Props = Record<string, unknown> & {
```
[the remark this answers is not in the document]
jakub-gonet: Definitely better, thanks!

Fixed in ed34cb6.

src/components/GestureComponents.tsx
jakub-gonet: Fixed in 62d66d6.

PR #1301, opened 2021-01-07
Specify minimal SDK version for enableTrackpadTwoFingerGesture

### LINKED ISSUE #1235: Latest release (1.9.0) breaks build
## Description

When running `npx react-native run-ios` the build fails with the following errors:

```
/Users/xxx/MyApp/node_modules/react-native-gesture-handler/ios/Handlers/RNPanHandler.m:212:18:
error: property 'allowedScrollTypesMask' not found on object of type 'RNBetterPanGestureRecognizer *'
      recognizer.allowedScrollTypesMask = UIScrollTypeMaskAll;
                 ^
/Users/xxx/MyApp/node_modules/react-native-gesture-handler/ios/Handlers/RNPanHandler.m:212:43:
error: use of undeclared identifier 'UIScrollTypeMaskAll'
      recognizer.allowedScrollTypesMask = UIScrollTypeMaskAll;
```

Downgrading to 1.8.0 solves the issue.

## Package versions

- React: 16.13.1
- React Native: 0.63.2
- React Native Gesture Handler: 1.9.0

KenCorma:
## Description

`@available` only works at run time, it will still break on compile if this version of iOS is not available. 

Added `__IPHONE_OS_VERSION_MAX_ALLOWED >= 130400` so that older versions of Xcode can compile.

## Test plan

Successfully build with Xcode 11.3.1

### DIFF
--- a/ios/Handlers/RNPanHandler.m
+++ b/ios/Handlers/RNPanHandler.m
@@ -206,7 +206,7 @@
   APPLY_FLOAT_PROP(failOffsetYStart);
   APPLY_FLOAT_PROP(failOffsetYEnd);
 
-#if !TARGET_OS_TV
+#if !TARGET_OS_TV && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130400
   if (@available(iOS 13.4, *)) {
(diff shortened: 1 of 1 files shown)


jakub-gonet:
Thanks!

PR #1307, opened 2021-01-18
Update Wix RN Navigation integration steps

danilobuerger:
## Description

When registering rnn components that are wrapped in providers, the third param needs to provide the concrete component.
See https://wix.github.io/react-native-navigation/api/component#registercomponent


jakub-gonet (approved):
Thanks!

PR #1310, opened 2021-01-19
Feature: minDistFromEdge, maxDistFromEdge

mrousavy:
## Description

This PR implements two new features for the `PanGestureHandler`:

* `minDistFromEdge`: A prop that specifies how far away from the view's edges the touch has to be so the gesture handler activates.
* `maxDistFromEdge`: A prop that specifies how far away from the view's edges the touch is allowed to be so the gesture handler activates.

Yes, `minDistFromEdge` is very similar to `hitSlop`. I've implemented it to make it clear that there is an inverse for `maxDistFromEdge`.

`maxDistFromEdge` is basically like the inverted `hitSlop`, if you have a `hitSlop` that insets the recognizable area by 50 on every size, the gesture handler only activates within the 50px inset on every size. `maxDistFromEdge` is the opposite, it only activates if the tap was outside of those 50px insets, so only on the edges.

## Test plan

* I've used the prop and verified that it doesn't activate if I start within 20px from the right edge, otherwise it activates:
    ```jsx
    <PanGestureHandler minDistFromEdge={{ top: 0, left: 0, bottom: 0, right: 20 }} />
    ```
* I've partially used the prop and verified that it doesn't activate if I start within 20px from the right edge, otherwise it activates:
    ```jsx
    <PanGestureHandler minDistFromEdge={{ right: 20 }} />
    ```
* I've removed the prop and verified that no side effects were introduced


mrousavy:
could you please take a look at this @jakub-gonet ? thanks ❤️

erkie:
I was looking for a way to not activate if the user touched on the edge, but then I found `hitSlop` (thanks to you for mentioning it!!! saved me so much time) I'm not sure what the difference with `minDistFromEdge` and `hitSlop` is?

jakub-gonet:
@mrousavy 
Could you merge master so we don't have conflicts and diff is slightly smaller than +18,371 −12,815? 😄 
Also, could you describe in more detail the differences between `minDistFromEdge` vs `hitSlop`?

jakub-gonet:
🏓

jakub-gonet:
@mrousavy 🏓

mrousavy:
Sorry, I'm really busy at the moment and don't have a use-case for this feature anymore.

If you want to rebase yourself, the code is here 🤗

jakub-gonet:
In that case, I'll close this one. If you want this feature merged in the future, please ping me.

PR #1320, opened 2021-01-25
Improve wording in LongPressGH's docs

jtalz:
made some english fixes and changed "default value" to agree with react native docs

### DIFF
--- a/docs/docs/handler-longpress.md
+++ b/docs/docs/handler-longpress.md
@@ -4,43 +4,43 @@ title: LongPressGestureHandler
 sidebar_label: LongPressGestureHandler
 ---
 
-A discrete gesture handler that activates when the corresponding view is pressed sufficiently long.
-When handler gets activated it will turn into [END](state.md#end) state when finger is released.
-The handler will fail to recognize if the finger is lifted before the [minimum required time](#mindurationms) or if the finger is moved further than the [allowable distance](#maxdist).
+A discrete gesture handler that activates when the corresponding view is pressed for a sufficiently long time.
+This handler's state will turn into [END](state.md#end) immediately after the finger is released.
+The handler will fail to recognize a touch event if the finger is lifted before the [minimum required time](#mindurationms) or if the finger is moved further than the [allowable distance](#maxdist).
 
 The handler is implemented using [UILongPressGestureRecognizer](https://developer.apple.com/documentation/uikit/uilongpressgesturerecognizer) on iOS and [LongPressGestureHandler](https://github.com/software-mansion/react-native-gesture-handler/blob/master/android/lib/src/main/java/com/swmansion/gesturehandler/LongPressGestureHandler.java) on Android.
 
 ## Properties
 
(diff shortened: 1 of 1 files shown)


jtalz:
Please review that the default values are indeed accurate

docs/docs/handler-longpress.md
```diff
@@ -4,43 +4,43 @@ title: LongPressGestureHandler
...
 
 ### `y`
 
-Y coordinate of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the view attached to the handler. Expressed in point units.
+Y coordinate, expressed in points, of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the view attached to the handler. 
 
 ### `absoluteX`
 
-X coordinate of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the root view. The value is expressed in point units. It is recommended to use it instead of [`x`](#x) in cases when the original view can be transformed as an effect of the gesture.
+X coordinate, expressed in points, of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the root view. It is recommended to use this instead of [`x`](#x) in cases when the view attached to the handler can be transformed as an effect of the gesture.
```
jakub-gonet, line 39:
```suggestion
X coordinate, expressed in points, of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the root view. It is recommended to use `absoluteX` instead of [`x`](#x) in cases when the view attached to the handler can be transformed as an effect of the gesture.
```

or "this property"

docs/docs/handler-longpress.md
```diff
@@ -4,43 +4,43 @@ title: LongPressGestureHandler
...
 
 ### `absoluteY`
 
-Y coordinate of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the root view. The value is expressed in point units. It is recommended to use it instead of [`y`](#y) in cases when the original view can be transformed as an effect of the gesture.
+Y coordinate, expressed in points, of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the root view. It is recommended to use this instead of [`y`](#y) in cases when the view attached to the handler can be transformed as an effect of the gesture.
```
jakub-gonet, line 43:
```suggestion
Y coordinate, expressed in points, of the current position of the pointer (finger or a leading pointer when there are multiple fingers placed) relative to the root view. It is recommended to use `absoluteY` instead of [`y`](#y) in cases when the view attached to the handler can be transformed as an effect of the gesture.
```
ditto

jakub-gonet (commented):
Looks good, left some suggestions.

docs/docs/handler-longpress.md
jtalz: I think we should stick with the convention used in TapGestureHandler doc and keet `absoluteX`

docs/docs/handler-longpress.md
jtalz: also yes (y)

jakub-gonet (approved):
Thanks!

PR #1323, opened 2021-01-26
Attach to RNRootViewGestureRecognizer instead of root view

WoLewicki:
## Description

Added change of registering `rootViews` and of calling `[touchHandler cancel]` for integration with [`react-native-screens`](https://github.com/software-mansion/react-native-screens) modal views.


ios/RNGestureHandlerManager.m
```diff
@@ -132,25 +128,35 @@ - (void)handleClearJSResponder
...
+    while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)]) parent = parent.superview;
+
+    UIView *rootView = [[parent performSelector:@selector(touchHandler)] view];
+  
+    if (rootView == nil) {
+      return;
     }
+  
+    for (RNRootViewGestureRecognizer *recognizer in _rootViewGestureRecognizers) {
+      // Many views can return the same touchHandler so we check if the one we want to register
```
kmagiera, line 143: This comment should be moved to the place where we extract `view` from `touchHandler` rather than here

ios/RNGestureHandlerManager.m
```diff
@@ -132,25 +128,35 @@ - (void)handleClearJSResponder
 
 #pragma mark Root Views Management
 
-- (void)registerRootViewIfNeeded:(UIView*)childView
+- (void)registerRootViewGestureRecognizerIfNeeded:(UIView*)childView
```
kmagiera, line 131: we are no longer interested in root views here but rather in views that have touchHandler attached. This naming is now misleading.

ios/RNGestureHandlerManager.m
```diff
@@ -132,25 +128,35 @@ - (void)handleClearJSResponder
...
-    if (rootContentView != nil && ![_rootViews containsObject:rootContentView]) {
-        RCTLifecycleLog(@"[GESTURE HANDLER] Initialize gesture handler for root view %@", rootContentView);
-        [_rootViews addObject:rootContentView];
-        RNRootViewGestureRecognizer *recognizer = [RNRootViewGestureRecognizer new];
-        recognizer.delegate = self;
-        rootContentView.userInteractionEnabled = YES;
-        [rootContentView addGestureRecognizer:recognizer];
+    while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)]) parent = parent.superview;
+
+    UIView *rootView = [[parent performSelector:@selector(touchHandler)] view];
```
kmagiera, line 136: same comment about the naming. This doesn't need to be a root view

ios/RNGestureHandlerManager.m
```diff
@@ -132,25 +128,35 @@ - (void)handleClearJSResponder
...
+  
+    if (rootView == nil) {
+      return;
     }
+  
+    for (RNRootViewGestureRecognizer *recognizer in _rootViewGestureRecognizers) {
```
kmagiera, line 142: doesn't look like we really need to check this table here, it should be enough to check if `view.gestureRecognizers` contains an object of tyle `RNRootViewGestureRecognizer` – this should be a simpler check as we wouldn't need to iterate over hash table

kmagiera (commented):
Looks good, I left a few comments inline and will let @jakub-gonet have the final call here

jakub-gonet (approved):
LGTM.

ios/RNGestureHandlerManager.m
```diff
@@ -132,25 +128,35 @@ - (void)handleClearJSResponder
 
 #pragma mark Root Views Management
 
-- (void)registerRootViewIfNeeded:(UIView*)childView
+- (void)registerViewWithGestureRecognizerAttachedIfNeeded:(UIView*)childView
```
WoLewicki, line 131:
```suggestion
- (void)registerViewWithGestureRecognizerAttachedIfNeeded:(UIView *)childView
```

jakub-gonet (approved):

PR #1347, opened 2021-02-10
Fix gestureHandlerRootHOC types

alpha0010:
## Description

With the most recent update, `gestureHandlerRootHOC()` gives errors like:
```
Argument of type 'typeof HomeScreen' is not assignable to parameter of type 'ComponentType<Record<string, unknown>>'.
  Type 'typeof HomeScreen' is not assignable to type 'FunctionComponent<Record<string, unknown>>'.
    Types of parameters 'props' and 'props' are incompatible.
```

## Test plan

`yarn ts-check`

The only change is typings which now allow my project to run clean.

### DIFF
--- a/src/gestureHandlerRootHOC.tsx
+++ b/src/gestureHandlerRootHOC.tsx
@@ -3,7 +3,9 @@ import { StyleSheet, StyleProp, ViewStyle } from 'react-native';
 import hoistNonReactStatics from 'hoist-non-react-statics';
 import GestureHandlerRootView from './GestureHandlerRootView';
 
-export default function gestureHandlerRootHOC<P extends Record<string, unknown>>(
+export default function gestureHandlerRootHOC<
+  P extends Record<string, unknown>
+>(
   Component: React.ComponentType<P>,
   containerStyles?: StyleProp<ViewStyle>
 ): React.ComponentType<P> {
--- a/src/gestureHandlerRootHOC.tsx
+++ b/src/gestureHandlerRootHOC.tsx
@@ -3,7 +3,7 @@ import { StyleSheet, StyleProp, ViewStyle } from 'react-native';
 import hoistNonReactStatics from 'hoist-non-react-statics';
 import GestureHandlerRootView from './GestureHandlerRootView';
 
-export default function gestureHandlerRootHOC<P>(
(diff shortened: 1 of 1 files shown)


jakub-gonet (commented):
Looks good, could you restrict P a little bit by writing `P extends object`?

jakub-gonet:
Also, could you provide some small piece of code that can be used to test this change?

alpha0010:
Sample code:
```tsx
import React from 'react';
import {Text, View} from 'react-native';
import {gestureHandlerRootHOC} from 'react-native-gesture-handler';
import {Navigation} from 'react-native-navigation';

function MyScreen(props: {componentId: string}) {
  return (
    <View>
      <Text>{props.componentId}</Text>
    </View>
  );
}
Navigation.registerComponent('my-screen', () =>
  gestureHandlerRootHOC(MyScreen),
);
```

jakub-gonet:
> Looks good, could you restrict P a little bit by writing P extends object?

Looks like the TS compiler doesn't like this, I changed `object` to `Record<string, unknown>`. It works on the example you provided, could you test it in your code?

alpha0010:
It sort of works.  If I factor out the props definition as a type, it works fine.
```tsx
type MyProps = {
  componentId: string;
};

function MyScreen(props: MyProps) {
  return (
    <View>
      <Text>{props.componentId}</Text>
    </View>
  );
}
```

But fails if I use an interface.  (Important if, for example, using `NavigationFunctionComponent` from RNN.)
```tsx
interface MyProps {
  componentId: string;
}

function MyScreen(props: MyProps) {
  return (
    <View>
      <Text>{props.componentId}</Text>
    </View>
  );
}
```
```
Argument of type '(props: MyProps) => Element' is not assignable to parameter of type 'ComponentType<Record<string, unknown>>'.
  Type '(props: MyProps) => Element' is not assignable to type 'FunctionComponent<Record<string, unknown>>'.
    Types of parameters 'props' and 'props' are incompatible.
      Property 'componentId' is missing in type 'PropsWithChildren<Record<string, unknown>>' but required in type 'MyProps'.
```

jakub-gonet:
All right, I will change it to P in the followup. Thanks for checking this.

PR #1364, opened 2021-02-15
Expand Jest mocks

TheWirv:
## Description

I have noticed that the mock was missing various components, which is why, up until now, I had to build my own mock that was building upon the one that was provided by RNGH. Now, that my tests broke because the path to the mock changed (to `src/`), I thought, why not add those components via PR instead of just "filing a complaint" via an issue.


jakub-gonet (commented):
Thanks. We recently introduced Bob so `__mocks__/` dirs are ignored when packaging. I think we need to extract mocks to the different file and provide them in `jestSetup.js`

TheWirv:
That makes sense. What about just having a `mock.ts` at root level? The user then can integrate however they want.

jakub-gonet:
We have jestSetup.js which has a similar function:
https://github.com/software-mansion/react-native-gesture-handler/blob/master/jestSetup.js

jakub-gonet:
I recently refactored our mocks so after moving your changes to [mocks.ts](https://github.com/software-mansion/react-native-gesture-handler/blob/master/src/mocks.ts) it should be alright to merge.

TheWirv:
@jakub-gonet Sorry for letting you wait so long, but now I finally merged the prior changes into the new mock file. Good work on the whole TS rewrite so far!

jakub-gonet (approved):
Looks alright, sorry for the delay.

PR #1366, opened 2021-02-16
Copy Directions enum to JS

### LINKED ISSUE #851: Throw informative error if gesture-handler isn't linked properly
If the user hasn't linked the library properly, currently the error messages aren't very useful.

e.g. If the library isn't linked, users usually get an error `null is not an object ( evaluating 'RNGestureHandlerModule.default.Direction')`. If the library is linked but no code was added to wrap the rootview, no error is thrown, but gestures don't work.

It'll be useful if the library showed a helpful error with a link to the documentation when it detects that it hasn't been linked properly.

### LINKED ISSUE #1081: fails with react-native-windows, cannot read property Direction of undefined
Hello, Excellent work by the team.

Please forgive me if my issue is not clear enough or does not follow the standard. I will provide further information if need be. Thanks in advance.

Here is my problem, I am running into this error message while trying out [react-native-windows](https://github.com/microsoft/react-native-windows) on a project.

I followed the instructions on the [react navigation webiste](https://reactnavigation.org/docs/getting-started) to get this installed on a bare react native project. All works well for android, but whenever I run `npx react-native run-windows` on windows, I get the following error message:

```
TypeError: Cannot read property 'Direction' of undefined
```

Project dependencies
```
"react": "16.11.0",
  "react-native": "0.62.0-0",
    "react-native-gesture-handler": "1.6.1",
    "react-native-reanimated": "1.9.0",
    "react-native-rss-parser": "1.4.0",
    "react-native-safe-area-context": "3.0.2",
    "react-native-screens": "2.8.0",
    "react-native-sqlite-2": "3.0.1",
    "react-native-webview": "10.2.3",
    "react-native-windows": "0.62.0-0"
```

### LINKED ISSUE #1324: MacOS error
TypeError: null is not an object (evaluating '_RNGestureHandlerModule.default.Direction')

    "react-native": "^0.63.4",
    "react-native-macos": "^0.63.4",
    "react-native-gesture-handler": "^1.9.0",

MacOS is not supported at this time？

jakub-gonet:
## Description

Fixes #1081.
Fixes #851.
Fixes #1324.

A common problem when importing RNGH is not finding native modules when running on not-yet-supported platforms like React Native for iOS or React Native for Windows. 

This PR changes three things:
- Copies Directions enum to JS and makes it a named export
- Adds an error message when the RNGH native module is not available
- Removes Directions from the RNGH Native Module type

Moving Directions into JS allows using it directly in tests. Also, State (which is defined in the Native Module, too) was already copied to JS.

Please note that importing mappings from native modules doesn't incur Bridge roundtrips, this is merely for tests and other platforms' convenience.

## Test plan

- [x] Tested on the Example app


kmagiera (approved):
Looks good, maybe put a link to installation instruction instead of pointing people to the issues page

src/RNGestureHandlerModule.ts
```diff
@@ -1,16 +1,18 @@
...
-  RIGHT: 1;
-  LEFT: 2;
-  UP: 4;
-  DOWN: 8;
-}>;
+if (RNGestureHandlerModule == null) {
+  console.error(
+    `react-native-gesture-handler module was not found. Make sure you're running your app on the native platform and your code is linked properly (cd ios && pod install && cd ..).
+
+    You can submit a bug report here: https://github.com/software-mansion/react-native-gesture-handler/issues`
```
jakub-gonet, line 8:
```suggestion
    For installation instructions, please refer to https://docs.swmansion.com/react-native-gesture-handler/docs/#installation
```

PR #1373, opened 2021-02-18
Complete types of wrapped Gesture Component methods

### LINKED ISSUE #1371: Missing type declarations for TextInput methods
## Description

Trying to call any [method](https://reactnative.dev/docs/textinput#methods) provided by the React Native `TextInput` component on the wrapped `TextInput` exported by this library causes a `Property [...] does not exist on type [...]. ts(2339)` error to be raised by TypeScript. The same problem may be occurring with the methods of `DrawerLayoutAndroid`, although I didn't test that.

The fix should be as easy as to add the four methods to the type declaration [here](https://github.com/software-mansion/react-native-gesture-handler/blob/5a97589d55fcaceabbf6e5c1c9828d6128670e88/src/components/GestureComponents.tsx#L47) in the same way it was done for other wrapped components like `ScrollView` or `FlatList`.

## Steps To Reproduce

1. Create a `ref` to the `TextInput` component exported by this library.
2. Try to call any method on it , for example `blur()` .

### Expected behavior
No TypeScript error, as the methods all exist and work when called.

### Actual behavior
TypeScript error raised because of missing type declarations.

## Minimal code example

```typescript
import { TextInput } from 'react-native-gesture-handler';

export const InputRefTest = () => {
	const ref = useRef<TextInput>(null);

	const handleOnFocus = () => {
		ref.current?.blur();
	};

	return <TextInput ref={ref} onFocus={handleOnFocus} />;
};
```
Error caused by this: 
```
Property 'blur' does not exist on type 'ForwardRefExoticComponent<TextInputProps & NativeViewGestureHandlerProps & RefAttributes<ComponentType<any>>>'.ts(2339)
```

## Package versions

- React: 16.13.1
- React Native: 0.63.4
- React Native Gesture Handler: 1.10.1

yfunk:
## Description
Fixes #1371 

The wrapped gesture components exported by this library (`TextInput`, `ScrollView`, etc.) were missing complete types for the methods they provide, which caused unjustified TypeScript errors to be raised and prevented IntelliSense from suggesting them.

I added all the methods I felt were relevant and documented in the official React Native Docs in the same way it was previously done with `ScrollView` and `FlatList`. I also adjusted the the way these definitions were typed to be consistent with each other.

## Test plan
I tested calling the methods on a `ref` to the component. No TypeScript errors were raised and IntelliSense suggested them.


jakub-gonet:
Could you test if this diff works?
```diff
diff --git a/src/components/GestureComponents.tsx b/src/components/GestureComponents.tsx
index 61790b3..ea1c357 100644
--- a/src/components/GestureComponents.tsx
+++ b/src/components/GestureComponents.tsx
@@ -25,14 +25,7 @@ export const ScrollView = createNativeWrapper<
 });
 // backward type compatibility with https://github.com/software-mansion/react-native-gesture-handler/blob/db78d3ca7d48e8ba57482d3fe9b0a15aa79d9932/react-native-gesture-handler.d.ts#L440-L457
 // eslint-disable-next-line @typescript-eslint/no-redeclare
-export type ScrollView = typeof ScrollView & {
-  scrollTo(
-    y?: number | { x?: number; y?: number; animated?: boolean },
-    x?: number,
-    animated?: boolean
-  ): void;
-  scrollToEnd(options?: { animated: boolean }): void;
-};
+export type ScrollView = RNScrollView;
 
 export const Switch = createNativeWrapper<RNSwitchProps>(RNSwitch, {
   shouldCancelWhenOutside: false,
@@ -40,11 +33,11 @@ export const Switch = createNativeWrapper<RNSwitchProps>(RNSwitch, {
   disallowInterruption: true,
 });
 // eslint-disable-next-line @typescript-eslint/no-redeclare
-export type Switch = typeof Switch;
+export type Switch = RNSwitch;
 
 export const TextInput = createNativeWrapper<RNTextInputProps>(RNTextInput);
 // eslint-disable-next-line @typescript-eslint/no-redeclare
-export type TextInput = typeof TextInput;
+export type TextInput = RNTextInput;
 
 export const DrawerLayoutAndroid = createNativeWrapper<
   PropsWithChildren<RNDrawerLayoutAndroidProps>
@@ -53,7 +46,7 @@ export const DrawerLayoutAndroid = createNativeWrapper<
 // @ts-ignore FIXME(TS) maybe this should be removed?
 DrawerLayoutAndroid.positions = { Left: 'left', Right: 'right' };
 // eslint-disable-next-line @typescript-eslint/no-redeclare
-export type DrawerLayoutAndroid = typeof DrawerLayoutAndroid;
+export type DrawerLayoutAndroid = RNDrawerLayoutAndroid;
 
 export const FlatList = React.forwardRef<RNFlatList<any>, RNFlatListProps<any>>(
   (props, ref) => (
```

I think this will be better than duplicating RN types. Note that `Flatlist` is unchanged, I just wanted to check if other components will work this way.

yfunk:
@jakub-gonet With this solution all methods exist on the components like they should, but there is a new TS error with their `ref` prop:

```typescript
import { TextInput } from 'react-native-gesture-handler';

export const InputRefTest = () => {
	const ref = useRef<TextInput>(null);

	const handleOnFocus = () => {
		ref.current?.blur();
	};

	return <TextInput ref={ref} onFocus={handleOnFocus} />;
};
```
```
Type 'RefObject<TextInput>' is not assignable to type '((instance: ComponentClass<any, any> | FunctionComponent<any> | null) => void) | RefObject<ComponentType<any>> | null | undefined'.
  Type 'RefObject<TextInput>' is not assignable to type 'RefObject<ComponentType<any>>'.
    Type 'TextInput' is not assignable to type 'ComponentType<any>'.
      Type 'TextInput' is not assignable to type 'FunctionComponent<any>'.
        Type 'TextInput' provides no match for the signature '(props: any, context?: any): ReactElement<any, any> | null'.ts(2322)
```
What seems to work is defining the types like this, although I'm not sure if this could cause problems elsewhere.
```typescript
export type TextInput = typeof RNTextInput & RNTextInput;
```

yfunk:
@jakub-gonet I changed my branch to implement the type definition mentioned in my last comment. From my testing this works well to provide all the methods that exist on each of the wrapped components.

I also updated the type definition for FlatList to incorporate the generic item type parameter. This should fix the issue of the item type in `renderItem` being `any` instead of being inferred. From my (basic) testing `FlatList` now behaves just like the RN component.

src/components/GestureComponents.tsx
```diff
@@ -53,7 +49,8 @@ export const DrawerLayoutAndroid = createNativeWrapper<
 // @ts-ignore FIXME(TS) maybe this should be removed?
 DrawerLayoutAndroid.positions = { Left: 'left', Right: 'right' };
```
yfunk, line 50: Is this still necessary after defining the type of `DrawerLayoutAndroid` as `typeof RNDrawerLayoutAndroid & RNDrawerLayoutAndroid`?

src/components/GestureComponents.tsx
```diff
@@ -63,24 +60,10 @@ export const FlatList = React.forwardRef<RNFlatList<any>, RNFlatListProps<any>>(
...
-);
+) as <ItemT = any>(
+  props: PropsWithChildren<RNFlatListProps<ItemT>> &
+    RefAttributes<RNFlatList<ItemT>>,
+  ref: ForwardedRef<RNFlatList<ItemT>>
+) => ReactElement | null;
 // eslint-disable-next-line @typescript-eslint/no-redeclare
-export type FlatList<ItemT> = React.ComponentType<
-  RNFlatListProps<ItemT> &
-    NativeViewGestureHandlerProps &
```
yfunk, line 70: I removed `NativeViewGestureHandlerProps` from the type definition, because they were causing TypeScript errors. I don't think they were needed, since from what I can see they were not passed to the `ScrollView` component.

yfunk:
@jakub-gonet I'm not sure if the `typeof RNTextInput & RNTextInput` approach is too hacky, but from my understanding it is the same as typing  `typeof RNTextInput & { // All methods of RNTextInput }`.

We could also type it like you suggested in your diff and just fix the error created by this. From what I can tell it occurs, because the type of the forwarded ref on `ComponentWrapper` in `createNativeWrapper` is typed as `React.ComponentType<any>` ([here](https://github.com/software-mansion/react-native-gesture-handler/blob/b2fb8ad68fc598ca411819d9e786a745e7e5a2ee/src/handlers/createNativeWrapper.tsx#L28)), when it should match the type we export for the wrapped component and use to create a ref to it (in the case of the text input this would be `RNTextInput`). I can't really think of another way to accomplish this than adding a second generic parameter to `createNativeWrapper` for that type, which isn't that clean either and could be a breaking change.

src/components/GestureComponents.tsx
jakub-gonet: Removing those will break users' code using things like `waitFor` or `simultanousHandlers`.

src/components/GestureComponents.tsx
jakub-gonet: Probably not.

jakub-gonet (changes requested):
Sorry for the delay.

We cannot simply change those components to have React Native types, because we add additional properties there, so we'll be missing props defined in [`NativeViewGestureHandlerProps`](https://github.com/software-mansion/react-native-gesture-handler/blob/master/src/handlers/NativeViewGestureHandler.ts#L4) which extends [`BaseGestureHandlerProps`](https://github.com/software-mansion/react-native-gesture-handler/blob/master/src/handlers/gestureHandlers.ts#L42-L46)

yfunk:
> We cannot simply change those components to have React Native types, because we add additional properties there, so we'll be missing props defined in [`NativeViewGestureHandlerProps`](https://github.com/software-mansion/react-native-gesture-handler/blob/master/src/handlers/NativeViewGestureHandler.ts#L4) which extends [`BaseGestureHandlerProps`](https://github.com/software-mansion/react-native-gesture-handler/blob/master/src/handlers/gestureHandlers.ts#L42-L46)

@jakub-gonet Are you talking about the `type` exported by `GestureComponents.tsx` for each component?
It seems I messed something up there during testing, it should actually be `typeof TextInput & RNTextInput`. 

The question is, whether the methods should be duplicated from RN like in the first version of this PR or included by creating an intersection type with the RN component (e.g. `RNTextInput`) like in this version. From my understanding the type of the RN component simply consists of all the public methods and instance variables exposed by the class, so it should be exactly what we need (although please correct me if I'm wrong).

jakub-gonet:
I think intersecting with the original types is a better approach, we're just forwarding those methods. It makes sense to avoid removing things that force us to keep our types in sync with RN types.

jakub-gonet:
Out of curiosity, why you closed this one and created a new PR?

yfunk:
> Out of curiosity, why you closed this one and created a new PR?

Mostly to keep it clean in case it gets merged. I made the changes from master and didn't want to force push again, because it would have removed all previous changes in this PR.

jakub-gonet:
I see. Next time, please merge master into your branch instead of rebasing onto master – rebasing removes commits hashes so any review that was made lose connection to commits and it's harder for us to check changes since the last review.

Creating a new PR basically forces us to review everything once again.

PR #1391, opened 2021-02-28
Prop for ending pinching gesture when one finger is lifted (iOS)

### LINKED ISSUE #1214: Change Pinch Gesture Handler Behavior to transition to END state after one finger is taken off screen and one left
## Description

After testing PinchGestureHandler in react-native-gallery-toolkit I found that on the real device Pinch gesture is active even though we have only a single finger on the screen. This breaks my focal point logic which is working as expected on Android.

### Screenshots

As soon as I release one finger - it jumps because now the focal point is not the point between fingers, but that single finger location.

But it should just transition to state "END" instead.

RPReplay_Final1603130602

## Steps To Reproduce

1. Run pinch example
2. Add log to onGestureEvent of PinchHandler

```
onGestureEvent={(evt) => console.log(evt.nativeEvent.numberOfFingers)}
```

3. Do the pinch then release one finger and move it
4. See that it fires even with 1 finger pressed after wi

### Expected behavior

The gesture should be in the state "end" as soon as the user releases one finger – just like on Android

### Actual behavior

The gesture remains active even when the user released one finger

Run the example app

## Package versions

- React: 16.13.1
- React Native: 0.63.3
- React Native Gesture Handler: 1.8.0

iqqmuT:
## Description

Fixes #1214.

Probably the maintainers don't want to alter the native iOS behavior with this PR, but this fork can be useful for those who want to avoid writing workarounds.

This fork adds a new boolean prop for PinchGestureHandler (iOS only): **twoPointersRequired**
* `false` (default): no change for behavior; lets [UIPinchGestureRecognizer](https://developer.apple.com/documentation/uikit/uipinchgesturerecognizer) end the gesture when the user lifts off **both** fingers from the view
* `true`: end the gesture when the user lifts off **one** finger from the view

## Test plan

Tested on iPhone 6s Plus.

```js
<PinchGestureHandler twoPointersRequired={true}>
  ...
</PinchGestureHandler>
```

### DIFF
--- a/ios/Handlers/RNPinchHandler.m
+++ b/ios/Handlers/RNPinchHandler.m
@@ -8,17 +8,69 @@
 
 #import "RNPinchHandler.h"
 
+#import <UIKit/UIGestureRecognizerSubclass.h>
+
+@interface RNBetterPinchGestureRecognizer : UIPinchGestureRecognizer
+
+@property (nonatomic) BOOL twoPointersRequired;
+
+- (id)initWithGestureHandler:(RNGestureHandler*)gestureHandler;
+
+@end
+
+@implementation RNBetterPinchGestureRecognizer {
+    __weak RNGestureHandler *_gestureHandler;
+}
+
+- (id)initWithGestureHandler:(RNGestureHandler*)gestureHandler
+{
+    if ((self = [super initWithTarget:gestureHandler action:@selector(handleGesture:)])) {
+        _gestureHandler = gestureHandler;
+        _twoPointersRequired = NO;
+    }
+    return self;
+}
+
+- (void)touchesEnded:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
+{
+    [super touchesEnded:touches withEvent:event];
+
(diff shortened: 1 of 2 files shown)


ios/Handlers/RNPinchHandler.m
```diff
@@ -8,18 +8,70 @@
...
+- (void)touchesEnded:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
+{
+    [super touchesEnded:touches withEvent:event];
+
+    // UIPinchGestureRecognizer ends pinching gesture when the user lifts BOTH fingers
+    // from the view. If twoPointersRequired prop is true, we end the gesture when the
+    // user lifts one finger from the view.
+    if (_twoPointersRequired) {
+        self.state = UIGestureRecognizerStateEnded;
+        [self reset];
```
iqqmuT, line 43: Not sure if calling `reset` is required.

j-piasecki:
Closing this as we actually implemented the opposite in https://github.com/software-mansion/react-native-gesture-handler/pull/2983

PR #1398, opened 2021-03-12
Add algolia search

jakub-gonet:
## Description

Add Algolia search to make navigating docs easier.


plandem:
@jakub-gonet is search actually working? For me, it's not working in any browsers, while everything is fine for reanimated. `Network` tab has failed requests with http-403 errors.

Screen Shot 2022-10-18 at 18 36 47

PR #1415, opened 2021-03-27
Prevent inject `@types/node`

soh335:
## Description

Currently, TS compiled GenericTouchable has `/// <reference types="node" />`.
It means project that depend to react-native-gesture-handler ( of course, react-navigation, too)
is injected `@types/node`. I think most developers don't expect it.

I know this p-r isn't good solution, but react-navigation also seems to use
similar solutions.

https://github.com/react-navigation/react-navigation/blob/038eb87c42564f9d733e6870826726d3fb0adaee/packages/stack/src/views/KeyboardManager.tsx#L24
https://github.com/react-navigation/react-navigation/blob/dd48fe9b158814cab272fb85c74fc7079e8cef27/packages/stack/src/views/Stack/Card.tsx#L328-L334

### DIFF
--- a/src/components/touchables/GenericTouchable.tsx
+++ b/src/components/touchables/GenericTouchable.tsx
@@ -57,7 +57,7 @@ interface InternalProps {
 
 // TODO: maybe can be better
 // TODO: all clearTimeout have ! added, maybe they shouldn't ?
-type Timeout = ReturnType<typeof setTimeout> | null | undefined;
+type Timeout = any;
 
 /**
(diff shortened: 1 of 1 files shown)


jakub-gonet:
Could you check if the #1420 patch fixes this issue for you?

soh335:
@jakub-gonet Thanks. Maybe it's solved my issue. ( I'm not good at typescript ecosystem ... )

jakub-gonet:
I'm going to close this one in favor of #1420 then.

soh335:
thanks !

PR #1420, opened 2021-04-01
Remove implicit inclusion of types

### LINKED ISSUE #1384: This module unconditionaly includes node.d.ts for RN and this bring incorrect typings in project.
## Description

With latest update to react-native-gesture-handler@1.10.2, I discover some types errors. For example `setInterval` return type changed from `number` to `NodeJS.Timeout`.
After running fresh typescript version with `--explainFiles` option I got this:
```
node_modules/@types/node/index.d.ts
  Type library referenced via 'node' from file 'node_modules/react-native-gesture-handler/lib/typescript/components/touchables/GenericTouchable.d.ts' with packageId '@types/node/index.d.ts@14.14.31'
```
It is clearly incorrect. RN code should not be same environment as node.js )

Actual core refers to `NodeJS.Timeout` here: https://github.com/software-mansion/react-native-gesture-handler/blob/1.10.2/src/components/touchables/GenericTouchable.tsx#L58-L60
```typescript
// TODO: maybe can be better
// TODO: all clearTimeout have ! added, maybe they shouldn't ?
type Timeout = null | NodeJS.Timeout | undefined;
```

Can I, as a response to TODO comment, suggest to use this instead?
```typescript
type Timeout = ReturnType<typeof setTimeout>
```

jakub-gonet:
## Description

Typescript includes types even if they're not explicitly imported. This behavior makes some declaration types in Gesture Handler include node's typings which isn't desired since React Native shadows some declarations from Node (e.g. [`setInterval`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/globals.d.ts#L12)).

We're utilizing [`types`](https://www.typescriptlang.org/tsconfig#types) property to prevent TS from including all definitions from `node_modules/@types`.

Hopefully, finally fixes #1384

### References

- [Discussion about _excluding_ some packages' definitions and TS maintainer response](https://github.com/microsoft/TypeScript/issues/18588#issuecomment-704482601)
- [PR in TS introducing opaque type definitions](https://github.com/microsoft/TypeScript/issues/31894) - this can prevent those type of issues in the future

## Test plan

`yarn bob build`ed `GenericTouchable.d.ts` before and after:

### Before

```js
/// <reference types="node" />
import { Component } from 'react';
import { StyleProp, ViewStyle, TouchableWithoutFeedbackProps } from 'react-native';
import { GestureEvent, HandlerStateChangeEvent } from '../../handlers/gestureHandlers';
// (...)
```

### After

```js
import { Component } from 'react';
import { StyleProp, ViewStyle, TouchableWithoutFeedbackProps } from 'react-native';
import { GestureEvent, HandlerStateChangeEvent } from '../../handlers/gestureHandlers';
// (...)
```

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,6 +4,7 @@
     "esModuleInterop": true,
     "jsx": "react-native",
     "lib": ["esnext"],
+    "types": [],
     "module": "esnext",
     "moduleResolution": "node",
     "resolveJsonModule": true,
@@ -12,6 +13,7 @@
     "target": "esnext",
     "declaration": true,
     "declarationMap": false,
+
     "noStrictGenericChecks": false,
     "forceConsistentCasingInFileNames": true,
     "noImplicitUseStrict": false,


soh335:
@jakub-gonet Can you release it ?

TimoGlastra:
Hi when is the next release scheduled? The node type inclusion is giving me clashes with Node.JS and React Native types. Seems like this would solve it :)

Issue #139, opened 2018-04-11
Doesn't work in a modal


martinezguillaume:
Hi !
I have a little issue on Android, `onGestureEvent` is not trigger when GestureHandler components is on a modal on Android. When I change the modal to a View, it works perfectly 👌

No problem on iOS

mordaha:
+1 
any workaround?

martinezguillaume:
I don’t use Modal for now, I just re-create it with a View in absolute position

csto:
+1

osdnk:
Hi, @martinezguillaume, @mordaha, @csto 
I did take a look on this issue and I suppose it's not a matter of our library, but RN Core.

mars-lan:
@osdnk is this fixable?

ParhamZare:
+1 gesture does not work on modal

ewendel:
I also ran into this bug today.

Via-profit:
Same problem. Gesture does not work on modal on **Android**.
[Expo for example](https://snack.expo.io/@vas-vas/rn-gesture-handler-issue-139)

Dmitrylolo:
Same for me. Tested react-native's PanResponder handlers on modal - works fine.

Dmitrylolo:
It even not works on react-navigation new screen. Ie 
```
const RootStack = createStackNavigator({
  App: {
    screen: AppNavigator,
  },
  UploadPhoto: {
    screen: UloadPhoto,
  },
  AddExercise: {
    screen: AddExercise,
  },
});
``` 
not works on UploadPhoto and AddExercise, when navigating from screen from AppNavigator stack

LaVielle:
I was using a standard React Native `Modal` and was experiencing this issue.
I worked around the problem by creating a new screen and displaying it as a modal. I'm using [react-native-navigation](https://wix.github.io/react-native-navigation), so:
```JavaScript
Navigation.showModal({
    component: {
      name: navRoutes.ImageModal,
      passProps: { image },
    },
  })
```
Gestures are working as expected on both iOS and Android, and I still get transparent background that I wanted from my original modal.

osdnk:
I have dug into it for a while.

1. It's a problem of RNGH. Sorry 😒 
2. It happens bc modals are not rendered below an RN root view. 
3. It's fixable by replacing modals' mechanism in a similar way as we do with an RNRootView by wrapping it with some extra logic.
[...]

PR #1493, opened 2021-06-29
Render nested GestureHandlerRootView so that it works in a Modal in Android

oxilor:
Detailed description with examples and videos: https://github.com/software-mansion/react-native-gesture-handler/issues/139#issuecomment-870554751

### DIFF
--- a/src/GestureHandlerRootView.android.tsx
+++ b/src/GestureHandlerRootView.android.tsx
@@ -1,34 +1,17 @@
 import * as React from 'react';
 import { PropsWithChildren } from 'react';
-import { View, requireNativeComponent } from 'react-native';
+import { requireNativeComponent } from 'react-native';
 
 const GestureHandlerRootViewNative = requireNativeComponent(
   'GestureHandlerRootView'
 );
 
-const GestureHandlerRootViewContext = React.createContext(false);
-
 type Props = PropsWithChildren<Record<string, unknown>>;
 
 export default function GestureHandlerRootView({ children, ...rest }: Props) {
   return (
-    <GestureHandlerRootViewContext.Consumer>
-      {(available) => {
-        if (available) {
-          // If we already have a parent wrapped in the gesture handler root view,
-          // We don't need to wrap it again in root view
-          // We still wrap it in a normal view so our styling stays the same
-          return <View {...rest}>{children}</View>;
-        }
-
-        return (
-          <GestureHandlerRootViewContext.Provider value>
-            <GestureHandlerRootViewNative {...rest}>
-              {children}
-            </GestureHandlerRootViewNative>
-          </GestureHandlerRootViewContext.Provider>
-        );
-      }}
-    </GestureHandlerRootViewContext.Consumer>
+    <GestureHandlerRootViewNative {...rest}>
+      {children}
+    </GestureHandlerRootViewNative>
   );
 }


DigitalZebra:
Possible alternative implementation would be to expose `GestureHandlerRootViewNative` for clients to use directly?

oxilor:
> Possible alternative implementation would be to expose `GestureHandlerRootViewNative` for clients to use directly?

One of the options is to add an additional optional parameter that determines whether the `GestureHandlerRootViewNative` component must be rendered forcibly (like `forcedRender: boolean`).

oxilor:
Tested:
[RNGH: nested GestureHandlerRootView renders as View](https://www.youtube.com/watch?v=ahcHiHNQjmk)

oxilor:
Is everything all right? Is it planned to merge?

urbanclap-admin:
@oxilor Thanks for raising this MR. We have been facing the same issue with nested Modals using `GestureHandlerRootView` as the inner nested Modal receives true for the `if(available)` condition but the root view does not actually exist as the parent and thus any gesture handler components like PanGestureHandler or touchables does not work. 
This is a great solution to keep the optimization as well as provide some control to the developer to override default behaviour.

@jakub-gonet please push this fix along as fast as possible as it is a major bug fix for the system but not a big change in terms of code.

oxilor:
@urbanclap-admin Do you know a case where it is necessary to render nested root view as a normal View? If there are no cases (in the beginning we need to consult with @jakub-gonet or may be you know any case), I think that it is better to remove this condition at all.

Otherwise, we can leave the current implementation, which allows a developer to control whether the root view must be rendered as `GestureHandlerRootViewNative` or not.

jakub-gonet (commented):
Hey, sorry for the delay.

> Why was it decided not to render the nested GestureHandlerRootView?

The problem with rendering nested GH root views is that they will create [different root helpers](https://github.com/software-mansion/react-native-gesture-handler/blob/79e4c34fabb211ff203fece3eae21fa3d48a1612/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootView.java#L51) which in turn will [create different orchestrators](https://github.com/software-mansion/react-native-gesture-handler/blob/79e4c34fabb211ff203fece3eae21fa3d48a1612/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.java#L59). This will prevent interaction between handlers across root views as [mentioned by @kmagiera](https://github.com/software-mansion/react-native-gesture-handler/pull/925#discussion_r370018179) and [in our docs](https://docs.swmansion.com/react-native-gesture-handler/docs/#for-library-authors):
> If you use props such as `shouldCancelWhenOutside`, `simultaneousHandlers`, `waitFor` etc. with gesture handlers, the handlers need to be mounted under a single `GestureHandlerRootView`. So it's important to keep the `GestureHandlerRootView` as close to the actual root view as possible.

That's why we render only one root view in React tree right now.

---

> The comment above this code says:
> > If we already have a parent wrapped in the gesture handler root view, we don't need to wrap it again in root view we still wrap it in a normal view so our styling stays the same.
>
> What styles stays the same (all properties, including styles, are passed to the normal View) and why?

It's [explained in PR linked above](https://github.com/software-mansion/react-native-gesture-handler/pull/925#discussion_r370056712). Basically we need consistent styling even if root view isn't applied so that's a difference between adding view and not adding it.

---

The problem with modals is that they're [rendered in the same React tree but different native view hierarchy](https://github.com/facebook/react-native/blob/a8a85034e1cc23a144e58a9fb3215afbaebedb19/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.java#L47-L61). When creating a new root view we check [if we did that earlier](https://github.com/software-mansion/react-native-gesture-handler/blob/79e4c34fabb211ff203fece3eae21fa3d48a1612/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootView.java#L42) and your change most likely doesn't show this info in logs due to modals but will if you use it without one.

I thought about the correct solution and it seems like the simplest one is to always wrap views in the root view (which will make HOC work) and remove the check for `mEnabled` in the native side (with an appropriate comment about modals in the following if). I haven't tested that though and I won't spend much time on RNGH in the few following weeks so feel free to experiment with that. If you come with an alternative solution, please share it with me - maybe it can be solved better.

EDIT: this won't solve the problem with the orchestrator, it probably should be shared between root views to correctly register and handle all gesture handlers.

satya164:
@jakub-gonet Wouldn't `hasGestureHandlerEnabledRootView` return `true` if there is already a root view? I had added this check natively as well because Expo apps already had a root view so we didn't want to wrap again. We can probably remove the check from JS side.

jakub-gonet:
> Wouldn't hasGestureHandlerEnabledRootView return true if there is already a root view?

If you're talking about a typical React hierarchy, then yes. But modals create their own hierarchy so there's no root view in their branch and this method returns `false`.

If we remove the check from the JS side then we need to remove `Log.i` telling that RNGH is already enabled (which will trigger when you have React Navigation + some random component using `GestureHandlerHOC` for example). But this isn't still solution to this problem since interactions between handlers above Modal and below modal won't work together. We need to change GestureHandler class to have backing view_s_ - basically, generalize it a bit. I'm fine with the first iteration that adds support for modals when you have nested root views in React hierarchy but we have to fix this problem eventually as well.

satya164:
> But this isn't still solution to this problem since interactions between handlers above Modal and below modal won't work together

I think this would be a very niche use case if someone even needs that.

Anyway, my main point was that the check on JS side isn't needed since we already check on native side and since it prevents modals from working.

PR #1504, opened 2021-07-14
Fix iteration over argument mapping on web

machouz:
## URGENT

`index of` was reading properties and indexes instead of reading only indexes which made the next line `const [key, value] = argMapping[index]` crash

Screen Shot 2021-07-14 at 17 06 21


jakub-gonet (approved):

jakub-gonet:
🍉 🍉 🍉

PR #1505, opened 2021-07-15
Add duration to long press event

j-piasecki:
### Description

Added duration property to the LongPressGestureHandler events. This change allows to check how long the entire event lasted.

Code I used to test it:
```js
import React, { Component } from 'react';

import {
  LongPressGestureHandler,
  LongPressGestureHandlerStateChangeEvent,

export class PressBox extends Component {
  private onHandlerStateChange = (
    event: LongPressGestureHandlerStateChangeEvent
  ) => {
    console.log(`Duration: ${event.nativeEvent.duration}`);
  };

  render() {
    return (
      <LongPressGestureHandler
        minDurationMs={600}>
          <View style={styles.box} />
      </LongPressGestureHandler>
    );
  }
}

  render() {
    return (
      <PressBox />
    );
  }
}

  box: {
    width: 150,
    height: 150,
    alignSelf: 'center',
    backgroundColor: 'plum',
    margin: 10,
    zIndex: 200,
  },
});
```

### DIFF
--- a/ios/Handlers/RNLongPressHandler.m
+++ b/ios/Handlers/RNLongPressHandler.m
@@ -12,12 +12,15 @@
 
 #import <React/RCTConvert.h>
 
+#import <mach/mach_time.h>
+
 @interface RNBetterLongPressGestureRecognizer : UILongPressGestureRecognizer {
-  double startTime;
+  uint64_t startTime;
+  uint64_t previousTime;
 }
 
 - (id)initWithGestureHandler:(RNGestureHandler*)gestureHandler;
-- (void)handle:(UIGestureRecognizer *)recognizer;
+- (void)handleGesture:(UIGestureRecognizer *)recognizer;
 - (NSUInteger) getDuration;
 
 @end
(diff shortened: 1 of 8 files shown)


ios/Handlers/RNLongPressHandler.m
```diff
@@ -24,12 +28,21 @@ @implementation RNBetterLongPressGestureRecognizer {
...
+  if ((self = [super initWithTarget:self action:@selector(handle:)])) {
     _gestureHandler = gestureHandler;
   }
   return self;
 }
 
+- (void)handle:(UIGestureRecognizer *)recognizer
+{
+  if (recognizer.state == UIGestureRecognizerStateBegan) {
+    startTime = [NSDate timeIntervalSinceReferenceDate];
```
jakub-gonet, line 40: You probably want to use [`mach_absolutetime`](https://stackoverflow.com/a/30363702) here.

jakub-gonet (commented):
If you have a code example you tested it on, please attach it to the PR description.

I wonder if calculating duration in `extractEventData` is a good idea since this effectively ties this method to state transitions. We can alternatively store it and send it only when the state transitions to FINISHED or FAILED.

ios/Handlers/RNLongPressHandler.m
j-piasecki: I have changed how the duration is calculated and now it is updated when the state changes and I have replaced `timeIntervalSinceReferenceDate` with `mach_absolute_time`.

jakub-gonet (approved):

jakub-gonet:
🍉 🍉 🍉

PR #1515, opened 2021-07-28
Convert ButtonViewManager to Kotlin

jakub-gonet:
## Test plan

Tested on Example App.


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -0,0 +1,276 @@
...
+          SELECTABLE_ITEM_BACKGROUND == attr -> {
+            R.attr.selectableItemBackground
+          }
+          SELECTABLE_ITEM_BACKGROUND_BORDERLESS == attr -> {
+            R.attr.selectableItemBackgroundBorderless
+          }
+          else -> {
+            context.resources.getIdentifier(attr, "attr", "android")
+          }
+        }
```
j-piasecki, line 272: We can move `attr` directly to `when` clause
```suggestion
        return when (attr) {
          SELECTABLE_ITEM_BACKGROUND -> {
            R.attr.selectableItemBackground
          }
          SELECTABLE_ITEM_BACKGROUND_BORDERLESS -> {
            R.attr.selectableItemBackgroundBorderless
          }
          else -> {
            context.resources.getIdentifier(attr, "attr", "android")
          }
        }
```

jakub-gonet: Fixed in e37404e.

j-piasecki (approved):

PR #1518, opened 2021-07-28
Convert base GestureHandler class to Kotlin

jakub-gonet:
## Test plan

Tested on Example App.


android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
```diff
@@ -0,0 +1,482 @@
...
+      }
+      if (hitSlopSet(height)) {
+        if (!hitSlopSet(padTop)) {
+          top = bottom - height
+        } else if (!hitSlopSet(padBottom)) {
+          bottom = top + height
+        }
+      }
+    }
+    return posX in left..right && posY >= top && posY <= bottom
```
j-piasecki, line 367: Did you have a reason for using range operator for one axis and not for the other, or is it just a mistake?

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
```diff
@@ -0,0 +1,482 @@
...
+  fun hasCommonPointers(other: GestureHandler<*>): Boolean {
+    for (i in mTrackedPointerIDs.indices) {
+      if (mTrackedPointerIDs[i] != -1 && other.mTrackedPointerIDs[i] != -1) {
+        return true
+      }
+    }
+    return false
+  }
+
+  fun setShouldCancelWhenOutside(shouldCancelWhenOutside: Boolean) = applySelf {
```
j-piasecki, line 85: For me it was not instantly obvious that this method (and others using `applySelf`) is returning a `ConcreteGestureHandlerT`. Maybe we should state is explicitly?
```suggestion
  fun setShouldCancelWhenOutside(shouldCancelWhenOutside: Boolean): ConcreteGestureHandlerT
    = applySelf {
```

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
jakub-gonet: This is a check for posY being out of range (inclusive).

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
jakub-gonet: Resolved in b565f2f.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
jakub-gonet: Resolved in 3ad3f8c.

j-piasecki (approved):

PR #1521, opened 2021-07-29
Convert helpers to Kotlin

jakub-gonet:
## Test plan

Tested on Example App.


android/lib/src/main/java/com/swmansion/gesturehandler/ViewConfigurationHelper.kt
```diff
@@ -0,0 +1,10 @@
+package com.swmansion.gesturehandler
+
+import android.view.View
+import android.view.ViewGroup
+
+interface ViewConfigurationHelper {
+  fun getPointerEventsConfigForView(view: View): PointerEventsConfig?
+  fun getChildInDrawingOrderAtIndex(parent: ViewGroup, index: Int): View?
```
j-piasecki, line 8: In the implementation the return types are not nullable, shouldn't they also be there?
```suggestion
  fun getPointerEventsConfigForView(view: View): PointerEventsConfig
  fun getChildInDrawingOrderAtIndex(parent: ViewGroup, index: Int): View
```

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
```diff
@@ -0,0 +1,134 @@
...
+      UiThreadUtil.runOnUiThread { tryCancelAllHandlers() }
+    }
+  }
+
+  companion object {
+    private const val MIN_ALPHA_FOR_TOUCH = 0.1f
+    private fun findRootViewTag(viewGroup: ViewGroup): ViewGroup {
+      UiThreadUtil.assertOnUiThread()
+      var parent: ViewParent? = viewGroup
+      while (parent != null && !(parent is ReactRootView || RNGHModalUtils.isDialogRootViewGroup(parent))) {
```
j-piasecki, line 125: Personally, I would change that expression, not sure if it is better though
```suggestion
      while (parent != null && parent !is ReactRootView && !RNGHModalUtils.isDialogRootViewGroup(parent)) {
```

jakub-gonet: Makes sense, fixed in 36b27e3.

android/lib/src/main/java/com/swmansion/gesturehandler/ViewConfigurationHelper.kt
jakub-gonet: I checked usages and they're not expecting nulls, that's the right call.
Fixed in 75fba25.

j-piasecki (approved):

PR #1525, opened 2021-07-30
Rewrite NativeView to Kotlin

jakub-gonet:
## Test plan

Tested on the example app.


android/lib/src/main/java/com/swmansion/gesturehandler/NativeViewGestureHandler.kt
```diff
@@ -0,0 +1,104 @@
...
+  override fun resetConfig() {
+    super.resetConfig()
+    shouldActivateOnStart = false
+    disallowInterruption = false
+  }
+
+  fun setShouldActivateOnStart(shouldActivateOnStart: Boolean): NativeViewGestureHandler {
+    this.shouldActivateOnStart = shouldActivateOnStart
+    return this
+  }
```
jakub-gonet, line 25:
```suggestion
  fun setShouldActivateOnStart(shouldActivateOnStart: Boolean) = apply {
    this.shouldActivateOnStart = shouldActivateOnStart
  }
```

j-piasecki (approved):

android/lib/src/main/java/com/swmansion/gesturehandler/NativeViewGestureHandler.kt
jakub-gonet: Fixed in be9c54b.

PR #1527, opened 2021-07-30
Convert PinchGH to Kotlin

jakub-gonet:
## Test plan

Tested on example app.


j-piasecki (commented):
I would rename the variables to match the style of previous pull requests.

j-piasecki (approved):

PR #1529, opened 2021-08-02
Convert RotationGH to Kotlin

jakub-gonet:
## Test plan

Tested on example app.


android/lib/src/main/java/com/swmansion/gesturehandler/RotationGestureDetector.kt
```diff
@@ -0,0 +1,125 @@
...
+    val secondPtX = event.getX(secondPointerIndex)
+    val secondPtY = event.getY(secondPointerIndex)
+    val vectorX = secondPtX - firstPtX
+    val vectorY = secondPtY - firstPtY
+    anchorX = (firstPtX + secondPtX) * 0.5f
+    anchorY = (firstPtY + secondPtY) * 0.5f
+
+    // Angle diff should be positive when rotating in clockwise direction
+    val angle = -atan2(vectorY.toDouble(), vectorX.toDouble())
+    rotation = if (java.lang.Double.isNaN(previousAngle)) {
```
j-piasecki, line 71:
```suggestion
    rotation = if (previousAngle.isNaN()) {
```

j-piasecki (approved):

PR #1533, opened 2021-08-03
Rewrite Orchestrator to Kotlin

jakub-gonet:
## Test plan

Tested on example app.


android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
```diff
@@ -0,0 +1,511 @@
...
+        handler.dispatchStateChange(GestureHandler.STATE_UNDETERMINED, GestureHandler.STATE_END)
+      }
+    }
+  }
+
+  private fun deliverEventToGestureHandlers(event: MotionEvent) {
+    // Copy handlers to "prepared handlers" array, because the list of active handlers can change
+    // as a result of state updates
+    val handlersCount = gestureHandlersCount
+    System.arraycopy(gestureHandlers, 0, preparedHandlers, 0, handlersCount)
```
j-piasecki, line 197:
```suggestion
    gestureHandlers.copyInto(preparedHandlers, 0, 0, handlersCount)
```

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
```diff
@@ -0,0 +1,511 @@
...
+    // We want to deliver events to active handlers first in order of their activation (handlers
+    // that activated first will first get event delivered). Otherwise we deliver events in the
+    // order in which handlers has been added ("most direct" children goes first). Therefore we rely
+    // on Arrays.sort providing a stable sort (as children are registered in order in which they
+    // should be tested)
+    Arrays.sort(preparedHandlers, 0, handlersCount, handlersComparator)
```
j-piasecki, line 203:
```suggestion
    preparedHandlers.sortWith(handlersComparator, 0, handlersCount)
```

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
```diff
@@ -0,0 +1,511 @@
...
+        val handler = it[i]
+        if (handler.isEnabled && handler.isWithinBounds(view, coords[0], coords[1])) {
+          recordHandlerIfNotPresent(handler, view)
+          handler.startTrackingPointer(pointerId)
+          found = true
+        }
+        i++
+      }
+    }
+    return found
```
j-piasecki, line 338: I'm not sure whether this change makes it possible for `ConcurrentModificationException` to occur or not (when attaching a new handler to a view). If that's the case, I think we can change it to use the range operator.
```suggestion
    var found = false
    handlerRegistry.getHandlersForView(view)?.let { handlers ->
      for (handler in handlers) {
        if (handler.isEnabled && handler.isWithinBounds(view, coords[0], coords[1])) {
          recordHandlerIfNotPresent(handler, view)
          handler.startTrackingPointer(pointerId)
          found = true
        }
      }
    }
    return found
```

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
```diff
@@ -0,0 +1,511 @@
...
+      var found = false
+      if (view is ViewGroup) {
+        found = extractGestureHandlers(view, coords, pointerId)
+      }
+      (recordViewHandlersForPointer(view, coords, pointerId)
+        || found || shouldHandlerlessViewBecomeTouchTarget(view, coords))
+    } else {
+      throw IllegalArgumentException(
+        "Unknown pointer event type: $pointerEvents")
+    }
```
j-piasecki, line 402:
```suggestion
    return when (viewConfigHelper.getPointerEventsConfigForView(view)) {
      PointerEventsConfig.NONE -> {
        // This view and its children can't be the target
        false
      }
      PointerEventsConfig.BOX_ONLY -> {
        // This view is the target, its children don't matter
        (recordViewHandlersForPointer(view, coords, pointerId)
          || shouldHandlerlessViewBecomeTouchTarget(view, coords))
      }
      PointerEventsConfig.BOX_NONE -> {
        // This view can't be the target, but its children might
        if (view is ViewGroup) {
          extractGestureHandlers(view, coords, pointerId)
        } else false
      }
      PointerEventsConfig.AUTO -> {
        // Either this view or one of its children is the target
        var found = false
        if (view is ViewGroup) {
          found = extractGestureHandlers(view, coords, pointerId)
        }
        (recordViewHandlersForPointer(view, coords, pointerId)
          || found || shouldHandlerlessViewBecomeTouchTarget(view, coords))
      }
    }
```

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
```diff
@@ -0,0 +1,511 @@
...
+    private fun isTransformedTouchPointInView(x: Float, y: Float, child: View) =
+      x in 0f..child.width.toFloat() && y in 0f..child.height.toFloat()
+
+    private fun shouldHandlerWaitForOther(handler: GestureHandler<*>, other: GestureHandler<*>): Boolean {
+      return handler !== other && (handler.shouldWaitForHandlerFailure(other)
+        || other.shouldRequireToWaitForFailure(handler))
+    }
+
+    private fun canRunSimultaneously(a: GestureHandler<*>?, b: GestureHandler<*>) =
+      a === b || a!!.shouldRecognizeSimultaneously(b) || b.shouldRecognizeSimultaneously(a)
```
j-piasecki, line 485: The only usage of this method is in the one right below it and that one has non-nullable arguments.
```suggestion
    private fun canRunSimultaneously(a: GestureHandler<*>, b: GestureHandler<*>) =
      a === b || a.shouldRecognizeSimultaneously(b) || b.shouldRecognizeSimultaneously(a)
```

jakub-gonet: Fixed in eaa3a0f.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
jakub-gonet: Fixed in 28f2023 and 
2d634e0.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
jakub-gonet: I went with the safe way here by using ranges.
Fixed in c4c41eb, 52b833d.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
jakub-gonet: Fixed in 1cdc2fb.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
jakub-gonet: Fixed in 1cdc2fb.

j-piasecki (approved):

PR #1555, opened 2021-08-11
Add try catch block to prevent crash for multitouch event

monikapencierzynska:
## Description

Workaround for the multitouch issue: `java.lang.IllegalArgumentException: pointerIndex out of range`

Solution based on:
https://github.com/facebook/react-native/issues/30320


RodolfoGS:
Hi @monikapencierzynska, what was the reason to close this PR? Could you fixed the crash using another approach? Thanks.

PR #1560, opened 2021-08-13
Restructure files

j-piasecki:
## Description

Extract generation of handler tags and the id to tag mapping from `createHandler` to separate file.
Extract allowed props of each gesture handler from `gestureHandlers` to separate file.
This change is a base for creating a new API that is interoperable with the old one.

## Test plan

Tested on the Example app


src/handlers/allowedProps.ts
```diff
@@ -0,0 +1,249 @@
...
+  'failOffsetXStart',
+  'failOffsetXEnd',
+] as const;
+
+export const nativeViewGestureHandlerProps = [
+  'shouldActivateOnStart',
+  'disallowInterruption',
+] as const;
+
+function validatePanGestureHandlerProps(props: PanGestureHandlerProps) {
```
jakub-gonet, line 82: I think this function should be in `createHandler.ts` file since it's used only there and its sole purpose is to validate Pan props.

jakub-gonet (changes requested):
tbh, the better way of structuring those files would be to split `gestureHandlers.ts` file into file with every handler we create (following the path of NativeViewGH and keep those prop array local to them (and exporting it if needed).

src/handlers/allowedProps.ts
j-piasecki: Since `gestureHandlers.ts` is being split into separate files for each handler, can we put those methods in the same file as `PanGestureHandler`? If I have the same idea about splitting, then that will be the only place it will be used.

jakub-gonet: Yup, I wrote this comment before comment on review.

j-piasecki: Done in e80d121cf21ee01911a14ec4ac83e6dd889ef436

src/handlers/FlingGestureHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+export interface FlingGestureHandlerProps
+  extends BaseGestureHandlerProps<FlingGestureHandlerEventPayload> {
+  direction?: number;
+  numberOfPointers?: number;
+}
+
+export type FlingGestureHandler = typeof FlingGestureHandler;
+// eslint-disable-next-line @typescript-eslint/no-redeclare -- backward compatibility; see description on the top of this file
```
jakub-gonet, line 26: This comment should point to gestureHandlerCommon.ts.

src/handlers/LongPressGestureHandler.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export interface LongPressGestureHandlerProps
+  extends BaseGestureHandlerProps<LongPressGestureHandlerEventPayload> {
+  minDurationMs?: number;
+  maxDist?: number;
+}
+
+export type LongPressGestureHandler = typeof LongPressGestureHandler;
+// eslint-disable-next-line @typescript-eslint/no-redeclare -- backward compatibility; see description on the top of this file
```
jakub-gonet, line 27: Same comment here.

src/handlers/PinchGestureHandler.ts
```diff
@@ -0,0 +1,26 @@
...
+  focalX: number;
+  focalY: number;
+  velocity: number;
+};
+
+export interface PinchGestureHandlerProps
+  extends BaseGestureHandlerProps<PinchGestureHandlerEventPayload> {}
+
+export type PinchGestureHandler = typeof PinchGestureHandler;
+// eslint-disable-next-line @typescript-eslint/no-redeclare -- backward compatibility; see description on the top of this file
```
jakub-gonet, line 18: And here.

src/handlers/RotationGestureHandler.ts
```diff
@@ -0,0 +1,26 @@
...
+  anchorX: number;
+  anchorY: number;
+  velocity: number;
+};
+
+export interface RotationGestureHandlerProps
+  extends BaseGestureHandlerProps<RotationGestureHandlerEventPayload> {}
+
+export type RotationGestureHandler = typeof RotationGestureHandler;
+// eslint-disable-next-line @typescript-eslint/no-redeclare -- backward compatibility; see description on the top of this file
```
jakub-gonet, line 18: .

src/handlers/handlersRegistry.ts
```diff
@@ -0,0 +1,7 @@
+export const handlerIDToTag: Record<string, number> = {};
```
jakub-gonet, line 1: We don't use it.

jakub-gonet (commented):
IMHO name `handlersRegistry` is a bit of a stretch since we don't register anything. Besides that, looks good.

src/handlers/FlingGestureHandler.ts
j-piasecki: Fixed in 4f58a3ff228fde7095fefb8cfec8843288fcb2af

src/handlers/LongPressGestureHandler.ts
j-piasecki: Fixed in 4f58a3ff228fde7095fefb8cfec8843288fcb2af

src/handlers/PinchGestureHandler.ts
j-piasecki: Fixed in 4f58a3ff228fde7095fefb8cfec8843288fcb2af

src/handlers/RotationGestureHandler.ts
j-piasecki: Fixed in 4f58a3ff228fde7095fefb8cfec8843288fcb2af

src/handlers/handlersRegistry.ts
j-piasecki: It's used in `createHandler.ts`.

jakub-gonet (approved):

PR #1572, opened 2021-08-16
Add Gestures with GestureBuilder

j-piasecki:
## Description

Add classes representing all basic gestures.
Add `GestureBuilder` for easy composition of gestures.

## Test plan

This code is not yet used.


src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,381 @@
...
+    this.handlerTag = getNextHandlerTag();
+    this.handlers = { ...this.handlers, handlerTag: this.handlerTag };
+
+    if (this.config.ref) {
+      this.config.ref.current = this;
+    }
+  }
+
+  build(): BuiltGesture {
+    let result = new BuiltGesture(this.prepare);
```
jakub-gonet, line 141:
```suggestion
    const result = new BuiltGesture(this.prepare);
```

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,381 @@
...
+
```
jakub-gonet, line 142: 🌵 No need to add additional whitespace here.

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  requireToFail(gesture: SimpleGesture): GestureBuilder {
+    this.log.push({ relation: Relation.RequireToFail, gesture: gesture });
+
+    return this;
+  }
+
+  build(): BuiltGesture {
+    let result = new BuiltGesture(this.prepare);
```
jakub-gonet, line 47:
```suggestion
    const result = new BuiltGesture(this.prepare);
```

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    result.gestures = [];
+
+    for (const pg of this.log) {
+      result.gestures.push(pg.gesture);
+    }
+
+    return result;
```
jakub-gonet, line 55:
```ts
result.gestures = this.log.map(prepared => prepared.gesture);
return result;
```

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  prepare = () => {
+    let simultaneous = [];
+    let after = [];
+    let waitFor = [];
```
jakub-gonet, line 61:
```suggestion
    const simultaneous = [];
    const after = [];
    const waitFor = [];
```

jakub-gonet: There are more variables with `let` that can be `const` but I won't add comments to them to decrease the signal to noise ratio 😄

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,381 @@
...
+    return result;
+  }
+
+  prepare = () => {
+    if (
+      this.config.requireToFail &&
+      !Array.isArray(this.config.requireToFail)
+    ) {
+      this.config.requireToFail = [this.config.requireToFail];
+    }
```
jakub-gonet, line 154:
```js
private toArray(x: unknown) {
	return [].concat(x);	
}

const config = this.config;

if (config.requireToFail !== undefined) {
	config.requireToFail = toArray(config.requireToFail);
}

// ...
```

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,381 @@
...
+    return this;
+  }
+
+  getAllowedProps() {
+    return Pan.allowedProps;
+  }
+
+  transformProps() {
+    return managePanProps(this.config);
+  }
```
jakub-gonet, line 278: This shouldn't be part of the API.

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,381 @@
...
+  private addDependency(
+    key: string,
+    gesture: SimpleGesture | React.RefObject<SimpleGesture>
+  ) {
+    if (this.config[key]) {
+      if (Array.isArray(this.config[key])) {
+        this.config[key].push(gesture);
+      } else {
+        this.config[key] = [this.config[key], gesture];
+      }
```
jakub-gonet, line 52: `this.config[key] = this.config[key] ? [].concat(this.config[key], gesture) : [gesture]`

It may be too dense so you can remove ternary if needed.

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,381 @@
...
+      this.config[key] = [gesture];
+    }
+  }
+
+  setRef(ref: React.RefObject<any>) {
+    this.setConfig('ref', ref);
+    return this;
+  }
+
+  setOnBegan(callback: (event: any) => void) {
```
jakub-gonet, line 63: `any`

jakub-gonet (changes requested):
Partial review.
Make sure you remove `any` and change `let` to `const` if possible.

src/handlers/gestureHandlerCommon.ts
```diff
@@ -29,6 +29,8 @@ export const baseGestureHandlerProps = [
   'onHandlerStateChange',
 ] as const;
 
+export const baseGestureHandlerWithMonitorProps = [...commonProps] as const;
```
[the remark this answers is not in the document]
j-piasecki: I ended up removing it in c207736e230eaea9e6e0261b51b19e1c411c7214

src/handlers/gestures/gesture.ts
```diff
@@ -0,0 +1,50 @@
...
+  Rotation,
+  Fling,
+  LongPress,
+  ForceTouch,
+} from './simpleGestures';
+
+export abstract class Gesture {
+  _requireToFail = [];
+  _after = [];
+  _simultaneousWith = [];
```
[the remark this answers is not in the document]
j-piasecki: They were intended more as 'package' properties. I think I have an idea on how to do without them so I removed them for now in 3cbebb608896a451b2905503150cbf974021507c.

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,126 @@
...
+  RequireToFail,
+}
+
+type PendingGesture = {
+  relation: Relation;
+  gesture: SimpleGesture;
+};
+
+export class GestureBuilder {
+  private log: Array<PendingGesture> = [];
```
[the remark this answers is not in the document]
j-piasecki: Fixed in dfde77640e3f6ffd4098264ae82f9911c45aa371 and fdabf483d5407f3ea1c8fb6f0c75309a533efe41

src/handlers/gestures/gestureBuilder.ts
j-piasecki: Fixed in fdabf483d5407f3ea1c8fb6f0c75309a533efe41

src/handlers/gestures/simpleGestures.ts
j-piasecki: Fixed in fdabf483d5407f3ea1c8fb6f0c75309a533efe41

src/handlers/gestures/gestureBuilder.ts
j-piasecki: Fixed in fdabf483d5407f3ea1c8fb6f0c75309a533efe41

src/handlers/gestures/gestureBuilder.ts
j-piasecki: Fixed in 6e24ae8e56c72a33ca7aec2546081e12507e2ad0

src/handlers/gestures/simpleGestures.ts
j-piasecki: Fixed in 6e24ae8e56c72a33ca7aec2546081e12507e2ad0

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,126 @@
...
+    this.log.push({ relation: Relation.Simultaneous, gesture: gesture });
+
+    return this;
+  }
+
+  exclusiveWith(gesture: SimpleGesture): GestureBuilder {
+    this.log.push({ relation: Relation.Exclusive, gesture: gesture });
+
+    return this;
+  }
```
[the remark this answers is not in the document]
j-piasecki: I think it's a great idea, added in 726ac49b835503aa77b6eac20656613ad0e93a0d

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,126 @@
...
+export class GestureBuilder {
+  private log: Array<PendingGesture> = [];
+
+  constructor(base: SimpleGesture) {
+    this.log.push({ relation: Relation.Exclusive, gesture: base });
+  }
+
+  simultaneousWith(gesture: SimpleGesture): GestureBuilder {
+    this.log.push({ relation: Relation.Simultaneous, gesture: gesture });
+
```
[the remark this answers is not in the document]
j-piasecki: Fixed in 726ac49b835503aa77b6eac20656613ad0e93a0d

src/handlers/gestures/simpleGestures.ts
j-piasecki: I think it's fine, replaced in 23b5ab772057790132673b0b74507c0e86d131d0

src/handlers/gestures/simpleGestures.ts
j-piasecki: Added in c110e1793d3e100c87080f9996d32922da44daef

src/handlers/gestures/simpleGestures.ts
j-piasecki: I removed `transformProps` and `allowedProps` in c207736e230eaea9e6e0261b51b19e1c411c7214. At the moment, only `Pan` gesture was making use of them, and this way we can update how the config is managed without sharing code with the old one.

src/handlers/gestures/simpleGestures.ts
j-piasecki: I added more types in 9c857d125e7dd1379d8535c868824e316252fa92.

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+    return this;
+  }
+
+  build(): BuiltGesture {
+    const result = new BuiltGesture(this.prepare);
+    result.gestures = this.pendingGestures.map((pending) => pending.gesture);
+    return result;
+  }
+
+  prepare = () => {
```
jakub-gonet, line 55: This should be a normal method for consistency.

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+    const simultaneous: number[] = [];
+    const waitFor: number[] = [];
```
jakub-gonet, line 57: Isn't that `enum Relation` instead of `number`?

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+      const newConfig = { ...pendingGesture.gesture.config };
+
+      if (newConfig.simultaneousWith) {
+        newConfig.simultaneousWith = [
+          ...newConfig.simultaneousWith,
+          ...simultaneous,
+        ];
+      } else {
+        newConfig.simultaneousWith = [...simultaneous];
+      }
```
jakub-gonet, line 72: We could add some small helper, like `overrideConfig(oldConfig, newConfig)` and use it:
```ts
newConfig.simultaneousWith = overrideConfig(newConfig.simultaneousWith, simultaneous)
```

`overrideConfig` maybe isn't the best name for it 🤷

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+
+type CommonGestureConfig = {
+  ref: React.RefObject<SimpleGesture>;
+  enabled: boolean;
+  minPointers: number;
+  shouldCancelWhenOutside: boolean;
+  hitSlop: HitSlopType;
+  requireToFail: (number | SimpleGesture | React.RefObject<SimpleGesture>)[];
+  simultaneousWith: (number | SimpleGesture | React.RefObject<SimpleGesture>)[];
+};
```
jakub-gonet, line 35: This could be shared with types from [here](https://github.com/software-mansion/react-native-gesture-handler/blob/297070cc80b078bfe7f10575400f8543f4d48162/src/handlers/gestureHandlerCommon.ts#L58-L60)

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+type HandlerCallbacks = {
+  handlerTag: number;
+  onBegan?: (event: HandlerStateChangeEventPayload) => void;
+  onStart?: (event: HandlerStateChangeEventPayload) => void;
+  onEnd?: (event: HandlerStateChangeEventPayload, success: boolean) => void;
+  onUpdate?:
+    | ((event: PanGestureHandlerEvent) => void)
+    | ((event: PinchGestureHandlerEvent) => void)
+    | ((event: RotationGestureHandlerEvent) => void)
+    | ((event: ForceTouchGestureHandlerEvent) => void);
```
jakub-gonet, line 58: Those are plain events or events wrapped with `{nativeEvent: eventData}`?

jakub-gonet: I'd like to unify types from here and gestureHandlerCommon to have them in one place. Also, please name this type since it's not clear if only handlers with continuous gestures should belong here or somewhere else.

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+    gesture: SimpleGesture | React.RefObject<SimpleGesture>
+  ) {
+    this.config[key] = this.config[key]
+      ? [].concat(this.config[key], gesture)
+      : [gesture];
+  }
+
+  private toArray(x: unknown) {
+    return [].concat(x);
+  }
```
jakub-gonet, line 80: 🌵 This function should be near `prepare` since we use it only there.

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+export class ForceTouch extends SimpleGesture {
+  public config: Partial<ForceTouchGestureConfig> = {};
+
+  constructor() {
+    super();
+
+    this.handlerName = 'ForceTouchGestureHandler';
+  }
+
+  setOnUpdate(callback: (event: ForceTouchGestureHandlerEventPayload) => void) {
```
jakub-gonet, line 410: You can use `override` keyword in TS to signify that.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+
+    if (this.config.ref) {
+      this.config.ref.current = this;
+    }
+  }
+
+  build(): BuiltGesture {
+    const result = new BuiltGesture(this.prepare);
+    result.gestures = [this];
+    return result;
```
jakub-gonet, line 164:
```suggestion
    const gestureConfig = new BuiltGesture(this.prepare);
    gestureConfig.gestures = [this];
    return gestureConfig;
```

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+}
+
+type TapGestureConfig = CommonGestureConfig & {
+  numberOfTaps: number;
+  maxDist: number;
+  maxDurationMs: number;
+  maxDelayMs: number;
+  maxDeltaX: number;
+  maxDeltaY: number;
+};
```
jakub-gonet, line 185: I don't like inline props, if we in the future add one we'll have to change it in multiple places. Could we unify those?

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+  onStart?: (event: HandlerStateChangeEventPayload) => void;
+  onEnd?: (event: HandlerStateChangeEventPayload, success: boolean) => void;
+  onUpdate?:
+    | ((event: PanGestureHandlerEvent) => void)
+    | ((event: PinchGestureHandlerEvent) => void)
+    | ((event: RotationGestureHandlerEvent) => void)
+    | ((event: ForceTouchGestureHandlerEvent) => void);
+};
+
+export abstract class SimpleGesture extends Gesture {
```
jakub-gonet, line 61: I don't like this name. Why is it `Simple`?

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+  private toArray(x: unknown) {
+    return [].concat(x);
+  }
+
+  setRef(ref: React.RefObject<SimpleGesture>) {
+    this.config.ref = ref;
+    return this;
+  }
+
+  setOnBegan(callback: (event: HandlerStateChangeEventPayload) => void) {
```
jakub-gonet, line 87: We could parametrize those with generic param. Check `HandlerStateChangeEvent` type or create your own here.

jakub-gonet (commented):
Left some more comments. We should focus on unifying types in one place, refactoring `gestureHandlerCommon` if necessary, renaming some things, and cleaning up code overall.

If you want, you may also split `simpleGestures` into multiple files to be consistent with `handlers/` dir.

src/handlers/gestures/simpleGestures.ts
```diff
@@ -0,0 +1,429 @@
...
+  | Partial<
+      Record<
+        'left' | 'right' | 'top' | 'bottom' | 'vertical' | 'horizontal',
+        number
+      >
+    >
+  | Record<'width' | 'left', number>
+  | Record<'width' | 'right', number>
+  | Record<'height' | 'top', number>
+  | Record<'height' | 'bottom', number>;
```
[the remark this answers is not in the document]
j-piasecki: Done in 4e99225f49479f38b3ff489e4d3ddbdd99d96c07

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+  private pendingGestures: PendingGesture[] = [];
+
+  constructor(base: SimpleGesture) {
+    this.addGesture({ relation: Relation.Exclusive, gesture: base });
+  }
+
+  private addGesture(gesture: PendingGesture): GestureBuilder {
+    this.pendingGestures.push(gesture);
+    return this;
+  }
```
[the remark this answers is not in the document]
j-piasecki: Moved in 29b3aa970e5df04c67d56af687ad3e1a82b9117d

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+  }
+
+  requireToFail(gesture: SimpleGesture): GestureBuilder {
+    this.pendingGestures.push({
+      relation: Relation.RequireToFail,
+      gesture: gesture,
+    });
+
+    return this;
```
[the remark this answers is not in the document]
j-piasecki: Fixed in d91574e8b5c3ccec9acc509f329f053f3cfa2a67

src/handlers/gestures/gestureBuilder.ts
j-piasecki: I used arrow function there because it's passed as an argument and I needed `this` binding. Can I just change other methods to be arrow functions or bind it explicitly (and which one would be better), or is passing it a bad idea?

src/handlers/gestures/gestureBuilder.ts
j-piasecki: No, those arrays store `handlerTags` of gestures with specified relation.

src/handlers/gestures/simpleGestures.ts
j-piasecki: Moved in 30f3fd1db328c92a552e394ebadbcb7ea119aeeb

src/handlers/gestures/simpleGestures.ts
j-piasecki: At the beginning, instead of the `GestureBuilder` there was a `ComplexGesture`, so `SimpleGesture` might have had more sense then. Would `BasicGesture` be better?

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+          break;
+        case Relation.RequireToFail:
+          waitFor.push(pendingGesture.gesture.handlerTag);
+          break;
+      }
+    }
+  };
+}
+
+export class BuiltGesture {
```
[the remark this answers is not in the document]
j-piasecki: Renamed in 179e9e4d8e7b6646b137064bd7afc332629d2129

src/handlers/gestures/gestureBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+      gesture: gesture,
+    });
+
+    return this;
+  }
+
+  build(): BuiltGesture {
+    const result = new BuiltGesture(this.prepare);
+    result.gestures = this.pendingGestures.map((pending) => pending.gesture);
+    return result;
```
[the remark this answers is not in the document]
j-piasecki: Fixed in 179e9e4d8e7b6646b137064bd7afc332629d2129

src/handlers/gestures/simpleGestures.ts
j-piasecki: Fixed in 179e9e4d8e7b6646b137064bd7afc332629d2129

src/handlers/gestures/gestureBuilder.ts
jakub-gonet: I see, so we can consider adding `Tags` suffix to them.

src/handlers/gestures/simpleGestures.ts
jakub-gonet: `BaseGesture`?

src/handlers/gestures/gestureBuilder.ts
j-piasecki: Renamed in 8c9bd4380dfe25f2174e83d174d816bbf49688a5

src/handlers/gestures/gesture.ts
```diff
@@ -0,0 +1,46 @@
...
+  LongPress,
+  ForceTouch,
+} from './simpleGestures';
+
+export abstract class Gesture {
+  abstract build(): BuiltGesture;
+
+  abstract initialize(): void;
+
+  abstract prepare(): void;
```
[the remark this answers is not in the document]
j-piasecki: Added JSDoc in 21ad38e94afaeba3fb34032ce620b2dd0c8bcebc

src/handlers/gestures/gestureBuilder.ts
j-piasecki: Fixed in 233fcb5cb3a59975ca32c2620746cf70a413de5f and b07ee00debe2e5d8c84898e4cd5f6d83d37268a6

src/handlers/gestures/gestureBuilder.ts
j-piasecki: Added in f181ba99141486b3c88755a4475c669ad0b6d3e5, is `extendRelation` name ok?

src/handlers/gestures/simpleGestures.ts
j-piasecki: Done (I think) in 57f965641192e5efd43bd69a41b622c939d2745c

src/handlers/gestures/simpleGestures.ts
j-piasecki: Changed in e6ab178da770372205b43669089ce73f8836288e

src/handlers/gestures/simpleGestures.ts
j-piasecki: Done in 712e64fc3e4ab072ac817af3700ba4918d277b12

src/handlers/gestures/simpleGestures.ts
j-piasecki: Those are plain events. I refactored it to use generics in 712e64fc3e4ab072ac817af3700ba4918d277b12 and it uses the same payload types as old handlers. Is that ok?

src/handlers/gestures/simpleGestures.ts
j-piasecki: Done in 57f965641192e5efd43bd69a41b622c939d2745c

src/handlers/gestures/gesture.ts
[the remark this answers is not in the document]
j-piasecki: Fixed circular dependency in d02c0b97b87d6165a845ae77edad61ecb955e85f

src/handlers/gestures/simpleGestures.ts
j-piasecki: This method is defined only in continuous gestures, not in the parent class.

src/handlers/PanGestureHandler.ts
```diff
@@ -41,8 +41,26 @@ export type PanGestureHandlerEventPayload = {
   velocityY: number;
 };
 
+interface CommonPanProperties {
+  minDist?: number;
+  avgTouches?: boolean;
+  enableTrackpadTwoFingerGesture?: boolean;
+}
```
jakub-gonet, line 48: Why do we need to extract it? It looks like its used only in PanGestureHandlerProps

src/handlers/gestures/gesture.ts
```diff
@@ -0,0 +1,289 @@
...
+      this.config.requireToFail = this.toArray(this.config.requireToFail);
+    }
+
+    if (this.config.simultaneousWith !== undefined) {
+      this.config.simultaneousWith = this.toArray(this.config.simultaneousWith);
+    }
+  }
+
+  private toArray(x: unknown) {
+    return [].concat(x as never);
```
jakub-gonet, line 180: Why we cast to never here?

src/handlers/gestures/gesture.ts
```diff
@@ -0,0 +1,289 @@
...
+    return this;
+  }
+
+  setHitSlop(hitSlop: HitSlop) {
+    this.config.hitSlop = hitSlop;
+    return this;
+  }
+
+  addSimultaneousGesture(
+    gesture: BaseGesture<any> | React.RefObject<BaseGesture<any>>
```
jakub-gonet, line 129: `any`

src/handlers/gestures/panGesture.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  setActiveOffsetY(offset: number | number[]) {
+    if (Array.isArray(offset)) {
+      this.config.activeOffsetYStart = offset[0];
+      this.config.activeOffsetYEnd = offset[1];
+    } else if (offset < 0) {
+      this.config.activeOffsetYStart = offset;
+    } else {
+      this.config.activeOffsetYEnd = offset;
+    }
```
jakub-gonet, line 34: I think we can extract those methods to something shared between Gestures and GestureHandlers but it's not that important right now.

jakub-gonet (approved):
I think we have unnecessary whitespaces in some places (like in setters inside builders). Other than that, looks great.

src/handlers/gestures/gesture.ts
j-piasecki: Fixed in 6da30cb388f462568017bb46206988886afd142d

src/handlers/gestures/gesture.ts
j-piasecki: Fixed in 6da30cb388f462568017bb46206988886afd142d

src/handlers/PanGestureHandler.ts
j-piasecki: It's also used in `PanGestureConfig`. In other handlers props are extending the config, but `PanGestureHandler` is the only one that makes use of `transformProps`. I copied the logic of that in c207736e230eaea9e6e0261b51b19e1c411c7214, and `PanGestureConfig` holds already transformed properties, while those in `PanGestureHandlerProps` are yet to be transformed, so I extracted the ones they shared to separate object.

PR #1585, opened 2021-08-20
Remove AndroidNativeExample

jakub-gonet:


j-piasecki:
If I didn't miss anything, at the moment there are two `example` directories. One with the `App.tsx` file and the other with actual examples.

j-piasecki (approved):

PR #1586, opened 2021-08-20
Remove unused implementations on Android

jakub-gonet:


j-piasecki:
Since only the last two commits are different from https://github.com/software-mansion/react-native-gesture-handler/pull/1585 which is already merged, wouldn't it be easier to extract them into separate PR instead of resolving conflicts?

PR #1589, opened 2021-08-23
Fix TS errors and handlers not attaching

j-piasecki:
## Description

Fixed TypeScript type errors
Removed unnecessary `setImmediate` that caused callback to be set too late

## Test plan

Tested on the example app.

### DIFF
--- a/src/handlers/gestures/gesture.ts
+++ b/src/handlers/gestures/gesture.ts
@@ -24,7 +24,9 @@ export type GestureType =
   | BaseGesture<ForceTouchGestureHandlerEventPayload>;
 
 export type GestureRef = number | GestureType | React.RefObject<GestureType>;
-export interface BaseGestureConfig extends CommonGestureConfig {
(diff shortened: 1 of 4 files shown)


src/handlers/gestures/gesture.ts
```diff
@@ -24,7 +24,9 @@ export type GestureType =
   | BaseGesture<ForceTouchGestureHandlerEventPayload>;
 
 export type GestureRef = number | GestureType | React.RefObject<GestureType>;
-export interface BaseGestureConfig extends CommonGestureConfig {
+export interface BaseGestureConfig
+  extends CommonGestureConfig,
+    Record<string, unknown> {
```
jakub-gonet, line 29: Why do we extend Record here?

j-piasecki: Because `filterConfig` expects config to be a `Record`.

jakub-gonet (approved):

PR #1592, opened 2021-08-23
Add conditional import of Reanimated

j-piasecki:
## Description

Adds Reanimated as an optional dependency.

## Test plan

Tested on an app with and without `react-native-reanimated` dependency.


jakub-gonet:
Lemme fix CI first and check if it builds.

jakub-gonet:
Could you merge master into new-api and this branch?

jakub-gonet (approved):

Issue #784, opened 2019-10-05
Nested touchable inside Swipeable


brunolemos:
### Expected Behavior

![Kapture 2019-10-05 at 16 34 43](https://user-images.githubusercontent.com/619186/66259912-209c5780-e78e-11e9-88f3-9a335a8c7d26.gif)

### Current Behavior

![Kapture 2019-10-05 at 16 57 32](https://user-images.githubusercontent.com/619186/66260162-40814a80-e791-11e9-9047-35393c58b0a4.gif)

### Description

I'm using the `Swipeable` component. 

To avoid triggering the `onPress` by mistake, I'm also using the `TouchableOpacity` from `react-native-gesture-handler`.

But I want to have another touchable inside of this card, nested.

When clicking at `#131` or `mentioned`, it should trigger only this touchable, like the first gif above. But it is also triggering the whole card touchable, like the second gif. 

How to achieve this?
Ideally the outer touchable should not even change opacity.
I tried using `e.preventDefault` but the outer touchable is firing before the inner nested touchable.

brunolemos:
Upgrading to `react-native@0.61+` and to the latest version of `react-native-gesture-handler` fixed the issue.

brunolemos:
Only iOS is working as expected, Android still has the wrong behavior. Both touchables get triggered.
@kmagiera Any workaround for this?

I noticed that the `onPress` handler doesn't receive an `event` argument. If they did, I could call `e.preventDefault()` once and check `e.isDefaultPrevented()` second as a workaround. 

https://github.com/kmagiera/react-native-gesture-handler/blob/533f8b042660d5d19eb54e4d99ea3b39a8a84339/touchables/GenericTouchable.js#L211

leteach:
Any fix for this? I am experiencing it as well. Upgraded our `react-native-gesture-handler` version to 1.4.1 and still seeing the same problem. Everything works fine in iOS but Android is trigger both onPress events for the two separate touchable events.

leteach:
I had to create a separate component for Android that imported the parent ToucableWithoutFeedback from react-native and the child TouchableOpacity from react-native-gesture-handler. That had to be conditionally rendered based one Platform.OS. It is hacky as hell but it works for now.

brunolemos:
[...]

PR #1601, opened 2021-08-26
Fix nested touchables on Android

j-piasecki:
## Description

This PR build upon https://github.com/software-mansion/react-native-gesture-handler/pull/486 in order to fix https://github.com/software-mansion/react-native-gesture-handler/issues/784.
Unfortunately I don't think a solution contained purely in `ButtonViewGroup` is possible because on Android touchables show feedback on the `BEGAN` state, and transition to that state happens in `NativeViewGestureHandler` before any logic in `ButtonViewGroup` is ran.
This change differs from the original one by two main things:
1. It fixes https://github.com/software-mansion/react-native-gesture-handler/pull/486#issuecomment-563310225 by canceling the handler if the `canStart` method returns `false`.
2. It adds logic making non-exclusive `ButtonViewGroup`s behave almost the same as they do on iOS, the difference being that touching a child prevents its ancestor `ButtonViewGroup`s from beginning. (But still works the other way around; touching ancestor doesn't prevent its descendants from beginning the gesture).

## Test plan

Added a new example and tested on the Example app.

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -73,15 +73,15 @@ const EXAMPLES: ExamplesSection[] = [
         name: 'Modals with nested GHRootViews - issue #139',
         component: NestedGestureHandlerRootViewWithModal,
       },
+      {
+        name: 'Nested Touchables - issue #784',
+        component: NestedTouchables as React.ComponentType,
+      },
       { name: 'Double pinch & rotate', component: DoublePinchRotate },
       { name: 'Double draggable', component: DoubleDraggable },
       { name: 'Rows', component: Rows },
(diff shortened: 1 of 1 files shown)


gp3gp3gp3:
Is there a release including this change?

j-piasecki:
Not yet, it will be included in the upcoming 2.0 release but we have no specific timeframe of when it will be.

cesarsousa3:
it hasn't been resolved I just updated in my app, before it worked fine and now I don't have the same result I have a RectButton inside it I have two more RectButtons and whenever I click on those two it takes on the ripple effect of the first RectButton, I'm not to be able to resolve as I had before that was right.

PR #1603, opened 2021-08-27
Fix modals

### LINKED ISSUE #139: Doesn't work in a modal

jakub-gonet:
## Description

Fixes #139

1. Removed RNGHModalUtils
	  
	  This class gave access to `ReactModalHostView.DialogRootViewGroup` in order to
	  call `#onChildStartedNativeGesture()` and use instanceof on it. Checks for
	  `ReactRootView` / `DialogRootViewGroup` can be unified by utilizing `RootView` (which
	  both of those classes implement).

2. Always render GestureHandlerRootView

	  This prevents gestures from not being delivered to RNGH when another
	  GestureHandlerRootView is used before modal (e.g. from React Navigation).
	  
	  Note: This isn't an ideal fix since a new orchestrator will be created, preventing modals from interacting with handlers outside of them. However, this use case is rare. Having gestures working in modals is already an improvement.

## Test plan

Tested on `nestedGHRootViewWithModal.tsx` example.


danilo-vieira:
Can anyone merge this? 👀

jcgertig:
@jakub-gonet any movement on this is a big blocker

hugoh59:
I still have the issue on 2.5.0 with PanGestureHandler inside a Modal.

PR #1605, opened 2021-08-27
Deprecate RNGestureHandlerEnabledRootView

jgonet:


docs/docs/getting-started.md
```diff
@@ -64,13 +87,22 @@ npm install --save react-native-gesture-handler
...
+[wix/react-native-navigation](https://github.com/wix/react-native-navigation)),
+you should follow [this separate guide](#with-wixreact-native-navigation) to get
+gesture handler library set up on Android. Ignore the rest of this step – it
+only applies to RN apps that use a standard Android project layout.
 
-##### Updating `MainActivity.java`
+#### Updating `MainActivity.java`
+
+:::caution
+From version 2.0.0 this step isn't needed. You should use
```
chakrihacker, line 103: is this reanimated v2?

jakub-gonet: Nope, the not-yet-released RNGH 2.

jakub-gonet: But thanks for reminding us that we should version our docs.

chakrihacker: cool, is their rc candidate to test v2?

jakub-gonet: Not yet, we don't know if we want to do rc release. You can pack master if you want to, though.

PR #1608, opened 2021-08-30
Add direction to Swipeable callbacks

jakub-gonet:
## Description

Added direction to `onSwipeable(Will)Open` & `onSwipeable(Will)Close`.

Should we deprecate `onSwipeable(Left/Right)Open` and similar callbacks?

## Test plan

Added console.logs to `AppleStyleSwipeableRow`


example/src/showcase/swipeable/AppleStyleSwipeableRow.tsx
```diff
@@ -96,7 +96,13 @@ export default class AppleStyleSwipeableRow extends Component {
         leftThreshold={30}
         rightThreshold={40}
         renderLeftActions={this.renderLeftActions}
-        renderRightActions={this.renderRightActions}>
+        renderRightActions={this.renderRightActions}
+        onSwipeableOpen={(direction) => {
+          console.log(`Opening swipeable to the ${direction}`);
```
j-piasecki, line 101: A very minor nitpick, could you change this message in some way, for example:
```suggestion
          console.log(`Opening swipeable from the ${direction}`);
```
I was confused for a moment when I swiped from left to right and the console log said that it was opening to the left.

src/components/Swipeable.tsx
```diff
@@ -102,12 +102,12 @@ export interface SwipeableProps
   /**
    * Called when action panel gets open (either right or left).
    */
-  onSwipeableOpen?: () => void;
+  onSwipeableOpen?: (direction: 'left' | 'right') => void;
```
j-piasecki, line 105: We already have a `Directions` enum, although it also provides `UP` and `DOWN` directions. Do you think using it here is fine or the fact that it has all four directions would be a problem?

j-piasecki (approved):
IMO we can deprecate those methods, they provide exactly the same functionality now.

src/components/Swipeable.tsx
jakub-gonet: Swipeable doesn't open to vertical directions so typing it with Directions is wrong here. I'd like to stay with the current typing.

Side note: we can move Directions to FlingGH definition file - from what I saw it's used only there.

example/src/showcase/swipeable/AppleStyleSwipeableRow.tsx
jakub-gonet: Sure, will do.

jakub-gonet: Fixed in 5438419.

PR #1625, opened 2021-09-07
Make Gesture.Simultaneous accept variable number of arguments

j-piasecki:
## Description

Current implementation of nesting gestures in pairs may suggest that the ordering of the gestures in `Simultaneous` matters, and it is not really necessary to limit it in this way. This change allows `Gesture.Simultaneous` to accept multiple gestures at once.

## Test plan

Updated and tested on the Example app.

### DIFF
--- a/src/handlers/gestures/gestureComposition.ts
+++ b/src/handlers/gestures/gestureComposition.ts
@@ -74,8 +74,7 @@ export class ComposedGesture extends Gesture {
 export class SimultaneousGesture extends ComposedGesture {
   prepare() {
     const simultaneousArray = this.gestures
-      .map((gesture) => gesture.toGestureArray())
-      .flat()
+      .flatMap((gesture) => gesture.toGestureArray())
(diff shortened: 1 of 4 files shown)


src/handlers/gestures/gestureComposition.ts
```diff
@@ -72,24 +72,19 @@ export class ComposedGesture extends Gesture {
...
-  constructor(first: Gesture, second: Gesture) {
-    super(first, second);
-  }
-
   prepare() {
-    const leftSide = this.gestures[0].toGestureArray();
-    const rightSide = this.gestures[1].toGestureArray();
+    const simultaneousArray = this.gestures
+      .map((gesture) => gesture.toGestureArray())
+      .flat()
```
jakub-gonet, line 78: `.flatmap()`

j-piasecki: Done in c6337844ec285e9632eead05b257fc869adfdff4

PR #1628, opened 2021-09-08
Add native gesture

j-piasecki:
## Description

Adds a `NativeGesture` object alongside `Gesture.Native()` method to incorporate `NativeViewGestureHandler` to the new API.

## Test plan

Tested on the example app


jakub-gonet (approved):
Needs docs.

PR #1637, opened 2021-09-10
Bottom Sheet example with new API

enterteg:
Adds bottom sheet example with new API. Need to be tested on Android.


examples/Example/src/showcase/bottomSheet/index.tsx
```diff
@@ -61,6 +61,7 @@ export class BottomSheet extends Component<
       { useNativeDriver: USE_NATIVE_DRIVER }
     );
     this.lastScrollY.addListener(({ value }) => {
+      console.log(value);
```
jakub-gonet, line 64: Should we have `console.log` here?

examples/Example/src/new_api/chat_heads/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
+import Animated, {
+  useAnimatedStyle,
+  useDerivedValue,
+  useSharedValue,
+  withSpring,
+} from 'react-native-reanimated';
+import { SafeAreaView } from 'react-native-safe-area-context';
+
+const HEADER_HEIGHT = 44; // let's assume it's always the same for ease of understanding ;)
```
jakub-gonet, line 12: Can't we get it from [navigation](https://reactnavigation.org/docs/elements/#useheaderheight)?

jakub-gonet (approved):
`examples/Example/src/new_api/drag_n_drop/index.tsx` should be removed.

examples/Example/src/new_api/chat_heads/index.tsx
enterteg: It wasn't not part of this PR. Removed this file.

examples/Example/src/new_api/bottom_sheet/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+    translationY.value = e.translationY;
+  };
+
+  const onHeaderHandlerEnd = (e: PanGestureHandlerEventPayload) => {
+    onHandlerEnd(e);
+  };
+
+  const onPanHandlerEnd = (e: PanGestureHandlerEventPayload) => {
+    onHandlerEnd(e);
+  };
```
jakub-gonet, line 56: Can't we inline those?

enterteg: Yeah previously logic was different, will update.

examples/Example/src/new_api/bottom_sheet/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+        <LoremIpsum words={200} />
+        <Animated.View style={[styles.bottomSheet, bottomSheetAnimatedStyle]}>
+          <GestureDetector gesture={headerGesture}>
+            <View style={styles.header} />
+          </GestureDetector>
+          <GestureDetector
+            gesture={Gesture.Simultaneous(
+              panGesture,
+              blockScrollUntilAtTheTop
+            )}>
```
j-piasecki, line 129: Here the `blockScrollUntilAtTheTop` gesture is attached to both gestures detectors. To just mark a relation without attaching it you can use `simultaneousWithExternalGesture`.

j-piasecki (commented):
If I recall correctly this example requires both https://github.com/software-mansion/react-native-gesture-handler/pull/1611 and https://github.com/software-mansion/react-native-gesture-handler/pull/1609 for calculating gesture translations relative to the window and not the view. It should be merged when they are merged (or at least the part that changes the way translations are calculated).

j-piasecki (approved):

victorzimnikov:
blockScrollUntilAtTheTop not working for android

PR #1650, opened 2021-09-22
Remove unnecessary null check from RNGestureHandlerModule on Android

j-piasecki:
## Description

It cannot be null, I checked. The `registry` is a `val` initialized at declaration.

## Test plan

Tested on the Example app


jakub-gonet (approved):
This check was strange since in the original code it was initialized in the same manner but also checked for null anyway:
- https://github.com/software-mansion/react-native-gesture-handler/blob/9c7664b7fa426f366c68155181082bcf52943125/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java#L452
- https://github.com/software-mansion/react-native-gesture-handler/blob/9c7664b7fa426f366c68155181082bcf52943125/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.java#L520

PR #1662, opened 2021-09-28
Update pan gesture to use the same set of properties as PanGestureHandler

j-piasecki:
## Description

Move all shared properties of PanGesture and PanGestureHandler to the CommonPanProperties interface. This allows PanGesture to utlize all (non-deprecated) properties available to the PanGestureHandler.
Added PanGesture config methods corresponding to the moved props.

## Test plan

Tested on the Example app


j-piasecki:
They were using the same native implementation as `activeOffsetX/Y` and `failOffsetX/Y`.

PR #1670, opened 2021-10-07
Prepare state management for pointer events

j-piasecki:
## Description

- Add common interface and protocol that will be shared between Gesture Handler and Reanimated.
- Add `setGestureState` method to allow managing state of the gesture handlers. This method will be used by `react-native-reanimated` to allow for synchronous state management inside worklets.
- Extracted pre- and post-activation logic to separate functions on Android. This allows for it to be executed just by calling `activate`.
- Reset `Pinch` and `Rotation` state on activation to prevent unexpected behavior with manual activation.

## Test plan

It compiles, needs to be tested when all the changes are merged together.

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
@@ -388,7 +388,9 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
 
   fun activate() {
     if (state == STATE_UNDETERMINED || state == STATE_BEGAN) {
+      beforeActivation()
       moveToState(STATE_ACTIVE)
+      afterActivation()
     }
   }
 
@@ -411,6 +413,8 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
   protected open fun onStateChange(newState: Int, previousState: Int) {}
   protected open fun onReset() {}
   protected open fun onCancel() {}
+  protected open fun beforeActivation() {}
+  protected open fun afterActivation() {}
   fun reset() {
     view = null
     orchestrator = null
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
@@ -386,11 +386,9 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
     }
   }
 
-  fun activate() {
+  open fun activate() {
     if (state == STATE_UNDETERMINED || state == STATE_BEGAN) {
-      beforeActivation()
       moveToState(STATE_ACTIVE)
-      afterActivation()
     }
   }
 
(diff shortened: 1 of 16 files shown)


android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.kt
```diff
@@ -244,14 +244,17 @@ class PanGestureHandler(context: Context?) : GestureHandler<PanGestureHandler>()
...
       } else if (shouldActivate()) {
-        // reset starting point
-        startX = lastX
-        startY = lastY
         activate()
       }
     }
   }
 
+  override fun beforeActivation() {
```
kmagiera, line 252: this seems like something we should be doing in "reset" logic, no? Why has this been moved here?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -1,18 +1,21 @@
 package com.swmansion.gesturehandler.react
 
 import android.content.Context
+import android.util.Log
```
kmagiera, line 4: remove

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
```diff
@@ -411,6 +413,8 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
   protected open fun onStateChange(newState: Int, previousState: Int) {}
   protected open fun onReset() {}
   protected open fun onCancel() {}
+  protected open fun beforeActivation() {}
```
kmagiera, line 416: This seem weirdly specific to activation only. We already have a bunch of patterns around placing state-change logic. E.g., we have methods like onCancel or onStateChange. I wonder if we could use one of these? Alternatively we can consider changing functions like activate to not be final and this way allow them to be overriden which would seem like a cleaner approach, no?

kmagiera (approved):
This looks mostly good. I wonder if we could make android related changes more cleaner w/o adding additional before/after callbacks, e.g. use the ones that are already there or allow activate() to be overriden?

android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.kt
j-piasecki: When the pan activates it has to "reset" its starting position so the translation is calculated relative to the activation point and not the point where the finger was first placed on the screen. Before it was called just before the `activate` in `onHandle` method.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Removed in https://github.com/software-mansion/react-native-gesture-handler/pull/1670/commits/6dcc8e79a56e8fb46c1ea1c3cef46d42c0648944

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
j-piasecki: I don't think it would be possible using one of the existing methods (besides adding the relevant code alongside every `activate` call), but making `activate` to not be final seems like a much better approach. I updated it in https://github.com/software-mansion/react-native-gesture-handler/pull/1670/commits/7db36d5dd2ba5fba4c1fca9f5cdbcc0da58e5705.

PR #1671, opened 2021-10-07
Add option to enable manual activation

j-piasecki:
## Description

Add `manualActivation` method to continuous gestures. When set to `true` the gesture will not activate by itself only by calling relevant method.

## Test plan

Needs to be tested when all parts are merged together.

### DIFF
--- a/src/handlers/gestureHandlerCommon.ts
+++ b/src/handlers/gestureHandlerCommon.ts
@@ -31,7 +31,10 @@ export const baseGestureHandlerProps = [
   'onHandlerStateChange',
 ] as const;
 
-export const baseGestureHandlerWithMonitorProps = commonProps;
+export const baseGestureHandlerWithMonitorProps = [
+  ...commonProps,
+  'manualActivation',
+];
 
 export interface GestureEventPayload {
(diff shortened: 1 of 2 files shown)


src/handlers/gestureHandlerCommon.ts
```diff
@@ -31,7 +31,10 @@ export const baseGestureHandlerProps = [
   'onHandlerStateChange',
 ] as const;
 
-export const baseGestureHandlerWithMonitorProps = commonProps;
+export const baseGestureHandlerWithMonitorProps = [
+  ...commonProps,
```
kmagiera, line 35: Not clear to me why we need separate const for the and can't just add this to commonProps but I'm assuming there is good reason

kmagiera (approved):
👍

src/handlers/gestureHandlerCommon.ts
j-piasecki: We could add it there but it's shared between gestures and gesture handlers, so the `manualActivation` prop would not be stripped from gesture handlers config if someone would add it there.

PR #1672, opened 2021-10-07
Add foundation for pointer events

j-piasecki:
## Description

Add JS implementation of pointer events and state management.

## Test plan

Needs to be tested when all the parts are merged together.


kmagiera (approved):
Looks good. The only comment I have is about the name "pointerEvent" which is used in a bunch of places. I think it may get confused with widely popular concept of CSS pointer events https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events that are also ported to RN https://reactnative.dev/docs/view#pointerevents – this concept is not that much related and does different things so I'd prefer to rename this. names like `onPointerDown` etc can stay as they don't have "event" in them, however "onPointerChange" suggest that we only get updates for one of the pointer and not all. If you think we can work on the naming still we can discuss this offline or you're free to come up with some names that'd fit the role better

PR #1673, opened 2021-10-07
Add pointer events on Android

j-piasecki:
## Description

Implementation of pointer events on Android.

## Test plan

Needs to be tested when all the parts are merged together.

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
@@ -331,14 +331,15 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
 
       for (i in 0 until event.pointerCount) {
         val pointerId = event.getPointerId(i)
+        val pointer = trackedPointers[pointerId] ?: continue
 
-        if (trackedPointers[pointerId]?.x != event.getX(i) || trackedPointers[pointerId]?.y != event.getY(i)) {
-          trackedPointers[pointerId]?.x = event.getX(i)
-          trackedPointers[pointerId]?.y = event.getY(i)
-          trackedPointers[pointerId]?.absoluteX = event.getX(i) + offsetX
-          trackedPointers[pointerId]?.absoluteY = event.getY(i) + offsetY
+        if (pointer.x != event.getX(i) || pointer.y != event.getY(i)) {
+          pointer.x = event.getX(i)
+          pointer.y = event.getY(i)
+          pointer.absoluteX = event.getX(i) + offsetX
+          pointer.absoluteY = event.getY(i) + offsetY
 
-          addPointerData(trackedPointers[pointerId]!!)
+          addPointerData(pointer)
         }
       }
     }
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
@@ -1,10 +1,15 @@
 package com.swmansion.gesturehandler
 
+import android.provider.Settings.Global.putInt
 import android.view.MotionEvent
 import android.view.MotionEvent.PointerCoords
 import android.view.MotionEvent.PointerProperties
 import android.view.View
+import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.UiThreadUtil
+import com.facebook.react.bridge.WritableArray
+import com.facebook.react.uimanager.PixelUtil
+import com.swmansion.gesturehandler.react.RNGestureHandlerPointerEvent
 import java.lang.IllegalStateException
 import java.util.*
 
@@ -25,6 +30,20 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
   var isEnabled = true
     private set
   var usesDeviceEvents = false
+
+  var pointerEventPayload: WritableArray? = null
+    private set
+  var pointerEventType = RNGestureHandlerPointerEvent.EVENT_UNDETERMINED
+    private set
+  var trackedPointersCount = 0
+    private set
+  private var trackedPointers: Array<PointerData?>? = null
+  var needsPointerData = false
+    set(value) {
+      field = value
+      trackedPointers = if (value) { Array(MAX_POINTERS_COUNT) { null } } else { null }
+    }
+
 
   private var hitSlop: FloatArray? = null
   var eventCoalescingKey: Short = 0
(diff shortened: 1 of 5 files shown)


android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
```diff
@@ -274,11 +298,112 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
...
+        val pointerId = event.getPointerId(event.actionIndex)
+
+        trackedPointers[pointerId] = PointerData(
+            pointerId,
+            event.getX(event.actionIndex),
+            event.getY(event.actionIndex),
+            event.getX(event.actionIndex) + offsetX,
+            event.getY(event.actionIndex) + offsetY,
+        )
+        trackedPointersCount++
```
kmagiera, line 319: Do we really need separate counter for this? Maybe we can use trackedPointersIDCount, which I believe should always be in sync. Unless we expect pointer tracking to be enabled while gesture is already running, which does not make sense in my opinion. I think we can also just use a fixed-size array of `PointerData` for `trackedPointers`  and only populate entries which have active pointers + when generating event we'd ignore `null` entries

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
```diff
@@ -245,17 +242,25 @@ class GestureHandlerOrchestrator(
...
+      handler.updatePointerData(event)
     }
-    event.setLocation(oldX, oldY)
-    // if event was of type UP or POINTER_UP we request handler to stop tracking now that
-    // the event has been dispatched
-    if (action == MotionEvent.ACTION_UP || action == MotionEvent.ACTION_POINTER_UP) {
-      val pointerId = event.getPointerId(event.actionIndex)
-      handler.stopTrackingPointer(pointerId)
+
+    if (!handler.isAwaiting || action != MotionEvent.ACTION_MOVE) {
```
kmagiera, line 249: This seem to be changing the event processing logic. What is the reason behind this change? We will no longer call `handle` in certain conditions and I wonder if it is not going to break something?

kmagiera (approved):
Looks mostly good, left a few comments inline, please address these prior to merge

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
j-piasecki: I wanted to send pointer events event if the handler was awaiting the activation of another handler so I couldn't do it in the `handle` method because it's not called in that case. So I slightly modified `deliverEventToGestureHandler` to handle pointer events there. I don't really see the case where `handle` wouldn't be called now but was called before. If you were referring to [this line](https://github.com/software-mansion/react-native-gesture-handler/pull/1673/files/2334e124b1f2d5e02fbc8ca473aa338f1827c0b0#diff-0183690c13c9a7694f9bad1e3e57c96de12db967f2aa3d8a5c02b41b76ef9c88R249) then it's doing the same check as the if I have removed above, just later.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
```diff
@@ -274,11 +298,112 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
...
+      pointerEventPayload = null
+
+      val offsetX = event.rawX - event.x
+      val offsetY = event.rawY - event.y
+
+      if (event.actionMasked == MotionEvent.ACTION_DOWN || event.actionMasked == MotionEvent.ACTION_POINTER_DOWN) {
+        pointerEventType = RNGestureHandlerPointerEvent.EVENT_POINTER_DOWN
+        val pointerId = event.getPointerId(event.actionIndex)
+
+        trackedPointers[pointerId] = PointerData(
```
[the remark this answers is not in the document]
j-piasecki: You're right, but in order to be consistent with how it's implemented on iOS, where only the new pointers are in the `DOWN` event, would sending two events in this case be a good solution? I.e. send `DOWN` up with just the data about the new pointer and then send the `MOVE` event with data about all pointers.

This would also apply to the `UP` event, right?

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
j-piasecki: There is one particular moment when they are not in sync (by design): currently `numberOfPointers` in pointer events represents number of pointers on the screen, so if you have one pointer on the screen and you lift it, you would get `UP` event with `numberOfPointers` set to zero. Should that be changed?

j-piasecki: As for the fixed array, done in https://github.com/software-mansion/react-native-gesture-handler/pull/1673/commits/e6ea1e68af716c619a34f032406b74219b9baad5 and https://github.com/software-mansion/react-native-gesture-handler/pull/1673/commits/fef255c2e23b9e795ad52bae965a1f30a65e387d.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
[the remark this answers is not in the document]
kmagiera: I don't think it matters that much if we send one or two events at the moment when new finger is placed on the screen. What's more important is for this to work consistently across platforms. So if that's how you want it I'm fine with it so long it works the same on Android and iOS

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
kmagiera: As for the UP event with numberOfPointers = 0 – it I think this is fine so long it works consistently. It is quite and edge case, as when there are no pointers and you send numberOfPointers = 1 one would expect the pointers data to also have an entry, but then what coordinates you put there? the pointer is not on screen so it has no position?

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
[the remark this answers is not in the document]
j-piasecki: Ok, I made it try sending two events for now: one down/up event and a move event with data about other pointers in https://github.com/software-mansion/react-native-gesture-handler/pull/1673/commits/06ae23ae8e9c6b42b728d43f551e9556daa41bc1.

PR #1674, opened 2021-10-07
Add pointer events on iOS

j-piasecki:
## Description

Implementation of pointer events on iOS.

## Test plan

Needs to be tested when all parts are merged together.


kmagiera (approved):
Looks good 👍

Issue #1679, opened 2021-10-14
Android Crash: Fatal Exception: java.lang.IllegalArgumentException: pointerIndex out of range


escowart:
## Description

```
Fatal Exception: java.lang.IllegalArgumentException: pointerIndex out of range
       at android.view.MotionEvent.nativeGetAxisValue(MotionEvent.java)
       at android.view.MotionEvent.getY(MotionEvent.java:2416)
       at android.widget.ScrollView.onTouchEvent(ScrollView.java:866)
       at com.facebook.react.views.scroll.ReactScrollView.onTouchEvent(ReactScrollView.java:323)
       at android.view.View.dispatchTouchEvent(View.java:14309)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3112)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2785)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3118)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2799)
[...]

PR #1681, opened 2021-10-14
Prevent a Android crash from dispatchTouchEvent

escowart:
## Description

try-catch block around call which intermittently throws an IllegalArgumentException

Resolves Issues:
  -  https://github.com/software-mansion/react-native-gesture-handler/issues/1679
  -  https://github.com/facebook/react-native/issues/30320

## Test plan

  - Installed the dependency
  - Cable load the app which uses the dependency onto a device
  - Try a variety of touch gestures


escowart:
I can't get the dependency to build because of the changes to Android :(

RodolfoGS:
@escowart I'm having this issue. Why you closed this PR? Could you fixed this with another approach?

PR #1689, opened 2021-10-18
Rename onBegan to onBegin

j-piasecki:
## Description

Rename `onBegan` to `onBegin` in order to be more consistent with naming.

## Test plan

Ran `npm pack` and tested on the Example app.


kmagiera (approved):
Let's do this :shipit:

PR #1690, opened 2021-10-19
Remove BetterDrawerLayout component

j-piasecki:
## Description

Remove BetterDrawerLayoutComponent in order not to force the users into using Reanimated and move to the Example app.

## Test plan

Tested on the Example app


jakub-gonet (commented):
It can be included in the examples. Could you resolve the conflicts?

PR #1696, opened 2021-10-22
Fix min distance on Pan with custom activation criteria

j-piasecki:
## Description

In https://github.com/software-mansion/react-native-gesture-handler/pull/1594 I have changed the default value of `minDist` with custom activation criteria from `Float.MAX_VALUE` to `Float.MIN_VALUE`. This fixed problems with `failOffset` set without `activeOffset` but broke `activeOffset` itself - in the `shouldActivate` method in PanGestureHandler the `minDist` prop is checked right after `activeOffset` and since the `minDist` was effectively 0 the handler would activate.

The simple fix would be to revert that PR, but I believe that setting only `failOffset` has some use cases (you could use it in tandem with `waitFor`) I changed the way props are parsed so that if only `failOffset` is set, the `minDist` will default to `Float.MIN_VALUE`, but if `activeOffset` or `minVelocity` is set the `minDist` will default to `Float.MAX_VALUE`.

In addition, I added a check in `setMinDist` in PanGestureHandler that sets the `minDistSq` to the default value if the method is called because of `failOffset`.

## Test plan

Tested on the Example app.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
@@ -146,65 +146,51 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) : ReactCont
     override fun configure(handler: PanGestureHandler, config: ReadableMap) {
       super.configure(handler, config)
       var hasCustomActivationCriteria = false
-      var minDistFromActivationCriteria = 0f
-      if (config.hasKey(KEY_PAN_FAIL_OFFSET_RANGE_X_START)) {
-        handler.setFailOffsetXStart(PixelUtil.toPixelFromDIP(config.getDouble(KEY_PAN_FAIL_OFFSET_RANGE_X_START)))
-        hasCustomActivationCriteria = true
-        minDistFromActivationCriteria = Float.MIN_VALUE
-      }
-      if (config.hasKey(KEY_PAN_FAIL_OFFSET_RANGE_X_END)) {
-        handler.setFailOffsetXEnd(PixelUtil.toPixelFromDIP(config.getDouble(KEY_PAN_FAIL_OFFSET_RANGE_X_END)))
-        hasCustomActivationCriteria = true
-        minDistFromActivationCriteria = Float.MIN_VALUE
-      }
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -146,59 +146,73 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) : ReactCont
...
+        minDistFromActivationCriteria = Float.MAX_VALUE
       }
 
       // PanGestureHandler sets minDist by default, if there are custom criteria specified we want
       // to reset that setting and use provided criteria instead.
       if (config.hasKey(KEY_PAN_MIN_DIST)) {
         handler.setMinDist(PixelUtil.toPixelFromDIP(config.getDouble(KEY_PAN_MIN_DIST)))
       } else if (hasCustomActivationCriteria) {
-        handler.setMinDist(Float.MIN_VALUE)
+        handler.setMinDist(minDistFromActivationCriteria)
```
kmagiera, line 215: As discussed I think it'd be ok to just throw here and also add a more descriptive message on the JS side when someone attempts to use minDist and activeOffset/failOffset together. As it'd be difficult to come up with a intuitive way of doing that it'd be safer to just disallow it completely as I think all the possible usecases here can be build using some other set of attributes.

j-piasecki: I added errors in the JS in https://github.com/software-mansion/react-native-gesture-handler/pull/1696/commits/41ccfc0797a13db1c7c4d0f9bf79e471e0745cb2, but I think we cannot throw in Android. Setting `minDist` to `Float.MAX_VALUE` here makes the handler not activate until `activeOffsetX/Y` condition is met. I'd say that showing error is ok for now and we can revisit the issue in the future if there are still problems with it. What do you think?

kmagiera: console.error is actually an equivalent to throw in production. That is this error would result in an app crash in production.

I don't quite understand the reason why we can't throw?

kmagiera (approved):

PR #1698, opened 2021-10-24
Update README.md

Macilias:
…s in there

## Description

The path to the example folder has obviously changed from `examples/Example `to `example`.
Besides it was unclear to me why the longer command `react-native run-android `should be favoured over the shorter form `yarn android`

## Test plan

I executed the updated commands and it worked


jakub-gonet (approved):
Thanks!

PR #1704, opened 2021-11-02
Replace fbjs with Lodash

j-piasecki:
## Description

Per suggestion of @jakub-gonet in https://github.com/software-mansion/react-native-gesture-handler/pull/1701#issuecomment-952745153, I replaced `fbjs/lib/areEqual` with `lodash/isEqual`.

## Test plan

Tested on the Example app.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -19,7 +19,6 @@
     "@react-navigation/native": "^5.8.10",
     "@react-navigation/stack": "^5.12.8",
     "expo": "^40.0.0",
-    "fbjs": "^3.0.0",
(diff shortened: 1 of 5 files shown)


jakub-gonet (commented):
Seems like `ua-parser-js` isn't removed from yarn.lock. Can we enforce a specific version that isn't affected by the last vulnerability?

j-piasecki:
I upgraded Expo and it resolved to the updated version.

jakub-gonet (approved):

PR #1706, opened 2021-11-02
Propose Expo integration for RNGestureHandlerEnabledRootView

Kudo:
## Description

Previously, to support gesture-handler in Expo SDK, we had gesture-handler builtin in our template even user don't use it. In newer SDK, we plan to support native modules as opt-in without extra setup; we proposed https://github.com/expo/expo/pull/14883 and it is for gesture-handler integration.

I know `RNGestureHandlerEnabledRootView` is going to deprecate in 2.0. Do you have any timeline of when will be 2.0 available? Otherwise, please consider having change in 1.x release. This pr is actually branched from tag `1.10.3`

## Test Expo integration

[demo repo](https://github.com/Kudo/ExpoAndroidWrapper) which based on expo sdk43 bare template with additional changes:
1. adapts https://github.com/expo/expo/pull/14883 patch because it doesn't publish yet.
2. remove `RNGestureHandlerEnabledRootView` from MainActivity.java
3. add this pr's patch by patch-package.

## Test classic RN app without Expo
makes sure not to break existing functionalities

```sh
$ npx react-native init RN066 --version 0.66
$ yarn add react-native-gesture-handler
# patch this pr's change into node_modules/react-native-gesture-handler
$ ./gradlew :app:assembleDebug
```

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'com.android.library'
+apply from: './expo/linking.gradle'
 
 def safeExtGet(prop, fallback) {
     rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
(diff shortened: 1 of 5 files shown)


Kudo:
close this since 2.0 is released 🚀 🔥 and `RNGestureHandlerEnabledRootView` setup is not necessary anymore.

PR #1717, opened 2021-11-12
Refactor manual activation on Android

j-piasecki:
## Description

Implements changes suggested in https://github.com/j-piasecki/react-native-gesture-handler/pull/1#pullrequestreview-802398161 - `activate` checks if the `manualActivation` is disabled before changing state, and `activateManually` is used when activating from worklet.

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
@@ -539,14 +539,20 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
   }
 
   open fun activate() {
-    if (state == STATE_UNDETERMINED || state == STATE_BEGAN) {
+    if (!manualActivation && (state == STATE_UNDETERMINED || state == STATE_BEGAN)) {
       moveToState(STATE_ACTIVE)
     }
   }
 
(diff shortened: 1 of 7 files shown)


kmagiera (approved):

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
```diff
@@ -539,14 +539,20 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
   }
 
   open fun activate() {
-    if (state == STATE_UNDETERMINED || state == STATE_BEGAN) {
+    if (!manualActivation && (state == STATE_UNDETERMINED || state == STATE_BEGAN)) {
```
kmagiera, line 542: I think I'd perhaps just add a parameter to `activate` that's called `force` and make it `false` by default. This should simplify things a bit, also allow to avoid calling `activate` from `activateManually` and also let overriding methods decide if they want to handle "force" parameter.

j-piasecki: Yeah it seems like a better idea, done in https://github.com/software-mansion/react-native-gesture-handler/pull/1717/commits/635ab9ddb01dcba83d669747c29fc48b6c5cdd1a.

PR #1718, opened 2021-11-12
Add custom gesture

j-piasecki:
## Description

Add custom gesture which utilizes pointer events and state management.


src/handlers/gestures/customGesture.ts
```diff
@@ -0,0 +1,11 @@
+import { ContinousBaseGesture } from './gesture';
+
+export class CustomGesture extends ContinousBaseGesture<Record<string, never>> {
```
kmagiera, line 3: What is the difference between using Continous vs Discrete here? Does it mean we can't make a discrete handler with "CustomGesture" class?

kmagiera (approved):
Looks good. I had a naming comment I already told you about: I think it'd make more sense to call the handler "Manual" rather than "Custom" – I think it better explains the purpose.

jakub-gonet (approved):

src/handlers/gestures/customGesture.ts
j-piasecki: You can make discrete gesture, you just need to control the states correctly. Making it inherit after `ContinousBaseGesture` allows for setting `onUpdate` callback.

PR #1736, opened 2021-11-25
Update types

j-piasecki:
## Description

Changes regarding types:
 - Rename `UnwrappedGestureHandlerEvent` to `GestureUpdateEvent` and export it
 - Rename `UnwrappedGestureHandlerStateChangeEvent` to `GestureStateChangeEvent` and export it
 - export `GestureType`
 - change gesture refs type to `React.RefObject<GestureType | undefined>`
 - add `React.RefObject<React.ComponentType | undefined>` to `GestureRef` not to show error when adding a gesture handler

## Test plan

Ran `tsc`

### DIFF
--- a/src/handlers/gestures/gesture.ts
+++ b/src/handlers/gestures/gesture.ts
@@ -3,8 +3,8 @@ import { ForceTouchGestureHandlerEventPayload } from '../ForceTouchGestureHandle
 import {
   HitSlop,
   CommonGestureConfig,
-  UnwrappedGestureHandlerStateChangeEvent,
-  UnwrappedGestureHandlerEvent,
+  GestureStateChangeEvent,
+  GestureUpdateEvent,
 } from '../gestureHandlerCommon';
 import { getNextHandlerTag } from '../handlersRegistry';
 import { LongPressGestureHandlerEventPayload } from '../LongPressGestureHandler';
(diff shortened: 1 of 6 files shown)


kmagiera (approved):
🚢 🇮🇹

src/handlers/gestures/gesture.ts
```diff
@@ -25,28 +25,28 @@ export type GestureType =
   | BaseGesture<ForceTouchGestureHandlerEventPayload>
   | BaseGesture<NativeViewGestureHandlerPayload>;
 
-export type GestureRef = number | GestureType | React.RefObject<GestureType>;
+export type GestureRef =
+  | number
+  | GestureType
+  | React.RefObject<GestureType | undefined>
+  | React.RefObject<any>; // allow adding a ref to a gesture handler
```
jakub-gonet, line 32: I wonder if we can constrain it to the React components instead of `any` here.

jakub-gonet (approved):
:shipit:

src/handlers/gestures/gesture.ts
j-piasecki: I think it's a good idea, done in https://github.com/software-mansion/react-native-gesture-handler/pull/1736/commits/256f5f4362c3f9d9810e66ad4ed073a7117b1c06.

PR #1749, opened 2021-12-06
Add onFinalize callback

j-piasecki:
## Description

Change `onEnd` callback to be called only when the gesture finishes in the `ACTIVE` state.
Add `onFinalize` callback to all gestures. It's called whenever the gestures finishes, if the gesture transitions from `ACTIVE` state, it will be called after `onEnd`.

This allows for simpler logic in callbacks, as `onEnd` can rely on logic in `onStart`, and `onFinalize` can rely on logic in `onBegin`.

## Test plan

Tested on the Example app, updated Reanimated example


docs/docs/api/gestures/base-gesture-callbacks.md
```diff
@@ -12,6 +12,10 @@ Set the `onStart` callback.
 
 Set the `onEnd` callback.
 
+### `onFinalize(callback)`
+
+Set the `onFinalize` callback.
```
kmagiera, line 17: Can we add a bit more explanation on what each callback does? Not only this one but also the remaining ones. e.g.:
onBegan – set the callback that is being called when given gesture handler starts receiving touches. At the moment of this callback the handler is not yet in an active state and we don't know yet if it will recognize the gesture at all.

kmagiera (approved):

docs/docs/api/gestures/base-gesture-callbacks.md
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1749/commits/dfec565565fc84f49b3f49cef4fcf442f7869048

PR #1762, opened 2021-12-15
Support for useReanimatedGestureHandler in Jest

piaskowyk:
## Description

I want to add support for useReanimatedGestureHandler in Jest tests and I want to make it user-friendly. So I added configuration for events, ready to use in Jest.

Related PR: https://github.com/software-mansion/react-native-reanimated/pull/2747

## Test plan
1. Apply patch - https://patch-diff.githubusercontent.com/raw/software-mansion/react-native-reanimated/pull/2747.patch
2. Run:
```
yarn test -i Event
```

## Example usage
```js
import { 
  fireGestureHandlerPan,
  ghTagEventMacro
} from '../src/jestUtils'

// ...

const eventHandler = useAnimatedGestureHandler({
  onStart: () => props.begin(),
  onActive: () => props.progress(),
  onEnd: () => props.end()
});

<TapGestureHandler onHandlerStateChange={eventHandler}>
  <Text {...ghTagEventMacro()}>TapGestureHandlerTest</Text>
</TapGestureHandler>

// ...

test('test fireGestureHandlerPan', () => {
  const begin = jest.fn();
  const { getByText } = render(<App begin={begin} ... />);

  firePanGestureHandler(
    getByText('PanGestureHandlerTest'),
    { 
      configBegin: { x: 1, y: 1 },
      configProgress: [{ x: 2, y: 2 }, { x: 3, y: 3 }],
      configEnd: { x: 4, y: 4 }
    },
  );
  
  expect(eventFunctions.begin).toHaveBeenCalledTimes(1);
});
```


src/jestUtils.ts
```diff
@@ -0,0 +1,270 @@
...
+const sendProgressEvent = (component: any, eventData?: Record<string, any>) => {
+  // simulate begin -> active state change
+  fireEvent(component, 'gestureHandlerStateChange', {
+    nativeEvent: {
+      oldState: 2,
+      state: 4,
+      ...eventData,
+    },
+  });
+};
```
j-piasecki, line 63: This method currently sends gesture updates in `gestureHandlerStateChange` events instead of `gestureHandlerEvent`. It should be split in two: one to simulate `BEGAN` -> `ACTIVE` state change, the other to simulate sending updates in the `ACTIVE` state.
```suggestion
const sendActivateEvent = (component: any, eventData?: Record<string, any>) => {
  // simulate begin -> active state change
  fireEvent(component, 'gestureHandlerStateChange', {
    nativeEvent: {
      oldState: 2,
      state: 4,
      ...eventData,
    },
  });
};

const sendProgressEvent = (component: any, eventData?: Record<string, any>) => {
  // simulate gesture update
  fireEvent(component, 'gestureHandlerEvent', {
    nativeEvent: {
      state: 4,
      ...eventData,
    },
  });
};
```

src/jestUtils.ts
```diff
@@ -0,0 +1,270 @@
...
+  configEnd?: Record<string, any>
+) => {
+  sendBeginEvent(component, { ...eventData, ...configBegin });
+  if (Array.isArray(configProgress)) {
+    for (const item of configProgress) {
+      sendProgressEvent(component, { ...eventData, ...item });
+    }
+  } else {
+    sendProgressEvent(component, { ...eventData, ...configProgress });
+  }
```
j-piasecki, line 90: Only the first item from `configProgress` should be sent as `gestureHandlerStateChange` event.
```suggestion
  if (Array.isArray(configProgress)) {
    for (const item of configProgress) {
      if (item === configProgress[0]) {
        sendActivateEvent(component, { ...eventData, ...item });
      } else {
        sendProgressEvent(component, { ...eventData, ...item });
      }
    }
  } else {
    sendActivateEvent(component, { ...eventData, ...configProgress });
  }
```

src/jestUtils.ts
```diff
@@ -0,0 +1,388 @@
...
+    sendFailEvent(handler, baseEventData, eventData?.onFail);
+  });
+
+  handlers.forEach((handler) => {
+    sendCancelEvent(handler, baseEventData, eventData?.onCancel);
+  });
+
+  handlers.forEach((handler) => {
+    sendFinishEvent(handler, baseEventData, eventData);
+  });
```
j-piasecki, line 239: There is no need to handle `onFinish` separately. `useAnimatedGestureHandler` implements a state machine that calls it when it receives a `StateChangeEvent` where `oldState` is either `ACTIVE` or `BEGAN`, and the `state` is not (so it must be `END`, `FAILED`, `CANCELLED` or `UNDETERMINED`).

src/handlers/gestures/GestureDetector.tsx
```diff
@@ -416,6 +417,28 @@ export const GestureDetector: React.FunctionComponent<GestureDetectorProps> = (
...
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      const reaGestureHandler = Reanimated.useAnimatedGestureHandler({
+        onStart: handlers.onBegin,
+        onActive: handlers.onUpdate,
+        onEnd: handlers.onEnd,
+        onFinish: handlers.onFinalize,
+      });
+      const handlerProperties = {
+        handlerType: handler.handlerName,
+        handlerTag: handler.handlerTag,
```
j-piasecki, line 433: This is a potential problem, as `handlerTag` is initialized in the `attachHandlers` method which is called in the `useEffect` hook. This has a value of `-1` here.

src/handlers/gestures/GestureDetector.tsx
```diff
@@ -416,6 +417,28 @@ export const GestureDetector: React.FunctionComponent<GestureDetectorProps> = (
...
+  // @ts-ignore @typescript-eslint/ban-ts-comment
+  if (process.env.JEST_WORKER_ID) {
+    for (const handler of gesture) {
+      const handlers = handler.handlers;
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      const reaGestureHandler = Reanimated.useAnimatedGestureHandler({
+        onStart: handlers.onBegin,
+        onActive: handlers.onUpdate,
+        onEnd: handlers.onEnd,
+        onFinish: handlers.onFinalize,
```
j-piasecki, line 429:
```suggestion
        // this one passes the context to the onBegin callback, but I think it's fine in tests
        onStart: handlers.onBegin,
        onActive: (event, context) => {
          // useAnimatedGestureHandler passes all events with state === 4 (ACTIVE) to the onActive
          // callback, event the StateChangeEvent from 2 (BEGAN) to 4. Since the StateChangeEvent
          // will always be the first in the new state we can just check a flag here.
          if (context.active) {
            handlers.onUpdate?.(event);
          } else {
            handlers.onStart?.(event);
            context.active = true;
          }
        },
        onEnd: (event, context) => {
          // in the new API onEnd, onFail and onCancel callbacks are merged into one, because
          // - from the user's perspective there is no meaningfull difference between gesture failing
          //   and being cancelled, so having separate callbacks for it makes no sense
          // - most of the time you want to clean up after the gesture in all of those callbacks,
          //   so there is no point in repeating the same logic three times
          // new onEnd callback receives a bool as the second argument which tells whether the gesture
          // has finished gracefully or not, and if you really need to know if it failed or was cancelled
          // you can chech the state property of event
          handlers.onEnd?.(event, true);
          context.success = true;
        },
        onCancel: (event, context) => {
          handlers.onEnd?.(event, false);
          context.success = false;
        },
        onFail: (event, context) => {
          handlers.onEnd?.(event, false);
          context.success = false;
        },
        onFinish: (event, context) => {
          // onFinalize, like onEnd, requires a flag that tells whether the gesture finished gracefully
          handlers.onFinalize?.(event, context.success ?? false);
        },
```

j-piasecki: I'm not sure if emulating the new API using the old one is the best approach.

src/jestUtils.ts
```diff
@@ -0,0 +1,388 @@
...
+  if (component.props.ghTagContainer !== undefined) {
+    component.props.ghTagContainer.push(handlerProperties);
+  }
+  if (Array.isArray(component.props.children)) {
+    for (const child of component.props.children) {
+      decorateChildrenWithTag(child, handlerProperties);
+    }
+  } else if (typeof component.props.children === 'object') {
+    decorateChildrenWithTag(component.props.children, handlerProperties);
+  }
```
j-piasecki, line 383: I think this may fail if you were to use your own component as a child of `GestureDetector`, then its inner structure wouldn't be decorated. In case of gesture handlers it shouldn't matter that much since they required a native view as a direct child.

PR #1766, opened 2021-12-16
Add prop for setting `handlerTag` in Jest environment

j-piasecki:
## Description

Add prop `testId` which allows to set `handlerTag` property of gesture handlers in Jest environments.
Add `withTestId` method to gesture config objects which behaves in the same way.

Add handler registry for Jest environment which allows to find every gesture configuration and gesture handler in tests using `handlerTag`.

This change simplifies creating tests for Gesture Handler as each gesture can have its tag set explicitly and `handlerTag` is necessary to deliver events to relevant handler.

## Test plan

Tested on a separate app

### DIFF
--- a/src/handlers/createHandler.ts
+++ b/src/handlers/createHandler.ts
@@ -154,9 +154,7 @@ export default function createHandler<
     constructor(props: T & InternalEventHandlers) {
       super(props);
       this.handlerTag =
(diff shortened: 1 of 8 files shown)


jakub-gonet:
Incorporated in #1844.

PR #1768, opened 2021-12-17
Check major version of Reanimated when looking for common interface

j-piasecki:
## Description

Also check major version of Reanimated alongside minor to avoid missing interface declaration with Reanimated 1.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -27,7 +27,7 @@ def shouldUseCommonInterfaceFromReanimated() {
         def json = new JsonSlurper().parseText(inputFile.text)
         def reanimatedVersion = json.version as String
         def (major, minor, patch) = reanimatedVersion.tokenize('.')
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -27,7 +27,7 @@ def shouldUseCommonInterfaceFromReanimated() {
         def json = new JsonSlurper().parseText(inputFile.text)
         def reanimatedVersion = json.version as String
         def (major, minor, patch) = reanimatedVersion.tokenize('.')
-        return Integer.parseInt(minor) >= 3
+        return Integer.parseInt(major) >= 2 && Integer.parseInt(minor) >= 3
```
tomekzaw, line 30: What about Reanimated 3.0? 🧐 😄

j-piasecki: Fair point, I'll leave that for future me to solve.

tomekzaw (approved):

PR #1781, opened 2021-12-23
Add information about change of value since the last event to continuous gestures.

j-piasecki:
## Description

- Add additive `changeX`, `changeY` values to the `PanGestureHandlerEventPayload`.
- Add multiplicative `change` value to the `PinchGestureHandlerEventPayload`
- Add additive `change` value to the `RotationGestureHandlerEventPayload`

## Test plan

Updated examples in the Example app and tested there.


j-piasecki:
Closing in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/1802.

PR #1791, opened 2021-12-29
Use `export type` when exporting types

j-piasecki:
## Description

At the moment some types from Gesture Handler 2.0 are exported using `export` keyword, which causes warnings on web since the types are stripped. This PR replaces it with `export type` which should fix the problem.


kmagiera (approved):

giautm:
Thank you, I just running to the same problem

PR #1793, opened 2021-12-29
Send first `onTouchesDown` after gesture begins

j-piasecki:
## Description

On Android touch events are sent before other events. This causes problems when the gesture is activated in the `onTouchesDown` callback. The gesture then goes from `UNDETERMINED` state to `ACTIVE` before it has a change to process the first event. For example, if done with `PanGesture`, the translation would start with values equal to the coordinates of the pointer.

This PR adds a workaround that sends the first touch event (which would be `TOUCH_DOWN`) only after the handler has processed the event (making sure the handler is initialized correctly, and `UNDETERMINED` -> `BEGAN` state change event is sent).

## Test plan

Tested on the Example app


kmagiera (approved):
👍

PR #1798, opened 2022-01-02
fix: android - onstart on pinch gesture handler

### LINKED ISSUE #553: PinchGestureHandler State.BEGAN inconsistency on Android/iOS
On iOS, the value of `focalY` is correct when `State.BEGAN`. However on Android, this value is always 0. Only in `State.ACTIVE` the `focalY` value is correct. Please let me know if I can help with this issue.
Interestingly enough this doesn't happen for the scale parameter.

For people having similar issues, here is a potential workaround (to adap to your needs because scale is not affected by this issue): 
```
const pinchBegan = Platform.OS === "ios"
      ? eq(state, State.BEGAN)
      : eq(diff(state), State.ACTIVE - State.BEGAN);
```

Possible duplicate: #546

intergalacticspacehighway:
## Description
This PR attempts to fix `onStart` callback on android with `PinchGestureHandler` and resolve the below inconsistency on android and iOS. Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/553
 
### On iOS
- onStart callback gets called during a Pinch (with two finger) gesture. 

### On Android
- onStart callback gets called even when a single finger is tapped and `focalX` and `focalY` are always 0 during onStart.

This PR can cause a breaking change for someone who's depending upon using `onStart` for single-finger gestures on Android (doesn't seem very likely with PinchGestureHandler, but I am not sure!)

## Test plan
- Use `useAnimatedGestureHandler` as shown below and print values of `focalX` and `focalY` or `numberOfPointers` in `onStart` callback. 
- Verify the consistency of `onStart` on Android and iOS.
- `onStart` should be called with 2 finger pinch (similar to iOS). 
- `focalX` and `focalY` should not be 0.

```jsx
    onStart(e, ctx: any) {
      console.log(e.focalX, e.focalY)
    },
});


      <Animated.View
      />
</PinchGestureHandler>
```

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.kt
@@ -40,6 +40,9 @@ class PinchGestureHandler : GestureHandler<PinchGestureHandler>() {
 
     override fun onScaleBegin(detector: ScaleGestureDetector): Boolean {
       startingSpan = detector.currentSpan
+      if (state == STATE_UNDETERMINED) {
+        begin()
+      }
       return true
     }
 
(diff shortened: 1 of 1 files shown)


wcandillon:
Thanks for looking into it. I can confirm the issue.

m-bert:
Hi! As far as I can see `Pinch` no longer activates on single pointer. Let me know if this should be reopened.

PR #1799, opened 2022-01-02
fix: android - Pinch gesture handler when one finger is lifted

intergalacticspacehighway:
## Description
- This PR attempts to make android PinchGestureHandler consistent with iOS's implementation. i.e. Do not move PinchHandler to end when a finger is lifted.
- This issue is discussed in detail over here https://github.com/software-mansion/react-native-gesture-handler/issues/1214

I agree with the notion of preserving the platform behaviour on iOS but in case of android there's no native pinch handler and some custom logic is already there in place, so I am not sure what behavior is correct. I am not sure whether this PR should be merged 😅 

## Approach
- End the handler on ACTION_UP. i.e. both fingers are lifted.
- Do not fire scale detector on ACTION_POINTER_UP. [Read more in the comment](https://github.com/software-mansion/react-native-gesture-handler/pull/1799#discussion_r777180518)

## Test plan

- We can test it with the below snippet. Pinch handler will only move to end state when both the fingers are lifted.
- [This gist](https://gist.github.com/intergalacticspacehighway/9e931614199915cb4694209f12bf6f11) can also be tested. With the changes in this PR, the gist code will run same in both the platforms.

```jsx
    onActive(e) {
      console.log("Active")
    },
    onEnd(e) {
      console.log("Ended")
    }
});


      <Animated.View
      />
</PinchGestureHandler>
```

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.kt
@@ -57,14 +57,13 @@ class PinchGestureHandler : GestureHandler<PinchGestureHandler>() {
       scaleGestureDetector = ScaleGestureDetector(context, gestureListener)
       val configuration = ViewConfiguration.get(context)
       spanSlop = configuration.scaledTouchSlop.toFloat()
-      begin()
     }
-    scaleGestureDetector?.onTouchEvent(event)
-    var activePointers = event.pointerCount
-    if (event.actionMasked == MotionEvent.ACTION_POINTER_UP) {
-      activePointers -= 1
+    
+    if(event.actionMasked != MotionEvent.ACTION_POINTER_UP){
+      scaleGestureDetector?.onTouchEvent(event)
     }
-    if (state == STATE_ACTIVE && activePointers < 2) {
+
+    if (state == STATE_ACTIVE && event.actionMasked == MotionEvent.ACTION_UP) {
       end()
     } else if (event.actionMasked == MotionEvent.ACTION_UP) {
       fail()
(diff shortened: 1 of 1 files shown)


android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.kt
```diff
@@ -59,12 +59,12 @@ class PinchGestureHandler : GestureHandler<PinchGestureHandler>() {
       spanSlop = configuration.scaledTouchSlop.toFloat()
       begin()
     }
-    scaleGestureDetector?.onTouchEvent(event)
-    var activePointers = event.pointerCount
-    if (event.actionMasked == MotionEvent.ACTION_POINTER_UP) {
-      activePointers -= 1
+  
+  if(event.actionMasked != MotionEvent.ACTION_POINTER_UP){
```
intergalacticspacehighway, line 63: This surely appears weird, but I am not sure If there's a better way to achieve this. This is done due to the below scenario.

User has 2 fingers and performing pinch -> One finger is lifted -> ACTION_POINTER_UP is fired -> event.pointerCount is still 2 -> scale detector sets the focal values to single finger -> the event received has new focus but the pointer count is still 2.

However, this again makes it consistent with iOS. iOS Pinch handler doesn't send events in case of a finger is lifted. Only when a single pointer is moved, it starts sending the event.

intergalacticspacehighway: One more way would be to override the numberOfPointers in PinchGestureHandler and reduce it by 1 when Action is Pointer up but then we'll have to make it `open` in GestureHandler 😓

intergalacticspacehighway:
Closing in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/1800

PR #1800, opened 2022-01-03
fix: android - Pinch gesture handler when one finger is lifted

intergalacticspacehighway:
## Description
- This PR attempts to make android PinchGestureHandler consistent with iOS's implementation. i.e. Do not move PinchHandler to `end` state when a finger is lifted.
- This issue is discussed in detail over here https://github.com/software-mansion/react-native-gesture-handler/issues/1214

I agree with the notion of preserving the platform behaviour on iOS but in case of android there's no native pinch handler and some custom logic is already there in place, so I am not sure what behavior is correct.

## Approach
- End the gesture on ACTION_UP. i.e. both fingers are lifted.
- Added a property in PinchGestureHandler to control the `numberOfPointer` during `ACTION_POINTER_UP`

## Test plan

- We can test it with the below snippet. Pinch handler will only move to end state when both the fingers are lifted.
- [This gist](https://gist.github.com/intergalacticspacehighway/9e931614199915cb4694209f12bf6f11) can also be tested. With the changes in this PR, the gist code will run same in both the platforms.

```jsx
    onActive(e) {
      console.log("Active")
    },
    onEnd(e) {
      console.log("Ended")
    }
});


      <Animated.View
      />
</PinchGestureHandler>
```

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/PinchGestureHandler.kt
@@ -61,15 +61,11 @@ class PinchGestureHandler : GestureHandler<PinchGestureHandler>() {
       spanSlop = configuration.scaledTouchSlop.toFloat()
       begin()
     }
-
-    activePointers = event.pointerCount;
-
     scaleGestureDetector?.onTouchEvent(event)
-
+    activePointers = event.pointerCount;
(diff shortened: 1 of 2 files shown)


wcandillon:
I can confirm that Android behaves differently from iOS there

m-bert:
This seems to work in both V2 and V3 API. Let me know if this should be reopened.

PR #1802, opened 2022-01-03
Add `onChange` callback to continuous gestures.

j-piasecki:
## Description

This PR adds `onChange` callback to continuous gestures. This callback is called just after `onUpdate` and carries additional information - change of value since the last event:
- For `PanGesture`: `changeX`, `changeY` - additive difference in translation
- For `PinchGesture`: `scaleChange` - multiplicative difference in scale
- For `RotationGesture`: `rotationChange` - additive difference in rotation
- For `ForceTouchGesture`: `forceChange` - additive difference in force

## Test plan

Updated examples and tested on the Example app.


kmagiera (approved):
🎉

PR #1807, opened 2022-01-05
Call `onChange` in `eventReceiver`

j-piasecki:
## Description

Call `onChange` in `eventReceiver` just after `onUpdate` to make it work without Reanimated.

## Test plan

Tested on the separate app.

### DIFF
--- a/src/handlers/gestures/eventReceiver.ts
+++ b/src/handlers/gestures/eventReceiver.ts
@@ -36,7 +36,7 @@ const dummyStateManager: GestureStateManagerType = {
   },
 };
 
-const lastUpdateEvent: GestureUpdateEvent[] = [];
+const lastUpdateEvent: (GestureUpdateEvent | undefined)[] = [];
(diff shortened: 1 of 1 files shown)


src/handlers/gestures/eventReceiver.ts
```diff
@@ -36,6 +36,8 @@ const dummyStateManager: GestureStateManagerType = {
   },
 };
 
+const lastUpdateEvent: GestureUpdateEvent[] = [];
```
kmagiera, line 39: Do we ever cleanup events stored for handler that are no longer in a tree? We perhaps should

j-piasecki: Changed in https://github.com/software-mansion/react-native-gesture-handler/pull/1807/commits/8ed3577f8640c5330fa52be049eef02aa4747b93. Events will be cleared after calling `onFinalize`.

PR #1817, opened 2022-01-11
Add notice for old API users

jakub-gonet:
## Description

We want to improve the discoverability of the new API. To do that we add `console.info` when creating a gesture component such as `<PanGestureHandler />`. Unfortunately, we can't prevent false positives when components are used inside libraries. To disable this info, you can use this snippet:
```tsx
import { LogBox } from 'react-native';

LogBox.ignoreLogs([
  "[react-native-gesture-handler] Seems like you\'re using an old API with gesture components, check out new Gestures system!",
]);
```

## Test plan

- Checked if warning doesn't appear in RNGH2 component (`Example/src/new_api/reanimated/index.tsx`) and appears in RNGH1 component (`Example/src/basic/bouncing/index.tsx`)
- Checked if a warning doesn't appear with `LogBox.ignoreLogs`

### DIFF
--- a/src/handlers/createHandler.ts
+++ b/src/handlers/createHandler.ts
@@ -124,13 +124,11 @@ type InternalEventHandlers = {
 
 let showedRngh2Notice = false;
 function showRngh2NoticeIfNeeded() {
-  if (__DEV__) {
-    if (!showedRngh2Notice) {
-      console.warn(
-        "[react-native-gesture-handler] Seems like you're using an old API with gesture components, check out new Gestures system!"
-      );
-      showedRngh2Notice = true;
-    }
+  if (!showedRngh2Notice) {
+    console.warn(
+      "[react-native-gesture-handler] Seems like you're using an old API with gesture components, check out new Gestures system!"
+    );
+    showedRngh2Notice = true;
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/handlers/createHandler.ts
```diff
@@ -162,6 +173,7 @@ export default function createHandler<
         }
         handlerIDToTag[props.id] = this.handlerTag;
       }
+      showRngh2NoticeIfNeeded();
```
kmagiera, line 176: wrap this call with `if (__DEV__)` – this will allow this code block to be completely removed from release packages

src/handlers/createHandler.ts
```diff
@@ -122,6 +123,16 @@ type InternalEventHandlers = {
   onGestureHandlerStateChange?: (event: any) => void;
 };
 
+let showedRngh2Notice = false;
+function showRngh2NoticeIfNeeded() {
+  if (!showedRngh2Notice && isDevEnvironment()) {
```
kmagiera, line 128: checking `isDevEnvironment` won't be necessary here if you apply the below comment

kmagiera (approved):
Can you remove changes to podfile/lock and send them in a separate PR ?

jakub-gonet:
Extracted lock updates to new PR: https://github.com/software-mansion/react-native-gesture-handler/pull/1818

PR #1842, opened 2022-01-24
Change log level from warn to info on new API notice

cm-nutrien:
## Description

This updates the notice about the new Gestures system API to use `console.info` instead of `console.warn` to match the expectation initially expressed in [PR #1817](https://github.com/software-mansion/react-native-gesture-handler/pull/1817). Mentioned also in #1831.

## Test plan

No specific plan other than verifying that the notice is logged as info and not warning.

### DIFF
--- a/src/handlers/createHandler.ts
+++ b/src/handlers/createHandler.ts
@@ -125,7 +125,7 @@ type InternalEventHandlers = {
 let showedRngh2Notice = false;
 function showRngh2NoticeIfNeeded() {
   if (!showedRngh2Notice) {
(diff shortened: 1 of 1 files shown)


jakub-gonet (commented):
IIRC, this was changed to warn, since info can't be suppressed. Could you verify that `LogBox.ignoreLogs()` still works?

cm-nutrien:
That makes sense and after working at it for a while, I can't say with certainty that LogBox will suppress it correctly if it's a `console.info` so I will close this PR.

PR #1844, opened 2022-01-25
Add ability to fire gesture events in testing

jakub-gonet:
## Description

Supersedes #1762.

**Needs to be synchronized with** https://github.com/software-mansion/react-native-reanimated/pull/2885.

This PR introduces a new test API `fireGestureHandler(handler, eventList)`, allowing to test handlers bound to a given gesture (or gesture handler).

`handler` can be one of two things:
1. a gesture component queried by `getByTestId` (`testID` must be used directly on handler, e.g. `<RotationGestureHandler testID="rotation">`),
2. a gesture opaque structure queried by `getByHandlerTestId` (`testID` can be set via `withTestId(name)`, e.g. `Gesture.Pan().withTestId('pan')`).

`eventList` contains events that should be emitted to provided `handler`. State change events must have `state` field. Subsequent `ACTIVE` events can have `state` field omitted. Every entry can have additional attributes for a given handler type (e.g. `translationX` for Pan).

`fireGestureHandler` is typed and can be parametrized with all Gestures and Gesture Handlers (e.g. `fireGestureHandler<PanGesture>`, `fireGestureHandler<LongPressGestureHandler>`).

### Example test

```jsx
import React from 'react';
import { render } from '@testing-library/react-native';
import { Text } from 'react-native';
import {
  GestureHandlerRootView,
  Gesture,
  GestureDetector,
  State,
} from '../index';
import { fireGestureHandler, getByGestureTestId } from 'gesture-handler';
import { cleanup } from '@testing-library/react-native';

beforeEach(cleanup);

const mockedEventHandlers = () => {
  return {
    begin: jest.fn(),
    start: jest.fn(),
    active: jest.fn(),
    end: jest.fn(),
    finish: jest.fn(),
  };
};

function SingleHandler({ handlers, treatStartAsUpdate }) {
  const pan = Gesture.Pan()
    .onBegin(handlers.begin)
    .onStart(treatStartAsUpdate ? handlers.active : handlers.start)
    .onUpdate(handlers.active)
    .onEnd(handlers.end)
    .onFinalize(handlers.finish)
    .withTestId('pan');

  return (
    <GestureHandlerRootView>
        <Text>v2 API test</Text>
      </GestureDetector>
    </GestureHandlerRootView>
  );
}

it('sends events with additional data to handlers', () => {
  const panHandlers = mockedEventHandlers();
  render(<SingleHandler handlers={panHandlers} treatStartAsUpdate />);
  fireGestureHandler<PanGesture>(getByGestureTestId('pan'), [
    { state: State.BEGAN, translationX: 0 },
    { state: State.ACTIVE, translationX: 10 },
    { translationX: 20 },
    { translationX: 20 },
    { state: State.END, translationX: 30 },
  ]);

  expect(panHandlers.active).toBeCalledTimes(3);
  expect(panHandlers.active).toHaveBeenLastCalledWith(
    expect.objectContaining({ translationX: 20 })
  );
});
```
 
 ## TODO
 - [x] Allow parametrizing the `fireGestureHandler` with handler, e.g. `fireGestureHandler<PanGesture>(...)` or `fireGestureHandler<PanGestureHandler>(...)`.
 - [x] Add tests for FAIL and CANCEL events.
 - [x] Implicitly add missing states if not specified.
 - [x] Add documentation
 
## Test plan

Ran `yarn test -i Events`.


src/jestUtils.ts
```diff
@@ -0,0 +1,359 @@
...
+    // @ts-ignore afesfa
+    return withPrevAndCurrent(this, mapFn);
+    /* eslint-enable */
+  });
+
+  const events = eventList
+    .map(fillMissingDefaultsFor({ handlerTag, handlerType }))
+    .withPrevAndCurrent(fillMissingActiveStateFields)
+    .withPrevAndCurrent(fillOldStateChanges)
+    .withPrevAndCurrent(validateStateTransitions)
```
jakub-gonet, line 321: Those functions are badly typed, we need to revise the internal handlers typing.

src/__tests__/Events.test.tsx
```diff
@@ -0,0 +1,384 @@
...
+
+    expect(handlers.begin).toBeCalledWith(
+      expect.objectContaining({ state: State.BEGAN }),
+      expect.any(Object)
+    );
+  });
+});
+
+describe('Using RNGH v2 gesture API', () => {
+  function SingleHandler({ handlers, treatStartAsUpdate }: any) {
```
j-piasecki, line 267: I think we can create a type for v2 similarly to how it looks for v1:
```ts
interface V2ApiProps {
  handlers: ReturnType<typeof mockedEventHandlers>;
  treatStartAsUpdate?: boolean;
}
```
Some functions will be unnecessary but `mockedEventHandlers` is used to generate handlers anyway.

src/jestUtils.ts
```diff
@@ -0,0 +1,359 @@
...
+    return {
+      oldState: State.UNDETERMINED,
+      ...currentEvent,
+    } as GestureHandlerTestEvent;
+  }
+
+  const isGestureStateEvent = previousEvent.state !== currentEvent.state;
+  if (isGestureStateEvent) {
+    return {
+      oldState: previousEvent.state,
```
j-piasecki, line 158: `previousEvent` may be null here, I'm not sure if ignoring that is intentional or not.
```suggestion
  const isGestureStateEvent = previousEvent?.state !== currentEvent.state;
  if (isGestureStateEvent) {
    return {
      oldState: previousEvent?.state,
```

src/jestUtils.ts
```diff
@@ -0,0 +1,359 @@
...
+
+function fillMissingActiveStateFields(
+  previousEvent: Omit<GestureHandlerTestEvent, 'oldState'> | null,
+  currentEvent: Omit<GestureHandlerTestEvent, 'state' | 'oldState'>
+) {
+  const isFirstEvent = previousEvent === null;
+  if (isFirstEvent) {
+    return currentEvent;
+  }
+  if (previousEvent.state === State.ACTIVE && !currentEvent['state']) {
```
j-piasecki, line 174: Same here, `previousEvent` may be null.
```suggestion
  if (previousEvent?.state === State.ACTIVE && !currentEvent['state']) {
```

jakub-gonet: Can it? I think we check for null above, right?
```tsx
const isFirstEvent = previousEvent === null;
if (isFirstEvent) {
  return currentEvent;
}
```

jakub-gonet: But `!currentEvent['state']` is buggy, it's true for `state === 0`. I'll change it.

src/__tests__/Events.test.tsx
jakub-gonet: Yes, I just forgot to type it properly.

> Some functions will be unnecessary but mockedEventHandlers is used to generate handlers anyway.

I don't understand this statement, could you expand on that?

src/jestUtils.ts
j-piasecki: Oh, you're right. I've missed it.

src/__tests__/Events.test.tsx
j-piasecki: `fail` and `cancel` can't be 1:1 mapped to the new API, they trigger `onEnd` with `false` as the second argument.

j-piasecki (commented):
Right now there are some require cycles. I think `isJest` function can be moved from `jestUtils.ts` to `utils.ts` as it's the only method imported from it by `createHandlers` and `handlersRegistry` (and that's the cause for the cycles).

src/__tests__/Events.test.tsx
jakub-gonet: Fixed in 0b089c0c.

src/jestUtils.ts
jakub-gonet: Property check fixed in 4d3dd6d5.

jakub-gonet:
Good point, fixed in 45f7d9eb.

src/__tests__/Events.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+    expect(panHandlers.active).toBeCalledTimes(3);
+    expect(panHandlers.active).toHaveBeenLastCalledWith(
+      expect.objectContaining({ x: 3, y: 13 })
+    );
+  });
+
+  it('fills BEGIN and END events for discrete handlers', () => {
+    const handlers = mockedEventHandlers();
+    render(<RacingTapAndPan handlers={handlers} treatStartAsUpdate />);
+    fireGestureHandler(getByGestureTestId('tap'), [{ x: 5 }]);
```
j-piasecki, line 467:
```suggestion
    fireGestureHandler<TapGesture>(getByGestureTestId('tap'), [{ x: 5 }]);
```

src/__tests__/Events.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+    expect(handlers.begin).toBeCalledTimes(1);
+    expect(handlers.end).toBeCalledTimes(1);
+  });
+
+  it('with FAILED event, fills BEGIN event for discrete handlers', () => {
+    const handlers = mockedEventHandlers();
+    render(<RacingTapAndPan handlers={handlers} treatStartAsUpdate />);
+    fireGestureHandler(getByGestureTestId('tap'), [{ state: State.FAILED }]);
```
j-piasecki, line 475:
```suggestion
    fireGestureHandler<TapGesture>(getByGestureTestId('tap'), [{ state: State.FAILED }]);
```

src/__tests__/Events.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+    expect(handlers.begin).toBeCalledTimes(1);
+    expect(handlers.end).toBeCalledTimes(1);
+    expect(handlers.end).toBeCalledWith(expect.anything(), false);
+  });
+
+  it('uses event data from first event in filled BEGIN, ACTIVE events', () => {
+    const handlers = mockedEventHandlers();
+    render(<RacingTapAndPan handlers={handlers} treatStartAsUpdate />);
+    fireGestureHandler<TapGesture>(getByGestureTestId('pan'), [{ x: 120 }]);
```
j-piasecki, line 484:
```suggestion
    fireGestureHandler<PanGesture>(getByGestureTestId('pan'), [{ x: 120 }]);
```

src/__tests__/Events.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+    expect(handlers.active).toHaveBeenNthCalledWith(
+      1,
+      expect.objectContaining({ x: 120 })
+    );
+  });
+
+  it('uses event data from last event in filled END events', () => {
+    const handlers = mockedEventHandlers();
+    render(<RacingTapAndPan handlers={handlers} treatStartAsUpdate />);
+    fireGestureHandler<TapGesture>(getByGestureTestId('pan'), [
```
j-piasecki, line 495:
```suggestion
    fireGestureHandler<PanGesture>(getByGestureTestId('pan'), [
```

src/__tests__/Events.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+    expect(handlers.end).toBeCalledWith(
+      expect.objectContaining({ x: 120 }),
+      false
+    );
+  });
+
+  it('uses event data filled events', () => {
+    const handlers = mockedEventHandlers();
+    render(<RacingTapAndPan handlers={handlers} treatStartAsUpdate />);
+    fireGestureHandler<TapGesture>(getByGestureTestId('pan'), [
```
j-piasecki, line 509:
```suggestion
    fireGestureHandler<PanGesture>(getByGestureTestId('pan'), [
```

src/__tests__/Events.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+    render(<RacingTapAndPan handlers={handlers} treatStartAsUpdate />);
+    fireGestureHandler<TapGesture>(getByGestureTestId('tap'));
+    expect(handlers.begin).toBeCalledTimes(1);
+    expect(handlers.end).toHaveBeenCalledTimes(1);
+  });
+
+  it("fills BEGIN, ACTIVE and END events when they're not present, for continuous handlers", () => {
+    const handlers = mockedEventHandlers();
+    render(<RacingTapAndPan handlers={handlers} treatStartAsUpdate />);
+    fireGestureHandler<TapGesture>(getByGestureTestId('pan'));
```
j-piasecki, line 535:
```suggestion
    fireGestureHandler<PanGesture>(getByGestureTestId('pan'));
```

src/jestUtils.ts
```diff
@@ -0,0 +1,491 @@
...
+  eventList: Partial<GestureHandlerTestEvent<ExtractConfig<THandler>>>[] = []
+): void {
+  const { emitEvent, handlerType, handlerTag } = getHandlerData(
+    componentOrGesture
+  );
+
+  let _ = fillMissingStatesTransitions(
+    eventList,
+    isDiscreteHandler(handlerType)
+  );
```
j-piasecki, line 459: Why `_` as a name?

jakub-gonet: In JS we don't have a pipe operator and this is basically a pipeline. Adding meaningful names would only obfuscate this code since we don't need intermediate results. Another approach is to modify the prototype (https://github.com/software-mansion/react-native-gesture-handler/pull/1844/commits/f28cb394c3156dd6a3341cae456f9a7ea54d2165) and chain it but it's rarely seen in JS codebases so I reverted to this one.

Andarius:
Hey @jakub-gonet  @j-piasecki , do you know when a release including `isJest` and `fireGestureHandler` will be uploaded?
Writing jest tests triggers a lot of warning and is limited for now (no way to test this for instance)
```ts
                        <View testID="background" style={[styles.background]} />
</GestureDetector>
```

j-piasecki:
Hi @Andarius! We have just released a new version that, among others, contains the testing API.

virdesai:
@jakub-gonet @j-piasecki is it expected that without Fabric enabled, the `RNGestureHandlerButton` no longer is getting assigned a testID when checking the snapshot in a jest test? when I comment out [this line](https://github.com/software-mansion/react-native-gesture-handler/commit/273fa03679cea7a648fb6f8639d36c4f58b8934a#diff-92425c02fdae54eeec59634f1410d3c7c6b1af060161a0a93708f9ec7dda8f97R443) it exists as I'd expect. from reading the docs, using `withTestId` doesn't seem to be made for this situation

PR #1847, opened 2022-01-26
Remove unused devDependencies

jakub-gonet:
## Description

Removed some unused dev packages.

## Test plan

Ran example app on iOS.


j-piasecki (commented):
Lint check failing suggests that some of them weren't unused. Keeping `eslint-import-resolver-babel-module` and adding `eslint-plugin-jest` seems to be enough to make it work.

Issue #1849, opened 2022-01-27
Composition Gesture Caused Huge UI-Frame Drops !


yaaliuzhipeng:
## Description

use composed gestures within virtualized list's item component and the item component take a state from the root component . huge ui frame dropped when the state changed. !!!

### Platforms

- ⏱ iOS  [haven't test yet]
- ❌ Android
- ⏱ Web [haven't test yet]

### Screenshots
[frame dropped screenshot](https://dd-static.jd.com/ddimg/jfs/t1/97145/2/21614/31536/61f2a32fE14b627c0/ec5ec48fb306dc69.png)
## Steps To Reproduce

1. see the example code below ( just copy and run it )

### Expected behavior
expect ui frame to stay at least 58fps
### Actual behavior
huge frame dropped , average 30 fps 
## Snack or minimal code example
```typescript
import React, { useMemo, useState } from 'react';
import { View, Text, FlatList } from 'react-native';
import { Gesture, GestureDetector, GestureHandlerRootView } from 'react-native-gesture-handler';

const Item = React.memo((props: any) => {
    const { size, currentIndex } = props;
    const singleTap = useMemo(() => Gesture.Tap()
        .numberOfTaps(1)
        .onEnd((event, success) => {
            if (success) {
                console.log('single tap');
            }
        }), [])
    const doubleTap = useMemo(() => Gesture.Tap()
        .numberOfTaps(2)
        .onEnd((event, success) => {
            if (success) {
                console.log('double tap');
            }
        }), [])
    const taps = useMemo(() => Gesture.Exclusive(doubleTap, singleTap), []);
    return (
        <GestureDetector gesture={taps}>
            <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center', width: size.width, height: size.height, borderWidth: 1 }}>
                <Text style={{ fontSize: 60 }}>{props.index}</Text>
            </View>
        </GestureDetector>
    )
});

const CaseComposeGestureTap = (props) => {

    const [layout, setLayout] = useState({ width: 0, height: 0 })
    const [data, setData] = useState(function () {
        let d: any[] = [];
        for (let i = 0; i < 8; i++) {
            d.push({ id: i + 1 })
        }
        return d;
    }())
    const [currentIndex, setCurrentIndex] = useState(0);

    const renderItem = ({ item, index }) => {
        return <Item index={index} item={item} size={layout} currentIndex={currentIndex} />
    }
    const onLayout = React.useCallback(event => {
[...]

PR #1856, opened 2022-01-31
Rebuild Reanimated event only when needed

j-piasecki:
## Description

At the moment, Reanimated event is rebuilt on every update which may cause a heavy performance hit in some cases. I believe that it's a leftover from pre-release Gesture Handler 2. This PR changes it so that the event will be rebuilt only when the gestures are reattached alleviating performance issues.

Should fix https://github.com/software-mansion/react-native-gesture-handler/issues/1849.

## Test plan

Tested on the Example app.


jakub-gonet (approved):

jakub-gonet:
Side note: we need some kind of perf tests to catch similar issues in the future.

Issue #1813, opened 2022-01-10
Xcode build error for AppleTV on 2.1.0


robbert040:
## Description

When trying to create a build for AppleTV with version 2.1.0 you wil run into the following issue

'UIRotationGestureRecognizer' is unavailable: not available on tvOS

### Screenshots

<img width="1048" alt="Screenshot 2022-01-10 at 11 14 13" src="https://user-images.githubusercontent.com/59539159/148749405-6fb3beeb-5091-4b08-8d75-043747bd3184.png">

## Steps To Reproduce

1. create a fresh react-native-tv-os project
2. Update the gesture handler to version 2.1.0
3. Build for AppleTV

### Expected behavior

The build will succeed it was working on version "react-native-gesture-handler": "^1.10.3"

### Actual behaviour

The build fails with an error

## Package versions

- React: "react": "17.0.1"
- React Native: npm:react-native-tvos@0.64.2-4
- React Native Gesture Handler: "react-native-gesture-handler": "^2.1.0"

todorone:
Confirming it, RNGH `1.x` works fine on tvOS and `2.x` does not.

PR #1857, opened 2022-01-31
Fix tvOS compilation errors

j-piasecki:
## Description

Fix tvOS compilation errors by excluding unsupported parts of the code when building for it.

Should fix https://github.com/software-mansion/react-native-gesture-handler/issues/1813.

## Test plan

Tested on a separate app.


jakub-gonet (approved):
LGTM, haven't tested though.

PR #1863, opened 2022-02-02
Update React Native version in the Example app to 0.67.2

j-piasecki:
## Description

Update React Native version in the Example app to 0.67.2
Make the app build on M1 Mac

## Test plan

Build the app


kmagiera (approved):

jakub-gonet (approved):
Tested on an intel mac, both platforms work.

I needed to update Podfile and yarn lock files though.

PR #1882, opened 2022-02-18
Fix TypeScript

j-piasecki:
## Description

Fixes TypeScript errors.

### DIFF
--- a/example/src/new_api/camera/index.tsx
+++ b/example/src/new_api/camera/index.tsx
@@ -20,7 +20,10 @@ export default function Home() {
   const [selectedFilter, setSelectedFilter] = useState(0);
   const [isRecording, setIsRecording] = useState(false);
   const [remainingTimeMs, setRemainingTimeMs] = useState(MAX_VIDEO_DURATION_MS);
-  const [recordingIntervalHandle, setRecordingIntervalHandle] = useState(-1);
+  const [
+    recordingIntervalHandle,
+    setRecordingIntervalHandle,
+  ] = useState<NodeJS.Timeout | null>(null);
 
   const filtersPanGesture = Gesture.Pan()
     .onUpdate((e) => {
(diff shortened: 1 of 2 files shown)


example/src/new_api/camera/index.tsx
```diff
@@ -20,7 +20,10 @@ export default function Home() {
   const [selectedFilter, setSelectedFilter] = useState(0);
   const [isRecording, setIsRecording] = useState(false);
   const [remainingTimeMs, setRemainingTimeMs] = useState(MAX_VIDEO_DURATION_MS);
-  const [recordingIntervalHandle, setRecordingIntervalHandle] = useState(-1);
+  const [
+    recordingIntervalHandle,
+    setRecordingIntervalHandle,
+  ] = useState<NodeJS.Timeout | null>(null);
```
jakub-gonet, line 26: IIRC `setTimeout` return value was different on the web and RN. Use `ReturnType<typeof setTimeout>` instead.

src/jestUtils.ts
```diff
@@ -172,10 +172,12 @@ function fillOldStateChanges(
     } as GestureHandlerTestEvent;
   }
 
-  const isGestureStateEvent = previousEvent.state !== currentEvent.state;
+  // At this point previousEvent is not null, but for some reason it breaks CI
+  // optional chaining used only to fix it
+  const isGestureStateEvent = previousEvent?.state !== currentEvent?.state;
```
jakub-gonet, line 177: Try upgrading TS to 4.5 in the example app (same as in the main package). It may resolve this issue.

example/src/new_api/camera/index.tsx
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1882/commits/08e2ff1a328b5da72dcf726cd2f236127c3fdef1.

src/jestUtils.ts
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1882/commits/ec3ee214654873c6c9304db755dc31f2b6afd648. Seems like it worked.

PR #1893, opened 2022-02-22
[Fabric] Gesture Handler for Fabric

tomekzaw:
## Description

This PR adds Fabric support to React Native Gesture Handler.

Co-authored by @j-piasecki.

### Main changes

* View Flattening with Fabric enabled (in some cases it may be necessary to add `collapsable={false}` in client apps)
* Old API & direct vs. device events (see [details](https://github.com/software-mansion/react-native-gesture-handler/pull/1893#discussion_r814109074))

### Building FabricExample app in release mode

<details>

The following errors may occur when building FabricExample app in release mode:

1. `Invalid regular expression: Quantifier has nothing to repeat`

Hermes does not support optional named groups in regular expressions but `ansi-styles` library uses them:

```js
const matches = /(?<colorString>[a-f\d]{6}|[a-f\d]{3})/i.exec(hex.toString(16));
```

As a workaround, you can comment out the implementation directly in `node_modules`.

2. `Undefined symbol: folly::SharedMutexImpl<false, void, std::__1::atomic, false, false, false>::deferredReaders`

As a workaround, you can comment out `use_flipper!()` in your Podfile.

</details>


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -17,32 +17,39 @@ import android.view.View.OnClickListener
 import android.view.ViewGroup
 import androidx.core.view.children
 import com.facebook.react.bridge.SoftAssertions
-import com.facebook.react.uimanager.PixelUtil
-import com.facebook.react.uimanager.ThemedReactContext
-import com.facebook.react.uimanager.ViewGroupManager
-import com.facebook.react.uimanager.ViewProps
+import com.facebook.react.module.annotations.ReactModule
+import com.facebook.react.uimanager.*
```
Ubax, line 21: Do you want to have `*` import here?

src/fabric/RNGestureHandlerRootView.tsx
```diff
@@ -0,0 +1,20 @@
...
+export default function GestureHandlerRootView({
+  children,
+  ...rest
+}: GestureHandlerRootViewProps) {
+  return (
+    <GestureHandlerRootViewNativeComponent {...rest}>
+      {children}
+    </GestureHandlerRootViewNativeComponent>
+  );
+}
```
Ubax, line 20: I think that `<GestureHandlerRootViewNativeComponent {...rest}>{childres}</...>` is the same as `<GestureHandlerRootViewNativeComponent {...props} />`. 

Is there a specific reason why you divide the props here?

Ubax (commented):
I added some comments. I didn't check the GH specific logic though

FabricExample/android/settings.gradle
```diff
@@ -1,9 +1,13 @@
 rootProject.name = 'FabricExample'
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 include ':app'
-includeBuild('../node_modules/react-native-gradle-plugin')
 
 if (settings.hasProperty("newArchEnabled") && settings.newArchEnabled == "true") {
+    includeBuild('../node_modules/react-native-gradle-plugin') {
+        dependencySubstitution {
+            substitute(module("com.facebook.react:react")).using(project(":"))
```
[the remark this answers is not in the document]
tomekzaw: Basically it's the same piece of code as in [react-native-screens](https://github.com/software-mansion/react-native-screens/blob/7c6e14b45834cf009c7069d46a528cf792d2a15e/FabricExample/android/settings.gradle#L4-L8) but moved inside `if`.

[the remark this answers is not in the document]
Ubax: ok

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
j-piasecki: I changed it to specific imports in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/89232814dc7aa64bda39cb1874016771a99ebfa0. There was a problem where I couldn't import a specific file but `*` made it work, it seems to be ok now.

src/fabric/RNGestureHandlerRootView.tsx
j-piasecki: It seems like the `fabric/...` wrappers weren't necessary at all. But your comment also applies to the older `GestureHandlerRootView.android.tsx`. I don't think there is a reason for it to be this way. I changed it to match `GestureHandlerRootView.tsx` in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/50f054f59637198cecad15875547887d0a6ac2e4.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -333,10 +348,10 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>() {
         SoftAssertions.assertNotNull(attr)
         return when (attr) {
           SELECTABLE_ITEM_BACKGROUND -> {
-            R.attr.selectableItemBackground
+            android.R.attr.selectableItemBackground
```
WoLewicki, line 351: why this is needed? `android.R` seems to be a bit different from `R`

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -518,83 +542,122 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?)
     handlerFactories.firstOrNull { it.type == handler.javaClass } as HandlerFactory<T>?
 
   private fun <T : GestureHandler<T>> onHandlerUpdate(handler: T, motionEvent: MotionEvent) {
+    // onUpdate
```
WoLewicki, line 545: what does this comment concern?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -518,83 +542,122 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?)
...
+
     if (handler.tag < 0) {
       // root containers use negative tags, we don't need to dispatch events for them to the JS
       return
     }
     if (handler.state == GestureHandler.STATE_ACTIVE) {
       val handlerFactory = findFactoryForHandler(handler)
 
-      if (handler.usesDeviceEvents) {
+      if (handler.actionType == GestureHandler.ACTION_TYPE_REANIMATED_WORKLET) {
```
WoLewicki, line 554: All those changes are needed due to different handling of events in Fabric? If so, maybe it would be good to give a quick explanation why is it like that now for other libraries maintainers, or is it something that will happen only in this library?

android/src/main/jni/RNGestureHandlerComponentsRegistry.cpp
```diff
@@ -0,0 +1,71 @@
...
+  delegate->buildRegistryFunction = buildRegistryFunction;
+  return instance;
+}
+
+void RNGestureHandlerComponentsRegistry::registerNatives() {
+  registerHybrid({
+      makeNativeMethod("initHybrid", RNGestureHandlerComponentsRegistry::initHybrid),
+  });
+  
+  // This is a temporary solution that allows components exported by the screens
```
WoLewicki, line 61: change all `screens` to `gesture-handler`

android/src/paper/java/com/swmansion/gesturehandler/ReactContextExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package com.swmansion.gesturehandler
+
+import com.facebook.react.bridge.ReactContext
+import com.facebook.react.uimanager.UIManagerModule
+import com.facebook.react.uimanager.events.Event
+
+fun ReactContext.dispatchEvent(event: Event<*>) {
+    this.getNativeModule(UIManagerModule::class.java)!!.eventDispatcher.dispatchEvent(event)
```
WoLewicki, line 8: Shouldn't `let` be used here to be `null-safe`?

ios/RNGestureHandlerButtonComponentView.mm
```diff
@@ -0,0 +1,48 @@
+#ifdef RN_FABRIC_ENABLED
```
WoLewicki, line 1: It is added to reduce the final code size on paper? If so, should we do the same but inverted for Fabric code?

src/utils.ts
```diff
@@ -26,14 +26,20 @@ export function withPrevAndCurrent<T, Transformed>(
...
 
-export function hasProperty(object: Record<string, unknown>, key: string) {
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function hasProperty(object: object, key: string) {
   return Object.prototype.hasOwnProperty.call(object, key);
 }
 
 export function isJestEnv(): boolean {
-  return !!process.env.JEST_WORKER_ID;
+  return hasProperty(global, 'process') && !!process.env.JEST_WORKER_ID;
```
WoLewicki, line 35: Is this change needed in this PR?

src/jestUtils.ts
```diff
@@ -60,6 +59,22 @@ import {
 import { State } from './State';
 import { hasProperty, withPrevAndCurrent } from './utils';
 
+// load fireEvent conditionally, so RNGH may be used in setups without testing-library
```
WoLewicki, line 62: Is this change needed in this PR?

src/utils.ts
WoLewicki, line 35: Is this change needed in this PR?

src/jestUtils.ts
WoLewicki, line 62: Is this change needed in this PR?

src/getShadowNodeFromRef.web.ts
```diff
@@ -0,0 +1,7 @@
+// Used by GestureDetector (unsupported on web at the moment) to check whether the
```
WoLewicki, line 1: I would add this comment to non-web version too.

example/ios/Podfile
```diff
@@ -6,7 +6,11 @@ platform :ios, '11.0'
 target 'Example' do
   config = use_native_modules!
 
-  use_react_native!(:path => config["reactNativePath"])
+  use_react_native!(
```
WoLewicki, line 9: Are all those changes of `example` project necessary in this PR? They seem to not concern it.

WoLewicki (commented):
Very good job, keep it going! I left some comments, please answer them :tada:

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Events sent by `onHandlerUpdate` are handled by the `onUpdate` callback of a gesture.

android/src/main/AndroidManifest.xml
```diff
@@ -1,3 +1,3 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.swmansion.gesturehandler.react">
+    package="com.swmansion.gesturehandler">
```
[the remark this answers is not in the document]
j-piasecki: IIRC it's due to a bug in codegen where it ignores the set package name. Because of that the `RNGestureHandlerPackage` was moved from `com.swmansion.gesturehandler.react` to `com.swmansion.gesturehandler`. This change makes it work on Paper.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
```diff
@@ -34,7 +34,7 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
     private set
   var isEnabled = true
     private set
-  var usesDeviceEvents = false
+  var actionType = 0
```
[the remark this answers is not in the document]
j-piasecki: On Paper, we used direct events for Animated & Reanimated and device events for JS callbacks. On Fabric, we needed more granular control over them:
- events for Reanimated (for now only on Paper using direct events, will need special treatment on Fabric)
- events for Animated with nativeDriver=true (direct events)
- events for Animated with nativeDriver=false (direct events on Paper, device events on Fabric)
- events for JS callbacks (device events)

android/src/paper/java/com/swmansion/gesturehandler/ReactContextExtensions.kt
j-piasecki: Should we be concerned about that? `UIManagerModule` shouldn't be `null` if everything is working correctly, no? But if that's the case, wouldn't `?.` work as well?

src/getShadowNodeFromRef.web.ts
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/a94c858205b10551ee7e2383f0bcf55d3d73f81f.

ios/RNGestureHandlerButtonComponentView.mm
j-piasecki: It implements `RCTRNGestureHandlerButtonViewProtocol` from codegen. This way it doesn't interfere with compilation on Paper where it's not used.

src/jestUtils.ts
j-piasecki: Not really, removed in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/af2e38b5d6a3996362bcbd9065fb8f54b2bdb894 in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/1895.

j-piasecki: Not really, removed in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/af2e38b5d6a3996362bcbd9065fb8f54b2bdb894 in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/1895.

src/utils.ts
j-piasecki: No, removed in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/1f7023072ff5d196ce6c479b44f67b6097e4fae8 in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/1896.

j-piasecki: No, removed in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/1f7023072ff5d196ce6c479b44f67b6097e4fae8 in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/1896.

FabricExample/src/ComponentsScreen.tsx
```diff
@@ -0,0 +1,193 @@
...
+  ScrollView,
+  Switch,
+  TextInput,
+  TouchableNativeFeedback,
+  TouchableOpacity,
+} from 'react-native-gesture-handler';
+import { StyleSheet, Text, View } from 'react-native';
+
+import { COLORS } from './colors';
+import React from 'react';
```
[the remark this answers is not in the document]
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/850d510e768f3d06c92b65f42e7d42684066c1e6.

FabricExample/src/FinalScreen.tsx
```diff
@@ -0,0 +1,55 @@
+import { Image, StyleSheet, Text, View } from 'react-native';
+
+import { COLORS } from './colors';
+import React from 'react';
```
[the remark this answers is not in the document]
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/850d510e768f3d06c92b65f42e7d42684066c1e6.

FabricExample/src/GestureCompositionScreen.tsx
```diff
@@ -0,0 +1,120 @@
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
+import { StyleSheet, Text, View } from 'react-native';
+
+import { COLORS } from './colors';
+import React from 'react';
```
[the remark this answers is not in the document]
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/850d510e768f3d06c92b65f42e7d42684066c1e6.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
tomekzaw: **1) Reanimated**

In the old architecture, Reanimated overwrites RCTEventDispatcher with REAEventDispatcher to intercept all direct events (see [here](https://github.com/software-mansion/react-native-reanimated/blob/a870aa28092322b627695f8cf2ea0dce4db34a53/ios/REAEventDispatcher.m#L8-L12)). This allows Gesture Handler to integrate with Reanimated without defining a hard dependency, simply by sending regular direct events.

In the new architecture, the event system was mainly ported to C++ so we can no longer overwrite RCTEventDispatcher. Instead, we need to dispatch events directly to Reanimated internals.

One more thing to note here is that in Reanimated for Paper there is no feedback if Reanimated did or did not handle specific event, so effectively they are handled twice (once by Reanimated and once by React Native). We plan to address this issue simultaneously with the adoption of Fabric.

**2) Animated.event with useNativeDriver: true**

Currently, `NativeAnimated[Turbo]Module` is a `RCTEventDispatcherObserver` (see [here](https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/React/CoreModules/RCTEventDispatcher.mm#L117-L119)). Basically, it listens to all events going through `RCTEventDispatcher` but cannot prevent them from being sent to JS side. That's why in such case, similarly as for Reanimated, we would like to invoke NativeAnimated[Turbo]Module directly instead of flooding the app with unnecessary events.

**3) Direct vs. device events**

On Paper, the old Gesture Handler API (e.g. `PanGestureHandler`) uses direct events in order to invoke JS callbacks from the native side. When designing API for Gesture Handler 2.0 it turned out to be a limitation, so the new API (i.e. `GestureDetector`) uses device events instead.

With Fabric, at some point we decided to use only device events, both for old and new API. That's because I've already managed to get device events working in Gesture Handler but also it was kinda unclear how to send direct events from Obj-C. It looks like for this purpose we need to use C++ EventEmitter that we can access from the provided ShadowNode. However, currently Gesture Handler is set up as a NativeModule (in Java/Obj-C) so we would need to migrate it first to C++ and use JSI in order to pass ShadowNodes from JS.

Effectively, using direct or device events should be no difference for the library user as this seems to be an internal implementation detail of our library. Nevertheless, we will research into ways of sending direct events like we used to do on Paper.

android/src/main/jni/RNGestureHandlerComponentsRegistry.cpp
tomekzaw: Done in 786902cbba0fd1278f16de76130d53365fd08cb7

example/ios/Podfile
tomekzaw: These changes restore the original Podfile from RN 0.67.2 template so we can build the app on our M1s. Will move them to a separate PR.

ios/RNGestureHandlerButtonComponentView.mm
tomekzaw: Another option would be to manually exclude these files in Podspec, either by using `*.{h,m}'` pattern for `source_files` or alternatively via `exclude_files` property, but we prefer the preprocessor conditionals.

RNGestureHandler.podspec
```diff
@@ -1,5 +1,12 @@
 require "json"
 
+fabric_enabled = ENV['RCT_NEW_ARCH_ENABLED']
+
+# folly_version must match the version used in React Native
+# See folly_version in react-native/React/FBReactNativeSpec/FBReactNativeSpec.podspec
+folly_version = '2021.06.28.00-v2'
```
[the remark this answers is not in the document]
tomekzaw: When it comes to configuration, we aimed to be compliant with react-native-screens, but your suggestion sounds like a good idea. We can also avoid overwriting `s.platforms`. Done in b745bbe68c06b1a789e6199ff41ad0619d153e9d. Would you accept an analogous PR to RNScreens?

FabricExample/android/settings.gradle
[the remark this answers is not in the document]
tomekzaw: FabricExample is meant to be used only with Fabric enabled so that makes no difference at all here. Let's wait until RN teams publishes an official tutorial to see what's the recommended configuration.

android/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
tomekzaw: To be precise, it looks like currently `codegenJavaPackageName` in `build.gradle` is never used. In particular, it does not seem to affect the package name of `RNGestureHandlerPackage` import when generating `PackageList.java` which is a part of auto-linking mechanism.

RNGestureHandler.podspec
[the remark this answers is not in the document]
WoLewicki: We are always open for PRs in RNScreens 😸

android/src/paper/java/com/swmansion/gesturehandler/ReactContextExtensions.kt
WoLewicki: I have no knowledge if it should be present there, but I think it should be handled in a better way than `!!`, by e.g. throwing an exception with descriptive message or something like that.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
WoLewicki: Maybe it would be nice to add this e.g. to the description of PR so it does not get lost?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
WoLewicki: So maybe it would be good to extend this comment a bit so it is more descriptive?

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandler.kt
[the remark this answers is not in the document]
WoLewicki: Since @tomekzaw gave more detailed description of this below, I think we can resolve this.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Good point, done in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/e51489fd2dc62a5e88f94fe0017bfec99081fc8a.

FabricExample/ios/FabricExample.xcodeproj/project.pbxproj
```diff
@@ -566,7 +566,7 @@
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "";
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = i386;
```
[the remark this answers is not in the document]
j-piasecki: I believe it was changes because of problems with building the app on M1, but it seems to be working correctly without it. Removed in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/5fee6b1760d9c9f060f45b26149ca335a36fc8f7.

FabricExample/ios/FabricExample/AppDelegate.mm
```diff
@@ -4,6 +4,8 @@
 #import <React/RCTBridge.h>
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTRootView.h>
+#import "MBFingerTipWindow.h"
```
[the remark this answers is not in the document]
j-piasecki: For now I've removed `MBFingerTipWindow` entirely as it was crashing the app in some cases. Will look into this in the future.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
j-piasecki: Yes, `android.R` was used before migration to Kotlin (https://github.com/software-mansion/react-native-gesture-handler/commit/48468306b48207b545b47f976dcdb9f69e43e546#diff-e614cdf726640239cd9b5e551cfc1c50eb1d6f078b8c6884bfa4fb321ed82e4dL209), I'm not sure why it was changed then.

android/src/paper/java/com/swmansion/gesturehandler/ReactContextExtensions.kt
j-piasecki: I wrapped it with a try-catch block in https://github.com/software-mansion/react-native-gesture-handler/pull/1893/commits/589e641049566b13c7aafc72d9a7fda4bec0306d.

android/src/main/java/com/swmansion/gesturehandler/RNGestureHandlerPackage.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.facebook.soloader.SoLoader
+import com.swmansion.gesturehandler.BuildConfig
+import com.swmansion.gesturehandler.react.RNGestureHandlerModule
+import com.swmansion.gesturehandler.react.RNGestureHandlerRootViewManager
+import com.swmansion.gesturehandler.react.RNGestureHandlerButtonViewManager
+
+class RNGestureHandlerPackage : ReactPackage {
+  override fun createNativeModules(reactContext: ReactApplicationContext): List<NativeModule> {
+    if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
+      // For Fabric, we load c++ native library here, this triggers screen's Fabric
```
kmagiera, line 16: "screens's"?

android/src/main/jni/OnLoad.cpp
```diff
@@ -0,0 +1,12 @@
+#include <fbjni/fbjni.h>
+
+#include "RNGestureHandlerComponentsRegistry.h"
+
+JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *) {
+  return facebook::jni::initialize(vm, [] {
+    // TODO: dvacca ramanpreet unify this with the way
```
kmagiera, line 7: why?

ios/RNGestureHandler.h
```diff
@@ -60,7 +54,7 @@ if (value != nil) { recognizer.prop = [RCTConvert type:value]; }\
 @property (nonatomic, readonly, nullable) UIGestureRecognizer *recognizer;
 @property (nonatomic, readonly, nullable) RNGestureHandlerPointerTracker *pointerTracker;
 @property (nonatomic) BOOL enabled;
-@property (nonatomic) BOOL usesDeviceEvents;
+@property (nonatomic, nonnull) NSNumber *actionType;
```
kmagiera, line 57: should be of `RNGestureHandlerActionType` type no?

ios/RNGestureHandlerActionType.h
```diff
@@ -0,0 +1,8 @@
+#import <Foundation/Foundation.h>
+
+typedef NS_ENUM(NSInteger, RNGestureHandlerActionType) {
```
kmagiera, line 3: I think it'd be ok to move it to GestureHandler.h

ios/RNGestureHandlerManager.mm
```diff
@@ -196,24 +210,82 @@ - (void)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
 
 #pragma mark Events
 
-- (void)sendTouchEvent:(RNGestureHandlerEvent *)event
+- (void)sendStateChangeEvent:(RNGestureHandlerStateChange *)event withActionType:(nonnull NSNumber *)actionType
 {
-    [_eventDispatcher sendEvent:event];
+    switch ([actionType integerValue]) {
```
kmagiera, line 215: should just be `actionType` here and there is no need to use integerValue (I think)

android/src/main/java/com/swmansion/gesturehandler/RNGestureHandlerPackage.kt
tomekzaw: Fixed in fe98498f3c9cdffd72a3b5803f82fe328bc33510

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -409,6 +417,21 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?)
     }
   }
 
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  fun install(): Boolean {
+    return try {
+      System.loadLibrary("rngesturehandler_modules")
```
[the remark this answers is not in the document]
tomekzaw: Done in f92834dc797947b31b89139ad4e6553443b7c184

android/src/main/jni/OnLoad.cpp
tomekzaw: Removed in 3f76094bf94748de6851c4cdde1db9674009024d

ios/RNGestureHandlerActionType.h
tomekzaw: We did it analogously to `RNGestureHandlerState.h`.

ios/RNGestureHandlerManager.mm
tomekzaw: Done in 3c381aa2075734797c5a997d6708843494d93dce

example/ios/Podfile
tomekzaw: Done in 58f752099ae1bbf53377258b61db4e9a8bd42a26, in favor of #1903

kmagiera (approved):

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
rashmi-priya: > However, currently Gesture Handler is set up as a NativeModule (in Java/Obj-C) so we would need to migrate it first to C++ and use JSI in order to pass ShadowNodes from JS.

> Why is Gesture Handler NativeModule (in Java/Obj-C) was not changed to TurboModule?

tomekzaw: Hey @rashmi-priya! The reason why we didn't migrate RNGH NativeModule to a TurboModule is the fact that NativeModules still work on the new architecture. The goal of the migration was to add Fabric support to RNGH 2.x, so the package itself supports all versions of RN 0.63.0+ as well as the old and the new architecture.

TurboModules are backward-compatible (meaning they work as NativeModules on the old architecture) but we couldn't just rewrite NativeModule into a TurboModule because it would not compile on older versions of RN (e.g. 0.64).

Another option was to conditionally enable NativeModule on Paper and TurboModule on Fabric. However, in this variant we would duplicate code and the secondary goal for this PR was to introduce as few changes as possible.

We decided to leave NativeModule as it is and address this issue in the future. Maybe now it's a good time to do so. 😄 

I hope this answers your questions.

rashmi-priya: Thanks for the clarification.

PR #1903, opened 2022-03-01
Restore original Podfile in example app

tomekzaw:
## Description

This PR restores the original Podfile from React Native 0.67.2 in the Example app, so we can build it on Apple M1.

## Test plan

1. Install pods
2. Run example app


j-piasecki (approved):
Builds on M1 as well as on Intel 👍

PR #1909, opened 2022-03-02
Rename `EventType` to `TouchEventType` to clear up the naming

j-piasecki:
## Description

Rename `EventType.ts` file to `TouchEventType.ts` and `EventType` to `TouchEventType` in code.

## Test plan

Ran `tsc`

### DIFF
--- a/src/handlers/gestureHandlerCommon.ts
+++ b/src/handlers/gestureHandlerCommon.ts
@@ -6,7 +6,7 @@ import * as React from 'react';
 import { Platform, findNodeHandle as findNodeHandleRN } from 'react-native';
 
 import { State } from '../State';
-import { EventType } from '../EventType';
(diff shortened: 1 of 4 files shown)


src/handlers/gestureHandlerCommon.ts
```diff
@@ -83,7 +83,7 @@ export type GestureTouchEvent = {
   handlerTag: number;
   numberOfTouches: number;
   state: ValueOf<typeof State>;
-  eventType: EventType;
+  eventType: TouchEventType;
```
tomekzaw, line 86: I guess we cannot rename this field to `touchEventType` because it would be a brekaing change in API, is that true?

j-piasecki: Correct.

tomekzaw (approved):
👍

PR #1910, opened 2022-03-02
fix missing testID when createHandler props do not have one assigned

virdesai:
## Description

Fix related to my comment [here](https://github.com/software-mansion/react-native-gesture-handler/pull/1844#issuecomment-1057400208) about `RNGestureHandlerButton` with `handlerType="NativeViewGestureHandler"` having missing testIDs when they were previously provided and assigned. I believe the testID on the `child` being cloned within `createHandler` was being removed by the testID being set within the render function when one was not even passed in. Now it will only assign a testID as a prop on the cloned element if a testID was provided to the handler

### DIFF
--- a/src/handlers/createHandler.ts
+++ b/src/handlers/createHandler.ts
@@ -494,7 +494,7 @@ export default function createHandler<
                 handlerTag: this.handlerTag,
               }
             : {}),
-          ...(this.props.testID ? { testID: this.props.testID } : {}),
+          testID: this.props.testID ?? child.props.testID,
           ...events,
         },
         grandChildren
--- a/src/handlers/createHandler.ts
+++ b/src/handlers/createHandler.ts
@@ -494,7 +494,7 @@ export default function createHandler<
                 handlerTag: this.handlerTag,
               }
             : {}),
-          testID: this.props.testID,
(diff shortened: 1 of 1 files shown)


src/handlers/createHandler.ts
```diff
@@ -494,7 +494,7 @@ export default function createHandler<
                 handlerTag: this.handlerTag,
               }
             : {}),
-          testID: this.props.testID,
+          ...(this.props.testID ? { testID: this.props.testID } : {}),
```
j-piasecki, line 497:
```suggestion
          testID: this.props.testID ?? child.props.testID,
```
HI! Thank you for taking the time to look into this ❤️ . I would really appreciate it if you could change this line to the other one you mentioned in your comment. The `??` operator returns the right value only when the left value is `null` on `undefined`, not any falsy value. While I don't think it will cause problems, I would rather be explicit about this.

virdesai: of course!

j-piasecki:
Thank you ❤️ !

vishalnarkhede:
Thanks for fixing this @virdesai :)

virdesai:
Lol no problem @vishalnarkhede

Issue #1913, opened 2022-03-03
[2.3.0] Cannot build on android


gabrielbull:
## Description

Version 2.3.0 does not build on android. Looks like it's importing code from the `ansi-styles` package which is a NodeJS package not compatible with Hermes.

### Platforms

- [ ] iOS
- [x] Android
- [ ] Web

## Steps To Reproduce

1. `./gradlew bundleRelease`

### Expected behavior

App builds

### Actual behavior

```
info Writing bundle output to:, /Users/gabriel/Code/pom/fom-native/android/app/build/generated/assets/react/development/release/index.android.bundle
info Writing sourcemap output to:, /Users/gabriel/Code/pom/fom-native/android/app/build/intermediates/sourcemaps/react/development/release/index.android.bundle.packager.map
info Done writing bundle output
info Done writing sourcemap output
/Users/gabriel/Code/pom/fom-native/android/app/build/generated/assets/react/development/release/index.android.bundle:71915:25: error: Invalid regular expression: Quantifier has nothing to repeat
          var matches = /(?<colorString>[a-f\d]{6}|[a-f\d]{3})/i.exec(hex.toString(16));
```

## Snack or minimal code example

```js
/**
 * @format
 */

import 'react-native-gesture-handler';
import { AppRegistry } from 'react-native';
import { name as appName } from './app.json';

const App = () => null;

AppRegistry.registerComponent(appName, () => App);
```

## Package versions

- React: 17.0.2
- React Native: 0.67.2
- React Native Gesture Handler: 2.3.0
- React Native Reanimated: 2.3.2

j-piasecki:
Hi! Unfortunately, I haven't been able to reproduce it on a fresh App. I managed to get the same error in our Example app in release mode, but it was due to the `devDependencies` of Gesture Handler also being bundled in this configuration. Installing dependencies for `react-native-gesture-handler` using `yarn --production --no-lockfile` solved it in that case.

If that doesn't help you, would you mind posting the output of `npm ls ansi-styles` and `yarn why ansi-styles`? Do you use any special configuration in your project (like having Gesture Handler source locally and linked to your project)?
And of course, a repository with a reproduction of this issue would be greatly appreciated.

beetlebum:
We have the same issue in our app, `0.67.3`, Reanimated `2.2.4`

```
├─┬ @babel/core@7.17.5
│ └─┬ @babel/code-frame@7.16.7
[...]

PR #1917, opened 2022-03-07
Remove `jestUtils` from the main `index.ts` and move it to separate `package.json`

j-piasecki:
## Description

Moves `jestUtils` to a separate package so it's not bundled when not actually used. This caused problems with `testing-library`, namely one of its dependencies - `ansi-styles`. It uses named capture groups in regex which is unsupported on Hermes.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/1913.

## Test plan

Run a new app in release configuration with and without `testing-library` installed.

### DIFF
--- a/docs/versioned_docs/version-2.3.0/api/test-api.md
+++ b/docs/versioned_docs/version-2.3.0/api/test-api.md
@@ -3,6 +3,10 @@ id: test-api
 title: Testing
 ---
 
+:::info
+If you want to use `fireGestureHandler` and `getByGestureTestId`, you need to import them from `react-native-gesture-handler/jest-utils` package.
+:::
+
 ## fireGestureHandler(gestureOrHandler, eventList)
 
 Simulates one event stream (i.e. event sequence starting with `BEGIN` state and ending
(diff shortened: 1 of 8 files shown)


docs/versioned_docs/version-2.3.0/api/test-api.md
```diff
@@ -3,6 +3,10 @@ id: test-api
 title: Testing
 ---
 
+:::info
+If you want to use `fireGestureHandler` and `getByGestureTestId`, you need to import them from `react-native-gesture-handler/jest-utils` package.
+:::
```
jakub-gonet, line 8: Change the newest md as well, please.

jakub-gonet (commented):
LGTM

docs/versioned_docs/version-2.3.0/api/test-api.md
j-piasecki: For now I have moved it to the newest one in https://github.com/software-mansion/react-native-gesture-handler/pull/1917/commits/7326e847e7d987e550d823273f3d615b2ae89535. It's kinda breaking change and I'm not sure whether we should release it as a patch.

PR #1925, opened 2022-03-11
Expand example code for PanGestureHandler

RWOverdijk:
## Description

I found it hard to get this example to work. It took me a while to figure out that I needed `GestureHandlerRootView` because it otherwise simply wouldn't do anything on android.

I also had to hunt down the needed imports.

Lastly, I made it a copy-paste solution that just works when dropped into App.tsx which is something I personally really like about example code.


j-piasecki (approved):
Thank you ❤️ !

Issue #1797, opened 2021-12-31
GestureDetector not detecting gestures or not firing any events


ernestasgobionis:
## Description

I'm having trouble with getting a gesture handler to work. I'm using the new API (2.0) for gestures, so I'm using GestureDetector with a Gesture.Pan(), and I can't make any events fire for the gesture. I'm using [Wix RN Navigation](https://github.com/wix/react-native-navigation) library, so this might be a cause of the issue?

## Steps To Reproduce

1. Register the screen component wrapping it with the provided HOC:
`Navigation.registerComponent(SCREENS.LOGIN.ROOT, () => gestureHandlerRootHOC(LoginScreen));`

2. The screen itself includes just one functional component which is present in the examples folder of the library:

`import React from 'react';
import { Button, StyleSheet, View } from 'react-native';
import { Gesture, GestureDetector } from 'react-native-gesture-handler';
import Animated, { useAnimatedStyle, useSharedValue, withSpring } from 'react-native-reanimated';

function Ball() {
    const isPressed = useSharedValue(false);
    const offset = useSharedValue({ x: 0, y: 0 });
    const start = useSharedValue({ x: 0, y: 0 });

    const animatedStyles = useAnimatedStyle(() => {
        console.log('ANIMATED STYLES --- ', offset.value.x, offset.value.y, isPressed.value);
        return {
            transform: [
                { translateX: withSpring(offset.value.x) },
                { translateY: withSpring(offset.value.y) },
                { scale: withSpring(isPressed.value ? 1.2 : 1) },
            ],
            backgroundColor: isPressed.value ? 'yellow' : 'blue',
        };
    });

    const gesture = Gesture.Pan()
        .onBegin(() => {
            'worklet';
            console.log('ONBEGIN');
            isPressed.value = true;
        })
        .onUpdate((e) => {
            'worklet';
            console.log('ONUPDATE');
            offset.value = {
                x: e.translationX + start.value.x,
                y: e.translationY + start.value.y,
            };
        })
        .onEnd(() => {
            'worklet';
            start.value = {
                x: offset.value.x,
                y: offset.value.y,
            };
        })
        .onFinalize(() => {
            isPressed.value = false;
        });

    return (
        <>
            <GestureDetector gesture={gesture}>
                <Animated.View style={[styles.ball, animatedStyles]} />
            </GestureDetector>
            <Button
                title="Test"
                onPress={() => {
                    isPressed.value = !isPressed.value;
[...]

PR #1929, opened 2022-03-14
Use device events when remote debugging is enabled

j-piasecki:
## Description

Don't rely on Reanimated for receiving events when using remote debugging, as it's not supported by Reanimated 2.

Should fix https://github.com/software-mansion/react-native-gesture-handler/issues/1797.

## Test plan

Run an app with Chrome remote debugging enabled and disabled.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -46,5 +46,6 @@ export function isFabric(): boolean {
 
 export function isRemoteDebuggingEnabled(): boolean {
   // react-native-reanimated checks if in remote debugging in the same way
+  // @ts-ignore global is available but node types are not included
   return !(global as any).nativeCallSyncHook || (global as any).__REMOTEDEV__;
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -45,5 +45,6 @@ export function isFabric(): boolean {
 }
 
 export function isRemoteDebuggingEnabled(): boolean {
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -43,3 +43,14 @@ export function isFabric(): boolean {
...
+export function isRemoteDebuggingEnabled(): boolean {
+  return (
+    // @ts-ignore WorkerGlobalScope available only when remote debugging
+    'undefined' !== typeof WorkerGlobalScope &&
+    // @ts-ignore importScripts available only when remote debugging
+    'function' === typeof importScripts &&
+    // @ts-ignore WorkerNavigator available only when remote debugging
+    navigator instanceof WorkerNavigator
+  );
+}
```
tomekzaw, line 56: Maybe let's do it the same way as in Reanimated? See [`isChromeDebugger`](https://github.com/software-mansion/react-native-reanimated/blob/688a244b3826b0f6cde52e9490eee5fe0d3e9bf9/src/reanimated2/PlatformChecker.ts#L7-L9)

tomekzaw (changes requested):

src/utils.ts
j-piasecki: Done in [b51acf5](https://github.com/software-mansion/react-native-gesture-handler/pull/1929/commits/b51acf592ae4669ecacff879207c0a5a94b865e9).

PR #1931, opened 2022-03-15
fixed a typo in example url

ghorbani-m:


j-piasecki (approved):
Thanks for catching it!

Issue #1869, opened 2022-02-07
Gesture Detector does not work in Expo Web


Karthik-B-06:
## Description
I tried to use the same example given [here](https://docs.swmansion.com/react-native-gesture-handler/docs/api/gestures/pan-gesture/). 

Works fine in iOS. But it does not work in Expo Web. 

### Platforms

- [x] iOS
- [ ] Android
- [x] Web

### Screenshots
iOS Recording
https://user-images.githubusercontent.com/35562287/152832813-1eb96324-f223-44f9-9c4c-41afc88dce7b.mov

## Steps To Reproduce

1. Create a new Expo App
2.  Add react-native gesture handler and try this example code available here. 

### Expected behavior

The Pan gesture should work across all Platforms.

### Actual behavior
Works well in iOS and Android but does not work on the Web.

## Snack or minimal code example

```
  const END_POSITION = 200;
  const onLeft = useSharedValue(true);
  const position = useSharedValue(0);

  const panGesture = Gesture.Pan()
    .onUpdate((e) => {
      if (onLeft.value) {
        position.value = e.translationX;
      } else {
        position.value = END_POSITION + e.translationX;
      }
    })
    .onEnd(() => {
      if (position.value > END_POSITION / 2) {
        position.value = withTiming(END_POSITION, { duration: 100 });
        onLeft.value = false;
      } else {
        position.value = withTiming(0, { duration: 100 });
        onLeft.value = true;
      }
    });

  const animatedStyle = useAnimatedStyle(() => ({
    transform: [{ translateX: position.value }],
  }));

<GestureDetector gesture={panGesture}>
    <AnimatedBox
      accessible
      accessibilityHint={'Hello World'}
      style={[tailwind.style('h-10 w-10 bg-red-300'), animatedStyle]}
    />
</GestureDetector>

```

## Package versions

- React:  "^17.0.2",
- React Native: "0.64.3",
- React Native Gesture Handler: "2.1.0"
- React Native Reanimated: "^2.2.0"

Issue #1840, opened 2022-01-24
GestureDetector not working on the web


thespacemanatee:
## Description
When I try to use `GestureDetector` on the web, I get the following error: `Uncaught TypeError: Cannot read properties of undefined (reading 'current')`

### Platforms

- [ ] iOS
- [ ] Android
- [x] Web

### Screenshots
<img width="1091" alt="Screenshot 2022-01-24 at 10 04 05 PM" src="https://user-images.githubusercontent.com/6837599/150797123-c64bb612-e996-4a28-a56e-bb21c0b8aa94.png">

## Steps To Reproduce

1. Try to use `GestureDetector` on the web
2. Start swiping
3. `Uncaught TypeError: Cannot read properties of undefined (reading 'current')`

### Expected behavior
Events should fire as per normal

### Actual behavior
Throws `Uncaught TypeError: Cannot read properties of undefined (reading 'current')` on the web

## Snack or minimal code example
```js
const gesture = Gesture.Pan()
  .onBegin(() => {
    offset.value.x = translateX.value;
  })
  .onUpdate(({ translationX }) => {
    const newX = offset.value.x + translationX;
    translateX.value = newX;
    rotate.value = newX / 10;
  })
  .onEnd(({ velocityX }) => {
    const dest = snapPoint(translateX.value, velocityX, snapPoints);
    translateX.value = withSpring(dest, { velocity: velocityX });
    rotate.value = withTiming(0, {}, () => {
      if (dest > 0) {
        runOnJS(onSwipeRight)();
      } else if (dest < 0) {
        runOnJS(onSwipeLeft)();
      }
    });
  })
  .enabled(enabled);

<View style={styles.container} pointerEvents="box-none">
  {Platform.OS === "web" ? ( // have to use the old PanGestureHandler for web
    <PanGestureHandler onGestureEvent={onGestureEvent} enabled={enabled}>
      <Animated.View style={style}>
        <TaskCard source={source} taskIndex={taskIndex} />
      </Animated.View>
    </PanGestureHandler>
  ) : (
    <GestureDetector gesture={gesture}>
      <Animated.View style={style}>
        <TaskCard source={source} taskIndex={taskIndex} />
      </Animated.View>
    </GestureDetector>
  )}
</View>
```

## Package versions

- React: `17.0.1`
- React Native: `0.64.3`
- React Native Gesture Handler: `^2.2.0`
- React Native Reanimated: `~2.3.1`

[...]

PR #1934, opened 2022-03-16
Basic web support for `GestureDetector`

j-piasecki:
## Description

Adds basic web support for `GestureDetector` using the existing web implementation.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/1869 and https://github.com/software-mansion/react-native-gesture-handler/issues/1840.

## Test plan

Tested on the Reanimated 2 playground app with updated Gesture Handler

### DIFF
--- a/src/handlers/gestures/GestureDetector.tsx
+++ b/src/handlers/gestures/GestureDetector.tsx
@@ -16,6 +16,7 @@ import {
   GestureTouchEvent,
   GestureUpdateEvent,
   GestureStateChangeEvent,
+  HandlerStateChangeEvent,
 } from '../gestureHandlerCommon';
 import {
   GestureStateManager,
@@ -35,6 +36,9 @@ import { ComposedGesture } from './gestureComposition';
 import { ActionType } from '../../ActionType';
 import { isFabric, tagMessage } from '../../utils';
 import { getShadowNodeFromRef } from '../../getShadowNodeFromRef';
+import { Platform } from 'react-native';
+import RNGestureHandlerModuleWeb from '../../RNGestureHandlerModule.web';
+import { onGestureHandlerEvent } from './eventReceiver';
 
 declare global {
   function isFormsStackingContext(node: unknown): boolean | null; // JSI function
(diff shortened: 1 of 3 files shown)


src/handlers/gestures/GestureDetector.tsx
```diff
@@ -35,6 +36,9 @@ import { ComposedGesture } from './gestureComposition';
 import { ActionType } from '../../ActionType';
 import { isFabric, tagMessage } from '../../utils';
 import { getShadowNodeFromRef } from '../../getShadowNodeFromRef';
+import { Platform } from 'react-native';
+import RNGestureHandlerModuleWeb from '../../RNGestureHandlerModule.web';
```
kmagiera, line 40: I'd prefer to avoid importing like this. In general `.web.ts` files should never be imported manually, as it is expected that they are specific for web and therefore can bring a ton of web-only dependencies (e.g. hammer.js) with them. I think it'd be best to have this abstracted in `RNGestureHandlerModule` directly (then the module would have separate impl for mobile/web)

src/web/GestureHandler.ts
```diff
@@ -182,7 +182,12 @@ abstract class GestureHandler {
         // onHandlerStateChange only
         handlerTag: this.handlerTag,
         target: this.ref,
-        oldState: this.oldState,
+        // send oldState only when the state was changed, or is different than ACTIVE
+        // GestureDetector relies on the presence of `oldState` to differentiate between
+        // update events and state change events
+        ...(state !== this.previousState || state != 4
```
kmagiera, line 188: can't we just make `oldState: undefined` if we don't want to include it? It'd look cleaner then destructing empty object imo

kmagiera (approved):
Left few inline comments, please address these prior to merging. Looks good otherwise 👍

src/web/GestureHandler.ts
j-piasecki: Done in [c81b618](https://github.com/software-mansion/react-native-gesture-handler/pull/1934/commits/c81b6188d58362283437b00c594b6d085df1314a)

src/handlers/gestures/GestureDetector.tsx
j-piasecki: This comes from the fact that the `RNGestureHandlerModule.attachGestureHandler()` has a different signature on web - it takes one more argument containing the ref with the callbacks. I don't see how it can be easily abstracted, other than adding an argument that will be ignored on native platforms.

I changed it in [07e620a](https://github.com/software-mansion/react-native-gesture-handler/pull/1934/commits/07e620ab2151c9d5a260dab0b24348dc42113658) to only import type (the same way it's already implemented in the `createHandler.tsx`). Is that enough or should I do it differently?

jonstuebe:
Works great for me! Can't wait for this to get merged

PR #1938, opened 2022-03-17
Add `activateAfterLongPress` modifier to Pan gesture

j-piasecki:
## Description

Since the release of Gesture Handler 2.0, it's been relatively easy to implement `LongPress` before `Pan` using `manualActivation` and touch events, however this approach adds a lot of unnecessary boilerplate code and may not seem obvious when tackling this problem.
This PR adds `activteAfterLongPress` modifier to `Pan` gesture, making it very easy to make `Pan` gesture activate only after `LongPress` of specified duration.

## Test plan

Tested on the Example app - added the modifier on existing examples utilizing `Pan` to see if they behave correctly.

### DIFF
--- a/android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.kt
+++ b/android/lib/src/main/java/com/swmansion/gesturehandler/PanGestureHandler.kt
@@ -1,6 +1,7 @@
 package com.swmansion.gesturehandler
 
 import android.content.Context
+import android.os.Handler
 import android.view.MotionEvent
 import android.view.VelocityTracker
 import android.view.ViewConfiguration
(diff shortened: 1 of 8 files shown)


kmagiera (approved):
Looks good 👍

TheFrostage:
@j-piasecki Hello Jakub! Thank you for this addition to this awesome package. 
Could you please help me with one thing real quick? In my use case I have a big (2000x2000) view that has circles positioned on top of it. I've implemented a pan gesture to move around my canvas and added pinch. Now I need to add a new layer to it. I need to be able to long press a circle and start changing its translation on the canvas. I don't think a second `Gesture.Pan()` with `activateAfterLongPress` would work. I think I need to re-write my pan with touch events & manual activation, but I can't find any good examples of this boilerplate that I'll need. 

**Can you help me with a piece of advice or maybe you could help me find some examples? I think my use-case is not super uncommon.**

j-piasecki:
@TheFrostage If I understood you correctly, [this discussion](https://github.com/software-mansion/react-native-gesture-handler/discussions/1826) may be helpful for you.

Issue #1883, opened 2022-02-18
NSInvalidArgumentException Crash


LeviWilliams:
## Description
App is crashing from RNGestureHandlerEvents.m while using gesture detector 

### Platforms

- [ ] iOS

## Stacktrace
Hardware Model:     iPad7,5 (Also various other models of iPhone)
Role:               Foreground
OS Version:         iOS 15.3.1

NSInvalidArgumentExceptionRNGestureHandlerEvents.m:248
*** -[__NSDictionaryM setObject:forKey:]: object cannot be nil (key: target)

0  CoreFoundation          ___exceptionPreprocess
1  libobjc.A.dylib         _objc_exception_throw
2  CoreFoundation          __CFThrowFormattedException
3  CoreFoundation          -[__NSDictionaryM setObject:forKey:].cold.2
4  CoreFoundation          -[__NSDictionaryM setObject:forKey:]
5  app -[RNGestureHandlerStateChange arguments] (RNGestureHandlerEvents.m:248:5)
6  app -[RNGestureHandlerManager sendStateChangeDeviceEvent:] (RNGestureHandlerManager.m:217:34)
7  app -[RNGestureHandler sendStateChangeEvent:] (RNGestureHandler.m)
8  app -[RNGestureHandler sendEventsInState:forViewWithTag:withExtraData:] (RNGestureHandler.m:223:9)
9  app -[RNGestureHandler handleGesture:] (RNGestureHandler.m:194:5)
10 app -[RNBetterTapGestureRecognizer triggerAction] (RNTapHandler.m:64:3)
11 app -[RNBetterTapGestureRecognizer reset] (RNTapHandler.m:175:5)
12 UIKitCore               -[UIGestureRecognizer _resetGestureRecognizer]
13 UIKitCore               ____UIGestureEnvironmentUpdateGestureGraph_block_invoke_2
14 UIKitCore               __UIGestureEnvironmentUpdate
15 CoreFoundation          ___CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__
16 CoreFoundation          ___CFRunLoopDoObservers
17 CoreFoundation          ___CFRunLoopRun
18 CoreFoundation          _CFRunLoopRunSpecific
19 GraphicsServices        _GSEventRunModal
20 UIKitCore               -[UIApplication _run]
21 UIKitCore               _UIApplicationMain
22 app main (main.m:14:12)

[...]

PR #1946, opened 2022-03-23
Don't send events on iOS when target is `nil`

j-piasecki:
## Description

Should fix https://github.com/software-mansion/react-native-gesture-handler/issues/1883.

I managed to (kind of and by accident) reproduce the crash mentioned in the issue when I was working on https://github.com/software-mansion/react-native-gesture-handler/pull/1912. After commenting out [those lines](https://github.com/software-mansion/react-native-gesture-handler/pull/1912/files#diff-0c167f367b6ea1c2605e4c3ec7d9ea7edd077697285273e4e5e5721c9248591cR22-R23), I tapped the view that had a `Pan` gesture attached with `manualActivation` set to true. After that going to a different screen caused the crash.

The issue comes from the fact that some gestures try to send events in their `reset` method to properly close the state flow, however this method is also called when the view is unmounted and all gesture recognizers get unbound from the view. This PR adds a simple check to see if the `reactTag` property is `nil` or not before sending the event. Events are send only when their target is not `nil` which prevents the crash.

## Test plan

Tested on the Example app, modified as explained above.

### DIFF
--- a/ios/RNGestureHandler.m
+++ b/ios/RNGestureHandler.m
@@ -189,6 +189,13 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer
 {
+    // it may happen that the gesture recognizer is reset after it's been unbound from the view,
+    // it that recognizer tried to send event, the app would crash because the target of the event
+    // would be nil.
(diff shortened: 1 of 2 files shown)


tomekzaw (approved):

LeviWilliams:
Awesome, thank you!!

PR #1952, opened 2022-03-24
Rewrite codegen config with TypeScript

j-piasecki:
## Description

Rewrite `RNGestureHandlerRootViewNativeComponent` and `RNGestureHandlerButtonNativeComponent` with TypeScript.

## Test plan

Tested on FabricExample and on a separate app.


src/fabric/RNGestureHandlerRootViewNativeComponent.ts
```diff
@@ -0,0 +1,12 @@
+// @ts-ignore TS being TS
+import codegenNativeComponentUntyped from 'react-native/Libraries/Utilities/codegenNativeComponent';
+import type { ViewProps, HostComponent } from 'react-native';
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+const codegenNativeComponent = codegenNativeComponentUntyped as <T extends {}>(
+  name: string
```
tomekzaw, line 7: It looks like it's unused.

src/fabric/RNGestureHandlerRootViewNativeComponent.ts
```diff
@@ -0,0 +1,12 @@
+// @ts-ignore TS being TS
```
tomekzaw, line 1: Let's add a note why we need to ignore these types with a TODO not to forget about it in the future.

j-piasecki: Done in [1151519](https://github.com/software-mansion/react-native-gesture-handler/pull/1952/commits/1151519599c4b31a3d0ed0eb22c888aa6b72ed53).

PR #1959, opened 2022-03-28
🔧🇫🇮 Fix TypeScript typing issues with `setTimeout` by removing NodeJS types

### LINKED ISSUE #1958: Type 'Timeout' is not assignable to type 'number'
## Description
After upgrading to the latest version, all the `setTimeout`s and `SetInterval`s types are broken. 
It turned out that somehow `RN-Gesture-Handler` picks the `Node` types as default instead of the `RN` types.
Also, these are related to this issue:
- https://github.com/software-mansion/react-native-gesture-handler/pull/1420

- https://github.com/software-mansion/react-native-gesture-handler/issues/1384

### Platforms

- [ ] iOS
- [ ] Android
- [ ] Web

### Screenshots
Screenshot 2022-03-28 at 11 55 41

## Steps To Reproduce

1. Clone the latest version
2. install `node_modules` for both root directory and the example project 
3. Add this snippet to the `App.tsx`
```ts
  const myTimer = React.useRef(0);

  React.useEffect(() => {
    myTimer.current = setTimeout(() => {
      console.log('Hello');
    }, 1000);
  }, []);
```
4. You will get this error:
```
Type 'Timeout' is not assignable to type 'number'
```
5. Go to root directory and open the `tsconfig.json`
6. Change the types value from `["node", "jest"]` to `[]`
7. The error will disappear 

### Expected behavior
It should respect RN types

### Actual behavior
It uses `node` types

## Package versions

- React: `17.0.2`
- React Native: `0.65.2`
- React Native Gesture Handler: `2.3.2`
- React Native Reanimated: `2.4.1`

henrikra:
## Description

This PR https://github.com/software-mansion/react-native-gesture-handler/pull/1844/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0 added `node` to the `types` array of `tsconfig`. This causes all of the React Native projects with `gesture-handler` as dependency to get Node typings to their project and node has conflicting types with Dom on `setTimeout`.

This PR gets rid of the `node` typing and adds a new file which adds missing types to gesture handler project.

Long story short:
🔴  We were requiring the whole NodeJS types just because we wanted to use `process.env.JEST_WORKER_ID `
✅  Fixed by not requiring NodeJS types and just typed out the `process.env.JEST_WORKER_ID`

Fixes #1958 

## Test plan

Run `yarn ts-check` and see that not TSC errors are produced.

### DIFF
--- a/src/handlers/gestures/GestureDetector.tsx
+++ b/src/handlers/gestures/GestureDetector.tsx
@@ -36,9 +36,9 @@ import { ActionType } from '../../ActionType';
 import { isFabric, tagMessage } from '../../utils';
 import { getShadowNodeFromRef } from '../../getShadowNodeFromRef';
 
-declare global {
-  function isFormsStackingContext(node: unknown): boolean | null; // JSI function
-}
+declare const global: {
+  isFormsStackingContext: (node: unknown) => boolean | null; // JSI function
+};
 
 const ALLOWED_PROPS = [
   ...baseGestureHandlerWithMonitorProps,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -32,7 +32,7 @@ export function hasProperty(object: object, key: string) {
 }
 
 export function isJestEnv(): boolean {
-  // @ts-ignore Do not use `@types/node` because it will break setTimeout function types in React Native projects
+  // @ts-ignore Do not use `@types/node` because it will prioritise Node types over RN types which breaks the types (ex. setTimeout) in React Native projects.
   return hasProperty(global, 'process') && !!process.env.JEST_WORKER_ID;
 }
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -32,6 +32,7 @@ export function hasProperty(object: object, key: string) {
 }
 
 export function isJestEnv(): boolean {
+  // @ts-ignore Do not use `@types/node` because it will break setTimeout function types in React Native projects
   return hasProperty(global, 'process') && !!process.env.JEST_WORKER_ID;
 }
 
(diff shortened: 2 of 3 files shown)


src/handlers/gestures/GestureDetector.tsx
```diff
@@ -36,10 +36,6 @@ import { ActionType } from '../../ActionType';
 import { isFabric, tagMessage } from '../../utils';
 import { getShadowNodeFromRef } from '../../getShadowNodeFromRef';
```
j-piasecki, line 38:
```suggestion
declare const global: {
  isFormsStackingContext: (node: unknown) => boolean | null; // JSI function
};
```

j-piasecki (commented):
Hi! Thanks for taking the time to look into this, however In my opinion a simple `ts-ignore` with an explanation [here](https://github.com/software-mansion/react-native-gesture-handler/blob/c74caf90c21182835f25e8609570da6bec4a266b/src/utils.ts#L35) instead of adding a new file would suffice in this case. Especially because this way the explanation would be in the same place as the usage.

henrikra:
@j-piasecki Done 🇫🇮 🥑

src/utils.ts
```diff
@@ -32,6 +32,7 @@ export function hasProperty(object: object, key: string) {
 }
 
 export function isJestEnv(): boolean {
+  // @ts-ignore Do not use `@types/node` because it will break setTimeout function types in React Native projects
```
hamidhadi, line 35: Hey @henrikra, 
Thanks for doing this! 🙌 
As it's not only breaking the `setTimeout` but also other stuff like `setInterval`, I would rewrite this comment in this  way:
```
Using `@types/node` will prioritise node types over RN types which can break the types in React Native projects.
```

j-piasecki (approved):

PR #1960, opened 2022-03-28
Remove warning about the old API

j-piasecki:
## Description

Reverts https://github.com/software-mansion/react-native-gesture-handler/pull/1817 and https://github.com/software-mansion/react-native-gesture-handler/pull/1822, removing the warning when using the old API.

Closes https://github.com/software-mansion/react-native-gesture-handler/issues/1831.

## Test plan

Checked Example app.

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -1,12 +1,5 @@
 import React from 'react';
-import {
-  Text,
-  View,
-  StyleSheet,
-  SectionList,
-  Platform,
-  LogBox,
-} from 'react-native';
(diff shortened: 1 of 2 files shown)


tomekzaw (approved):

chakravarthy-mm:
Does this mean, RNGH gonna support the old API for v2?

hirbod:
The old API was ALWAYS supported, the only change that has to be done is to wrap your app with a `GestureHandlerRootView`. v1 API is fully working with v2 since the beginning @chakravarthy-mm

PR #1961, opened 2022-03-28
Add GitHub Actions for FabricExample

tomekzaw:
## Description

This PR adds a workflow matrix to run CI jobs both for example and FabricExample apps. These workflows test whether `react-native-gesture-handler` builds without errors using Fabric architecture on both Android and iOS.

## Test plan

- [ ] Ensured that CI passes


j-piasecki:
I would change `Test example apps TypeScript and Lint` to `Test TypeScript and Lint`.

FabricExample/src/ViewFlatteningScreen.tsx
```diff
@@ -58,13 +62,13 @@ function InnerNotFlattenedParentCollapsable() {
...
   const tap = Gesture.Tap().onStart(() =>
-    console.log(window.performance.now(), 'tap!')
+    console.log(performance.now(), 'tap!')
   );
 
   return (
     <>
       <Text>Parent view not flattened, collapsable=false</Text>
-      <Text>(structure doesn't change)</Text>
+      <Text>(structure doesn&apos;t change)</Text>
```
j-piasecki, line 71: What was wrong with `'`?

j-piasecki (approved):

FabricExample/src/ViewFlatteningScreen.tsx
tomekzaw: Don't know and at this point I'm too afraid to ask.

PR #1962, opened 2022-03-29
Updated Tap Gesture Handler Documentation

ajatdarojat45:
## Description

Since `maxDurationMs` function is undefined I have updated it to `maxDuration` in the gesture handler documentation description and example.

## Test plan
I just changed the documentation and checked the markdown preview.


j-piasecki (approved):
Thank you for fixing it ❤️

PR #1963, opened 2022-03-29
Pass accessibilityActions properties

pooyahrtn:
## Description
`accessibilityActions` and `onAccessibilityAction` properties weren't passed to the internal `Animated.View` component, making the component not supporting the accessibility actions.

## Test plan

* Passing `accessibilityActions` and `onAccessibilityAction` worked as expected with this PR.


j-piasecki (approved):
Thank you for taking the time to do this ❤️ .

PR #1965, opened 2022-03-30
Update documentation for wix/react-native-navigation

Yrsafam:
## Description

Hello! Thank you for your hard work. In version 2, the section with information about the wix\react-native-navigation conflict was removed, but I ran into the problem of gestures not working properly. After looking in the documentation of version 1, I found a solution for myself that was not mentioned in version 2.

## Test plan

I just changed the documentation and checked the markdown preview.

### DIFF
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -123,6 +123,40 @@ buildscript {
 
 The minimal version of the Kotlin plugin supported by RNGH is `1.4.10`.
 
+
+### With [wix/react-native-navigation](https://github.com/wix/react-native-navigation)
+
+If you are using a native navigation library like [wix/react-native-navigation](https://github.com/wix/react-native-navigation) you need to follow a different setup for your Android app to work properly. The reason is that both native navigation libraries and Gesture Handler library need to use their own special subclasses of `ReactRootView`.
+
+Instead of changing Java code you will need to wrap every screen component using `gestureHandlerRootHOC` on the JS side. This can be done for example at the stage when you register your screens. Here's an example:
+
+```js
+import { gestureHandlerRootHOC } from 'react-native-gesture-handler';
+import { Navigation } from 'react-native-navigation';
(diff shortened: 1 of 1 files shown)


docs/docs/installation.md
```diff
@@ -123,6 +123,40 @@ buildscript {
 
 The minimal version of the Kotlin plugin supported by RNGH is `1.4.10`.
 
+
+### With [wix/react-native-navigation](https://github.com/wix/react-native-navigation)
+
+If you are using a native navigation library like [wix/react-native-navigation](https://github.com/wix/react-native-navigation) you need to follow a different setup for your Android app to work properly. The reason is that both native navigation libraries and Gesture Handler library need to use their own special subclasses of `ReactRootView`.
+
+Instead of changing Java code you will need to wrap every screen component using `gestureHandlerRootHOC` on the JS side. This can be done for example at the stage when you register your screens. Here's an example:
```
j-piasecki, line 131:
```suggestion
If you are using a native navigation library like [wix/react-native-navigation](https://github.com/wix/react-native-navigation) you need to make sure that every screen is wrapped with `GestureHandlerRootView` (you can do this using `gestureHandlerRootHOC` function). This can be done for example at the stage when you register your screens. Here's an example:
```

Gesture Handler no longer needs the subclass of `ReactRootView`.
With version 2.0 `GestureHandlerEnabledRootView` was deprecated and using `GestureHandlerRootView` component is now necessary in all cases, and that was the main reason why it was removed. This may not have been the best idea, since the configuration in this case is still a bit different.

j-piasecki (commented):
Thank you for taking the time for this ❤️ .

Could you please move this fragment above the Android section? I think it's better not to split the native platforms.

docs/docs/installation.md
Yrsafam: Yes, I agree with you, I have made adjustments

j-piasecki (approved):

PR #1967, opened 2022-03-30
Fix props on `GestureHandlerButton`

j-piasecki:
## Description

- Fix codegen configuration: mark all props as optional and set default values for `exclusive` and `enabled`
- Add `updateProps` method for `GestureHandlerButton` on iOS that passes props to the underlying `Button` implementation

## Test plan

Tested on FabricExample


tomekzaw (approved):
👍

PR #1979, opened 2022-04-04
Fix gestures not attaching to views on iOS when using Fabric

j-piasecki:
## Description

when gestures are created and attached to views, on iOS the operations are batched and executed on the UI thread. The method we are using currently is run too early - the batched queue is flushed before the gestures are created on the JS side, effectively making the app unresponsive.

This PR adds a method that allows flushing the queue from the JS side, after all the operations are scheduled.

## Test plan

Tested on Example and FabricExample apps

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
@@ -407,10 +407,6 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?)
   fun handleClearJSResponder() {
   }
 
-  @ReactMethod
-  fun flushQueuedHandlers() {
-  }
-
   override fun setGestureHandlerState(handlerTag: Int, newState: Int) {
     registry.getHandler(handlerTag)?.let { handler ->
       when (newState) {
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
@@ -407,6 +407,10 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?)
   fun handleClearJSResponder() {
   }
 
+  @ReactMethod
+  fun flushQueuedHandlers() {
+  }
+
   override fun setGestureHandlerState(handlerTag: Int, newState: Int) {
     registry.getHandler(handlerTag)?.let { handler ->
       when (newState) {
--- a/ios/RNGestureHandlerModule.mm
+++ b/ios/RNGestureHandlerModule.mm
@@ -173,7 +173,7 @@ RCT_EXPORT_METHOD(handleClearJSResponder)
     }];
 }
 
-RCT_EXPORT_METHOD(flushQueuedHandlers)
+RCT_EXPORT_METHOD(flushOperations)
 {
     if (_operations.count == 0) {
         return;
--- a/ios/RNGestureHandlerModule.mm
+++ b/ios/RNGestureHandlerModule.mm
@@ -173,6 +173,22 @@ RCT_EXPORT_METHOD(handleClearJSResponder)
     }];
 }
 
+RCT_EXPORT_METHOD(flushQueuedHandlers)
+{
+    if (_operations.count == 0) {
+        return;
+    }
+
+    NSArray<GestureHandlerOperation> *operations = _operations;
+    _operations = [NSMutableArray new];
+
+    [self.bridge.uiManager addUIBlock:^(__unused RCTUIManager *manager, __unused NSDictionary<NSNumber *, UIView *> *viewRegistry) {
+        for (GestureHandlerOperation operation in operations) {
+            operation(self->_manager);
+        }
+    }];
+}
+
 - (void)setGestureState:(int)state forHandler:(int)handlerTag
 {
   RNGestureHandler *handler = [_manager handlerWithTag:@(handlerTag)];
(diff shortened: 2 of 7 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -407,6 +407,10 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?)
   fun handleClearJSResponder() {
   }
 
+  @ReactMethod
+  fun flushQueuedHandlers() {
+  }
```
tomekzaw, line 412: Would it be possible to define this method as no-op in pure JS, similarly as for web implementation?

ios/RNGestureHandlerModule.mm
```diff
@@ -173,6 +173,22 @@ - (void)setBridge:(RCTBridge *)bridge
     }];
 }
 
+RCT_EXPORT_METHOD(flushQueuedHandlers)
```
tomekzaw, line 176: I'm thinking of renaming this method to something like `flushGestureHandlerOperations` or `flushOperations` 🤔

src/handlers/gestureHandlerCommon.ts
```diff
@@ -180,3 +181,17 @@ export function findNodeHandle(
   if (Platform.OS === 'web') return node;
   return findNodeHandleRN(node);
 }
+
+let scheduledHandlersQueueFlush:
+  | ReturnType<typeof requestAnimationFrame>
+  | undefined = undefined;
+
+export function scheduleHandlersQueueFlush() {
```
tomekzaw, line 189: Once `flushHandlersQueue` method is renamed, don't forget to propagate this change here as well.

I'm also thinking if the pattern of the name of this function should be `scheduleFlushSomethingQueue` or `scheduleSomethingQueueFlush`.

Personally I like the first option better since it will also appear in the search results for `flushSomethingQueue` (at least for case-insensitive search).

src/handlers/gestureHandlerCommon.ts
```diff
@@ -180,3 +181,17 @@ export function findNodeHandle(
   if (Platform.OS === 'web') return node;
   return findNodeHandleRN(node);
 }
+
+let scheduledHandlersQueueFlush:
+  | ReturnType<typeof requestAnimationFrame>
+  | undefined = undefined;
```
tomekzaw, line 187: Can we use `null` as the default value instead? Then we can use `?` from TypeScript instead of `| undefined`.

src/handlers/gestureHandlerCommon.ts
```diff
@@ -180,3 +181,17 @@ export function findNodeHandle(
   if (Platform.OS === 'web') return node;
   return findNodeHandleRN(node);
 }
+
+let scheduledHandlersQueueFlush:
```
tomekzaw, line 185: Is there anything we can do to explicitely state that this variable holds the `id` returned from `requestAnimationFrame`? The name of this variable is very similar to the name of the method and I'm not sure if I like this or not 😄

src/handlers/gestureHandlerCommon.ts
```diff
@@ -180,3 +181,17 @@ export function findNodeHandle(
...
+  | ReturnType<typeof requestAnimationFrame>
```
tomekzaw, line 186: In fact, it's just a `number`, but I like the fact that your approach will work even if they change it to something else.

tomekzaw (changes requested):

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: [Yes, it would](https://github.com/software-mansion/react-native-gesture-handler/pull/1979/commits/d583ce848e7ff49a22b1ee64de9e9b222f6b4912).

ios/RNGestureHandlerModule.mm
j-piasecki: I went with [flushOperations](https://github.com/software-mansion/react-native-gesture-handler/pull/1979/commits/da1592520ebf7f94e9a060a68b5bbe83efc6d2a1), as this method will be called on `RNGestureHandlerModule` anyway. IMO it's self-explanatory that it relates to GH.

src/handlers/gestureHandlerCommon.ts
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1979/commits/da1592520ebf7f94e9a060a68b5bbe83efc6d2a1.

src/handlers/gestureHandlerCommon.ts
j-piasecki: TS didn't really like the idea of using `?` here, but using `null` instead of `undefined` is better. Done in https://github.com/software-mansion/react-native-gesture-handler/pull/1979/commits/325994e3843a700f3f01ae6f1c5c251f2c692c64.

src/handlers/gestureHandlerCommon.ts
j-piasecki: [Yes, there is](https://github.com/software-mansion/react-native-gesture-handler/pull/1979/commits/da1592520ebf7f94e9a060a68b5bbe83efc6d2a1#diff-88a2920cc09e756b9842ecad2212fb5750c715ffd5eb6f3deea58f0c3bab788aR185).

src/handlers/gestureHandlerCommon.ts
j-piasecki: I learned the hard way that `setTimeout` and `setInterval` have different return types on web and in RN. I assumed this would also be the case here. I think it's not much of a problem, it helps with understanding what this variable holds (I think) and will get compiled out anyways.

tomekzaw: I agree.

tomekzaw (approved):
👏

tomekzaw (approved):
👏

PR #1987, opened 2022-04-06
Implement sending events directly to Reanimated on Fabric

j-piasecki:
## Description

Adds support for sending events directly to Reanimated when using Fabric.

## Test plan

Tested on Reanimated FabricExample


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -589,10 +589,8 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?)
   private fun <T : Event<T>>sendEventForReanimated(event: T) {
     // Delivers the event to Reanimated.
     if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
-      // TODO: send event directly to Reanimated
       // This is already supported in Reanimated with Fabric but let's wait until the official release.
```
tomekzaw, line 592:
```suggestion
      // Send event directly to Reanimated
```

tomekzaw (approved):
👍

PR #1988, opened 2022-04-07
Drop all gesture handlers on invalidate

j-piasecki:
## Description

For some reason scheduled `dropGestureHandler` calls weren't executed on app refresh and all recognizers would stay attached to their respective views, alongside the newly created ones. This PR solves this problem by dropping all recognizers when the module is invalidated.

[Android already has a similar mechanism](https://github.com/software-mansion/react-native-gesture-handler/blob/8afdda06c532af957b1466760f11a28c973ce605/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt#L457).

## Test plan

Tested on FabricExample app.

### DIFF
--- a/ios/RNGestureHandlerRegistry.h
+++ b/ios/RNGestureHandlerRegistry.h
@@ -14,5 +14,6 @@
 - (void)registerGestureHandler:(nonnull RNGestureHandler *)gestureHandler;
 - (void)attachHandlerWithTag:(nonnull NSNumber *)handlerTag toView:(nonnull UIView *)view withActionType:(RNGestureHandlerActionType)actionType;
 - (void)dropHandlerWithTag:(nonnull NSNumber *)handlerTag;
+- (void)dropAllHandlers;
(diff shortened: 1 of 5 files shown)


ios/RNGestureHandlerRegistry.h
```diff
@@ -14,5 +14,6 @@
 - (void)registerGestureHandler:(nonnull RNGestureHandler *)gestureHandler;
 - (void)attachHandlerWithTag:(nonnull NSNumber *)handlerTag toView:(nonnull UIView *)view withActionType:(RNGestureHandlerActionType)actionType;
 - (void)dropHandlerWithTag:(nonnull NSNumber *)handlerTag;
+- (void)dropAllHandlers;
```
tomekzaw, line 17: I'd suggest renaming this method to `dropAllGestureHandlers`.

kmagiera (approved):

Issue #1720, opened 2021-11-14
@types/react-native should be a dependency


alamothe:
Since this package's types depend on `@types/react-native`, it should be a dependency of this package. Please check: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Currently, the problem will be observed with Yarn PNP (which is strict about this) where your typings are unusable.

j-piasecki:
Since 0.71 types are bundled with React Native, so it shouldn't be necessary anymore.

PR #1990, opened 2022-04-07
Move `@types/react-native` to dependencies

j-piasecki:
## Description

Moves `@types/react-native` to dependencies from devDependencies.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/1720.


j-piasecki:
Since 0.71 types are bundled with React Native.

Issue #386, opened 2018-12-15
Bug with permissions dialog and position


ahce:
## Description

- Position bug: The drawer is behind the button:
![image](https://user-images.githubusercontent.com/12452003/50048330-d4b9de00-00a7-11e9-9942-69bac2bc6146.png)
- Position fix: Set zIndex
![image](https://user-images.githubusercontent.com/12452003/50048337-121e6b80-00a8-11e9-8e1f-9d2b9bcf9a4a.png)
- Permission dialog bug: The drawer in close mode is visible behind the dialog, this occur in special situations
  - Form 1:
    - First: press next up to value 2
    - Second: focus the textinput and open keyboard
    - Third: press next again
  - Form 2:
    - First: press next or back button repeatedly

**Steps**
![image](https://user-images.githubusercontent.com/12452003/50048367-e18b0180-00a8-11e9-9caf-14e16736d214.png)

**Result**
![image](https://user-images.githubusercontent.com/12452003/50048359-c0c2ac00-00a8-11e9-877a-111635f3c5a0.png)

## Note
This bugs don't occur with react-native-drawer-layout-polyfill.
In the demo, comment and uncomment the imports for switch the drawers.
![image](https://user-images.githubusercontent.com/12452003/50048410-10ee3e00-00aa-11e9-873b-8572bdd9cbb6.png)

## Environment
- react: 16.6.3
- react-native: 0.57.8
- react-native-gesture-handler: 1.0.12
- phone: Nokia 5 Android 8.1

## Demo
[RNGH-DrawerLayout-Bug](https://github.com/ahce/RNGH-DrawerLayout-Bug)

osdnk:
@brentvatne

ahce:
@brentvatne Thanks for your reply.

1. Yes, the problem is the elevation, where zIndex is the fix.

2. Special situations are those reported in the steps of Form 1 and 2 where the drawer is closed.

Please test my repo [RNGH-DrawerLayout-Bug](https://github.com/ahce/RNGH-DrawerLayout-Bug)

ahce:
@brentvatne the bug when drawer is closed and is visible behind the alert, don't occur with useNativeAnimations set in false.
It is possible to set useNativeAnimations to false by default?

osdnk:
@ahce, I don't want to change `useNativeAnimations` to `false`. It's not a point of this library. Will try to figure it out

ahce:
[...]

PR #2005, opened 2022-04-13
Use constant value for translation when `DrawerLayout` is idle

j-piasecki:
## Description

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/386.

`AnimatedInterpolation` was causing problems on Android when the activity was paused for some reason (like permission or share dialog). For some reason it was changing its value on pause, opening the drawer, and again on resume closing the drawer. This PR changes `DrawerLayout` so that it uses constant value instead of interpolation when the drawer is idle.

## Test plan

Tested on the Example app

### DIFF
--- a/src/components/DrawerLayout.tsx
+++ b/src/components/DrawerLayout.tsx
@@ -155,6 +155,8 @@ export type DrawerLayoutState = {
   touchX: Animated.Value;
   drawerTranslation: Animated.Value;
   containerWidth: number;
+  drawerState: DrawerState;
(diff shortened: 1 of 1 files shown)


linxianxi:
Someone review?

kmagiera (approved):

linxianxi:
When will the package be released?

j-piasecki:
@linxianxi This change is included in Gesture Handler 2.4.2.

PR #2010, opened 2022-04-16
Make sure that react-native-macos is supported by using the web fallback

### LINKED ISSUE #1927: react-native-gesture-handler does not work with react-native-macos anymore
## Description

Hi!
This lib used to work with react-native-macos, but since 2.0.0 it's crashing because [it's trying to call a native module function (which doesn't exist for macos target)](https://github.com/software-mansion/react-native-gesture-handler/blob/9569b56f31aed5c9e0bafc6d193e7accf0414883/src/handlers/createHandler.ts#L91).

Reusing the web version seems to be working, what do you think about reusing it?
FY it's logging a lot of warnings, but looks like this is [intended behavior](https://github.com/software-mansion/react-native-gesture-handler/blob/9569b56f31aed5c9e0bafc6d193e7accf0414883/src/RNGestureHandlerModule.web.ts#L25-L30). I don't see them in react-native-web though.

### Platforms

- [x] macOS (react-native-macos)

## Steps To Reproduce

1. Create a react-native-macos project
2. Wrap `<GestureHandlerRootView>`
3. Run application and click somewhere

### Expected behavior

No error should be seen.

### Actual behavior

App is crashes on call to `handleSetJSResponder` because there is no native module for macOS target.

```JSX
import React from "react";
import { AppRegistry } from "react-native";
import { GestureHandlerRootView } from "react-native-gesture-handler";
import { name as appName } from "./app.json";

function App() {
  return (
    <GestureHandlerRootView />
  );
};

AppRegistry.registerComponent(appName, () => App);
```

## Package versions

- React: 17.0.2 but shouldn't matter
- React Native: 0.67.3 but shouldn't matter
- React Native macOS: 0.66.31 but shouldn't matter
- React Native Gesture Handler: 2.3.1
- React Native Reanimated: 1.13.3 but shouldn't matter

hsjoberg:
## Description

Hi, this adds "support" for react-native-macos by using the web fallback, I say support in quotes because it would be nicer if gestures actually could be used on macOS, nonetheless this PR makes sure that you can at least run react-native-macos apps using react-native-gesture-handler (crashes without this fix), real support could be added later on.
Fixes #1927.

### DIFF
--- a/src/RNGestureHandlerModule.macos.ts
+++ b/src/RNGestureHandlerModule.macos.ts
@@ -1,0 +1,62 @@
+import { ActionType } from './ActionType';
+import { Direction } from './web/constants';
+import FlingGestureHandler from './web/FlingGestureHandler';
+import LongPressGestureHandler from './web/LongPressGestureHandler';
+import NativeViewGestureHandler from './web/NativeViewGestureHandler';
+import * as NodeManager from './web/NodeManager';
+import PanGestureHandler from './web/PanGestureHandler';
+import PinchGestureHandler from './web/PinchGestureHandler';
+import RotationGestureHandler from './web/RotationGestureHandler';
+import TapGestureHandler from './web/TapGestureHandler';
+
+export const Gestures = {
+  PanGestureHandler,
+  RotationGestureHandler,
+  PinchGestureHandler,
+  TapGestureHandler,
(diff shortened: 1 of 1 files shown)


j-piasecki (approved):
Thanks ❤️ !

PR #2012, opened 2022-04-18
Add ndkVersion to Android project

maximzhemerenko:
## Description

When I build Android app (react native 0.68.1) using arm based MacBook Pro M1, I receive 'Unknown host CPU architecture: arm64' error. This happens because library uses default ndkVersion, but not custom ndkVersion from 'ext'. I propose include this line in the build.gradle file to use globally specified ndkVersion if exists.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -51,6 +51,15 @@ repositories {
 android {
     compileSdkVersion safeExtGet("compileSdkVersion", 28)
 
+    // Used to override the NDK path/version on internal CI or by allowing
+    // users to customize the NDK path/version from their root project (e.g. for M1 support)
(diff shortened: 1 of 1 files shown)


j-piasecki:
Hi! Could you do it similarly to how it's implemented in [React Native](https://github.com/facebook/react-native/blob/23b6240b256b2afb0e5efb2cef1ce27cbd0fcdf8/ReactAndroid/build.gradle#L257-L262), including `ndkPath`?

maximzhemerenko:
@j-piasecki thank you for suggestion, I changed implementation like it made in react-native

j-piasecki (approved):
Thank for taking the time to do this ❤️ .

PR #2015, opened 2022-04-19
Support monorepos with new architecture

oblador:
## Description

When enabling the new architecture in a monorepo/yarn workspace where the `react-native` package is hoisted `RNGH` fails due to incorrect paths to `react-native`/`react-native-codegen` packages. 

This PR vendors the `findNodeModulePath` from RN core and applies it to this project. 

## Test plan

Tested the changes in our project via `patch-package`.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,4 +1,5 @@
 import groovy.json.JsonSlurper
+import java.nio.file.Paths
 
 buildscript {
     def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['RNGH_kotlinVersion']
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -148,9 +163,9 @@ dependencies {
 
 if (isNewArchitectureEnabled()) {
     react {
-        reactRoot = rootProject.file("../node_modules/react-native/")
+        reactNativeDir = rootProject.file(findNodeModulePath(rootProject.rootDir, "react-native") ?: "../node_modules/react-native/")
```
j-piasecki, line 166: What's the reason for changing `reactRoot` to `reactNativeDir`?

oblador: reactRoot is deprecated, see https://github.com/facebook/react-native/pull/33142

j-piasecki (approved):
Thanks for taking the time to fix this ❤️ .

WoLewicki:
@oblador do you happen to know if such problem is not also present in `react-native-screens` repo?

oblador:
Not sure, I don't use that library.

PR #2018, opened 2022-04-20
Fix `simultaneousHandlers` prop on `NativeViewGestureHandler` on iOS

j-piasecki:
## Description

It seems like to make two `UIGestureRecognizers` work simultaneously, both calls of `shouldRecognizeSimultaneouslyWithGestureRecognizer` should return `YES` (should `A` be recognized alongside `B` and vice versa). This isn't much of a problem in case of all gestures but `NativeViewGestureHandler`, which most often is used implicitly via components exported by Gesture Handler. In this case one can access `simultaneousHandlers` prop on the wrapped component, but cannot obtain the ref to the `NativeViewGestureHandler` (only to the view).

This PR makes `shouldRecognizeSimultaneouslyWithGestureRecognizer` symmetric, so only one gesture needs to be marked as simultaneous with the other

## Test plan

Tested on the Example app

### DIFF
--- a/ios/RNGestureHandler.h
+++ b/ios/RNGestureHandler.h
@@ -77,6 +77,7 @@ if (value != nil) { recognizer.prop = [RCTConvert type:value]; }\
 - (void)sendEvent:(nonnull RNGestureHandlerStateChange *)event;
 - (void)sendTouchEventInState:(RNGestureHandlerState)state
                  forViewWithTag:(nonnull NSNumber *)reactTag;
+- (NSArray<NSNumber *> *)getSimultaneousHandlers;
 
 @end
 
--- a/ios/RNGestureHandler.h
+++ b/ios/RNGestureHandler.h
@@ -77,7 +77,6 @@ if (value != nil) { recognizer.prop = [RCTConvert type:value]; }\
 - (void)sendEvent:(nonnull RNGestureHandlerStateChange *)event;
 - (void)sendTouchEventInState:(RNGestureHandlerState)state
                  forViewWithTag:(nonnull NSNumber *)reactTag;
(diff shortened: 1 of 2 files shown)


ios/RNGestureHandler.h
```diff
@@ -77,6 +77,7 @@ if (value != nil) { recognizer.prop = [RCTConvert type:value]; }\
 - (void)sendEvent:(nonnull RNGestureHandlerStateChange *)event;
 - (void)sendTouchEventInState:(RNGestureHandlerState)state
                  forViewWithTag:(nonnull NSNumber *)reactTag;
+- (NSArray<NSNumber *> *)getSimultaneousHandlers;
```
kmagiera, line 80: maybe change to readonly prop "simultaneousHandlers" ?

kmagiera (approved):
Looks good 👍

ios/RNGestureHandler.m
```diff
@@ -390,14 +395,21 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
+    
+    RNGestureHandler *handler = [RNGestureHandler findGestureHandlerByRecognizer:otherGestureRecognizer];
+    if (handler != nil) {
+        if ([_simultaneousHandlers count]) {
             for (NSNumber *handlerTag in _simultaneousHandlers) {
                 if ([handler.tag isEqual:handlerTag]) {
                     return YES;
                 }
             }
+        } else if ([handler getSimultaneousHandlers]) {
```
kmagiera, line 407: Ot just make `handler->_simultaniousHandlers` here

j-piasecki: Done in fe4b2ea.

PR #2028, opened 2022-04-25
Schedule calling `dropAllGestureHandlers` on main queue

j-piasecki:
## Description

Instead of directly calling `dropAllGestureHandlers` in the `invalidate`, dispatch it to be called on the main queue.
Why not `[bridge.uiManager addUIBlock ...]` you may ask. Well, it appears to be cleared on the invalidation as well 😞 .

## Test plan

Check if the warning appears.


tomekzaw (approved):
👍

Issue #2029, opened 2022-04-25
Bug: type issues with FunctionComponent


hirbod:
## Description

Your exports use React.FunctionComponent (which is discouraged). It adds implicit children prop which got removed in latest react types. so I need to downgrade @types/react to 17.x.x.

All my usages of `RectButton`, `GestureDetector` etc are reporting type issues now.

### Platforms

- [x] iOS
- [x] Android
- [x] Web

### Screenshots
<img width="941" alt="Bildschirmfoto 2022-04-25 um 11 46 37" src="https://user-images.githubusercontent.com/504909/165067372-9cff8843-7451-4435-a272-1e6a32c4f74e.png">
<img width="1122" alt="Bildschirmfoto 2022-04-25 um 11 46 26" src="https://user-images.githubusercontent.com/504909/165067377-a1472042-f7c8-45db-a0aa-8e414d03e28c.png">

## Steps To Reproduce

1. Upgrade react types and import a component
2. Have issues

### Expected behavior

No type issues

### Actual behavior

Having type issues

## Snack or minimal code example

The best way is not using FC anymore.

## Package versions

- React: 0.17
- React Native: 0.64.3
- React Native Gesture Handler: 2.4.1
- React Native Reanimated: 2.7.0

PR #2030, opened 2022-04-25
Fix compatibility with `@types/react^18.0.0`

j-piasecki:
## Description

Fix compatibility with `@types/react^18.0.0`:
- Remove `React.FunctionComponent` type from `GestureDetector`
- Add explicit `children` props to components as suggested in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210
- Update types in the Example app

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2029.


hirbod:
It is also happening on RectButton, BorderlessButton etc

kmagiera (approved):

PR #2031, opened 2022-04-27
Update  scroll view class check in `RNNativeViewHandler` on Fabric

j-piasecki:
## Description

When `NativeViewGestureHandler` is attached to a `ScrollView` on iOS, it sets its `delaysContentTouches` property to `YES`. This is done by checking if the view the gesture is attached to is `RCTScrollView` class, which wraps `UIScrollView`, but on Fabric the `UIScrollView` is wrapped with `RCTScrollViewComponentView`. This PR makes it so that the former is used on Paper and the later on Fabric.

## Test plan

Tested on Example and FabricExample apps. Added a breakpoint inside the ifs to see if the code inside is run.


tomekzaw (approved):
👍

PR #2036, opened 2022-04-28
Default import `package.json` in `src/utils`

nandorojo:
This change silences the following error from Webpack:

```sh
warn  - ../../node_modules/react-native-gesture-handler/lib/module/utils.js
Should not import the named export 'version'.'split' (imported as 'rnVersion') from default-exporting module (only default export is available soon)
```

## Description

Changes `import { version as rnVersion }` to `import pack`.

## Test plan

Patched it in my app.


j-piasecki (approved):
Thank you!

Issue #1242, opened 2020-11-27
Increasing hitSlop does not increase touch area


LouiseBC:
## Description

I added hitSlop to buttons on iOS and it works perfectly, but the touch area isn't increased on Android.
It seems to be an issue with react-native-gesture-handler: We're using a `BorderlessButton`, but when I replace it with `TouchableOpactiy` from react-native it works just fine. Using RNGH's TouchableOpacity once again doesn't work, and besides that it seems to calculate different touch areas.

### Screenshots         
![Screenshot 2020-11-27 at 09 55 18](https://user-images.githubusercontent.com/21264029/100430230-c8619c80-3096-11eb-8cd0-eb2ce352575a.png)
This is what the touch area looks like using TouchableOpacity from RN, or using BorderlessButton from RNGH. When using TouchableOpacity from RN the touch area indicated here works. When using BorderlessButton only the red area (padding) is touchable.

![Screenshot 2020-11-27 at 09 55 27](https://user-images.githubusercontent.com/21264029/100430391-052d9380-3097-11eb-9177-ff65aca7ac91.png)
This is what the touch area looks like using TouchableOpacity from RNGH

## Snack or minimal code example

```
    <TouchableOpacity    // or BorderlessButton
      ...
      hitSlop={{top: 16, bottom: 16, left: 8, right: 0}}>
      ...
    </TouchableOpacity>
```

## Package versions

- React: 16.13.1
- React Native: 0.63.3
- React Native Gesture Handler: 1.9.0

Thavry:
Any news about this issue?

VictorPullzz:
Any news about this issue?

leepowellnbs:
Same issue - this seems to work correctly for RN's implementation of TouchableOpacity, but not RNGH in Android.

PR #2039, opened 2022-05-05
Fix `hitSlop` on Gesture Handler buttons

j-piasecki:
## Description

This PR does two things:
- (iOS) sets `hitTestEdgeInsets` on `RNGestureHandlerButton` when Fabric is enabled
- (Android) sets `clipChildren` to `false` on `ButtonViewGroup`

The second change makes the button behavior consistent with iOS where its content isn't clipped. It also fixes https://github.com/software-mansion/react-native-gesture-handler/issues/1242 because of [this line](https://github.com/software-mansion/react-native-gesture-handler/blob/1217039146ddcae6796820b5ecf19d1ff51af837/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt#L449), which tried to extract handlers from view when it's not clipping children.

## Test plan

Tested on the Example app.


j-piasecki:
@hirbod It should be possible already, because the [`hitSlop` property is transformed when it's not an object](https://github.com/software-mansion/react-native-gesture-handler/blob/b9904f69d225b9a40607795ff4c3fb1575f009b9/src/handlers/gestureHandlerCommon.ts#L156-L158). I tried it on a simple touchable and it worked fine, although the types seem to be wrong, I'll look into it.

lightrow:
can this please be looked at? This is a major annoyance

magrinj:
@j-piasecki Are you able to get this merged soon ? I'm facing the same issue on my side too mostly on Android devices, it's really annoying and frustrating for the users

m-bert (approved):

DenisDov:
seems not to be fixed on Android

PR #2053, opened 2022-05-16
Upgrade Gesture Handler and FabricExample to support RN 0.69

j-piasecki:
## Description

- upgrade FabricExample to use React Native 0.69.0
- fix GH compatibility with RN 0.69

Also adds `mkdirp` as a dev dependency to FabricExample because it's used by some RN script but it's not anywhere in the dependencies and CI was failing because of that.

## Test plan

Build and check FabricExample


kmagiera (approved):
Unless that's already done maybe worth submitting these patches to safe-area and screens repos?

Issue #2056, opened 2022-05-18
[Web] No handler for tag undefined error when having Gesture.exclusive(PanGesture, TapGesture)


itsramiel:
## Description

I am trying to have an Exclusive Gesture between Tap and Pan Gestures. When I pass the Exclusive Gesture to GestureDetector's "gesture" prop, I get the following error in the console: 

```
Uncaught Error: No handler for tag undefined
    getHandler NodeManager.ts:9
    waitFor GestureHandler.ts:514
    ensureConfig GestureHandler.ts:513
    updateGestureConfig GestureHandler.ts:118
    updateGestureHandler RNGestureHandlerModule.web.ts:53
    _loop GestureDetector.tsx:143
    run setImmediate.js:40
    runIfPresent setImmediate.js:69
    onGlobalMessage setImmediate.js:109
    installPostMessageImplementation setImmediate.js:114
    js setImmediate.js:169
    js setImmediate.js:186
    js bundle.js:4755
```

As well as the whole page becomes unresponsive. Buttons become unpressable.
This seems to only occur when having both gestures Exclusively not Simultaneously. 

### Platforms

- [ ] iOS
- [ ] Android
- [x] Web

### Screenshots

![image](https://user-images.githubusercontent.com/80689446/168978437-48012038-4bf4-4d05-a5a7-f89bb16e38e7.png)

## Steps To Reproduce

1. In react native web. Add a pan as well as a tap gesture in Gesture.Exclusive and pass it to a GestureDetector

### Expected behavior

I should not be getting that error and page should still be responsive

### Actual behavior

I get the error shown in the description

## Snack or minimal code example

Here is a repro: https://github.com/ramiel1999/gesture-handler-pan-tap/tree/tap-pan-causes-handle-tag-error
```

export default function App() {
  const translateX = useSharedValue(0);
  const context = useSharedValue(0);

  const rStyle = useAnimatedStyle(() => {
    return { transform: [{ translateX: translateX.value }] };
  }, []);

  const tapGesture = Gesture.Tap().onEnd(() => (translateX.value = withRepeat(withTiming(100), 2, true)));
  const panGesture = Gesture.Pan()
    .onStart(() => {
      console.log("onStart");
      context.value = translateX.value;
    })
    .onChange((e) => {
      translateX.value = e.translationX + context.value;
    })
    .onEnd(() => {
      console.log("onEnd");
      context.value = translateX.value;
    });

  return (
[...]

Issue #2057, opened 2022-05-18
[Web] onStart event for Pan and Tap gestures not invoked


itsramiel:
## Description

I am trying to run code when onStart event is invoked for Pan or Tap gesture. Both of them seem to not work independently of each other. 

### Platforms

- [ ] iOS
- [ ] Android
- [x] Web

### Screenshots
pan:
![image](https://user-images.githubusercontent.com/80689446/168982686-e5eae910-fc84-4e61-b333-8f5427a6712a.png)
tap:
![image](https://user-images.githubusercontent.com/80689446/168983835-7359ea92-05a4-4e23-85dc-cac716265a15.png)

## Steps To Reproduce

1. Pass a tap/pan gesture to GestureDetector and run some code in the onStart event's callback

### Expected behavior

I expect the onStart callback to be called

### Actual behavior

onStart callback is not being called

## Snack or minimal code example

pan:
repro repo: https://github.com/ramiel1999/gesture-handler-pan-tap/tree/pan-start-state-not-called-on-web
minimal code: 
```
export default function App() {
  const translateX = useSharedValue(0);
  const context = useSharedValue(0);

  const rStyle = useAnimatedStyle(() => {
    return { transform: [{ translateX: translateX.value }] };
  }, []);

  const panGesture = Gesture.Pan()
    .onStart(() => {
      console.log("onStart");
      context.value = translateX.value;
    })
    .onChange((e) => {
      console.log("onChange");
      translateX.value = e.translationX + context.value;
    })
    .onEnd(() => {
      console.log("onEnd");
      context.value = translateX.value;
    });

  return (
    <GestureHandlerRootView style={styles.container}>
      <GestureDetector gesture={panGesture}>
        <Animated.View style={[styles.box, rStyle]} />
      </GestureDetector>
    </GestureHandlerRootView>
```

tap:
repro repo: https://github.com/ramiel1999/gesture-handler-pan-tap/tree/tap-start-state-not-called-on-web
minimal code: 

```
export default function App() {
  const translateX = useSharedValue(0);

  const rStyle = useAnimatedStyle(() => {
    return { transform: [{ translateX: translateX.value }] };
  }, []);

  const tapGesture = Gesture.Tap()
    .onStart(() => {
      console.log("onStart");
      translateX.value = withTiming(100);
    })
    .onEnd(() => {
      console.log("onEnd");
      translateX.value = withTiming(-100);
    });

  return (
[...]

PR #2058, opened 2022-05-18
Fix `Gesture.Exclusive` and `onStart` not being called on web

j-piasecki:
## Description

- moves checking if the state has changed to happen before `previousState` gets overwritten, which makes GestureHandler correctly add `oldState` prop to events
- updates `waitFor` check so that it handles getting only `handlerTag`, which is the case with the new API

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2056 and https://github.com/software-mansion/react-native-gesture-handler/issues/2057.

## Test plan

Tested on the Example app.

### DIFF
--- a/src/web/GestureHandler.ts
+++ b/src/web/GestureHandler.ts
@@ -511,9 +511,13 @@ function ensureConfig(config: Config): Required<Config> {
   }
   if ('waitFor' in config) {
     props.waitFor = asArray(config.waitFor)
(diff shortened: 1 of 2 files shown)


KrasniqiR:
Also ran into this issue and would be grateful if the solution could be merged provided it passes checks.

j-piasecki:
Gesture Handler 2.10.0 uses the new implementation by default, where this problem doesn't exist.

Issue #2062, opened 2022-05-18
Gesture.Tap().maxDuration(...) not working on web


elliotwaite:
## Description

Using `maxDuration()` on a Tap gesture doesn't seem to work on web.

### Platforms

- [ ] iOS
- [ ] Android
- [x] Web

### Screenshots

![screen-recording](https://user-images.githubusercontent.com/1767836/169109204-50e0e77c-8a01-4ade-bfe0-2e8f14e73321.gif)

## Steps To Reproduce

Run the code example below on the web then tap and hold on the red circle.

### Expected behavior

For it to behave the same as on native where if the tap is held longer than the `maxDuration` value, the `onTouchesCancelled` callback is called followed by the `onFinalize` callback being called.

### Actual behavior

On web, the tap can be held forever without ever being canceled.

## Snack or minimal code example

```
import { Platform, StyleSheet, Text } from 'react-native'
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,
} from 'react-native-gesture-handler'
import Animated, {
  useAnimatedStyle,
  useSharedValue,
} from 'react-native-reanimated'

export default function App() {
  const isPressed = useSharedValue(false)
  const wasCancelled = useSharedValue(false)
  const wasFinalized = useSharedValue(false)

  const tapGesture = Gesture.Tap()
    .maxDuration(500)
    .onBegin(() => {
      isPressed.value = true
      wasCancelled.value = false
      wasFinalized.value = false
    })
    .onTouchesCancelled(() => {
      wasCancelled.value = true
    })
    .onFinalize(() => {
      isPressed.value = false
      wasFinalized.value = true
    })

  const wasCancelledAnimatedStyles = useAnimatedStyle(() => ({
    opacity: wasCancelled.value ? 1 : 0,
  }))

  const wasFinalizedAnimatedStyles = useAnimatedStyle(() => ({
    opacity: wasFinalized.value ? 1 : 0,
  }))

  const ballAnimatedStyles = useAnimatedStyle(() => ({
    backgroundColor: isPressed.value ? 'yellow' : 'red',
  }))

  return (
    <GestureHandlerRootView style={styles.container}>
      <Text style={styles.text}>
        {Platform.OS === 'web' ? 'Web Version' : 'Native Version'}
      </Text>
      <Animated.View style={wasCancelledAnimatedStyles}>
        <Text style={styles.text}>`onTouchesCancelled()` was called.</Text>
      </Animated.View>
[...]

PR #2065, opened 2022-05-19
Make `Tap` respect `maxDuration` on web

j-piasecki:
## Description

This PR changes the logic of `TapGestureHandler` on web so that it sends `CANCELLED` event when the view is pressed too long by utilizing both timers defined in `TapGestureHandler`.

(Mostly) fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2062

## Test plan

Tested on the Example app


j-piasecki:
Fixed by the new implementation.

Issue #2066, opened 2022-05-19
Pan behaves differently on web


elliotwaite:
## Description

The native version of Pan calls `onBegin` when pressed, and then `onUpdate` when panned beyond the `minDistance` value.

The web version of Pan calls `onBegin` and `onUpdate` at the same time, after being panned. Also, the web version seems to ignore small `minDistance` values. For example, setting `minDistance(0)` seems to have no effect (as can be seen in the demo below).

### Platforms

- [ ] iOS
- [ ] Android
- [x] Web

### Screenshots

Here's an animated GIF of it (you might have to press the play button in the top right to watch it):

![pan-screen-recording](https://user-images.githubusercontent.com/1767836/169351815-036cdd37-8e9d-45d6-8fe5-9684f288dbbc.gif)

## Steps To Reproduce

1. Run the code below.
2. Click the box and drag it to the side.

### Expected behavior

For the web and native versions to behave the same.

### Actual behavior

The web version behaves differently.

## Snack or minimal code example

```
import { Platform, StyleSheet, Text, View } from 'react-native'
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,
} from 'react-native-gesture-handler'
import Animated, {
  useAnimatedStyle,
  useSharedValue,
} from 'react-native-reanimated'

export default function App() {
  const isActive = useSharedValue(false)
  const startX = useSharedValue(0)
  const translateX = useSharedValue(0)
  const onBeginWasCalled = useSharedValue(false)
  const onUpdateWasCalled = useSharedValue(false)

  const gesture = Gesture.Pan()
    .minDistance(0)
    .onBegin(() => {
      isActive.value = true
      startX.value = translateX.value
      onBeginWasCalled.value = true
    })
    .onUpdate((e) => {
      translateX.value = startX.value + e.translationX
      onUpdateWasCalled.value = true
    })
    .onFinalize(() => {
      isActive.value = false
      onBeginWasCalled.value = false
      onUpdateWasCalled.value = false
    })

  const onBeginAnimatedStyles = useAnimatedStyle(() => ({
    opacity: onBeginWasCalled.value ? 1 : 0,
  }))

  const onUpdateAnimatedStyles = useAnimatedStyle(() => ({
    opacity: onUpdateWasCalled.value ? 1 : 0,
  }))

  const boxAnimatedStyles = useAnimatedStyle(() => ({
    transform: [{ translateX: translateX.value }],
[...]

PR #2068, opened 2022-05-20
Fix `minDist` not working for `Pan` on web and improve the state flow of `Pan` gesture

j-piasecki:
## Description

This PR updates the `Pan` gesture on web:
- adds `threshold` to hammer config, mapping to `minDist` which was previously ignored
- removes `getState` which was overwriting state of the first `ACTIVE` event to `BEGAN`
- adds logic responsible for sending `BEGAN` event when the input starts
- adds logic responsible for sending `FAILED` event when the gesture finishes without activating

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2066.

## Test plan

Tested on the Example app.

### DIFF
--- a/src/web/PanGestureHandler.ts
+++ b/src/web/PanGestureHandler.ts
@@ -220,8 +220,8 @@ class PanGestureHandler extends DraggingGestureHandler {
       this.previousState !== State.BEGAN &&
       this.previousState !== State.ACTIVE
     ) {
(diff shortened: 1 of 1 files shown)


burakgormek:
I have the same issue as #2066 and can confirm that this PR fixes it.

tomekzaw (approved):

PR #2075, opened 2022-05-25
Update to 2.2.1

jonwu:


j-piasecki:
Hi! Are you sure this PR should be there 😅 ? If so, could you provide a bit more context about it?

Issue #2060, opened 2022-05-18
feat: add `longPress` to RectButton/BaseButton/BorderlessButton


hirbod:
## Description
I think it would be great of your Touchables would implement `longPress`. Used often in FlatLists and right now I am forced to wrap those with `GestureDetector` to wrap longPress detection around them. Would be great if this could be done in a more convenient way like all the other touchables

### Platforms

- [x] iOS
- [x] Android
- [ ] Web

## Steps To Reproduce

1.  No function/settings to longPress a RectButton etc.

### Expected behavior

Should have longPress

### Actual behavior

Don't have longPress

## Package versions

- React: 0.17.2
- React Native: 0.68.2
- React Native Gesture Handler: 2.4.2
- React Native Reanimated: 2.8.0

hirbod:
Thanks so much for this! This will make thing so much easier in the future

PR #2083, opened 2022-06-01
Add `onLongPress` callback to Gesture Handler buttons

j-piasecki:
## Description

This PR adds `onLongPress` to Gesture Handler buttons as suggested in https://github.com/software-mansion/react-native-gesture-handler/issues/2060, implemented in the same way as on Gesture Handler Touchables. The long press delay is customizable using `delayLongPress` prop.

## Test plan

Tested on the Example app.

### DIFF
--- a/docs/docs/api/components/buttons.mdx
+++ b/docs/docs/api/components/buttons.mdx
@@ -59,7 +59,7 @@ function that gets triggered when the button gets pressed (analogous to `onPress
 
 ### `onLongPress`
 
-function that gets triggered when the putton gets pressed for at least `delayLongPress` milliseconds.
+function that gets triggered when the button gets pressed for at least `delayLongPress` milliseconds.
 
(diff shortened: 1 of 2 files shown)


docs/docs/api/components/buttons.mdx
```diff
@@ -57,6 +57,10 @@ function that gets triggered when button changes from inactive to active and vic
 
 function that gets triggered when the button gets pressed (analogous to `onPress` in `TouchableHighlight` from RN core).
 
+### `onLongPress`
+
+function that gets triggered when the putton gets pressed for at least `delayLongPress` milliseconds.
```
tomekzaw, line 62:
```suggestion
function that gets triggered when the button gets pressed for at least `delayLongPress` milliseconds.
```

tomekzaw (changes requested):

tomekzaw (approved):

Issue #2094, opened 2022-06-06
[Android]The real default value of LongPressGestureHandler maxDist property is not 10 points


b1tjoy:
## Description
According to RNGH2 [documentation](https://docs.swmansion.com/react-native-gesture-handler/docs/gesture-handlers/api/longpress-gh#maxdist), the default value of maxDist is 10, expressed in points. But when long press and move finger within the initial 500ms less than 10 points on Android, LongPressGestureHandler will move to FAILED state.

The member variable which holds value for that inside LongPressGestureHandler is maxDistSq, meaning maxDist square, but expressed in pixels.

maxDistSq is initialized with:
https://github.com/software-mansion/react-native-gesture-handler/blob/94d0b52eba8db13aa185b38e43c5d9ed936d1ced/android/lib/src/main/java/com/swmansion/gesturehandler/LongPressGestureHandler.kt#L22-L23

But when we set prop maxDist of LongPressGestureHandler in React Native, maxDistSq is set to square of maxDist:
https://github.com/software-mansion/react-native-gesture-handler/blob/94d0b52eba8db13aa185b38e43c5d9ed936d1ced/android/lib/src/main/java/com/swmansion/gesturehandler/LongPressGestureHandler.kt#L32-L35

Google Pixel 5 for example, ppi is approximately 440, density independent pixel is 440 / 160 = 2.75, and the wrong default maxDistSq is set to `10 * density = 10 * 2.75 = 27.5`.
```JSX
maxDistSq: 27.5
maxDistPixel: Math.sqrt(27.5) = 5.244, expressed in pixels
maxDist: 5.244 / 2.75 = 1.907, expressed in points, MUCH SMALLER THAN 10 POINTS
```
when we move finger more than 1.907 points, the LongPressGestureHandler will fail.
https://github.com/software-mansion/react-native-gesture-handler/blob/94d0b52eba8db13aa185b38e43c5d9ed936d1ced/android/lib/src/main/java/com/swmansion/gesturehandler/LongPressGestureHandler.kt#L66-L72

The default maxDistSq should set to:
```Kotlin
defaultMaxDistSq = setMaxDist(DEFAULT_MAX_DIST_DP * context.resources.displayMetrics.density)
maxDistSq = defaultMaxDistSq
```

### Platforms
The issue has been verified on Android. After check RNGH2 source code, web may also be impacted.
- [ ] iOS
- [x] Android
- [x] Web

### Screenshots

[...]

PR #2095, opened 2022-06-08
Fix default max distance for long press on Android

j-piasecki:
## Description

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2094

As described in the issue, the `defaultMaxDistSq` property of `LongPressGestureHandler` was calculated in the wrong way - it was not squared. This resulted in long press gesture effectively  failing even during slightest moves if the `maxDist` prop was not configured.

## Test plan

Tested on the snack from the issue.


parasharrajat:
+1 for this. Waiting for merge.

tomekzaw (approved):

roryabraham:
Woohoo 🎉

PR #2097, opened 2022-06-09
Wrap Gesture Handler button with `View` and transform styles

j-piasecki:
## Description

There are many issues with styling buttons provided by Gesture Handler, mostly on Android where the class responsible for button extends `ViewGroup` instead of `ReactViewGroup` as it changes the behavior of onTouchEvent.

The best solution to those issues is wrapping the button with a `View` and styling the outer view instead. This PR attempts to make it happen auto-magically, which should make the buttons look as expected.

The main problem is correctly transforming the provided styles into two separate objects that when applied to two composed views give the same effect as the original styles applied to just one view.

## Test plan

For now, made sure that the changes don't break the Example app.

### DIFF
--- a/src/components/GestureButtons.tsx
+++ b/src/components/GestureButtons.tsx
@@ -6,6 +6,7 @@ import {
   StyleSheet,
   StyleProp,
   ViewStyle,
+  View,
 } from 'react-native';
 
 import createNativeWrapper from '../handlers/createNativeWrapper';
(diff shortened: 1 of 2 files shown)


src/components/GestureButtons.tsx
```diff
@@ -154,19 +157,79 @@ export class BaseButton extends React.Component<BaseButtonProps> {
...
+    paddingRight,
+    paddingStart,
+    paddingTop,
+    paddingVertical,
+    ...rest
+  } = StyleSheet.flatten(style) ?? {};
+
+  const fillParent =
+    // parent has its size specified
+    hasOneOfProperties(rest, ['width', 'height', 'flex']) ||
```
tomekzaw, line 205: Why not `'width' in rest || 'height' in rest || 'flex' in rest`?

j-piasecki:
I don't think this one is going places

PR #2104, opened 2022-06-14
fix: use different touchHandler on Fabric

WoLewicki:
## Description

On new architecture, the `RCTTouchHandler` has been replaced with `RCTSurfaceTouchHandler` ( see e.g. https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/React/Fabric/Mounting/ComponentViews/Modal/RCTFabricModalHostViewController.mm#L15). Old `cancel` method is not available there but same behavior is achieved through calling `setEnabled:NO` and then `setEnabled:YES`. ( see https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/React/Base/RCTTouchHandler.m#L362)

### DIFF
--- a/ios/RNGestureHandlerManager.mm
+++ b/ios/RNGestureHandlerManager.mm
@@ -215,12 +215,12 @@
 
 #ifdef RN_FABRIC_ENABLED
     RCTSurfaceTouchHandler *touchHandler = [viewWithTouchHandler performSelector:@selector(touchHandler)];
-    [touchHandler setEnabled:NO];
-    [touchHandler setEnabled:YES];
 #else
     RCTTouchHandler *touchHandler = [viewWithTouchHandler performSelector:@selector(touchHandler)];
-    [touchHandler cancel];
 #endif
+    [touchHandler setEnabled:NO];
+    [touchHandler setEnabled:YES];
+
 }
 
 #pragma mark Events


ios/RNGestureHandlerManager.mm
```diff
@@ -211,8 +213,14 @@ - (void)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
     if ([gestureRecognizer.view isKindOfClass:[UIScrollView class]]) return;
 
+#ifdef RN_FABRIC_ENABLED
+    RCTSurfaceTouchHandler *touchHandler = [viewWithTouchHandler performSelector:@selector(touchHandler)];
+    [touchHandler setEnabled:NO];
+    [touchHandler setEnabled:YES];
+#else
     RCTTouchHandler *touchHandler = [viewWithTouchHandler performSelector:@selector(touchHandler)];
     [touchHandler cancel];
+#endif
```
j-piasecki, line 223:
```suggestion
#ifdef RN_FABRIC_ENABLED
    RCTSurfaceTouchHandler *touchHandler = [viewWithTouchHandler performSelector:@selector(touchHandler)];
#else
    RCTTouchHandler *touchHandler = [viewWithTouchHandler performSelector:@selector(touchHandler)];
#endif
    // Cancel the touchHandler
    [touchHandler setEnabled:NO];
    [touchHandler setEnabled:YES];
```

Since `[touchHandler cancel]` does exactly the same thing, wouldn't it be cleaner this way to show that the same action is taken regardless of the `touchHandler`? (This is more of a question than a suggestion)

Issue #1067, opened 2020-05-12
ScrollView breaks RefreshControl on Android


janicduplessis:
When using ScrollView from rngh RefreshControl no longer works. This happens because on Android RefreshControl works by wrapping the ScrollView with a SwipeRefreshLayout component. This component should interact with the rngh gesture system since it needs to be able to interrupt and recognize simultaneously the ScrollView it wraps.

This can be kind of accomplished by wrapping RefreshControl with `createNativeWrapper` on Android. This is the first part of the hack patch I have to fix this:

GestureComponents.js
```diff
     });
   },
+  get RefreshControl() {
+    if (Platform.OS === 'android') {
+      return memoizeWrap(ReactNative.RefreshControl, {
+        disallowInterruption: true,
+        shouldCancelWhenOutside: false,
+      });
+    } else {
+      return ReactNative.RefreshControl;
+    }
+  },
   get Switch() {
     return memoizeWrap(ReactNative.Switch, {
```

The problem then is that ScrollView sets `disallowInterruption` to true, which means it cannot get interrupted or recognize with another gesture handler. Setting `disallowInterruption` to false makes RefreshControl work but causes other issues like nested ScrollView will both scroll at the same time.

I tried playing with adding `simultaneousHandlers` or `waitFor` to the RefreshControl associated with the ScrollView but wasn't able to get it working. Seems like `disallowInterruption` takes priority over that.

So at this point I was mostly looking for a hack to get it working so I came up with:

NativeViewGestureHandler.java
```diff
 import android.view.ViewGroup;
 
+import androidx.swiperefreshlayout.widget.SwipeRefreshLayout;
+
 public class NativeViewGestureHandler extends GestureHandler<NativeViewGestureHandler> {
 
   private boolean mShouldActivateOnStart;
@@ -48,7 +50,7 @@ public class NativeViewGestureHandler extends GestureHandler<NativeViewGestureHa
       }
     }
 
-    boolean canBeInterrupted = !mDisallowInterruption;
+    boolean canBeInterrupted = !shouldDisallowInterruptionBy(handler);
     int state = getState();
     int otherState = handler.getState();
 
@@ -62,9 +64,19 @@ public class NativeViewGestureHandler extends GestureHandler<NativeViewGestureHa
     return state == STATE_ACTIVE && canBeInterrupted;
   }
 
+  private boolean shouldDisallowInterruptionBy(GestureHandler handler) {
[...]

PR #2116, opened 2022-07-02
Fix RefreshControl on Android

janicduplessis:
## Description

Adds a hook to fix interaction between SwipeRefreshLayout and ScrollView. Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/1067.

## Test plan

Tested in an app that pull to refresh works properly after this fix.


janicduplessis:
Fixed in https://github.com/software-mansion/react-native-gesture-handler/pull/2137

PR #2124, opened 2022-07-13
Mention automatic workletization of callbacks in the docs

j-piasecki:
## Description

Add section about automatic workletization of gesture callbacks in the documentation alongside some examples.

### DIFF
--- a/docs/docs/introduction.md
+++ b/docs/docs/introduction.md
@@ -36,7 +36,7 @@ This should allow you to migrate your codebase from the gesture handlers to gest
 
 ### Automatic workletization of gesture callbacks
 
-Reanimated's Babel plugin is setup in a way that automatically marks callbacks passed to gestures in the configuration chain as worklets, this means that as long as all your callbacks are defined in a single chain, you don't need to add a `'worklet';` directive at the beginning of the functions. Here is an example that will be automatically workletized:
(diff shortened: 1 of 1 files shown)


docs/docs/introduction.md
```diff
@@ -34,6 +34,37 @@ Similarly, if you want to make a gesture simultaneous with (or wait for failure
 
 This should allow you to migrate your codebase from the gesture handlers to gestures smoothly and at your own pace. Just keep in mind that the gesture handlers cannot have the GestureDetector as their direct child, as it's a functional component.
 
+### Automatic workletization of gesture callbacks
+
+Reanimated's Babel plugin is setup in a way that automatically marks callbacks passed to gestures in the configuration chain as worklets, this means that as long as all your callbacks are defined in a single chain, you don't need to add a `'worklet';` directive at the beginning of the functions. Here is an example that will be automatically workletized:
```
tomekzaw, line 39:
```suggestion
Reanimated's Babel plugin is setup in a way that automatically marks callbacks passed to gestures in the configuration chain as worklets. This means that as long as all your callbacks are defined in a single chain, you don't need to add a `'worklet';` directive at the beginning of the functions. Here is an example that will be automatically workletized:
```

tomekzaw (changes requested):

tomekzaw (approved):

Issue #2091, opened 2022-06-04
[WEB] Pan Gesture and Native Gesture do not work simultaneously!


gorhom:
## Description

While working on supporting web for the Bottom Sheet, i noticed that the `Pan` gesture handler fails unexpectedly when interacting with a scroll view that is wrapped with `Native` gesture, this behaviour works just fine on mobile platforms ( iOS & Android ).

### Platforms

- [ ] iOS
- [ ] Android
- [x] Web

## Steps To Reproduce

- Run the provided snack
- Check the logs 
  - you will notice that the pan gesture fails at the very beginning of the interaction

### Expected behavior

Pan gesture and the scroll view should be active while interacting

### Actual behavior

Pan gesture fails early while the scroll view still active.

## Snack or minimal code example

https://snack.expo.dev/@gorhom/gh-web-issue

## Package versions

- React: 17.0.2
- React Native: 0.68.2
- React Native Gesture Handler: 2.2.1
- React Native Reanimated: 2.8.0

AAlvarez90:
I don't know if this issue is for the web or not but it doesn't work on IOS either. I am using expo 45 with those versions.
This is a simple example that won't work:
```import React  from 'react';
import { Dimensions, StyleSheet, View } from 'react-native';
import { PanGestureHandler, GestureHandlerRootView } from 'react-native-gesture-handler';
import Animated, { 
  useSharedValue,
  useAnimatedStyle,
  useAnimatedGestureHandler,
} from 'react-native-reanimated';

const {
  width: SCREEN_WIDTH
} = Dimensions.get('window');

const SIZE = 100.0;

const MagicCircle = () => {

    const x = useSharedValue(0);
    const y = useSharedValue(0);

    const gestureEvent = useAnimatedGestureHandler({
        onStart: (event, ctx) => {
          ctx.x = x.value;
          ctx.y = y.value;
        },
        onActive: (event, ctx) => {
          console.log(event);
            x.value = ctx.x + event.translateX;
            y.value = ctx.x + event.translateY;
        },
        onEnd: (event) => {

        }
    });

    const rStyle = useAnimatedStyle(() => {
        return {
            transform: [{translateX: x.value},{translateY: y.value} ]
        }
    });

    return (
      <GestureHandlerRootView style={styles.container} pointerEvents="none">
        <PanGestureHandler onGestureEvent={gestureEvent}>
          <Animated.View style={[styles.square, rStyle]} />
        </PanGestureHandler>
[...]

PR #2125, opened 2022-07-14
Make Gesture Handler on web respect `simultaneousHandlers` prop when used on a `ScrollView`

j-piasecki:
## Description

Should fix https://github.com/software-mansion/react-native-gesture-handler/issues/2091

HammerJS (which is used by Gesture Handler on web) by default tries to use [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events), which can be [cancelled](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointercancel_event) when the browser starts recognizing its own gestures (e.g. scrolling). Because of that all gestures would fail upon receiving `pointercancel` event. This is not the case when using [touch events](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events), which do not get cancelled.
This PR forces Hammer to use touch events when a gesture is simultaneous with a `NativeViewGestureHandler` and keeps its default behavior in other cases. This way the gesture keeps receiving events event when a browses is handling a gesture be itself.
Unfortunately, I don't think we can accomplish feature-parity between web and native platforms using Hammer. For example, one of the missing things is the possibility to prevent the scroll from starting.

## Test plan

Tested on the Example app and on the code from the issue.

### DIFF
--- a/src/web/GestureHandler.ts
+++ b/src/web/GestureHandler.ts
@@ -269,8 +269,7 @@ abstract class GestureHandler {
   shouldUseTouchEvents(config: Config) {
     return (
       config.simultaneousHandlers
-        ?.map((handler) => handler.isNative)
-        ?.reduce((prev, current) => prev || current, false) ?? false
+        ?.some((handler) => handler.isNative) ?? false
     );
   }
 
--- a/src/web/GestureHandler.ts
+++ b/src/web/GestureHandler.ts
@@ -268,8 +268,7 @@ abstract class GestureHandler {
 
   shouldUseTouchEvents(config: Config) {
     return (
-      config.simultaneousHandlers
-        ?.some((handler) => handler.isNative) ?? false
+      config.simultaneousHandlers?.some((handler) => handler.isNative) ?? false
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/web/GestureHandler.ts
```diff
@@ -61,6 +62,12 @@ abstract class GestureHandler {
     return `${this.name}${this.gestureInstance}`;
   }
 
+  // a simple way to check if GestureHandler is NativeViewGestureHandler, since importing it
+  // here to use instanceof woulf cause import cycle
```
tomekzaw, line 66:
```suggestion
  // here to use instanceof would cause import cycle
```

src/web/GestureHandler.ts
```diff
@@ -259,18 +266,37 @@ abstract class GestureHandler {
     }
   }
 
+  shouldUseTouchEvents(config: Config) {
+    return (
+      config.simultaneousHandlers
+        ?.map((handler) => handler.isNative)
+        ?.reduce((prev, current) => prev || current, false) ?? false
```
tomekzaw, line 273:
```suggestion
        ?.some((handler) => handler.isNative) ?? false
```

tomekzaw (changes requested):

tomekzaw:
Oops, looks like the changes I've suggested are not formatted properly.

tomekzaw (approved):

PR #2131, opened 2022-07-18
Display a more expressive error message when Gesture Handler receives more than one child

j-piasecki:
## Description

Display a more expressive error message when a Gesture Handler receives more than one child.

## Test plan

Tested on the Example app.

### DIFF
--- a/src/handlers/createHandler.tsx
+++ b/src/handlers/createHandler.tsx
@@ -469,42 +469,9 @@ export default function createHandler<
 
       this.propsRef.current = events;
 
+      let child: any = null;
       try {
-        const child: any = React.Children.only(this.props.children);
-        let grandChildren = child.props.children;
-        if (
-          __DEV__ &&
-          child.type &&
-          (child.type === 'RNGestureHandlerButton' ||
-            child.type.name === 'View' ||
-            child.type.displayName === 'View')
-        ) {
-          grandChildren = React.Children.toArray(grandChildren);
-          grandChildren.push(
-            <PressabilityDebugView
-              key="pressabilityDebugView"
-              color="mediumspringgreen"
-              hitSlop={child.props.hitSlop}
-            />
-          );
-        }
-
-        return React.cloneElement(
-          child,
-          {
-            ref: this.refHandler,
-            collapsable: false,
-            ...(isJestEnv()
-              ? {
-                  handlerType: name,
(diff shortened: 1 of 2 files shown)


src/handlers/createHandler.tsx
```diff
@@ -469,41 +469,49 @@ export default function createHandler<
...
-          child.type.displayName === 'View')
-      ) {
-        grandChildren = React.Children.toArray(grandChildren);
-        grandChildren.push(
-          <PressabilityDebugView
-            key="pressabilityDebugView"
-            color="mediumspringgreen"
-            hitSlop={child.props.hitSlop}
-          />
+      try {
```
tomekzaw, line 472: I wonder if it is possible to convert this enormous try/catch into something like
```ts
let child: any; // TODO: fix type
try {
  child = React.Children.only(this.props.children);
} catch (e) {
  throw new Error(...);
}
```
or in functional style:
```ts
let child: any = (() => {
  try {
    return React.Children.only(this.props.children);
  } catch (e) {
    throw new Error(...);
  }
})();
```
Obviously, the first variant has better performance.

Thanks to this change we won't get weird error messages if some other part of the code fails.

tomekzaw (changes requested):

src/handlers/createHandler.tsx
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/2131/commits/8933f4eb7bea0608e6eacc2775184c73bcfb3285.

src/handlers/gestures/GestureDetector.tsx
```diff
@@ -633,18 +633,26 @@ class Wrap extends React.Component<{
...
-    // correct viewTag to attach to.
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const child: any = React.Children.only(this.props.children);
-    return React.cloneElement(
-      child,
-      { collapsable: false },
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-      child.props.children
-    );
+    try {
```
tomekzaw, line 636: Same story here

tomekzaw (changes requested):

tomekzaw (approved):

PR #2132, opened 2022-07-19
Validate children of GestureDetector to check if there is only one native view that can be a target of the gestures

j-piasecki:
## Description

This PR adds a validation step during rendering of `GestureDetector` that traverses the Fiber tree to check if there is more that one view that could be a candidate to become the gesture target. If there are more the error message will be displayed.
For example, this is not ok:
```mermaid
graph TD;
    Wrap-->NativeView1;
    Wrap-->NativeView2;
```
but this is fine:
```mermaid
graph TD;
    Wrap-->NativeView1;
    NativeView1-->NativeView2;
    NativeView1-->NativeView3;
```

## Test plan

Tested on the Example app.


tomekzaw (approved):
👏

Issue #2143, opened 2022-07-26
RNGH example repo on branch 2.5.0 closes app without error on drag and drop example


xddq:
## Description
I had a weird issue with a click closing the app instantly without an error when trying out the 2.0 Gesture system. I could not figure out my error and tried running the example repo.
Basicly the same issue happens inside the drag-and-drop example repo here. It just closes without an error on click.

### Platforms

- [ ] iOS
- [x] Android
- [ ] Web

## Steps To Reproduce

1. Clone this repo, install deps and run the example on the android simulator.
2. Go to "Drag and Drop"
3. Click any box once

### Expected behavior
- nothing happens

### Actual behavior
- app closes without error message in metro

## Snack or minimal code example
- https://github.com/software-mansion/react-native-gesture-handler/tree/2.5.0/example
- tag 2.5.0 and ./example folder

## Package versions

- React: 17.0.2
- React Native: 0.68.0
- React Native Gesture Handler: 2.5.0
- React Native Reanimated: 2.6.0

j-piasecki:
Hi! It's most likely caused by the callback of the gesture not being worklet or you trying to call a non-worklet function from the callback.

Here's description of when the callbacks are, and when they aren't, automatically workletized: https://github.com/software-mansion/react-native-gesture-handler/issues/1821#issuecomment-1057005397. In case you want to declare a function outside the gesture chain and simply pass it as an argument, you need to add `'worklet'` directive at the beginning of the function.

If you need to call a non-worklet function from a worklet, you need to wrap it with [`runOnJS`](https://docs.swmansion.com/react-native-reanimated/docs/api/miscellaneous/runOnJS/).

PR #2144, opened 2022-07-26
Add `runOnJS` where necessary in Drag and Drop example

j-piasecki:
## Description

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2143

Drag and drop example was trying to run non-worklet functions on UI thread, this resulted in the app crashing in some cases. This PR wraps those functions in `runOnJS`, preventing crashes.

## Test plan

Checked if the example app crashes

### DIFF
--- a/example/src/new_api/drag_n_drop/DragAndDrop.tsx
+++ b/example/src/new_api/drag_n_drop/DragAndDrop.tsx
@@ -12,7 +12,6 @@ import {
   PanGestureHandlerEventPayload,
 } from 'react-native-gesture-handler';
 import {
-  runOnJS,
   useSharedValue,
   withSpring,
   withTiming,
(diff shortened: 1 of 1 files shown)


tomekzaw (changes requested):
plz use `.runOnJS()` modifier instead

example/src/new_api/drag_n_drop/DragAndDrop.tsx
```diff
@@ -200,12 +200,21 @@ function DragAndDrop<T extends { id: string }>({
...
-    }
-  });
+  const dragGesture = Gesture.Pan()
+    .runOnJS(true)
+    .onUpdate((e) => {
+      onUpdateHandler(e);
+    })
+    .onEnd(() => {
+      onDragEnd();
+    });
```
tomekzaw, line 210: Why not ...?
```suggestion
    .onUpdate(onUpdateHandler)
    .onEnd(onDragEnd);
```

tomekzaw (approved):

PR #2148, opened 2022-07-28
fix: remove explicit folly version

kacperkapusciak:
## Description

As a version of folly can change between versions of React-native we need to remove an explicit version of `folly_version` from our libraries to let RN infer the correct version.

Change based on https://github.com/th3rdwave/react-native-safe-area-context/commit/eb111ff4eebc8d06ca01e7d12d9661147d03da20

## Changes

Removed explicit folly version from RNGestureHandler.podspec

## Test code and steps to reproduce

Build FabricExample/

## Checklist

- [ ] Ensured that CI passes


j-piasecki (approved):
❤️

PR #2154, opened 2022-08-05
Possibly fix Windows builds failing because of long paths

j-piasecki:
## Description

This PR adds `NDK_APP_SHORT_COMMANDS=true` argument to the ndk build on Android.

Possibly fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2122.


j-piasecki:
NDK has been replaced with CMake

PR #2155, opened 2022-08-05
Detect duplicate instances

piaskowyk:
## Description

PR contains the fix for duplicate instances of the library in project dependencies. Detailed description in PR to Reanimated repo - https://github.com/software-mansion/react-native-reanimated/pull/3441

## Test plan

```
npx react-native init tester
cd tester
yarn add react-native-gesture-handler
yarn add <lib with strong dependency to Gesture Handler>
// try to build app
```

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -57,7 +57,7 @@ If you use `npm` you should add [`overrides` property](https://docs.npmjs.com/cl
 }
 ```
 
-After that you need to run you package manager again
+After that you need to run your package manager again
 ```bash
 yarn
 ```
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -33,3 +33,35 @@ To make it easier for us to understand your issue and to be able to approach it 
 - `Native` gesture may not conform to the standard state flow due to platform specific workarounds to incorporate native views into RNGH.
 - Keep in mind that `Touchables` from RNGH are rendering two additional views that may need to be styled separately to achieve desired effect (`style` and `containerStyle` props).
 - In order for the gesture composition to work, all composed gestures must be attached to the same `GestureHandlerRootView`.
+
+### Multiple versions of Gesture Handler were detected
+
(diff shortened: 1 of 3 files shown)


docs/docs/troubleshooting.md
```diff
@@ -33,3 +33,35 @@ To make it easier for us to understand your issue and to be able to approach it
...
+```
+
+If you use `npm` you should add [`overrides` property](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides).
+```json
+"overrides": {
+  "react-native-gesture-handler": <Gesture Handler version>
+}
+```
+
+After that you need to run you package manager again
```
j-piasecki, line 60:
```suggestion
After that you need to run your package manager again
```

docs/docs/troubleshooting.md
```diff
@@ -33,3 +33,35 @@ To make it easier for us to understand your issue and to be able to approach it
 - `Native` gesture may not conform to the standard state flow due to platform specific workarounds to incorporate native views into RNGH.
 - Keep in mind that `Touchables` from RNGH are rendering two additional views that may need to be styled separately to achieve desired effect (`style` and `containerStyle` props).
 - In order for the gesture composition to work, all composed gestures must be attached to the same `GestureHandlerRootView`.
+
+### Multiple versions of Gesture Handler were detected
+
+This error usually happens when in your project there exists more than one instance of Gesture Handler. It can occur when some of your dependency has installed Gesture Handler inside their own `node_modules` instead of using it as a peer dependency. In this case two different versions of Gesture Handler JS module try to install the same Native Module. You can resolve this problem manually by modifying your `package.json` file.
```
j-piasecki, line 39:
```suggestion
This error usually happens when in your project there exists more than one instance of Gesture Handler. It can occur when some of your dependencies have installed Gesture Handler inside their own `node_modules` instead of using it as a peer dependency. In this case two different versions of Gesture Handler JS module try to install the same Native Module. You can resolve this problem manually by modifying your `package.json` file.
```

j-piasecki (changes requested):

j-piasecki (approved):

Issue #2138, opened 2022-07-24
focalX and focalY are wrong on android only


wood1986:
## Description

I am trying to implement pinch to zoom and I pretty sure my math is correct. But the focalX and focalY are wrong on android only

Android
![android](https://user-images.githubusercontent.com/5212215/180639025-16401a93-905d-42f9-9d1a-d8841b3965e8.gif)

iOS
![ios](https://user-images.githubusercontent.com/5212215/180639033-b442d225-586f-4044-ad2b-759a4e24ebe5.gif)

On android, the first pinch-to-zoom behave "correct". The last pinch-to-zoom must be wrong. I tap two points and the coordinates are `x = 194.86 y = 351.61` and `x = 194.32 y = 396.36` . Then I pinch and the focal point is `x = 195.93 y = 464.81`. I expect the y-coord should be between two taps point. Now it is outside. `351.61 <= 464.81 <= 396.36 ❌ ❌ ❌ ❌ `. x-coord has the same issue.

When you take a look iOS, it behaves correctly `313 <= 351 <= 374 ✅✅✅✅`

I think the bug is on either react-native or react-native-gesture-handler. I do not think it is react or react-native-reanimated bug

### Platforms

- [ ] iOS
- [x] Android
- [ ] Web

## Steps To Reproduce

1. `git clone https://github.com/wood1986/pinch-bug.git`
2. `yarn android`

### Expected behavior

`tap[0].x <= focalX <= tap[1].x`
`tap[0].y <= focalY <= tap[1].y`

### Actual behavior

`focalX <= tap[0].x or tap[1].x <= focalX`
`focalY <= tap[0].y or tap[1].y <= focalY`

## Package versions

- React: 18.0.0
- React Native: 0.69.2
- React Native Gesture Handler: 2.5.0
- React Native Reanimated: 2.9.1

wood1986:
https://github.com/software-mansion/react-native-gesture-handler/blob/3ff7674fe67041d5ca6e80caf768fb62908c11cc/android/lib/src/main/java/com/swmansion/gesturehandler/ScaleGestureDetector.java#L285-L292

if I change it to getRawX() amd getRawY(), it behaves correctly

j-piasecki:
Hi! I tried to reproduce it on a simple view but it seems to be working fine in that case. Would you mind sharing the repository with the application you used to make a recordings?

wood1986:
In the repro step, I have included my sample code. 

[...]

PR #2156, opened 2022-08-08
Transform entire `MotionEvent` instead of single pointer only

j-piasecki:
## Description

Gesture Handler on Android intercepts events via the `GestureHandlerRootView`, which means that we need to transform the touch coordinated by ourselves to check if it's inside the view. Similarly, we need to transform the event when dispatching it to one of the gesture handlers in order to correctly calculate the changes and send touch events.

The current method works well until there is more than one pointer at once on the screen. When there is more than one pointer, only the first one will be correctly transformed, while others will simply be moved by the offset calculated when transforming the first pointer.

This PR changes the logic when delivering events to gesture handlers - the whole event is transformed, and handlers receive both events - transformed and untransformed one. This way things like touch events and `NativeViewGestureHandler` which rely on the location in the coordinate space of the view keep working, and other gestures get to work in the coordinate space of the root view.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2138.

## Test plan

Tested on the Example app.


wood1986 (approved):
I tested in my app. it works. But I do not know other area

wood1986:
Hey @j-piasecki as this PR has been a month, May I know when you merge into master and publish a new version?

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
```diff
@@ -305,7 +296,7 @@ class GestureHandlerOrchestrator(
       }
     }
 
-    event.setLocation(oldX, oldY)
+    event.recycle()
```
kmagiera, line 299: are you sure it is safe to recycle event here? It seem like `transormEventToViewCoords` may sometimes create a new event that should be recycled and other times it may reuse the one provided to it. The latter case matches the previous behavior of this method and we never called recycle here. Not saying this is bad but just look suspicious.

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
```diff
@@ -329,19 +320,71 @@ class GestureHandlerOrchestrator(
...
-      return
+  /**
+   * Transforms an event in the coordinates of wrapperView into the coordinate space of the received view.
+   *
+   * This modifies and returns the same event as it receives
+   *
+   * @param view - view to which coordinate space the event should be transformed
+   * @param event - event to transform
+   */
+  fun transformEventToViewCoords(view: View?, event: MotionEvent): MotionEvent {
```
kmagiera, line 331: not sure about the naming of this method. It suggests that it mutates the provided object while sometimes it doesn't. Also, it looks like it mutates the event in a very few scenarios. Unless I don't understand this well it appears like it only mutates if the view is the wrapperView (which happens only at the very top of the view hierarchy) and when view is `null` which I don't think should ever happen?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -1,6 +1,7 @@
 package com.swmansion.gesturehandler.react
 
 import android.content.Context
+import android.graphics.PointF
```
kmagiera, line 4: revert

kmagiera (commented):
Most of the changes look good. Accepting to unblock. Please take a look at my comments to GestureHandlerOrchestrator.kt as some of the changes there look a bit suspicious

kmagiera (approved):
well, forgot to check 'approve' with my last comment

wood1986:
Thanks. May I know when it will be released? Is it a patch or minor or major release?

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
j-piasecki: It traverses the view hierarchy upwards until it finds `wrapperView` (which is an instance of `GestureHandlerRootView`) or `null` (in case there is no root view we don't want to crash the app) and then comes back recursively transforming the event using the same matrices that are used to transform views.

It will not mutate the provided event in the following cases:
- `null` is passed as an argument for `view`
- all the views on the path from the `wrapperView` to the provided view aren't transformed (their matrix is identity) and are positioned at the top-left corner of their parents

As for the name, I wanted to make It obvious that it doesn't return a new object but may transform the one provided as an argument so it's not safe to use if the untransformed event is also needed. Semantically I believe it's fine since the only case where the event is not transformed is when the coordinate space of the provided view is the same as the wrapper view.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/2156/commits/f4056390b05eea7d82f001fb6e8d4a1b16b4f8e9

android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt
j-piasecki: `transormEventToViewCoords` never creates a new event, it always transforms the one it receives but I can see how the code of the method might've suggested otherwise. To be hones I don't remember why I've done it like that (it might've been a result of some logic that used to be there), but I've simplified it in https://github.com/software-mansion/react-native-gesture-handler/pull/2156/commits/78c37cd5f7984ca0d7590d76c00aef7513a71574 so it should be much clearer that it doesn't create a new event.

That said, It should be safe to recycle there since `transformEventToViewCoords` receives (and returns transformed) a copy of the source event: https://github.com/software-mansion/react-native-gesture-handler/blob/78c37cd5f7984ca0d7590d76c00aef7513a71574/android/lib/src/main/java/com/swmansion/gesturehandler/GestureHandlerOrchestrator.kt#L255

j-piasecki:
@wood1986 Just a heads up, GH 2.7.0 has been released yesterday.

PR #2173, opened 2022-08-18
Make grep pattern for detecting multiple installations stronger

kkafar:
## Description

Previous pattern caused problems when library was used directly from GitHub (not downloaded from package repository) as there were multiple `package.json` files with `"/react-native-gesture-handler/"` in their path (example applications, etc.).

This commit narrows down the search to look only for "main package.json" files.

## Test plan

Install `react-native-gesture-handler` directly from repository and see that it works after applying this patch.

### DIFF
--- a/RNGestureHandler.podspec
+++ b/RNGestureHandler.podspec
@@ -4,7 +4,7 @@ fabric_enabled = ENV['RCT_NEW_ARCH_ENABLED']
 
 isUserApp = File.exists?(File.join(__dir__, "..", "..", "node_modules", "react-native", "package.json"))
 if isUserApp
-  libInstances = %x[find ../../ -name "package.json" | grep "/react-native-gesture-handler/"]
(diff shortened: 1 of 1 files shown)


j-piasecki (approved):

tomekzaw (approved):

tomekzaw:
Good catch! 👏 Can we also have:
* Android 🙏 
* analogous PR to Reanimated (it has only one `package.json` but let's keep this chunk of code identical) 🙏

kkafar:
@tomekzaw 

I partially fixed Android implementation here in [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2174).

Also I issued [PR for Android](https://github.com/software-mansion/react-native-reanimated/pull/3493) in `react-native-reanimated` repo. However it seems to me that [iOS implementation](https://github.com/software-mansion/react-native-reanimated/blob/f310408c0fdb39b14c178dd1909bf9a82f8af657/RNReanimated.podspec#L49) is fine there.

PR #2177, opened 2022-08-22
Touch events web

Warus15:
## Description

Main change in this PR is adding support for TouchEvents. With this change, all touches are handled by TouchEvents, while pointers such as mouse or pen are still being recognized as PointerEvents. `EventManager` now has two subclasses, one for touch and one for pointer events. There is also change in logic of detecting whether we should call `onPointerDown` or `onPointerAdd` (and the same for their counterparts). Each event manager tracks how many pointers are currently being handled. With that change, gesture handlers no longer have to decide which method to use.

There are also some minor changes, like adding activation after long press for PanGestureHandler and modfications of NativeViewGestureHandler (especially changes in activation criterias), fail after given distance in TapGestureHandler and fix of RotationGestureHandler

## Test plan

Tested on example app


src/web/handlers/GestureHandler.ts
```diff
@@ -35,6 +32,7 @@ export default abstract class GestureHandler {
   protected awaiting = false;
   protected active = false;
   protected shouldResetProgress = false;
+  protected pointerType: string | null = null;
```
j-piasecki, line 35: This can potentially be an enum

src/web/handlers/TapGestureHandler.ts
```diff
@@ -61,8 +61,8 @@ export default class TapGestureHandler extends GestureHandler {
       this.maxDeltaY = this.config.maxDeltaY;
     }
 
-    if (this.config.maxDistSq !== undefined) {
-      this.maxDistSq = this.config.maxDistSq;
+    if (this.config.maxDist !== undefined) {
+      this.maxDist = this.config.maxDist;
     }
```
j-piasecki, line 66:
```suggestion
    if (this.config.maxDistSq !== undefined) {
      this.maxDistSq = this.config.maxDist * this.config.maxDist;
    }
```

Right now, in the `shouldFail` function, you're comparing squared distance to non-squared one

src/web/tools/EventManager.ts
```diff
@@ -1,186 +1,103 @@
...
-      offsetY: event.offsetY,
-      pointerId: event.pointerId,
-      eventType: eventType,
-      pointerType: event.pointerType,
-      buttons: event.buttons,
-      time: event.timeStamp,
-    };
-  }
-
   public isPointerInBounds({ x, y }: { x: number; y: number }): boolean {
```
j-piasecki, line 66: This function is also in the `GestureHandler`, maybe moving it out to some `utils` makes sense?

src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -320,9 +326,43 @@ export default class GestureHandlerOrchestrator {
     );
   }
 
+  // This function is called when handler receives touchdown event
+  // If handler is using mouse or pen as a pointer and any handler receives touch event,
+  // mouse/pen event dissappears - it doesn't send onPointerCancel nor onPointerUp (and others)
+  // This became a problem because handler was left at active state without any signal to end or fail
+  // To handle this, when new touch event is received, we loop through active handlers and check which type of
+  // pointer they're using
+  // If there are any handler with mouse/pen as a pointer, we cancel them
```
j-piasecki, line 335:
```suggestion
  // pointer they're using. If there are any handler with mouse/pen as a pointer, we cancel them
```

src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -320,9 +326,43 @@ export default class GestureHandlerOrchestrator {
...
+      }
+
+      if (handler === currentHandler) {
+        // Handler that received touch event should have its pointer tracker reseted
+        // This allows handler to smoothly change from mouse/pen to touch
+        // The drawback is, that when we try to use mouse/pen one more time, it doesn't send onPointerDown at the first time
+        // so it is required to click two times to get handler to work
+        //
+        // However, handler will receive mannualy created onPointerEnter that is triggered in EventManager in onPointerMove method.
+        // There may be possibility to use that fact to make handler respond properly to first mouse click
```
j-piasecki, line 356:
```suggestion
        // Handler that received touch event should have its pointer tracker reset
        // This allows handler to smoothly change from mouse/pen to touch
        // The drawback is, that when we try to use mouse/pen one more time, it doesn't send onPointerDown at the first time
        // so it is required to click two times to get handler to work
        //
        // However, handler will receive manually created onPointerEnter that is triggered in EventManager in onPointerMove method.
        // There may be possibility to use that fact to make handler respond properly to first mouse click
```

src/web/tools/PointerEventManager.ts
```diff
@@ -0,0 +1,125 @@
...
+    });
+
+    this.view.addEventListener('pointerup', (event: PointerEvent): void => {
+      if (event.pointerType === 'touch') {
+        return;
+      }
+
+      if (this.activePointersCounter === 0) {
+        return;
+      }
```
j-piasecki, line 37: I'd add a comment explaining why it may happen

src/web/tools/PointerEventManager.ts
```diff
@@ -0,0 +1,125 @@
...
+      target.releasePointerCapture(adaptedEvent.pointerId);
+      this.markAsOutOfBounds(adaptedEvent.pointerId);
+
+      if (--this.activePointersCounter > 0) {
+        adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_UP;
+        this.onPointerRemove(adaptedEvent);
+        return;
+      }
+
+      this.onPointerUp(adaptedEvent);
```
j-piasecki, line 51:
```suggestion
      if (--this.activePointersCounter > 0) {
        adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_UP;
        this.onPointerRemove(adaptedEvent);
      } else {
        this.onPointerUp(adaptedEvent);
      }
```

This may be more of a personal preference, but this looks more obvious to me. We can talk about it if you have a different opinion.

src/web/tools/PointerEventManager.ts
```diff
@@ -0,0 +1,125 @@
...
+      target.setPointerCapture(adaptedEvent.pointerId);
+      this.markAsInBounds(adaptedEvent.pointerId);
+
+      if (++this.activePointersCounter > 1) {
+        adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_DOWN;
+        this.onPointerAdd(adaptedEvent);
+        return;
+      }
+
+      this.onPointerDown(adaptedEvent);
```
j-piasecki, line 27:
```suggestion
      if (++this.activePointersCounter > 1) {
        adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_DOWN;
        this.onPointerAdd(adaptedEvent);
      } else {
        this.onPointerDown(adaptedEvent);
      }
```

Same as above

src/web/tools/TouchEventManager.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        this.markAsInBounds(adaptedEvent.pointerId);
+
+        if (++this.activePointersCounter > 1) {
+          adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_DOWN;
+          this.onPointerAdd(adaptedEvent);
+          continue;
+        }
+
+        this.onPointerDown(adaptedEvent);
```
j-piasecki, line 31:
```suggestion
        if (++this.activePointersCounter > 1) {
          adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_DOWN;
          this.onPointerAdd(adaptedEvent);
        } else {
          this.onPointerDown(adaptedEvent);
        }
```

See above

src/web/tools/TouchEventManager.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        this.markAsOutOfBounds(adaptedEvent.pointerId);
+
+        if (--this.activePointersCounter > 0) {
+          adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_UP;
+          this.onPointerRemove(adaptedEvent);
+          continue;
+        }
+
+        this.onPointerUp(adaptedEvent);
```
j-piasecki, line 92:
```suggestion
        if (--this.activePointersCounter > 0) {
          adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_UP;
          this.onPointerRemove(adaptedEvent);
        } else {
          this.onPointerUp(adaptedEvent);
        }
```

See above

j-piasecki (changes requested):
Looks good 🚀, left some comments though

src/web/tools/PointerEventManager.ts
Warus15: Could be, but in this case I agree, that the ones You suggest look better

j-piasecki (approved):

PR #2180, opened 2022-08-23
Use angle-bracket imports

tomekzaw:
## Description

This PR replaces bad double-quotes imports (which break Swift interop) with good angle-bracket imports.

The forbidden pattern is: `#import "RCT`

See also:
* https://github.com/software-mansion/react-native-reanimated/pull/3150
* https://github.com/software-mansion/react-native-screens/pull/1572
* https://github.com/react-native-svg/react-native-svg/pull/1848

## Test plan

Check if Example and FabricExample apps compile successfully.


tsapeta (approved):
Thank you 🙏 💚

j-piasecki (approved):

Issue #2160, opened 2022-08-09
<RectButton> becomes unresponsive when used within a List


nujhong:
## Description

Using a **RectButton** in a List component (**FlatList**/**SectionList**) will become unresponsive in certain scenarios. I am not sure whether this is an incorrect usage of this library or it's something new with RN. _This wasn't an issue in production before._

- It's observed that the [event handler](https://github.com/software-mansion/react-native-gesture-handler/blob/main/src/components/GestureButtons.tsx#L140) is not called when it's unresponsive
- ~~It's observed that when **RectButton** is unresponsive, it's also unresponsive to [swipe up to close the android app](https://support.google.com/android/answer/9079646?hl=en#:~:text=Close%20one%20app%3A%20Swipe%20up,Swipe%20from%20left%20to%20right.)~~
- It's observed that using the same example, it cannot be reproduced on Snack
- Finally, setting `shouldActivateOnStart=true` on **RectButton** seems to solve the issue 
- Additionally, I initially thought it's because I'm not using **RectButton** inside RNGH's **ScrollView** so I've added a **SectionList** export via `patch-package`. This also seems to solve the issue other than setting `shouldActivateOnStart=true`.

### Platforms

- [ ] iOS
- [x] Android
- [ ] Web

### Screenshots

https://user-images.githubusercontent.com/23186058/183584450-ceebbea3-9298-4fb4-abcf-b0bf30b29cd5.mov

https://user-images.githubusercontent.com/23186058/183584650-a8e130da-a4db-4100-94f3-fcbbed814f91.mov

## Steps To Reproduce
**Using RN's SectionList**

1. Scroll multiple times to the bottom of the list
2. All <RectButton> instances become unresponsive
3. Navigate to "Solution" by pressing the button in Header
4. All <RectButton shouldActivateOnStart> instances are still responsive (however the ripple effect is gone)
5. Exit and reload the app and navigate straight to "Solution" and repeat **Step 1**
6. Everything works fine

---
**Using re-exported SectionList by overriding its ScrollViewComponent with RNGH's ScrollView via `renderScrollComponent`**

1. Apply the package by running `yarn patch-package`
2. Scroll multiple times to the bottom of the list
3. Everything works fine

### Expected behavior

- **RectButton** does not become unresponsive when used within a List component on Android

### Actual behavior

- **RectButton** becomes unresponsive when used within a List component Android

## Snack or minimal code example

[...]

PR #2187, opened 2022-08-29
Fix nested buttons on Android and buttons becoming unresponsive in a list

j-piasecki:
## Description

Should fix https://github.com/software-mansion/react-native-gesture-handler/issues/2160

This PR changes the logic of responder by splitting it in two: one for sound and one for touch exclusivity. Button can become the touch responder when `setPressed` method is called with `pressed` set to `true` and it also is the touch responder. The sound responder is freed in the `performClick` method - after that point `setPressed` should only be called with `pressed` set to `false` for existing pointers. This prevents the sound effect from being played more than one time when pressing on nested buttons.

It also adds more ways to free the responder: it reverts a change from a while back that frees responder on gesture end and frees the responder when button holding it receives `POINTER_CANCEL` event. Because of that, the responder should always be freed correctly, preventing the buttons from becoming unresponsive.

## Test plan

Tested on the example app (using the new and existing example) and on the code from the issue.


hirbod:
Looks like the fix works @j-piasecki - can we get this shipped quickly please? But I am still not sure if it fixed the issue in general. Its happen sporadically and I can't say with 100% confidence that it has fixed all occurrences , because casual RectButtons have been breaking as well. (even when not nested etC)

tomekzaw (approved):
👍

PR #2190, opened 2022-08-30
Remove event parameter

Warus15:
## Description

This PR removes event from parameters of all state managing functions. Doing this brings us a step closer to achieve working `ManualGestureHandler`, because right now it's impossible to move it to any state without passing event, to which this handler doesn't have access.

## Test plan

Tested on example app


src/web/handlers/GestureHandler.ts
```diff
@@ -125,67 +125,67 @@ export default abstract class GestureHandler {
...
+  public end() {
     if (
       this.currentState === State.BEGAN ||
       this.currentState === State.ACTIVE
     ) {
-      this.moveToState(State.END, event);
+      this.moveToState(State.END);
       this.view.style.cursor = 'auto';
     }
+    // this.currentState = State.UNDETERMINED;
```
j-piasecki, line 188:
```suggestion
```

src/web/tools/PointerTracker.ts
```diff
@@ -107,12 +115,49 @@ export default class PointerTracker {
   public getVelocityY(pointerId: number): number {
     return this.trackedPointers.get(pointerId)?.velocityY as number;
   }
-  public getLastX(pointerId: number): number {
-    return this.trackedPointers.get(pointerId)?.lastX as number;
+
+  /**
+   * Returns X coordinate of last tracked pointer
```
j-piasecki, line 120:
```suggestion
   * Returns X coordinate of last moved pointer
```

src/web/tools/PointerTracker.ts
```diff
@@ -107,12 +115,49 @@ export default class PointerTracker {
...
+      return this.trackedPointers.get(pointerId)?.lastX as number;
+    } else {
+      return this.trackedPointers.get(this.lastMovedPointerId)?.lastX as number;
+    }
   }
-  public getLastY(pointerId: number): number {
-    return this.trackedPointers.get(pointerId)?.lastY as number;
+
+  /**
+   * Returns Y coordinate of last tracked pointer
```
j-piasecki, line 141:
```suggestion
   * Returns Y coordinate of last moved pointer
```

j-piasecki (approved):
LGTM, please update the comments before merging

PR #2191, opened 2022-08-30
Adding state manager on web

Warus15:
## Description

This PR brings gesture state manager to new web implementation. With this change, `ManualGestureHandler` is now officially supported on web.

## Test plan

Tested on example app


src/web/handlers/GestureHandler.ts
```diff
@@ -185,6 +185,7 @@ export default abstract class GestureHandler {
       this.moveToState(State.END);
       this.view.style.cursor = 'auto';
     }
+    // this.currentState = State.UNDETERMINED;
```
j-piasecki, line 188:
```suggestion
```

j-piasecki (changes requested):

j-piasecki (approved):

PR #2206, opened 2022-09-07
Add support for enable

Warus15:
## Description

This PR adds support for `enabled` prop. Now when enabled is set to false, handler won't send events. If handler was `active` and enabled is set to false, handler fails and if its state was `began`, handler will cancel.

## Test plan

Tested on example app


src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -89,8 +95,13 @@ export default class GestureHandlerOrchestrator {
   public onHandlerStateChange(
     handler: GestureHandler,
     newState: State,
-    oldState: State
+    oldState: State,
+    forceStateChange?: boolean
```
j-piasecki, line 99:
```suggestion
    sendIfDisabled?: boolean
```
I'd name it to be more descriptive, as it's used only for sending events to disabled handlers. (I'm not really convinced of the one I've proposed in case you have better ideas)

j-piasecki (changes requested):

j-piasecki (approved):

PR #2207, opened 2022-09-07
Pointer events to touch events mapping

Warus15:
## Description

This PR adds mapping of pointer events to touch events. This change allows us to send touch events with pointers such as mouse or stylus, therefore making `ManualGestureHandler` usable with these pointers

## Test plan

Tested on example app


src/web/handlers/GestureHandler.ts
```diff
@@ -394,62 +397,93 @@ export default abstract class GestureHandler {
...
-        event.allTouches[i].identifier
-      );
+    const trackerData = this.tracker.getData();
 
-      // When we simultaneously add 2 (or more) pointers to handler, both allTouches and changedTouches contains information
-      // about all pointers. Because first pointer tries to send event before second was added to tracker, mapped id of second pointer will be NaN.
-      // To avoid crashes, we check whether current id is NaN and, if that's the case, we return undfined and don't send event.
+    // This if handles edge case where all pointers have been cancelled
+    // When pointercancel is triggered, reset method is called. This means that tracker will be reset after first pointer being cancelled
+    // The problem is, that handler will receive another pointercancel event from rest oh the pointers
```
j-piasecki, line 409:
```suggestion
    // The problem is, that handler will receive another pointercancel event from the rest of the pointers
```

src/web/handlers/GestureHandler.ts
```diff
@@ -394,62 +397,93 @@ export default abstract class GestureHandler {
...
-        absoluteX: event.allTouches[i].clientX,
-        absoluteY: event.allTouches[i].clientY,
+        x: element.lastX - rect.left,
+        y: element.lastY - rect.top,
+        absoluteX: element.lastX,
+        absoluteY: element.lastY,
+      });
+    });
+
+    // Each pointer send its own event, so we want changed touches to contain only the pointer that has changed.
```
j-piasecki, line 427:
```suggestion
    // Each pointer sends its own event, so we want changed touches to contain only the pointer that has changed.
```

src/web/tools/TouchEventManager.ts
```diff
@@ -23,7 +23,9 @@ export default class TouchEventManager extends EventManager {
           !isPointerInBounds(this.view, {
             x: adaptedEvent.x,
             y: adaptedEvent.y,
-          })
+          }) ||
+          //@ts-ignore touchType field does exist
+          event.changedTouches[i].touchType === 'stylus'
```
j-piasecki, line 28: I'd comment why you're skipping stylus

src/web/handlers/GestureHandler.ts
```diff
@@ -394,62 +397,93 @@ export default abstract class GestureHandler {
...
-        absoluteX: event.changedTouches[i].clientX,
-        absoluteY: event.changedTouches[i].clientY,
-      });
+    let numberOfTouches: number = all.length;
+
+    if (
+      event.eventType === EventTypes.UP ||
+      event.eventType === EventTypes.ADDITIONAL_POINTER_UP
+    ) {
+      --numberOfTouches;
```
j-piasecki, line 476: I assume this is to be compatible with native implementation, if so I'd mention it in comment

example/src/new_api/manualGestures/index.tsx
```diff
@@ -82,7 +82,7 @@ export default function Example() {
       }
     })
     .onTouchesCancelled((e, manager) => {
-      for (const touch of e.changedTouches) {
+      for (const touch of e.allTouches) {
```
j-piasecki, line 85: In case of cancel, wouldn't both arrays contain exactly the same pointers?

j-piasecki (changes requested):

example/src/new_api/manualGestures/index.tsx
Warus15: At first they didn't and that's why I changed that. Then I slightly changed cancel behavior and right now they do contain both. However I left that change to emphasize that we do cancel every pointer

j-piasecki (approved):

PR #2210, opened 2022-09-08
Nested touchables fix

Warus15:
## Description

This PR fixes nested touchables example (unfortunately exclusive prop still doesn't work as it should). Also, with changes in module file, combo example can be opened on web.

In orchestrator there's change in order of if statements - that's necessary, because now handlers check if they should be cancelled by others before they reach `makeActive` method. If the order wasn't flipped, overlapping handlers would be canceled, even though they shouldn't be (that was tested on overlapping parents example).

Other minor change is setting `disallowInterruption` prop on `ScrollView` to false by default. This solves issue where pan was canceled by scroll in combo example.

## Test plan

Tested on example app


src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -73,13 +73,24 @@ export default class GestureHandlerOrchestrator {
     if (this.hasOtherHandlerToWaitFor(handler)) {
       this.addAwaitingHandler(handler);
     } else if (
+      this.shouldActivate(handler) &&
```
j-piasecki, line 76: Have you checked if handlers not activated because of this check are being correctly canceled/are failing?

src/RNGestureHandlerModule.web.ts
```diff
@@ -88,13 +90,22 @@ export default {
...
   attachGestureHandler(
     handlerTag: number,
-    newView: number,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    newView: any,
     _actionType: ActionType,
     propsRef: React.RefObject<unknown>
   ) {
+    if (
+      !(newView instanceof HTMLElement || newView instanceof React.Component)
```
j-piasecki, line 99: Now that I think about it, I'm not sure whether `svg` inherits after `HTMLElement`. Could you check that?

src/web/tools/GestureHandlerOrchestrator.ts
Warus15: Good point, they should fail/be canceled right away.

j-piasecki (approved):

PR #2212, opened 2022-09-09
Cache averages in PointerTracker

Warus15:
## Description

This PR adds caching average coordinates in `PointerTracker`. Some handlers (like `pan`) send average values in native event. The problem is, that these events are being sent after pointers are removed from tracker. This means, that tracker may return NaN as averege, when received event was `pointerup`. In that case, cached value should be returned.

## Test plan

Tested on example app


src/web/tools/PointerTracker.ts
```diff
@@ -156,11 +171,17 @@ export default class PointerTracker {
...
   }
 
+  // Some handlers use these methods to send average values in native event.
+  // This may happen when pointers have already been removed from tracker (i.e. pointerup event).
+  // In situation when NaN would be sent as a response, we return cached value.
+  // That prevents handlers from crashing
   public getLastAvgX(): number {
-    return this.getSumX() / this.trackedPointers.size;
+    const avgX: number = this.getSumX() / this.trackedPointers.size;
+    return avgX ? avgX : this.cachedAverages.x;
```
j-piasecki, line 180: IMO, explicit `isNaN` is better, this would also send cached value if the calculated average is 0, same in other places.

j-piasecki (approved):

PR #2217, opened 2022-09-12
Add `Dependencies` section to the issue template

j-piasecki:
## Description

Adds `Dependencies` section to the issue template. Having this knowledge will make it easier to reproduce issues.


j-piasecki:
I think we should be enforcing reproducers more instead of asking for details.

PR #2225, opened 2022-09-16
Replace `jcenter` with `mavenCentral`

j-piasecki:
## Description

https://developer.android.com/studio/build/jcenter-migration

## Test plan

Build the app


kkafar (approved):
Looks good!

PR #2234, opened 2022-09-21
Android TalkBack fix

m-bert:
This PR solves issue when buttons don't activate after enabling TalkBack. When TalkBack is enabled, buttons don't receive MotionEvents, hence they don't go through orchestrator. However, `performClick` method is still being called. That allows us to manually activate required handlers.

These changes should also allow navigating with keyboard and fix activation of nested buttons.


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -326,11 +330,35 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
...
+          receivedKeyEvent -> {
+            if (soundResponder == this) {
+              receivedKeyEvent = false
+            }
+            gestureHandlerRootView?.activateNativeHandlers(this)
+          }
+          soundResponder == this -> {
+            soundResponder = null
+          }
+        }
```
j-piasecki, line 360:
```suggestion
        if (soundResponder == this) {
          soundResponder = null
        } else if (receivedKeyEvent || context.isScreenReaderOn()) {
          if (soundResponder == this) {
            receivedKeyEvent = false
          }
          
          getGestureHandlerRootView()?.activateNativeHandlers(this)
        }
```

I'm not sure if this is more readable in any way, but it should avoid traversing the view hierarchy in the most common branch.

j-piasecki (changes requested):

android/lib/src/main/java/com/swmansion/gesturehandler/Extensions.kt
```diff
@@ -9,3 +11,11 @@ val ReactContext.deviceEventEmitter: DeviceEventManagerModule.RCTDeviceEventEmit
...
 val ReactContext.UIManager: UIManagerModule
     get() = this.getNativeModule(UIManagerModule::class.java)!!
+
+fun Context.isScreenReaderOn(): Boolean {
+    val am = getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager
+    if (am.isEnabled) {
+        return true
+    }
+    return false
+}
```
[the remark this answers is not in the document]
Warus15: First one is definitely better. And for the second one, I'm not used to that Kotlin syntax so I didn't use it, but it looks ok.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -370,6 +397,20 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
       // by default Viewgroup would pass hotspot change events
     }
 
+    private fun getGestureHandlerRootView(): RNGestureHandlerRootView? {
```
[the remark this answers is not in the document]
Warus15: I agree

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
Warus15: I wouldn't say that it's more readable, but avoiding unnecessary traversing sounds like a good reason to make that change.

j-piasecki (approved):

android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -127,6 +127,10 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
     this.hook = defaultHook
   }
 
+  fun markAsInBounds(value: Boolean) {
+    isWithinBounds = value
+  }
```
j-piasecki, line 132: Maybe changing it to something like this:
```suggestion
  fun withMarkedAsInBounds(closure: () -> Unit) {
    isWithinBounds = true
    closure()
    isWithinBounds = false
  }
```

(name tdb) and passing the state change chain in closure would be better? Then, this method could be moved to base `GestureHandler`, and `isWithinBounds` setter could stay private. Not sure though.

m-bert: Looks like a good idea, done in [5b6db76](https://github.com/software-mansion/react-native-gesture-handler/pull/2234/commits/5b6db7632b91de26c558aa7a169fac6b6b01639f)

m-bert: I've also tried to simplify a little `performClick` method. With this change I was able to navigate through the app using keyboard. Have a look at [8193cc8](https://github.com/software-mansion/react-native-gesture-handler/pull/2234/commits/8193cc8079d2df7a678b6242764aac449b78737d) and tell me what you think.

Issue #2233, opened 2022-09-21
Undefined is not an object (evaluating 'ref._reactInternals.elementType)


ucheNkadiCode:
### Description

I'm using Gesture handler 2.6.1 and receiving this error from GestureDetector.tsx. I'm also using

   "react": "16.13.1",
    "react-dom": "16.13.1",
    "react-native": "~0.63.4",

```javascript
function validateDetectorChildren(ref: any) {
  // finds the first native view under the Wrap component and traverses the fiber tree upwards
  // to check whether there is more than one native view as a pseudo-direct child of GestureDetector
  // i.e. this is not ok:
  //            Wrap
  //             |
  //            / \
  //           /   \
  //          /     \
  //         /       \
  //   NativeView  NativeView
  //
  // but this is fine:
  //            Wrap
  //             |
  //         NativeView
  //             |
  //            / \
  //           /   \
  //          /     \
  //         /       \
  //   NativeView  NativeView
  if (__DEV__ && Platform.OS !== 'web') {
    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
    const wrapType = ref._reactInternals.elementType;
    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
    let instance = RNRenderer.findHostInstance_DEPRECATED(ref)
      ._internalFiberInstanceHandleDEV;
```

For whatever reason, const wrapType is causing an error. I am using Typescript but I doubt this is the issue

### Steps to reproduce

I followed setup of putting GestureHandlerRootView at the start of the app. I am using react-navigation which I'm unsure if this causes problems.

Then I am building the app and as soon as I go to a page that is using react-native-swipeable-item which wraps Gestures underneath, I am getting this error. 

No clue how to fix it.

### Snack or a link to a repository

No link

### Gesture Handler version

2.6.1

### React Native version

~0.63.4

### Platforms

iOS

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Device

Real device

### Device model

iphone 12

### Acknowledgements

Yes

### Current Workaround

I simply commented out all of the code in this function and the error went away.

PR #2238, opened 2022-09-22
Use `_reactInternalFiber` instead of `_reactInternals` on RN 0.63

j-piasecki:
## Description

This PR fixes an error caused by the `_reactInternals` property missing on ref objects on React Native 0.63. Instead, the `_reactInternalFiber` prop should be used on that version.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2233.

## Test plan

Add `GestureDetector` in an app based on React Native 0.63.


tomekzaw (approved):

mxmkhv:
Introducing REACT_NATIVE_VERSION  the check now causes crash in dev mode. 

Simulator Screen Shot - iPhone 13 - 2022-10-24 at 11 15 35

For some reason it returns 0 for both minor and major values.

Screenshot 2022-10-26 at 17 22 33

PR #2240, opened 2022-09-22
Make handlers send touches cancel on gesture end

Warus15:
## Description

This PR adds minor change to web version. Now when gesture handlers move to `end/failed/cancelled` state, they also trigger onTouchesCancelled callback. This will happen, when gesture has ended, but there are still remaining pointers on handler.

## Test plan

Tested on example app

### DIFF
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -563,9 +563,7 @@ export default abstract class GestureHandler {
     const { onGestureHandlerEvent }: PropsRef = this.propsRef
       .current as PropsRef;
 
-    if (cancelEvent) {
-      invokeNullableMethod(onGestureHandlerEvent, cancelEvent);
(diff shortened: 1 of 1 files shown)


src/web/handlers/GestureHandler.ts
```diff
@@ -507,6 +516,58 @@ export default abstract class GestureHandler {
...
+      },
+      timeStamp: Date.now(),
+    };
+
+    const { onGestureHandlerEvent }: PropsRef = this.propsRef
+      .current as PropsRef;
+
+    if (cancelEvent) {
+      invokeNullableMethod(onGestureHandlerEvent, cancelEvent);
+    }
```
j-piasecki, line 568: Why is it inside if? It looks like `cancelEvent` is always truthy.

j-piasecki (approved):
🚀

Issue #2239, opened 2022-09-22
Fatal Exception: java.lang.NullPointerException


vishaljackett:
### Description

Fatal Exception: java.lang.NullPointerException
       at com.swmansion.gesturehandler.GestureHandler.isWithinBounds(GestureHandler.kt:605)
       at com.swmansion.gesturehandler.GestureHandler.handle(GestureHandler.kt:352)
       at com.swmansion.gesturehandler.GestureHandlerOrchestrator.deliverEventToGestureHandler(GestureHandlerOrchestrator.kt:269)
       at com.swmansion.gesturehandler.GestureHandlerOrchestrator.deliverEventToGestureHandlers(GestureHandlerOrchestrator.kt:215)
       at com.swmansion.gesturehandler.GestureHandlerOrchestrator.onTouchEvent(GestureHandlerOrchestrator.kt:44)
       at com.swmansion.gesturehandler.react.RNGestureHandlerRootHelper.dispatchTouchEvent(RNGestureHandlerRootHelper.kt:95)
       at com.swmansion.gesturehandler.react.RNGestureHandlerRootView.dispatchTouchEvent(RNGestureHandlerRootView.kt:37)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3049)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2693)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3049)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2693)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3049)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2693)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3049)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2693)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3049)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2693)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3049)
       at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2693)
[...]

PR #2241, opened 2022-09-22
Fix `NullPointerException` when evaluating `hitSlop` on Android

j-piasecki:
## Description

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2239.

Based on the location of the crash from the issue this seems to be a concurrency issue. Using `?.let` instead of non-null assertions should resolve the problem, as the used variable will be guaranteed not to change while the block is being executed.

## Test plan

I don't really know how to reproduce it, but I also don't see what could be the other solution, so... trust me?


Warus15 (approved):
lgtm

PR #2252, opened 2022-10-03
Fix simultaneous gesture when it receives composed gestures as arguments

j-piasecki:
## Description

Right now `Simultaneous` gesture transforms received gestures into the array of non-composed gestures (flattening it if necessary) and makes every gesture simultaneous with every gesture in the array. It's not a good approach because `Exclusive` also exists and putting an exclusive gesture into `Simultaneous` meant that gestures meant to be exclusive are now also simultaneous, resulting in different behavior on different platforms.

This PR changes the behavior so it respects groups, meaning exclusive gestures will not be made simultaneous with each other.

## Test plan

Tested on the example app

### DIFF
--- a/src/handlers/gestures/gestureComposition.ts
+++ b/src/handlers/gestures/gestureComposition.ts
@@ -69,14 +69,24 @@ export class ComposedGesture extends Gesture {
 
 export class SimultaneousGesture extends ComposedGesture {
   prepare() {
-    const simultaneousArray = this.gestures
-      .flatMap((gesture) => gesture.toGestureArray())
-      .concat(this.simultaneousGestures);
+    // this piece of magic works something like this:
+    // for every gesture in the array
+    const simultaneousArrays = this.gestures.map((gesture) =>
+      // we take the array it's in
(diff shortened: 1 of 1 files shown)


src/handlers/gestures/gestureComposition.ts
```diff
@@ -69,14 +69,24 @@ export class ComposedGesture extends Gesture {
...
-      .flatMap((gesture) => gesture.toGestureArray())
-      .concat(this.simultaneousGestures);
+    // this piece of magic works something like this:
+    // for every gesture in the array
+    const simultaneousArrays = this.gestures.map((gesture) =>
+      // we take the array it's in
+      this.gestures
+        // and make a copy without it
+        .filter((x) => x !== gesture)
+        // then we flatmap the restul to get list of raw (not composed) gestures
```
Warus15, line 79:
```suggestion
        // then we flatmap the result to get list of raw (not composed) gestures
```

Warus15 (approved):
Looks fine

PR #2262, opened 2022-10-07
fix a potential race on component unmount

phryneas:
## Description

This fixes a potential race condition where the component unmounted before the callback passed into `setImmediate` in `updateHandlers` could execute. 

That would lead to a "No handler for tag X" error, since the handler would not have been registered any more at that point in time.

There might also be other race conditions because of that `setImmediate`, but this is the only one I could reproduce.

## Test plan

I encountered this in our project code and debugged it with a [replay recording](https://www.replay.io/). 
I might be able to provide that replay if contacted on a private side channel for verification purposes, but I assume the problem & solution are probably already understandable just by the explanation above.

### DIFF
--- a/src/handlers/gestures/GestureDetector.tsx
+++ b/src/handlers/gestures/GestureDetector.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useRef, RefObject } from 'react';
 import {
   GestureType,
   HandlerCallbacks,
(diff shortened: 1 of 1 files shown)


src/handlers/gestures/GestureDetector.tsx
```diff
@@ -246,6 +247,7 @@ function updateHandlers(
   // and handlerTags in BaseGesture references should be updated in the loop above (we need to wait
   // in case of external relations)
   setImmediate(() => {
+    if (!mountedRef.current) return;
```
j-piasecki, line 250: Just one small nit

```suggestion
    if (!mountedRef.current) {
      return;
    }
```

j-piasecki (changes requested):
Thanks for fixing this ❤️!

j-piasecki (approved):

PR #2269, opened 2022-10-11
Comparison for reference

rajatbarman:


j-piasecki:
Hi, could you share some context on this PR? I don't see a particular reason for it.

Issue #2265, opened 2022-10-10
Gesture.Manual().shouldCancelWhenOutside(true) doesn't work on iOS or web


elliotwaite:
### Description

Using `Gesture.Manual().shouldCancelWhenOutside(true)` doesn't seem to work on iOS or web. As can be seen in this demo:

![ezgif-5-9f11e5bccb](https://user-images.githubusercontent.com/1767836/194941944-1523e54e-32da-4165-95ba-a98166469efd.gif)

UPDATE: I did some more testing to compare how Tap, LongPress, and Manual work when using `shouldCancelWhenOutside(true)`.

On Android it works correctly for all three gestures (Tap, LongPress, and Manual):
![ezgif-android](https://user-images.githubusercontent.com/1767836/195099430-76bf8f01-1edc-4456-b5cf-55f78224e09e.gif)

On iOS it works correctly for Tap and LongPress but not Manual:
![ezgif-ios](https://user-images.githubusercontent.com/1767836/195100765-bd6e28e2-c776-493a-8758-dc90b6de33a9.gif)

On the web:
* It doesn't work for Tap.
* It works for LongPress if the press starts near the center of the element, but it doesn't work if the press starts near the edge of the element.
* It doesn't work for Manual.
![ezgif-web](https://user-images.githubusercontent.com/1767836/195102560-20502673-79ea-49ea-8525-e81699291756.gif)

### Steps to reproduce

You can test out the Android and iOS versions using [this Snack demo](https://snack.expo.dev/@elliotwaite/rngh-shouldcancelwhenoutside), but I haven't figured out how to get the web version to work on Snack yet. If I try to use `enableExperimentalWebImplementation(true)`, it says it's not a function. Any suggestions? However, I tested the web version on my local machine as can be seen in the animated GIFs.

### Snack or a link to a repository

https://snack.expo.dev/@elliotwaite/rngh-shouldcancelwhenoutside

### Gesture Handler version

2.7.0

### React Native version

0.69.0

### Platforms

Android, iOS, Web

### Acknowledgements

Yes

elliotwaite:
@Warus15, okay, nice, thanks for the update.

Warus15:
Hi @elliotwaite! 

I've also managed to solve this problem on iOS, could you have a look at [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2274) and check whether it indeed solves your issue?

elliotwaite:
[...]

PR #2271, opened 2022-10-11
Make handlers respect `shouldCancelWhenOutside`

Warus15:
## Description

This PR solves web part of #2265 issue. I've modified `onPointerOut` method to act similar as on android. Right now, if pointer leaves handler and `shouldCancelWhenOutside` is set to `true`, handler goes to `cancel` state if it was `active`, or to `fail`, if it was in `began`

This change also affects `ManualGestureHandler`, which now cancels properly.

## Test plan

Tested on example app


src/web/handlers/FlingGestureHandler.ts
```diff
@@ -32,6 +32,10 @@ export default class FlingGestureHandler extends GestureHandler {
 
     this.enabled = enabled;
 
+    if (this.config.shouldCancelWhenOutside !== undefined) {
+      this.setShouldCancelWhenOutside(this.config.shouldCancelWhenOutside);
+    }
```
j-piasecki, line 37: This is repeated for every handler (I believe Pan already had it), cannot it be moved to the base class?

j-piasecki (approved):

PR #2272, opened 2022-10-12
Add module file for Windows

### LINKED ISSUE #2266: [Windows c#] TypeError: Cannot read properties of undefined (reading 'handleSetJSResponder')
### Description

"Windows application, Visual Studio 2019, Developed by c#"
I had updated the react-native version to 0.69.5, react-native-windows version to 0.69.9, @react-navigation/native5.x->6.x
how solve this bug.
Screenshot 2022-10-11 134918
Screenshot 2022-10-11 135004

### Steps to reproduce

1. react navigation upgrade to last version(using native, stack, drawer)
2. react-native-gesture-handler install 2.5.0 between 2.7.0 versions
3. yarn run windows

https://github.com/software-mansion/react-native-gesture-handler/issues/2213

### Gesture Handler version

2.5.0-2.7.0

### React Native version

0.69.5

### Platforms

iOS

### JavaScript runtime

_No response_

### Workflow

React Native (without Expo)

### Architecture

_No response_

### Build type

Debug mode

### Device

Real device

### Device model

Windows 11 UWP

### Acknowledgements

Yes

Warus15:
## Description

This PR should fix #2266. It adds `RNGestureHandlerModule.windows.ts`, which should solve problem of undefined module on Windows.

## Test plan

Since we don't have computer with windows to test that, you have to trust me.

### DIFF
--- a/src/RNGestureHandlerModule.windows.ts
+++ b/src/RNGestureHandlerModule.windows.ts
@@ -1,0 +1,140 @@
+import React from 'react';
+
+import { ActionType } from './ActionType';
+import { isExperimentalWebImplementationEnabled } from './EnableExperimentalWebImplementation';
+
+//GestureHandlers
+import InteractionManager from './web/tools/InteractionManager';
+import NodeManager from './web/tools/NodeManager';
+import PanGestureHandler from './web/handlers/PanGestureHandler';
+import TapGestureHandler from './web/handlers/TapGestureHandler';
+import LongPressGestureHandler from './web/handlers/LongPressGestureHandler';
+import PinchGestureHandler from './web/handlers/PinchGestureHandler';
+import RotationGestureHandler from './web/handlers/RotationGestureHandler';
+import FlingGestureHandler from './web/handlers/FlingGestureHandler';
+import NativeViewGestureHandler from './web/handlers/NativeViewGestureHandler';
+import ManualGestureHandler from './web/handlers/ManualGestureHandler';
+
+//Hammer Handlers
+import * as HammerNodeManager from './web_hammer/NodeManager';
+import HammerNativeViewGestureHandler from './web_hammer/NativeViewGestureHandler';
+import HammerPanGestureHandler from './web_hammer/PanGestureHandler';
+import HammerTapGestureHandler from './web_hammer/TapGestureHandler';
+import HammerLongPressGestureHandler from './web_hammer/LongPressGestureHandler';
+import HammerPinchGestureHandler from './web_hammer/PinchGestureHandler';
+import HammerRotationGestureHandler from './web_hammer/RotationGestureHandler';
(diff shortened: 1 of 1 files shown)


src/RNGestureHandlerModule.windows.ts
```diff
@@ -0,0 +1,140 @@
...
+  TapGestureHandler: HammerTapGestureHandler,
+  LongPressGestureHandler: HammerLongPressGestureHandler,
+  PinchGestureHandler: HammerPinchGestureHandler,
+  RotationGestureHandler: HammerRotationGestureHandler,
+  FlingGestureHandler: HammerFlingGestureHandler,
+};
+
+export default {
+  handleSetJSResponder(tag: number, blockNativeResponder: boolean) {
+    console.warn('handleSetJSResponder: ', tag, blockNativeResponder);
```
j-piasecki, line 52: I think this is removed in the macOS file. If that's the case, you can remove it from here.

I don't recall more differences, but I think we would want to copy the macOS file instead of web.

src/RNGestureHandlerModule.windows.ts
```diff
@@ -0,0 +1,140 @@
...
+  },
+  handleClearJSResponder() {
+    console.warn('handleClearJSResponder: ');
```
j-piasecki, line 55: Same as above.

j-piasecki (changes requested):

src/RNGestureHandlerModule.windows.ts
Warus15: Both methods are present in `.macos` file, do You mean to remove only warns? Either way, I can copy `.macos` instead of `.web`, it shouldn't make that much of a difference

j-piasecki: Yes, I meant the warns. I don't remember other differences between the macOS and the web file, if there aren't more then just removing those warns is fine.

j-piasecki (approved):

Aashu-Dubey:
Hello,

I recently tried an implementation that uses gesture-handler, in react-native-windows and running that I got this error
image

and it is at [this line](https://github.com/software-mansion/react-native-gesture-handler/blob/main/src/RNGestureHandlerModule.windows.ts#L99).

Now if I comment the `newView instanceof HTMLElement ||` part of the if condition (through node_modules), I don't get any more errors and component is properly rendered.

image

Now even though you mentioned you don't have a windows computer, but still if you need a repro. where I faced the issue, then it can be found [here](https://github.com/Aashu-Dubey/youtube/tree/react-native/rn_youtube), though make sure to change the condition [here](https://github.com/Aashu-Dubey/youtube/blob/react-native/rn_youtube/App.tsx#L7) to use the `ToolbarReanimated` solution instead.

Not related to this, but if you really run the project, then you might notice some other issues with gesture-handler support on windows, as gesture callbacks used for the demo don't really work on windows.

PR #2274, opened 2022-10-13
Fix `ManualGestureHandler` shouldCancelWhenOutside on iOS

Warus15:
## Description

This PR adds changes in `ManualGestureHandler` on iOS. It should now properly react to `shouldCancelWhenOutside` prop. This PR should solve part of #2265 issue.

## Test plan

Tested on example app

### DIFF
--- a/ios/Handlers/RNManualHandler.m
+++ b/ios/Handlers/RNManualHandler.m
@@ -2,7 +2,7 @@
 
 @interface RNManualRecognizer : UIGestureRecognizer
 
-- (id)initWithGestureHandler:(RNGestureHandler*)gestureHandler;
+- (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler;
 
 @end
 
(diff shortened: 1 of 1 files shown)


ios/Handlers/RNManualHandler.m
```diff
@@ -35,6 +35,16 @@ - (void)touchesMoved:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
 {
   [super touchesMoved:touches withEvent:event];
   [_gestureHandler.pointerTracker touchesMoved:touches withEvent:event];
+
+  if ([self shouldFail]) {
+    if (_gestureHandler.state == RNGestureHandlerStateActive) {
+      self.state = UIGestureRecognizerStateFailed;
+    } else if (_gestureHandler.state == RNGestureHandlerStateBegan) {
+      self.state = UIGestureRecognizerStateCancelled;
+    }
```
j-piasecki, line 44: It may be better to stay consistent and rely on the `UIGestureRecognizer`'s state here, like so: https://github.com/software-mansion/react-native-gesture-handler/blob/234ec684e8cc3e6e65a2e42f5ed3ae72c4d6e681/ios/Handlers/RNPanHandler.m#L122-L124

Warus15: Sure, I've missed that I left `RNGestureHandler` states in ifs

j-piasecki (approved):

Issue #2263, opened 2022-10-07
Gesture callback differences between platforms


elliotwaite:
### Description

For certain gestures, the callbacks that are called, or the order in which they are called, differs between platforms. I listed the difference I've found below. The tables show the order of events from top to bottom.

Tap gesture (if pressed quickly):
|iOS|Android|
|-|-| 
| **onTouchesDown** | **onBegin** |
| **onBegin** | **onTouchesDown** |
| onTouchesUp | onTouchesUp |
| onStart | onStart |
| onEnd | onEnd |
| onFinalize | onFinalize |

Tap gesture (if pressed and held):
|iOS|Android|
|-|-| 
| **onTouchesDown** | **onBegin** |
| **onBegin** | **onTouchesDown** |
| onTouchesCancelled | onTouchesCancelled |
| onFinalize | onFinalize |
| **onFinalize (a second time)** | |

LongPress gesture (if pressed quickly):
|iOS|Android|
|-|-| 
| **onTouchesDown** | **onBegin** |
| **onBegin** | **onTouchesDown** |
| onTouchesUp | onTouchesUp |
| onFinalize | onFinalize |

LongPress gesture (if pressed and held):
|iOS|Android|
|-|-| 
| **onTouchesDown** | **onBegin** |
| **onBegin** | **onTouchesDown** |
| onStart | onStart |
| onTouchesUp | onTouchesUp |
| onEnd | onEnd |
| onFinalize | onFinalize |

Fling gesture (when a fling is detected):
|iOS|Android|
|-|-| 
| onBegin | onBegin |
| onTouchesDown | onTouchesDown |
| onTouchesMove (possibly multiple) | onTouchesMove (possibly multiple) |
| onStart | onStart |
| **onEnd** | **onTouchesCancelled** |
| **onFinalize** | **onEnd** |
| **onTouchesCancelled** | **onFinalize** |

Pinch gesture, Rotation gesture, and Native gesture (when pressed):
|iOS|Android|
|-|-| 
| | **onBegin** |
| onTouchesDown | onTouchesDown |
| onTouchesUp | onTouchesUp |
| | **onFinalize** |

Manual gesture (when pressed). In the provided Snack, this only happens on the first press because the gesture is never finalized:
|iOS|Android|
|-|-| 
| **onTouchesDown** | **onBegin** |
| **onBegin** | **onTouchesDown** |
| onTouchesUp | onTouchesUp |

### Steps to reproduce

[...]

PR #2278, opened 2022-10-14
Fix `onBegin` and `onTouchesDown` event ordering on iOS

j-piasecki:
## Description

This PR updates the ordering of events for `Fling` and `Pan` gestures so that the `onTouchesDown` callback is invoked before `onBegin`.

Fixes part of https://github.com/software-mansion/react-native-gesture-handler/issues/2263.

## Test plan

Tested on the code from the issue and on the example app.


Warus15 (approved):
LGTM

PR #2279, opened 2022-10-14
Fix `onBegin` and `onTouchesDown` events ordering on Android

j-piasecki:
## Description

This PR changes the ordering of `onBegin` and `onTouchesDown` events on Android. `onBegin` was invoked first, as some gestures required additional setup when they started tracking the first pointer, which is happening after the touch events are dispatched. The workaround for that was to send `onTouchesDown` later but it resulted in inconsistent behavior between platforms.
This behavior is now updated to defer sending the first `onTouchesDown` to a point in time where the gesture is initialized, but `onBegin` hasn't been called yet - just before sending the `UNDETERMINED` -> `BEGIN` state change event. This way the correct behavior is kept alongside the correct event ordering without the need to modify every handler.

Fixes part of https://github.com/software-mansion/react-native-gesture-handler/issues/2263.

## Test plan

Tested on the example app and on the code from the issue.


j-piasecki:
Superseded by https://github.com/software-mansion/react-native-gesture-handler/pull/2283.

Issue #2284, opened 2022-10-20
shouldCancelWhenOutside default behavior doesn't agree with docs


elliotwaite:
### Description

The docs for `shouldCancelWhenOutside()` mention:
> Default value of this property is different depending on the handler type. Most handlers' shouldCancelWhenOutside property defaults to false except for the [LongPressGesture](https://docs.swmansion.com/react-native-gesture-handler/docs/api/gestures/long-press-gesture) and [TapGesture](https://docs.swmansion.com/react-native-gesture-handler/docs/api/gestures/tap-gesture) which default to true.

However, the default value for `Tap` and `LongPress` seems to be `false`, as can be seen in this demo:

![default](https://user-images.githubusercontent.com/1767836/196841329-59e17d2d-5181-4c08-a8bf-ed90a355cc37.gif)

So far I've only tested it on Android and iOS. I'm not sure what the default behavior is on the web.

### Steps to reproduce

Load up the provided Snack in Android or iOS and test out the gestures as seen in the demo to see which gestures get canceled when the press is held and the cursor is dragged outside the buttons.

### Snack or a link to a repository

https://snack.expo.dev/@elliotwaite/rngh-shouldcancelwhenoutside-default-behavior

### Gesture Handler version

2.5.0

### React Native version

0.69.0

### Platforms

Android, iOS

### Acknowledgements

Yes

Warus15:
Hi @elliotwaite 👋

I've checked web version and it seems to work fine. However it would be great if you could also have a look at that, maybe I've missed something.

Also, I've prepared [PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2286) that should fix this issue on iOS and Android.

elliotwaite:
Okay, sounds good. With my current setup, it's difficult to test out the web version, but if I test it out in the future and I notice that it's not working as would have expected, I'll reopen this issue.

Thanks for the quick PR!

Warus15:
Hi! As this issue seems to be solved, I'll close it. However if you'll find out that something is not working properly, feel free to reopen it.

elliotwaite:
Sounds good. Thanks again.

PR #2286, opened 2022-10-20
Fix default `shouldCancelWhenOutside` on `Tap` and `LongPress`

Warus15:
## Description

This PR should fix problem mentioned in #2284. `TapGestureHandler` and `LongPressGestureHandler` are now cancelling by default after pointer leaves underlying view, even if `shouldCancelWhenOutside` prop is not passed, as the docs say.

## Test plan

Tested on example app on iOS and Android


j-piasecki:
IMO doing it in JS is a cleaner approach as it allows to consolidate the default configuration in one place.

For the old API you're able to specify the default config: https://github.com/software-mansion/react-native-gesture-handler/blob/8cd9775e530f342f61d00831f66afe0b3f7b310d/src/handlers/TapGestureHandler.ts#L91 and in the new API you're able to modify relevant config fields in the constructor: https://github.com/software-mansion/react-native-gesture-handler/blob/8cd9775e530f342f61d00831f66afe0b3f7b310d/src/handlers/gestures/tapGesture.ts#L13

I think this way you should also be able to remove this line https://github.com/software-mansion/react-native-gesture-handler/blob/8cd9775e530f342f61d00831f66afe0b3f7b310d/src/web/handlers/TapGestureHandler.ts#L36

The same applies to long press.

Warus15:
You're right @j-piasecki, it is better to put that in default config in JS. I've done this, so please have a look at that when you have time.

src/handlers/TapGestureHandler.ts
```diff
@@ -88,5 +88,7 @@ export const TapGestureHandler = createHandler<
     ...baseGestureHandlerProps,
     ...tapGestureHandlerProps,
   ] as const,
-  config: {},
+  config: {
+    shouldCancelWhenOutside: true,
```
j-piasecki, line 92: This should also be done for long press.

j-piasecki (commented):
In`RNGestureHandler.m` there is a lot related to formatting but the imports for Tap and Long press are still there. I didn't look thoroughly at it so there may be more problems. I think undoing the changes there is the simplest solution.

Warus15:
Ok, since this PR got really messed up by this iOS linter changes (which i didn't notice while reverting other commit), I've created new one. When you'll have time, please take a look at #2297, which does what this PR is meant to do, but without any crazy formatting stuff. I'll also close this one.

PR #2296, opened 2022-10-28
Reformat iOS codebase

j-piasecki:
## Description

This PR reformats the iOS codebase to be consistent.

The intention was to drag this in time - when someone modifies an iOS file, it would be formatted so the history would be less impacted. But it turns out that it also formats files while merging branches, which is pretty irritating. Turns out you can straight up [ignore certain commits](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) when using `git blame` - this solves the problem IMO. (I just hope I've set it up correctly - I'm not sure whether it will work after merging as the merge commit will have a different hash)

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,3 @@
+# .git-blame-ignore-revs
+# Reformatted iOS codebase
+93e0d1d41c0ce985327363c230453846df42d439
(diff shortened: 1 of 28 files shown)


Warus15 (approved):
Good idea to use ignore in blame

PR #2313, opened 2022-11-09
Support React Native 0.71

j-piasecki:
## Description

This PR adds support for new architecture on React Native 0.71, which most notably includes the usage of prefabs in native builds.

It also removes most of the dependencies of `FabricExample` app, so upgrades in the future should be easier.

PR #2314, opened 2022-11-09
Update finding the `react-native` package path

j-piasecki:
## Description

This PR updates the code responsible for finding the `react-native` path on Android to support more cases. The code is borrowed from `react-native-reanimated` so it should be working correctly.


j-piasecki:
Included in https://github.com/software-mansion/react-native-gesture-handler/pull/2313

Issue #2312, opened 2022-11-08
Touchables stop working on Android after using the app for a while


gabrieldonadel:
### Description

Touchables stop working on Android after using the app for a while. I was able to reproduce this with my logcat open and when pressing on the Touchables element an error saying "Invalid ID 0x00000XXX" is thrown

![image](https://user-images.githubusercontent.com/11707729/200573812-f1dabb69-fe17-42dc-8f33-75b79d4bf080.png)

### Steps to reproduce

1. Using an Android device navigate around the app and try to press on many things at the same time
2. At some point touchable elements that use react-native-gesture-handler will stop working (this may take up to ~10 minutes to repro)

### Snack or a link to a repository

https://snack.expo.dev/@gabrieldonadel/touchables-bug

### Gesture Handler version

2.8.0

### React Native version

0.69.7

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Real device

### Device model

Samsung Galaxy S10 (Android 12)

### Acknowledgements

Yes

j-piasecki:
@gabrieldonadel Hi! I believe that https://github.com/software-mansion/react-native-gesture-handler/pull/2315 fixes the problem, at least I wasn't able to reproduce it anymore with this fix applied. I would really appreciate it if you could check whether it solves the problem for you.

gabrieldonadel:
> @gabrieldonadel Hi! I believe that #2315 fixes the problem, at least I wasn't able to reproduce it anymore with this fix applied. I would really appreciate it if you could check whether it solves the problem for you.

@j-piasecki your fix looks good to me, I've tested it on my Apps and seems to resolve the problem, thx!

simon-abbott:
I can confirm that this is an issue in v2.7.0 through v2.8.0, and that the fix in 04f22dc8178595a3c8afb5260470649a711cee5a works (tested via [patch-package](https://www.npmjs.com/package/patch-package)). When can we expect a new release with this fix in it?

---

In case it becomes a problem again and/or anyone wants to revisit this fix further, here is a small example that reproduces the issue 100% of the time:
```tsx
<View>
  <View style={{ position: 'absolute' }}>
    <TouchableOpacity
      onPress={() => {
        console.log('Button 1 pressed');
      }}
    >
      <Text>Button 1</Text>
    </TouchableOpacity>
  </View>
[...]

PR #2315, opened 2022-11-09
Fix buttons getting locked in some cases on Android

j-piasecki:
## Description

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2312

Buttons rely on the NativeViewGestureHandler that wraps them for some of the state management, mainly for freeing responder which acts as a lock to prevent more than one exclusive buttons from getting pressed at the same time. This proved problematic as this method was not always called.

This PR fixes it by also calling `tryFreeingResponder` in the `performClick` method, which is called by Android, so I assume we can rely on it.

## Test plan

Tested on the code from snack.


gabrieldonadel (approved):
LGTM, tested on Galaxy S10 Android 12 and works perfectly!

tomekzaw (approved):

Issue #2333, opened 2022-11-28
react-native-gesture-handler will have many instances in pnpm project


chj-damon:
### Description

I'm trying to use `rnx-kit` and `pnpm` to configure my project. it works fine when I didn't add `react-native-gesture-handler` in my project. 
Things get wierd after `pnpm i react-native-gesture-handler`. It will fail my project as RNGH tries to assertNoMultipleInstances and apparently there are two instances:
![image](https://user-images.githubusercontent.com/4902684/204298732-27c7bbac-6fd9-497c-ae6f-a3436369834f.png)

### Steps to reproduce

clone my project

run script: pnpm android:dev and it will fail

### Snack or a link to a repository

https://github.com/chj-damon/rnxkitDemo/tree/pnpm

### Gesture Handler version

2.8.0

### React Native version

0.69.7

### Platforms

Android

### Build type

Debug mode

### Device

Android emulator

### Acknowledgements

Yes

PR #2334, opened 2022-11-29
fix RNGH not working in pnpm project

chj-damon:
in pnpm project, there will be many RNGH instances since some package has depend on RNGH, pnpm will generate such folder, which will break `noMultipleInstancesAssertion` function


chj-damon:
@kmagiera any advice for this PR?

chj-damon:
https://github.com/software-mansion/react-native-gesture-handler/issues/2333

j-piasecki (approved):
Thanks for taking the time to do the PR ❤️

PR #2335, opened 2022-12-01
GestureDetector: more checks for mounted component

phryneas:
## Description

This builds on top of #2262 - I encountered more cases where a component that was immediately remounting was running into race conditions, so I added checks to every `setImmediate` call.


src/handlers/gestures/GestureDetector.tsx
```diff
@@ -148,6 +150,9 @@ function attachHandlers({
   // use setImmediate to extract handlerTags, because all refs should be initialized
   // when it's ran
   setImmediate(() => {
+    if (!mountedRef.current) {
+      return;
+    }
```
phryneas, line 155: This one didn't happen to me (yet), but I added it for good measure - we should probably never assume in an `setImmediate` call that the component did not unmount before.

PR #2339, opened 2022-12-04
docs: explain `useNativeAnimations` for use cases

fivecar:
## Description

In cases where users use `LayoutAnimation` to smooth transitions upon swiping an item (e.g. perhaps swipe left deletes this Swipeable from a list), the non-open version of the Swipeable will be animated out (as opposed to the open version). This ends up looking bad/amatuer. Turns out this can be fixed by simply setting `useNativeAnimations` to false in those cases so that RN's `LayoutAnimation` shows the opened version of the Swipeable exiting.

## Test plan

View doc markdown


m-bert:
Hi! Old `Swipeable` component has been deprecated. `ReanimatedSwipeable` does not have this property. Given that I'm closing this PR, let me know if you think we should reopen it.

fivecar:
Yah... if it's 3+ years later, the task is likely useless.

PR #2354, opened 2022-12-20
Migrate Gesture Handler to TurboModule

j-piasecki:
## Description

Make Gesture Handler a TurboModule instead of a Native Module 🚀 

## Test plan

Tested on the Example and FabricExample app. Needs to be tested on RN 0.67 - the lowes still supported by RNGH


src/fabric/NativeRNGestureHandlerModule.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Int32 } from 'react-native/Libraries/Types/CodegenTypes';
+
+export interface Spec extends TurboModule {
+  handleSetJSResponder: (tag: Int32, blockNativeResponder: boolean) => void;
+  handleClearJSResponder: () => void;
+  createGestureHandler: (
+    handlerName: string,
+    handlerTag: Int32,
+    // Record<> is not supported by codegen
+    // eslint-disable-next-line @typescript-eslint/ban-types
```
WoLewicki, line 11: What is the exact type to be used there?

j-piasecki: `Record<string, unknown>` was used in the native module.

package.json
```diff
@@ -146,7 +146,10 @@
   ],
   "codegenConfig": {
     "name": "rngesturehandler_codegen",
-    "type": "components",
-    "jsSrcsDir": "./src/specs"
+    "type": "all",
+    "jsSrcsDir": "./src/specs",
```
WoLewicki, line 150: Sometimes we use `fabric` and sometimes we use `specs` 😅  But `specs` sound better since turbomodules are not fabric.

example/ios/Podfile.lock
```diff
@@ -784,4 +784,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: c0ab0cb5a538e5f3dac07a771e01a60888d07b58
 
-COCOAPODS: 1.14.2
+COCOAPODS: 1.12.1
```
WoLewicki, line 787: Why did it drop? xd

apple/RNGestureHandlerModule.mm
```diff
@@ -123,55 +125,55 @@ - (void)setBridge:(RCTBridge *)bridge
...
-  return @true;
+  [self.bridge
+      dispatchBlock:^{
+        RCTCxxBridge *cxxBridge = (RCTCxxBridge *)self.bridge;
+        auto runtime = (jsi::Runtime *)cxxBridge.runtime;
+        decorateRuntime(*runtime);
+      }
+              queue:RCTJSThread];
+
+  return nil;
```
WoLewicki, line 136: You should return `true`  here since I need this information for bridgeless mode.

src/specs/NativeRNGestureHandlerModule.ts
```diff
@@ -0,0 +1,26 @@
...
+  ) => void;
+  attachGestureHandler: (
+    handlerTag: Int32,
+    newView: Int32,
+    actionType: Int32
+  ) => void;
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  updateGestureHandler: (handlerTag: Int32, newConfig: Object) => void;
+  dropGestureHandler: (handlerTag: Int32) => void;
+  install: () => Int32;
```
WoLewicki, line 22: boolean would be better here.

package.json
```diff
@@ -146,7 +146,10 @@
   ],
   "codegenConfig": {
     "name": "rngesturehandler_codegen",
```
WoLewicki, line 148: I don't like this name tbh. We have a convention of rather `rngesturehandler`

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -408,17 +424,18 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
...
+      try {
+        SoLoader.loadLibrary("gesturehandler")
+        val jsContext = reactApplicationContext.javaScriptContextHolder!!
+        decorateRuntime(jsContext.get())
+      } catch (exception: Exception) {
+        Log.w("[RNGestureHandler]", "Could not install JSI bindings.")
+      }
     }
+
+    return 0.0
```
WoLewicki, line 438: boolean here also please 😄

android/src/main/java/com/swmansion/gesturehandler/RNGestureHandlerPackage.kt
```diff
@@ -1,21 +1,84 @@
...
+  override fun getViewManagerNames(reactContext: ReactApplicationContext?) =
+    viewManagers.keys.toList()
+
+  override fun getViewManagers(reactContext: ReactApplicationContext?): MutableList<ModuleSpec> =
+    viewManagers.values.toMutableList()
+
+  override fun createViewManager(
+    reactContext: ReactApplicationContext?,
+    viewManagerName: String?
+  ): ViewManager<*, out ReactShadowNode<*>>? {
```
WoLewicki, line 45: Why those shenanigans with `ReactShadowNode` are needed?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -355,8 +352,20 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
     throw JSApplicationIllegalArgumentException("Invalid handler name $handlerName")
   }
 
-  @ReactMethod
```
WoLewicki, line 358: Are you sure you should remove the `@ReactMethod`? Will it work correctly on old arch?

WoLewicki (commented):
I left some comments, other than them, LGTM 🚀

package.json
j-piasecki: What do you mean? I think the only other place where `fabric` appears is inside `android` directory and contains code specific to the new-arch. I guess we could rename that to `newarch` and `paper` to `oldarch`. LMK what you think.

apple/RNGestureHandlerModule.mm
j-piasecki: Done

src/specs/NativeRNGestureHandlerModule.ts
j-piasecki: Done

package.json
j-piasecki: IIRC there were conflicts with `rngesturehandler` name, hence the `_codegen` suffix.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Done

android/src/main/java/com/swmansion/gesturehandler/RNGestureHandlerPackage.kt
j-piasecki: They are not

example/ios/Podfile.lock
j-piasecki: That's a really good question

package.json
WoLewicki: I meant `./src/specs` where in e.g. `RNScreens` we have https://github.com/software-mansion/react-native-screens/tree/336efc0c715c1312ec4f9d8bd5709cc1d884709b/src/fabric

j-piasecki: Ah, it was `fabric` here as well but I changed it in https://github.com/software-mansion/react-native-gesture-handler/pull/2461.

`specs` makes way more sense I think - as you noted, turbomodules are not fabric.

example/ios/Podfile.lock
j-piasecki: Fixed

j-piasecki:
Ok, I've made some changes regarding the annotations - turns out that React Native 0.67 doesn't really like the fact that the annotations were in the parent class of the module, and those methods aren't visible from JS. On 0.73 (old arch) it wasn't a problem.

Because of this, I've restored the annotations in the module and removed them from the spec file. This makes it work on both 0.73 and 0.67, so by induction it must also work on everything in between. The only possible problem with this is on the new arch the annotations are in two places - inside the spec generated by codegen and inside the module implementation itself. It seems to work for me on 0.73 new arch, but it may not work in the future though I guess it's a future us problem.

WoLewicki (approved):
LGTM 🚀

PR #2359, opened 2022-12-27
Added disableSelection prop

s77rt:
## Description

Added `disableSelection` prop

This prop is mostly useful for `PanGestureHandler` especially on Safari.
Because Safari changes the cursor to `text` once you start dragging which is undesirable.
Passing `disableSelection` fixes the issue.

```
<PanGestureHandler onGestureEvent={props.onGesture} disableSelection>
    ...
</PanGestureHandler>
```

## Test plan

Tested on https://github.com/Expensify/App/issues/13688

### DIFF
--- a/docs/docs/api/gestures/gesture-detector.md
+++ b/docs/docs/api/gestures/gesture-detector.md
@@ -29,3 +29,7 @@ Starting with Reanimated-2.3.0-beta.4 Gesture Handler will provide a [StateManag
 ### `userSelect` (**web only**)
 
 This parameter allows to specify which `userSelect` property should be applied to underlying view. Possible values are `"none" | "auto" | "text"`. Defaults to `"none"`.
+
+### `disableSelection` (**web only**)
+
+This parameter disables selection. If enabled `onselectstart` returns `false`. Defaults to `false`.
--- a/docs/docs/gesture-handlers/api/common-gh.md
+++ b/docs/docs/gesture-handlers/api/common-gh.md
@@ -67,6 +67,10 @@ Specifying `width` or `height` is useful if we only want the gesture to activate
 
 This parameter allows to specify which `userSelect` property should be applied to underlying view. Possible values are `"none" | "auto" | "text"`. Defaults to `"none"`.
 
+### `disableSelection` (**web only**)
+
+This parameter disables selection. If enabled `onselectstart` returns `false`. Defaults to `false`.
+
 ### `onGestureEvent`
 
 Takes a callback that is going to be triggered for each subsequent touch event while the handler is in an [ACTIVE](../basics/state.md#active) state. Event payload depends on the particular handler type. Common set of event data attributes is documented [below](#event-data) and handler specific attributes are documented on the corresponding handler pages. E.g. event payload for [`PinchGestureHandler`](./rotation-gh.md#event-data) contains `scale` attribute that represents how the distance between fingers changed since when the gesture started.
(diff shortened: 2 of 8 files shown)


j-piasecki:
I don't quite get why this is necessary. I'm not that familiar with web standards, so it may be a stupid question, but what's the difference between `onuserselect = () => false` and CSS `user-select: none`, which is used by HammerJS (used to back up the old web implementation of GH)?

s77rt:
@j-piasecki I can't say I know the exact difference between the two, but it seems as if Safari is ignoring the CSS rule.
https://github.com/WebKit/WebKit/blob/3c53378a77e57141a3b53f344e4bc1c14f7cef39/Source/WebCore/page/EventHandler.cpp#L1553-L1562

Regarding our case, we have found some workarounds for the issue on Safari but it would be nice to have this as a feature as well if it's not much of trouble.

By the way, the issue seems to be fixed using the new experimental web implementation but we are not going for that solution now.

s77rt:
@j-piasecki Sorry for the ping, can you help us get this merged asap?

jatinsonijs:
@j-piasecki IMO we don't require new prop to handle this case. We already have userSelect prop if we pass userSelect: none it will prevent text selection but in safari we are facing problem related to on drag cursor change to I-Beam. I have added more details in PR plz have a look:
https://github.com/software-mansion/react-native-gesture-handler/pull/2381

s77rt:
@jatinsonijs We do require the of use of an explicit prop for two reasons:
1. `onselectstart` (JS) and `uset-select` (CSS) are not the same. I see you have used that in your PR and this will introduce a major regression. I will post the details on your PR. 
2. Compatibility / No breaking changes. Everything will stay the same for every current implementation.  The user should explicitly opt for `disableSelection`

jatinsonijs:
@s77rt Yes plz, I don't think so both are different if user-select is none its mean content is not selectable. and we are preventing default behaviour. in web-hammer we are not using user-select property so I have make them user have to explicitly pass user-select: 'none'. undefined will consider are onselectstart => true

s77rt:
@jatinsonijs I have provided the regression on your PR. In web-hammer we don't use `user-select` but in web we do. In web-hammer you are asking the user to provide a prop that do something but you do something else.
When the user provide `user-select=none` he expects `user-select=none` and not `onselectstart = () => false`.

jatinsonijs:
Yes css can be override, but both props are still confusing as a consumer userSelect: 'none' and disableSelection seem to be same, I think consumer should handler it on there side instead of in this package.

s77rt:
I have explained the use of `disableSelection` on the docs.

@jatinsonijs You may be right about handling this outside this lib, but as I user of `react-native-gesture-handler` I would expect a way to solve the issue without the need to write extra code. Especially that everytime I would have to write that custom code it would revolve around `<PanGestureHandler />` for that I think it's better to have this as a feature where we can solve the issue by simply passing `<PanGestureHandler disableSelection />`

j-piasecki:
> Yes css can be override, but both props are still confusing as a consumer userSelect: 'none' and disableSelection seems to be same, I think consumer should handle it on there side instead of in this package.

I agree with the names being confusing. On top of that, from the linked issue it seems like using `setPointerCapture` also solves the problem, and the new web implementation already uses it, so the prop would effectively do nothing there.

s77rt:
@j-piasecki That's right, but the new web implementation is still experimental. That new prop would only be useful where using the experimental implementation is not an option.

j-piasecki:
I see that the underlying issue has been fixed, so I'll close this PR.

PR #2381, opened 2023-01-17
Prevent i beam on safari if userSelect is none

jatinsonijs:
## Description

Safari use I-Beam cursor on drag on most of the elements, exclude some native element like button, anchor tag etc...
We are using gesture on div that's way safari show I-Beam cursor on drag.
We already have userSelect prop which we can use to prevent text select. This PR extending the use of userSelect prop, if userSelect is none prevent safari default behaviour to show I-Beam cursor on drag.

As indicate in Safari webkit source code:
https://github.com/WebKit/WebKit/blob/3c53378a77e57141a3b53f344e4bc1c14f7cef39/Source/WebCore/page/EventHandler.cpp#L1553

What we are trying to fix:

https://user-images.githubusercontent.com/114683042/212810714-5615bc84-94d8-44f4-bfc1-6ec825170789.mp4

As in video on drag slider cursor changing to I-Beam cursor even when userSelect passed as none.

## Test plan

Tested on Expensify app in https://github.com/Expensify/App/issues/13688

### DIFF
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -83,7 +83,7 @@ export default abstract class GestureHandler {
       this.view.style['userSelect'] = this.config.userSelect;
     }
 
-    this.view.onselectstart = () => !!this.config.userSelect && this.config.userSelect !== 'none';
+    this.view.onselectstart = () => this.config.userSelect !== 'none';
   }
 
   private addEventManager(manager: EventManager): void {
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -82,6 +82,9 @@ export default abstract class GestureHandler {
       this.view.style['webkitUserSelect'] = this.config.userSelect;
       this.view.style['userSelect'] = this.config.userSelect;
     }
+    if (this.view.style['userSelect'] == 'none') {
+      this.view.onselectstart = () => false;
+    }
   }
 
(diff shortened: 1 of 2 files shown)


s77rt:
As stated in my PR here https://github.com/software-mansion/react-native-gesture-handler/pull/2359
We should not couple the `onselectstart` (JS) with `uset-select` (CSS) as they are not the same thing and do not behave the same way, doing so will cause a regression throughout the entire application. Instead we should use an explicit prop that the user can opt for so he knows what to expect.

Consider this example:
```html
<!DOCTYPE html>
<html>
<body>
<div id="a" style="user-select: none">
	<span>text inside a</span>
    <div id="b" style="user-select: all">
        <span>text inside b</span>
    </div>
</div>
</body>
</html>
```

- You can't select text a
- But you can select text b

Now if execute this code:
`document.querySelector('#a').onselectstart = () => false;`

- You won't be able to select any <-- the regression

jatinsonijs:
@s77rt same situation will also happen with disableSelection as we cannot override inner view to perform text selection.

Can you plz provide an example with Panhandler where userSelect: 'none' will not work but disableSelection will work

s77rt:
@jatinsonijs Yes, same thing with `disableSelection` but the user would be aware of that. Your solution will apply that to every current implementation.

jatinsonijs:
@s77rt your example not working I cannot select "But you can select text b"

Issue #2383, opened 2023-01-18
```GestureDetector``` not working in android when used inside a nested TopTab from “@react-navigation/material-top-tabs” 

### ISSUE THREAD

**Issue by hrithik73** [software-mansion/react-native-gesture-handler#2383]
### Description

```GestureDetector``` is not working when used inside a nested  ```createMaterialTopTabNavigator from “@react-navigation/material-top-tabs”``` on android it works when there is only one tab not in nested also, work perfectly fine in ios 

I have wrapped the root of my app with GestureHandlerRootView 
Note: - I have tried swipeEnabled: false in nested tab and wrapping the Rootview to different places in navigato

### Steps to reproduce

step 1:- git clone https://github.com/hrithik73/rngh-toptab-issue.git
step 2:- yarn 
step 3:- yarn android

### Snack or a link to a repository

https://github.com/hrithik73/rngh-toptab-issue.git

### Gesture Handler version

2.6.1

### React Native version

0.70.1

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

Pixel_4 (Android 12)

### Acknowledgements

Yes

**Comment by ttkien** [comment]
I am facing the same issue that using Swipeable in the tab mentioned above cannot work too

**Comment by j-piasecki** [comment]
The PR above should enable you to fix the issue. Using it, you will be able to wrap each nested screen with `<GestureHandlerRootView style={{flex: 1}} active>`. This way, when the `TopTab.Navigator` tries to prevent its parent from handling touch, the child GH root view will handle it instead.

Keep in mind that you won't be able to define interactions (`simultaneous`, `exclusive`) between gestures attached under different root views.

**Comment by hrithik73** [comment]
Hi @j-piasecki ,
Sorry for the late response, I am new to OSS, can you please provide any reference/tutorial on how to install from a PR
> @hrithik73 Can you try installing Gesture Handler from the aforementioned PR to check if it helps?

**Comment by j-piasecki** [comment]
@hrithik73 You could install it from the branch:
```
yarn add https://github.com/software-mansion/react-native-gesture-handler.git#@jpiasecki/root-view-always-active
```
Or from a specific commit:
```
yarn add https://github.com/software-mansion/react-native-gesture-handler.git#b9f236cfa180e8d42ca9d64c95db60c5c3aa327d
```
[...]

PR #2401, opened 2023-02-01
Allow `GestureHandlerRootView` to be manually made active

j-piasecki:
## Description

This PR allows users to override whether a particular `GestureHandlerRootView` is enabled or not. By default, only the topmost root view would be active, allowing all gestures to interact with each other.

The problem with that approach is that if any of the native views underneath called`requestDisallowInterceptTouchEvent`, all gestures would be canceled. It comes from the fact that `requestDisallowInterceptTouchEvent` calls are propagated up the view hierarchy and we expect `GestureHandlerRootView` to be relatively close to the top.

The simple solution would be to add another root view underneath the view that may call `requestDisallowInterceptTouchEvent` to prevent gestures from being canceled, this, however, resulted in two problems:
- the inner root view would not be enabled
- after adding the option to enable it manually, both root views would start handling the event, resulting in duplicated event streams

I've fixed the second problem by adding checks at the beginning of `extractGestureHandlers` and `traverseWithPointerEvents` to ensure that we're not extracting gestures attached under another enabled root view.

This allows for solving the problem described in https://github.com/software-mansion/react-native-gesture-handler/issues/2383.

## Test plan

Tested on the reproduction from https://github.com/software-mansion/react-native-gesture-handler/issues/2383

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
@@ -569,13 +569,16 @@ class GestureHandlerOrchestrator(
     extractGestureHandlers(wrapperView, tempCoords, pointerId, event)
   }
 
+  private fun shouldIgnoreSubtreeIfGestureHandlerRootView(view: View) =
+    view is RNGestureHandlerRootView && view != wrapperView && view.isRootViewEnabled()
+
   private fun extractGestureHandlers(
     viewGroup: ViewGroup,
     coords: FloatArray,
     pointerId: Int,
     event: MotionEvent,
   ): Boolean {
-    if (viewGroup is RNGestureHandlerRootView && viewGroup != wrapperView && viewGroup.isRootViewEnabled()) {
+    if (shouldIgnoreSubtreeIfGestureHandlerRootView(viewGroup)) {
       // When we encounter another active root view while traversing the view hierarchy, we want
       // to stop there so that it can handle the gesture attached under it itself.
       // This helps in cases where a view may call `requestDisallowInterceptTouchEvent` (which would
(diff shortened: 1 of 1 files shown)


m-bert (approved):
LGTM

lpatrun:
Can we expect this to be merged? It's approved after all 
🙈

j-piasecki:
Actually, I'm not sure if this is the best solution to this problem. I'll try to get back to it in the near future to see if I can find other approaches to this.

lpatrun:
@j-piasecki thank you for the quick response

javaing:
Still has same problem in 2024, any suggestions? tnanks

justinkx:
@j-piasecki Issue is still reproducible in Nested Tabs  with "@react-navigation/material-top-tabs": "7.3.7"
Do you think this approach is ready to be merged ?

j-piasecki:
As I mentioned in the previous message, this is more of a workaround than a solution, hence I changed the prop name with an `unstable_` prefix. The proper fix will require some under-the-hood changes (which we are actively working on) and removal of APIs that will not make use of the new implementation.

Given that it will take time to make it possible to implement a proper fix, I think we can merge this as a temporary solution. Keep in mind that this API **will** be removed in the future, and **will** cause unexpected behavior when trying to reference gestures attached under different root views in gesture relations.

packages/react-native-gesture-handler/src/specs/RNGestureHandlerRootViewNativeComponent.ts
```diff
@@ -1,6 +1,8 @@
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 import type { ViewProps } from 'react-native';
 
-interface NativeProps extends ViewProps {}
+export interface NativeProps extends ViewProps {
```
m-bert: What about `RootViewProps` / `RootViewNativeProps`?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -568,6 +575,17 @@ class GestureHandlerOrchestrator(
     pointerId: Int,
     event: MotionEvent,
   ): Boolean {
+    if (viewGroup is RNGestureHandlerRootView && viewGroup != wrapperView && viewGroup.isRootViewEnabled()) {
```
m-bert: It is the same condition as below, but with `viewGroup` instead of `view`. Maybe we could extract it to a helper function. WDYT?

m-bert (approved):
I trust you 🫣

Issue #2402, opened 2023-02-01
(Web) `setImmediate` is not defined


nandorojo:
### Description

 
<img width="817" alt="Screenshot 2023-02-01 at 3 51 28 PM" src="https://user-images.githubusercontent.com/13172299/216160709-bfcc3f35-179e-49aa-b7c2-c2154388ab57.png">

`GestureDetector` breaks on Next.js for using `setimmediate`. This same thing was fixed on Reanimated by switching to `requestAnimationFrame`

### Steps to reproduce

Use `GestureDetector` in Next.js causes this. It's an old [bug](https://github.com/expo/expo/issues/7996) from Reanimated.

The fix is here: https://github.com/software-mansion/react-native-reanimated/pull/1521/files

### Snack or a link to a repository

Let me know if this is needed.

### Gesture Handler version

2.9.0

### React Native version

RNW 0.18

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

nandorojo:
Here's a patch for anyone who needs it:

```patch
diff --git a/node_modules/react-native-gesture-handler/lib/module/handlers/gestures/GestureDetector.js b/node_modules/react-native-gesture-handler/lib/module/handlers/gestures/GestureDetector.js
index baf1ca3..0c9f50b 100644
--- a/node_modules/react-native-gesture-handler/lib/module/handlers/gestures/GestureDetector.js
+++ b/node_modules/react-native-gesture-handler/lib/module/handlers/gestures/GestureDetector.js
@@ -23,6 +23,8 @@ import { RNRenderer } from '../../RNRenderer';
 import { isExperimentalWebImplementationEnabled } from '../../EnableExperimentalWebImplementation';
 const ALLOWED_PROPS = [...baseGestureHandlerWithMonitorProps, ...tapGestureHandlerProps, ...panGestureHandlerProps, ...panGestureHandlerCustomNativeProps, ...longPressGestureHandlerProps, ...forceTouchGestureHandlerProps, ...flingGestureHandlerProps];
 
+const scheduleUpdate = Platform.OS == 'web' ? requestAnimationFrame : setImmediate;
+
 function convertToHandlerTag(ref) {
   if (typeof ref === 'number') {
     return ref;
@@ -84,7 +86,7 @@ function attachHandlers({
   // when it's ran
 
 
-  setImmediate(() => {
[...]

PR #2404, opened 2023-02-02
Use `requestAnimationFrame` on web instead of `setImmediate`

j-piasecki:
## Description

Replace `setImmediate` with `requestAnimationFrame` on the web.

Closes https://github.com/software-mansion/react-native-gesture-handler/issues/2402.

## Test plan

Tested on the Example app.

### DIFF
--- a/src/handlers/gestures/GestureDetector.tsx
+++ b/src/handlers/gestures/GestureDetector.tsx
@@ -68,6 +68,9 @@ export type GestureConfigReference = {
   useReanimatedHook: boolean;
 };
 
+const scheduleUpdate =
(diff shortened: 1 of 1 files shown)


nandorojo:
thanks for the quick fix!

tomekzaw (approved):

gorhom:
@j-piasecki thanks for the fix, any eta when its going to be released :)

janicduplessis:
This is causing an issue for me still during SSR since requestAnimationFrame will not be defined.

Using this patch for now https://github.com/janicduplessis/react-native-gesture-handler/commit/39440a3f0384cba343d19c81a034d88ff972f19f. Not sure if that is good enough to upstream.

Issue #2380, opened 2023-01-16
Swipeable horizontal in a vertical Flatlist: too sensitive


pierroo:
### Description

I have a vertical FlatList of items, each of those being wrapped in a Swipeable (horizontal).
It works well (except the performance issue already mentioned in other topics), however the issue is that sometimes upon scrolling, it accidentally triggers the horizontal scroll of those swipeable items.

It happens as soon as we scroll the flatlist too "diagonal" (thus not 100% vertical, which is rarely natural)

Any way to reduce the sensitivity of triggering the swipeable?
Or at least, to make sure it only happens on a truly horizontal gesture within the swipeable?

I already tried with all the frictions parameters, with Flatlist imported from RNGH, etc

### Steps to reproduce

1. Create a flatlist
2. each render item should be a swipeable from RNGH
3. scroll vertically, but sometimes with a more "diagonal" angle (even just 25degrees is enough to trigger the horizontal)
4. it triggers horizontal swipeable by mistake

### Snack or a link to a repository

(https://snack.expo.dev/YHU664VVC)

### Gesture Handler version

2.8.0

### React Native version

0.69.7

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

all

### Acknowledgements

Yes

j-piasecki:
Could you try the above PR and check if the new `dragOffsetLeft` and `dragOffsetRight` props are enough to solve your problem?

pierroo:
It works perfectly, thank you so much!
Any idea when it will be merged into an actual release, or should I patch-package for now?

j-piasecki:
I'd guess it would be released in GH 2.10.0, probably in one or two weeks (although no promises there). I don't see a problem with using patch-package till then.

VictorioMolina:
@j-piasecki Is there any other way to fix this problem without upgrading to 2.10.0? Because of my dependencias (expo sdk 47) I cannot use the `dragOffsetRight` property (upgrading react-native-gesture-handler > 2.9.0 breaks the EAS compilation for Expo 47).

j-piasecki:
You can also use [`patch-package`](https://www.npmjs.com/package/patch-package) to apply the changes from the PR linked above in your project.

anoopmm:
[...]

PR #2408, opened 2023-02-06
Add `dragOffsetFromLeftEdge` and `dragOffsetFromRightEdge` props to `Swipeable`

j-piasecki:
## Description

This PR adds `dragOffsetLeft ` and `dragOffsetRight` props to GH's `Swipeable` component. Those props to customize how much the user must swipe from each edge for the gesture to activate.

Closes https://github.com/software-mansion/react-native-gesture-handler/issues/2380

## Test plan

Create the `Swipeable` component, set the new props, and check that they work.

### DIFF
--- a/src/components/Swipeable.tsx
+++ b/src/components/Swipeable.tsx
@@ -74,6 +74,18 @@ export interface SwipeableProps
   rightThreshold?: number;
 
   /**
+   * Distance that the panel must be dragged from the left edge to be considered
+   * a swipe. The default value is 10.
+   */
+  dragOffsetLeft?: number;
+
+  /**
+   * Distance that the panel must be dragged from the right edge to be considered
+   * a swipe. The default value is 10.
+   */
+  dragOffsetRight?: number;
+
+  /**
    * Value indicating if the swipeable panel can be pulled further than the left
    * actions panel's width. It is set to true by default as long as the left
    * panel render method is present.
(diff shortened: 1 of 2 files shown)


marf:
Could this change please merged into master and released?

Thank you!

src/components/Swipeable.tsx
```diff
@@ -481,7 +499,7 @@ export default class Swipeable extends Component<
 
     return (
       <PanGestureHandler
-        activeOffsetX={[-10, 10]}
+        activeOffsetX={[-dragOffsetRight, dragOffsetLeft]}
```
tomekzaw, line 502: why right and then left? does "left" mean left edge of the screen or swiping to the left from the right edge of the screen? can we rename the props to make the names more explicit?

j-piasecki: In this case `dragOffsetRight` means the amount of drag right-to-left required for the gesture to start, and `dragOffsetLeft` means the amount of drag left-to-right.

The `right` is first, as `activeOffsetX` specifies a dead zone on the left and on the right. Since `dragOffsetRight` means dragging RTL, the dead zone needs to be on the left.

Would `dragOffsetRTL` and `dragOffsetLTR` be better (or non-abbreviated `dragOffsetRightToLeft` and `dragOffsetLeftToRight`)?

tomekzaw: `RTL` and `LTR` are strongly associated with the direction of the text, what about `dragOffsetFromLeftEdge` and `dragOffsetFromRightEdge`?

tomekzaw (approved):

Issue #2326, opened 2022-11-18
[Android] Cannot remove ripple from any GH touchables or buttons. There is an opacity limit


bfricka:
### Description

On Android, I cannot remove the ripple from `TouchableWithoutFeedback`, `RectButton`, etc. I can customize the color and opacity of the ripple, but the opacity seems to have a hard limit.

For example, `rgba(0,0,0,1)` shows a full black ripple, `rgba(0,0,0,0.5)`, shows a 50% opaque black ripple, but `rgba(0,0,0,0)` (or anything below 0.5), seems to not have any effect on reducing the opacity.

Since there is no way to disable the ripple, and there's no way to lower the opacity to 0, these touchables always show a ripple.

In my case, I'm trying to work through cancellation behavior in GH. It's frustratingly difficult, for example, to manage nested touchables. Using regular RN components, I can easily avoid triggering a parent touchable, by wrapping the area that I want to be independent in a `Pressable`. This behavior is very tricky in GH. There are some props that help (e.g. `disallowInterruption`, `shouldActivateOnStart`), but these can interfere with scrolling.

Anyways, that's why I'm trying to use a touchable without a ripple.

I was using 2.6.1, but updated to try to resolve the ripple issue and see if the nested touchable issues were fixed or at least more tractable.

### Steps to reproduce

1. Import a touchable from GH
2. Set the ripple color to transparent, or some low opacity value
3. For some touchables, transparent is the default, but this shows 50% black. For all touchables, neither the opacity nor "transparent", disable the ripple.

### Snack or a link to a repository

https://github.com/bfricka/rngh-ripple

### Gesture Handler version

2.8.0

### React Native version

0.67.5

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Real device

### Device model

Pixel 4a Android 12

### Acknowledgements

Yes

bfricka:
FWIW, I've done some investigation on this. React Native `Pressable` without a ripple config defined (or with `null` values for all config fields), does not trigger the ripple effect because it simply removes the drawable and does not dispatch the commands to the `RippleDrawable`. Here's the relevant code:

- [Pressable passes viewRef and ripple config to `useAndroidRippleForView`](https://github.com/facebook/react-native/blob/v0.70.6/Libraries/Components/Pressable/Pressable.js#L204)
[...]

PR #2418, opened 2023-02-16
Fix ripple always showing on Touchables when on Android 13

j-piasecki:
## Description

On Android 13 the [default ripple behavior changed](https://github.com/aosp-mirror/platform_frameworks_base/commit/475648ae0250621d62b855546351bc4f6192c417) from limiting ripple's alpha channel to max 128 to enforcing at minimum 128. This resulted in GH's touchables showing ripple when they shouldn't as they relied on setting the ripple color to `transparent` to prevent it.

This PR provides a simple fix that returns `null` instead of `RippleDrawable` when the ripple color is set to `transparent`. Thanks, @bfricka for the investigation!

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2326

## Test plan

Check the `Touchables` example on Android 13 and below

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -234,12 +234,18 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
       }
     }
 
-    private fun createSelectableDrawable(): Drawable {
+    private fun createSelectableDrawable(): Drawable? {
       // TODO: remove once support for RN 0.63 is dropped, since 0.64 minSdkVersion is 21
       if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
         context.theme.resolveAttribute(android.R.attr.selectableItemBackground, resolveOutValue, true)
         @Suppress("Deprecation")
         return resources.getDrawable(resolveOutValue.resourceId)
+      }
+
+      // Since Android 13, alpha channel in RippleDrawable is clamped between [128, 255]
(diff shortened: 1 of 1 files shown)


todorone:
@j-piasecki what could be done to push this?

On Android 13 all RNGH touchables look really ugly bc of ripple effect(RN's are ok)

tomekzaw (approved):

fiznool:
Should this have been marked as a breaking change? `Build.VERSION_CODES.TIRAMISU` isn't available unless you are using Android build tools 33 or above. Hence, a project using anything below build tools 33 will experience a compile error after upgrading to 2.10.0.

gogolubev:
I found a similar issue but on Android 11. The difference is that you need to disable TouchableOpacity immediately after the user taps on it. Also, the user must quickly tap on TouchableOpacity several times.

```typescript
function App(): JSX.Element {
  const [isDisabled, setIsDisabled] = React.useState(false);
  const onPress = () => {
    setIsDisabled(true);
    setTimeout(() => {
      setIsDisabled(false);
    }, 1000);
  };
  return (
    <GestureHandlerRootView style={{flex: 1}}>
      <SafeAreaView style={styles.container}>
        <TouchableOpacity
          disabled={isDisabled}
          onPress={onPress}
          style={[styles.button, isDisabled && styles.buttonDisabled]}>
          <Text style={styles.buttonText}>Press me</Text>
        </TouchableOpacity>
      </SafeAreaView>
    </GestureHandlerRootView>
  );
}
```


https://github.com/software-mansion/react-native-gesture-handler/assets/22788908/ca79bc0e-1ef3-4a90-851c-d2c96d572da5



So, if I replace
```kotlin
if (rippleColor == Color.TRANSPARENT && Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
    return null
}
```
with
```kotlin
if (rippleColor == Color.TRANSPARENT) {
    return null
}
```
everything works fine.

PR #2105, opened 2022-06-15
Returning Swipeable reference

franck-nadeau:
## Description

The following code allows me to do the following two things:

1. Close any previously opened Swipeable views when one is opened

````
  const openedRow = useRef();
  function renderItem({item: payable, index, separators}) {
    return (
      <Swipeable
        onSwipeableWillOpen={() => openedRow.current?.close()} // Close a previously opened row
        onSwipeableOpen={(_, swipeable) => (openedRow.current = swipeable)} // Keep a link to the currenlty
````

2. Close a Swipeable view when it is being deleted.
````
        renderRightActions={(_, __, swipeable) => {
          return (
             <TouchableOpacity
                onPress={() => {
                  swipeable.close(); // Close the Swipeable when the user presses the button
````

Resolves this issue: https://github.com/software-mansion/react-native-gesture-handler/issues/764

## Test plan

Made these changes in my own app and used patch-package. It has been working fine for my needs.

PR #2426, opened 2023-02-27
Returns Swipeable reference on renderLeftActions

ntocampos:
## Description

https://github.com/software-mansion/react-native-gesture-handler/pull/2105 returned the Swipeable reference, but only for `renderRightActions`.

This MR aims to return that same reference on `renderLeftActions` also. The use case is the same.

## Test plan

I've been using my fork on a few personal projects and it's working as expected.


j-piasecki (approved):
Thanks!

PR #2442, opened 2023-03-24
Fix Gesture Handlers getting stuck due to missing `up` event when using the new web implementation

j-piasecki:
## Description

Sometimes the view didn't receive the `pointerup` event when moving the pointer really fast, even though it was captured. This was causing a problem where the gesture would never be finished, preventing every other handler from activating.

Funny thing is, despite not receiving the `pointerup` event, `lostpointercapture` event was still there. I used this fact to fill the missing `up` event, so the gesture can correctly handle it.

I've only been able to reproduce it in Chrome, when using external display.

## Test plan

Tested on the Example app, specifically the `Draggable` example.

### DIFF
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -352,6 +352,9 @@ export default abstract class GestureHandler {
     if (this.config.needsPointerData) {
       this.sendTouchEvent(event);
     }
+
+    this.cancel();
+    this.reset();
   }
   protected onPointerOutOfBounds(event: AdaptedEvent): void {
     this.tryToSendMoveEvent(true);
--- a/src/web/tools/PointerEventManager.ts
+++ b/src/web/tools/PointerEventManager.ts
@@ -125,19 +125,18 @@ export default class PointerEventManager extends EventManager {
     this.view.addEventListener(
       'lostpointercapture',
       (event: PointerEvent): void => {
-        const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.UP);
+        const adaptedEvent: AdaptedEvent = this.mapEvent(
+          event,
+          EventTypes.CANCEL
+        );
 
         if (this.trackedPointers.has(adaptedEvent.pointerId)) {
           // in some cases the `pointerup` event is not fired, but `lostpointercapture` is
-          // we simulate the `pointerup` event here to make sure the gesture handler stops tracking it
-          this.trackedPointers.delete(adaptedEvent.pointerId);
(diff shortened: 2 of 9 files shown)


src/web/tools/PointerEventManager.ts
```diff
@@ -115,7 +119,28 @@ export default class PointerEventManager extends EventManager {
...
+      (event: PointerEvent): void => {
+        const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.UP);
+
+        if (this.trackedPointers.has(adaptedEvent.pointerId)) {
+          // in some cases the `pointerup` event is not fired, but `lostpointercapture` is
+          // we simulate the `pointerup` event here to make sure the gesture handler stops tracking it
+          this.trackedPointers.delete(adaptedEvent.pointerId);
+
+          if (--this.activePointersCounter > 0) {
+            adaptedEvent.eventType = EventTypes.ADDITIONAL_POINTER_UP;
```
kmagiera, line 136: Shouldn't we treat this event similarly to how we process CANCEL on Android and iOS? The issue with this approach is that sending UP event may result in some action being triggered (e.g. click). I think what could happen here is that the pointer may be lost due to the browser window changing focus, or some alert being presented?

j-piasecki: Makes sense to cancel the gesture in that case, I've updated the PR (also I've unified how `pointercancel` event is handled by the handlers).

> I think what could happen here is that the pointer may be lost due to the browser window changing focus, or some alert being presented?

In this specific case, it was happening when the pointer was moving really fast. The event stream would end without receiving either `up` or `cancel` events.

src/web/handlers/GestureHandler.ts
```diff
@@ -352,6 +352,9 @@ export default abstract class GestureHandler {
     if (this.config.needsPointerData) {
       this.sendTouchEvent(event);
     }
+
+    this.cancel();
+    this.reset();
```
tomekzaw, line 357: What about `this.end();`  from `RotationGestureHandler`?

src/web/tools/PointerEventManager.ts
```diff
@@ -8,6 +8,8 @@ import EventManager from './EventManager';
 import { isPointerInBounds } from '../utils';
 
 export default class PointerEventManager extends EventManager {
+  private trackedPointers: Set<number> = new Set<number>();
```
tomekzaw, line 11: will it work?
```suggestion
  private trackedPointers = new Set<number>();
```

tomekzaw (approved):

PR #2443, opened 2023-03-24
Change how velocity is calculated on the new web implementation

j-piasecki:
## Description

Changes how pointer velocity is calculated on the web. Previously, it was using only the last two events which wasn't the best solution as the speed of the pointer just before it's released may not represent the speed of the swipe.

This changes it to be calculated from the last 5 events, with the possibility to easily change it in the future (although some work may be required to handle quick direction changes correctly).

I've also removed a lot of magic numbers and replaced them with one magic formula 😅.

## Test plan

Use the `Velocity test` example added to the app

### DIFF
--- a/example/src/new_api/velocityTest/index.tsx
+++ b/example/src/new_api/velocityTest/index.tsx
@@ -1,0 +1,69 @@
+import { StyleSheet, View } from 'react-native';
+import Animated, {
+  measure,
+  useAnimatedRef,
+  useAnimatedStyle,
+  useSharedValue,
+  withDecay,
+} from 'react-native-reanimated';
+
+import React from 'react';
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
+
+const BOX_SIZE = 120;
+
+export default function App() {
+  const aref = useAnimatedRef() as React.RefObject<View>;
+  const offsetX = useSharedValue(0);
+  const offsetY = useSharedValue(0);
+
+  const pan = Gesture.Pan()
+    .onChange((event) => {
+      offsetX.value += event.changeX;
+      offsetY.value += event.changeY;
+    })
+    .onFinalize((event) => {
+      const size = measure(aref)!;
+
+      offsetX.value = withDecay({
+        velocity: event.velocityX,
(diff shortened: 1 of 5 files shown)


example/src/new_api/velocityTest/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+        velocity: event.velocityY,
+        clamp: [
+          -size.height / 2 + BOX_SIZE / 2,
+          size.height / 2 - BOX_SIZE / 2,
+        ],
+      });
+    });
+
+  const animatedStyles = useAnimatedStyle(() => ({
+    transform: [{ translateX: offsetX.value }, { translateY: offsetY.value }],
```
tomekzaw, line 43: can we split it into separate lines?

example/src/new_api/velocityTest/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }));
+
+  return (
+    <View style={styles.container} ref={aref}>
```
tomekzaw, line 47: wow, I didn't know that you can use animated refs on non-animated views, thanks! 🤯

example/src/new_api/velocityTest/index.tsx
j-piasecki: Prettier doesn't like your idea 😞.

example/src/new_api/velocityTest/index.tsx
j-piasecki: Ah yes, the totally-on-purpose animated ref on non-animated view. Happy to help.

src/web/tools/CircularBuffer.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.actualSize = Math.min(this.actualSize + 1, this.bufferSize);
+  }
+
+  public get(at: number): T {
+    if (this.actualSize === this.bufferSize) {
+      let index = (this.index + at) % this.bufferSize;
+      if (index < 0) {
+        index += this.bufferSize;
+      }
+
```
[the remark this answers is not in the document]
tomekzaw: idk, this is the only newline in all methods and it looks kinda redundant

PR #2446, opened 2023-03-27
Remove unnecessary `throw`

j-piasecki:
## Description

Removes redundant `throw` as it was inside `try...catch` anyway.

### DIFF
--- a/src/handlers/gestures/reanimatedWrapper.ts
+++ b/src/handlers/gestures/reanimatedWrapper.ts
@@ -34,7 +34,6 @@ try {
     // @ts-ignore Make sure the loaded module is actually Reanimated, if it's not
     // reset the module to undefined so we can fallback to the default implementation
     Reanimated = undefined;
-    throw new Error('react-native-reanimated is not found');
   }
 
   if (!Reanimated.setGestureState) {


src/handlers/gestures/reanimatedWrapper.ts
```diff
@@ -34,7 +34,6 @@ try {
     // @ts-ignore Make sure the loaded module is actually Reanimated, if it's not
     // reset the module to undefined so we can fallback to the default implementation
     Reanimated = undefined;
-    throw new Error('react-native-reanimated is not found');
   }
 
   if (!Reanimated.setGestureState) {
```
tomekzaw, line 39: Without `throw new Error` here the execution will continue. In the previous line, we assigned `Reanimated = undefined` and in the following line we have a check `!Reanimated.setGestureState`. If `Reanimated` is undefined, this check will fail.

j-piasecki: Good thing you've caught it 😅. I've put the code that followed the throw into the else branch so it should be clearer now.

tomekzaw: Honestly, the current flow looks even more complicated. What about the following?

```ts
try {
  Reanimated = require('react-native-reanimated');
} catch (e) {
  // When 'react-native-reanimated' is not available we want to
  // quietly continue
}

if (!Reanimated?.useSharedValue) {
  // @ts-ignore Make sure the loaded module is actually Reanimated, if it's not
  // reset the module to undefined so we can fallback to the default implementation
  Reanimated = undefined;
}

if (!Reanimated?.setGestureState) {
  Reanimated.setGestureState = () => {
    'worklet';
    console.warn(
      tagMessage(
        'Please use newer version of react-native-reanimated in order to control state of the gestures.'
      )
    );
  };
}

export { Reanimated };
```

j-piasecki: What about something like this:

```jsx
try {
  Reanimated = require('react-native-reanimated');
} catch (e) {
  // When 'react-native-reanimated' is not available we want to quietly continue
  // @ts-ignore TS demands the variable to be initialized
  Reanimated = undefined;
}

if (!Reanimated?.useSharedValue) {
  // @ts-ignore Make sure the loaded module is actually Reanimated, if it's not
  // reset the module to undefined so we can fallback to the default implementation
  Reanimated = undefined;
} else if (!Reanimated?.setGestureState) {
  // The loaded module is Reanimated but it doesn't have the setGestureState defined
  Reanimated.setGestureState = () => {
    'worklet';
    console.warn(
      tagMessage(
        'Please use newer version of react-native-reanimated in order to control state of the gestures.'
      )
    );
  };
}

export { Reanimated };
```

In your snippet, the execution continues after the first `if` and we're back where we've started 😅.

tomekzaw (approved):

src/handlers/gestures/reanimatedWrapper.ts
tomekzaw: Also, this is a nice place to use IIFE and early returns.

Issue #2449, opened 2023-03-31
Android build fails with Gradle Configuration Cache


pvegh:
### Description

Using Configuration Cache our build seems to fail with this error:

```
Could not evaluate onlyIf predicate for task ':react-native-gesture-handler:assertNoMultipleInstances'.
> Could not evaluate spec for 'Task satisfies onlyIf closure'.

```

The problem seems to be something in Gesture Handler not being correctly defined around here:

tasks.preBuild {
//    dependsOn assertionTask // this fixes the problem so it must be with assertionTask not being compatible
}

### Steps to reproduce

1. enable CC with:
org.gradle.unsafe.configuration-cache=true
2. Run app in Android Studio

And the first launch seems to work, but the second launch throws above exception. So when the config cache would be used the problem occures.

### Snack or a link to a repository

cant

### Gesture Handler version

2.9.0

### React Native version

0.71.4

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

pvegh:
Sorry I can't build a reproducer. It is as simple as enable build cache and keep using a project with RNGH.
Also using Gradle 7.6, it might help.

artemis-prime:
The workaround is just to delete the cache.  I'm not sure what a fix would mean.

pvegh:
Deleting the config cache defies the purpose of enabling it, so it's not an option.

The fix will be a mix of using APIs that supported in that environment, supporting the caching logic decide what can be reused and what needs to be re-run, and to stop using things that completely break the build when CC is enabled.

As a first step changes should be made to simply not break the whole build (by the looks of it it's what #2453 is doing).
Later some actual caching could be implemented, if applicable, to save time for every execution.

Documentation:
https://docs.gradle.org/current/userguide/configuration_cache.html

PR #2453, opened 2023-04-05
Make single instance assertion work with Gradle Configuration Cache

j-piasecki:
## Description

Makes `assertNoMultipleInstances` task not break when [Gradle Configuration Cache](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:intro) is enabled. 

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2449.

## Test plan

Enable configuration cache by adding `org.gradle.unsafe.configuration-cache=true` in `gradle.properties` file and build the app twice. Do it once without this patch and once with it.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,5 +1,7 @@
 import groovy.json.JsonSlurper
-import java.nio.file.Paths
+
+import javax.inject.Inject
+import java.nio.file.Files
 
 buildscript {
     def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['RNGH_kotlinVersion']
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -115,10 +93,23 @@ file("$REACT_NATIVE_DIR/ReactAndroid/gradle.properties").withInputStream { react
 def REACT_NATIVE_VERSION = reactProperties.getProperty("VERSION_NAME")
 def REACT_NATIVE_MINOR_VERSION = REACT_NATIVE_VERSION.startsWith("0.0.0-") ? 1000 : REACT_NATIVE_VERSION.split("\\.")[1].toInteger()
 
+
+def objects = objects
```
tomekzaw, line 97: why? BTW why this PR is different than https://github.com/software-mansion/react-native-reanimated/pull/4350/files?

j-piasecki: The RHS `objects` is actually a getter, and assigning its result to a variable makes it possible to capture in closure.

As for the difference between this and Reanimated, it does more complex checks, together with checking multiple locations. I can make it the same as in Reanimated.

tomekzaw: Thank you for the explanation! Yeah, let's move the implementation of this task into a class as in Reanimated (`abstract class NoMultipleInstancesAssertionTask extends DefaultTask`).

PR #2455, opened 2023-04-05
Add `Hover` gesture

j-piasecki:
## Description

Introduce a new gesture: `Hover`. It works exactly how you would expect based on the name 😄.
It supports hovering with a mouse & stylus on all platforms and makes it easy to add [Pointer interactions](https://developer.apple.com/documentation/uikit/pointer_interactions) to a view on iOS (via `withFeedback` method).

The API is identical to all other gestures, you can simply create the configuration object and set the callbacks:
```jsx
const gesture = Gesture.Hover()
  .onBegin(() => {
    console.log('hover begin');
  })
  .onFinalize(() => {
    console.log('hover finalize');
  })
```

## Test plan

This PR adds two examples: `Hover` and `HoverableIcons`, which can be used to verify that the gesture works correctly quickly.


kesha-antonov:
Nice feature !
Would be useful on macOS app

android/src/main/java/com/swmansion/gesturehandler/core/HoverGestureHandler.kt
```diff
@@ -0,0 +1,114 @@
...
+
+      handler!!.postDelayed(finishRunnable, 4)
+    } else if (!isWithinBounds) {
+      finish()
+    } else if (event.action == MotionEvent.ACTION_HOVER_MOVE || event.action == MotionEvent.ACTION_HOVER_ENTER) {
+      if (this.state == STATE_UNDETERMINED) {
+        begin()
+        activate()
+      }
+    }
```
m-bert, line 100: Here I would also change it to avoid chaining if...else (and maybe combine last if statement with its inner if, but I'm not so sure about this one).

```suggestion
if (event.action == MotionEvent.ACTION_HOVER_EXIT) {
  if (handler == null) {
    handler = Handler(Looper.getMainLooper())
  }

  handler!!.postDelayed(finishRunnable, 4)
  return
}

if (!isWithinBounds) {
  finish()
  return
} 

if (event.action == MotionEvent.ACTION_HOVER_MOVE || event.action == MotionEvent.ACTION_HOVER_ENTER) {
  if (this.state == STATE_UNDETERMINED) {
      begin()
      activate()
    }
}
```

android/src/main/java/com/swmansion/gesturehandler/core/HoverGestureHandler.kt
```diff
@@ -0,0 +1,114 @@
...
+    } else if (rootView == view) {
+      return false
+    } else if (rootView is ViewGroup) {
+      for (i in 0 until rootView.childCount) {
+        val child = viewConfigHelper.getChildInDrawingOrderAtIndex(rootView, i)
+        return isViewDisplayedOverAnother(view, other, child) ?: continue
+      }
+    }
+
+    return null
```
m-bert, line 40: I wouldn't make if...else chain since the if statements have returns inside

```suggestion
if (rootView == other) {
  return true
} 
if (rootView == view) {
  return false
} 

if (rootView is ViewGroup) {
  for (i in 0 until rootView.childCount) {
    val child = viewConfigHelper.getChildInDrawingOrderAtIndex(rootView, i)
    return isViewDisplayedOverAnother(view, other, child) ? : continue
  }
}

return null
```

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -464,15 +464,17 @@ class GestureHandlerOrchestrator(
...
+  private fun recordViewHandlersForPointer(view: View, coords: FloatArray, pointerId: Int, event: MotionEvent): Boolean {
     var found = false
     handlerRegistry.getHandlersForView(view)?.let {
       synchronized(it) {
         for (handler in it) {
           if (handler.isEnabled && handler.isWithinBounds(view, coords[0], coords[1])) {
-            recordHandlerIfNotPresent(handler, view)
-            handler.startTrackingPointer(pointerId)
-            found = true
+            if (event.action !in listOf(MotionEvent.ACTION_HOVER_EXIT, MotionEvent.ACTION_HOVER_ENTER, MotionEvent.ACTION_HOVER_MOVE) || handler is HoverGestureHandler) {
```
m-bert, line 473: Since this change introduces nested ifs I would change the outer one by negating condition and using `continue`. I think it will be easier to read.

m-bert (approved):
LGTM. Love the idea of introducing new gesture!

m-bert (approved):

src/web/tools/PointerEventManager.ts
```diff
@@ -122,6 +110,26 @@ export default class PointerEventManager extends EventManager {
...
+    });
+
+    this.view.addEventListener('pointerleave', (event: PointerEvent): void => {
+      if (event.pointerType === PointerType.TOUCH) {
+        return;
+      }
+
+      const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.LEAVE);
+
+      this.onPointerMoveOut(adaptedEvent);
```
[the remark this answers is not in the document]
j-piasecki: I think we can rename the methods in the future, but for now I've left a comment to explain why it's done this way: https://github.com/software-mansion/react-native-gesture-handler/pull/2455/commits/f35b2b2a99918e5cdd03e785623843824676c4a4

android/src/main/java/com/swmansion/gesturehandler/core/HoverGestureHandler.kt
j-piasecki: I've tried to use [`when`](https://kotlinlang.org/docs/control-flow.html#when-expression) and in this case it doesn't look that bad (https://github.com/software-mansion/react-native-gesture-handler/pull/2455/commits/1a95954dbff666c337357771e9e989dfb47ee176)

android/src/main/java/com/swmansion/gesturehandler/core/HoverGestureHandler.kt
j-piasecki: Same as above, tried using `when` but in this case it might have ended up worse than before (https://github.com/software-mansion/react-native-gesture-handler/pull/2455/commits/1a95954dbff666c337357771e9e989dfb47ee176), what do you think?

android/src/main/java/com/swmansion/gesturehandler/core/HoverGestureHandler.kt
m-bert: Yes, it looks much better now 🎉

android/src/main/java/com/swmansion/gesturehandler/core/HoverGestureHandler.kt
m-bert: In that case I think that leaving if's with returns (like in my suggestion) looks a bit clearer. Also the case with `ViewGroup` looks better in if version.

m-bert: In that case I think that leaving if's with returns (like in my suggestion) looks a bit clearer. Also the case with `ViewGroup` looks better in if version.

m-bert: In that case I think that leaving if's with returns (like in my suggestion) looks a bit clearer. Also the case with `ViewGroup` looks better in if version.

j-piasecki: Agreed - https://github.com/software-mansion/react-native-gesture-handler/pull/2455/commits/c498d0593765d8bfa3482b7f9a0ebb8249517fae

emzet93:
Question: are you guys planning to add `onHoverIn` and `onHoverOut` support for your Touchables components? Currently I'm using `TouchableWithoutFeedback` because it's super fast, however for web i'm still forced to use react native `Pressable` because of lack of hover support in touchables from rngh.

Issue #2351, opened 2022-12-17
Using RefreshControl from react-native-gesture-handler on Android causes flatlist scroll to press underlying touchables


dorbaruch2:
### Description

Hello, 

I'm encountering an issue when using the RefreshControl component exported from react-native-gesture-handler. 
I have a FlatList that renders a list of cards. The cards in the list are each wrapped with TouchableOpacity and should be clickable.
When I import both the FlatList and TouchableOpacity components from react-native-gesture-handler I am able to properly scroll the list without pressing the cards during scrolling. The problem starts when I also try to implement pull-to-refresh. Using react-native's RefreshControl causes the pull-to-refresh to not work at all. 
On the other hand, when using RefreshControl from react-native-gesture-handler the pull-to-refresh does work, but when I scroll the cards in the list keep getting pressed as well during scrolling. 

This is only an issue on Android, on iOS using RefreshControl from react-native-gesture-handler works perfectly and cards and not being pressed during scrolling.
Any help on this matter would be much appreciated.

### Steps to reproduce

1. Use FlatList from react-native-gesture-handler
2. Render TouchableOpacity from react-native-gesture-handler inside the FlatList items
3. Use RefreshControl from react-native-gesture-handler

### Snack or a link to a repository

https://snack.expo.dev/@dorkhe/gesture-handler-refresh-control-issue?platform=android

### Gesture Handler version

2.8.0

### React Native version

0.70.6

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

j-piasecki:
Hi! Could you check if the following patch solves the problem for you (and doesn't break the behavior in any way)?

```diff
diff --git a/node_modules/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt b/node_modules/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
index 2c5f4b3..c9527fb 100644
--- a/node_modules/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/node_modules/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -64,7 +64,7 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
[...]

PR #2457, opened 2023-04-18
Fix Gesture Handler's buttons activating after scroll when using `RefreshControl`

j-piasecki:
## Description

Closes https://github.com/software-mansion/react-native-gesture-handler/issues/2351

`RefreshControl` from GH is set up so that `ScrollView` wrapped with it should wait for it to fail. This way the scroll handler does not cancel the `RefreshControl`. The problem with this approach is that the `RefreshControl` handler always stays active preventing scroll from activating.

This PR adds a workaround for this problem using `SwipeRefreshLayoutHook`, which will cancel the `RefreshControl` if the `ScrollView` is not at the top during the swipe.

## Test plan

Tested on the snack from https://github.com/software-mansion/react-native-gesture-handler/issues/2351

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -64,7 +64,7 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
   private var shouldCancelWhenOutside = false
   var numberOfPointers = 0
     private set
-  private var orchestrator: GestureHandlerOrchestrator? = null
+  protected var orchestrator: GestureHandlerOrchestrator? = null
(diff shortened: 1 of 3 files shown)


janicduplessis:
@j-piasecki What do you think of the approach I used here https://github.com/software-mansion/react-native-gesture-handler/pull/2116?

Do you think it has the same drawback mentioned here https://github.com/software-mansion/react-native-gesture-handler/pull/2457/files#diff-b16ffbe3727ccf7c385e51ce84075f61ea7133b7db675d7191a0be902d1b2b49R243?

TBH its been too long since I worked on that so I don't remember exactly hehe.

tomekzaw (approved):

j-piasecki:
@janicduplessis I've tried it briefly and it seems to suffer from the same problem that this one fixes, although the root cause may be different.

PR #2461, opened 2023-04-20
Remove conditional imports for native components

j-piasecki:
## Description

Removes the conditional `requires` based on the React Native version.

Note that this change makes the library incompatible with React Native 0.63, where `codegenNativeComponent` is not available.

## Test plan

Test the Example and FabricExample apps.

### DIFF
--- a/src/components/GestureHandlerButton.tsx
+++ b/src/components/GestureHandlerButton.tsx
@@ -1,10 +1,5 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
-import { HostComponent, requireNativeComponent } from 'react-native';
+import { HostComponent } from 'react-native';
 import { RawButtonProps } from './GestureButtons';
-import { shouldUseCodegenNativeComponent } from '../utils';
-
-const RNGestureHandlerButtonNativeComponent = shouldUseCodegenNativeComponent()
-  ? require('../fabric/RNGestureHandlerButtonNativeComponent').default
-  : requireNativeComponent('RNGestureHandlerButton');
+import RNGestureHandlerButtonNativeComponent from '../fabric/RNGestureHandlerButtonNativeComponent';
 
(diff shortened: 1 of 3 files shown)


src/components/GestureHandlerButton.tsx
```diff
@@ -1,10 +1,5 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
-import { HostComponent, requireNativeComponent } from 'react-native';
+import { HostComponent } from 'react-native';
 import { RawButtonProps } from './GestureButtons';
-import { shouldUseCodegenNativeComponent } from '../utils';
-
-const RNGestureHandlerButtonNativeComponent = shouldUseCodegenNativeComponent()
-  ? require('../fabric/RNGestureHandlerButtonNativeComponent').default
-  : requireNativeComponent('RNGestureHandlerButton');
+import RNGestureHandlerButtonNativeComponent from '../fabric/RNGestureHandlerButtonNativeComponent';
```
tomekzaw, line 3: just wondering if the directory name `fabric` still makes sense

j-piasecki: Yeah, I guess renaming it to `codegen` would be a good idea.

tomekzaw: in many places i've seen name `components` which kind of makes sense but in our case `src/components/components` would be weird. as for `codegen`, it's ok but it's not true for Paper (although `codegenNativeComponent` is also a scam), but maybe `native` instead?

j-piasecki: One more thing that comes to mind is `specs`, but if not then `native` sounds good.

tomekzaw: we have a winner here

tomekzaw (approved):

PR #2467, opened 2023-04-25
Replace `setImmediate` and `requestAnimationFrame` with `queueMicrotask`

j-piasecki:
## Description

- `setImmediate` is not available on web & when using server-side-rendering
- `requestAnimationFrame` is not available when using server-side-rendering

## Test plan

Test on Example and FabricExample apps.


tomekzaw (approved):

hirbod:
I see signs of impact thanks to @slorber :D

Issue #2482, opened 2023-05-17
[2.10.0] Swipeable swiape don't prevent onPress


Bardiamist:
### Description

I use Swipeable and wrapped content into TouchableOpacity. After update react-native-gesture-handler from 2.9.0 to 2.10.0 swipe don't prevents onPress of TouchableOpacity

### Steps to reproduce

Write Swipeable content in TouchableOpacity

### Snack or a link to a repository

https://snack.expo.dev/@bardiamist/swipeable-onpress

### Gesture Handler version

2.10.0

### React Native version

0.71.8

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 14

### Acknowledgements

Yes

GoktuqCan:
Hey @Bardiamist I had the same issue. Solved it by using **BaseButton** exported from react-native-gesture-handle. You can also use the **BaseButton** if you want the opacity effect.
https://docs.swmansion.com/react-native-gesture-handler/docs/api/components/buttons#basebutton

TexxUkr:
To fix the issue it needs to use TouchableOpacity from react-native-gesture-handler rather than react-native.
in the problem code we have
`import { Text, View, StyleSheet, TouchableOpacity, Animated } from 'react-native';
import { RectButton, Swipeable } from 'react-native-gesture-handler';`
to fix the issue it needs to be
`import { Text, View, StyleSheet, Animated } from 'react-native';
import { RectButton, Swipeable, TouchableOpacity } from 'react-native-gesture-handler';`

dannybit:
I am using the TouchableOpacity from react-native-gesture-handler but I am still seeing the same issue

PierreCapo:
@dannybit on my side it works well if you place the `TouchableOpacity` from `react-native-gesture-handler` as first child of the `Swipeable` component 👍

PR #2483, opened 2023-05-17
Fix operation scheduling on iOS

j-piasecki:
## Description

https://github.com/software-mansion/react-native-gesture-handler/pull/2467 replaces calls to `setImmediate` and `requestAnimationFrame` with `queueMicrotask`. This introduces a bug, where on iOS the `flushOperations` method would get called too early.

This PR changes `flushOperations` to only be used on the new architecture, where it's actually required, and adds a retry mechanism for attaching handlers to native views. It also updates parts of the new arch-related code to function properly - finding the recognizer responsible for the touch responder and attaching the root view handler.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2482

## Test plan

Test on the Example and FabricExample apps.

### DIFF
--- a/ios/RNGestureHandlerManager.mm
+++ b/ios/RNGestureHandlerManager.mm
@@ -199,8 +199,9 @@ constexpr int NEW_ARCH_NUMBER_OF_ATTACH_RETRIES = 25;
 #ifdef RCT_NEW_ARCH_ENABLED
   UIView *touchHandlerView = childView;
 
-  while (touchHandlerView != nil && ![touchHandlerView isKindOfClass:[RCTSurfaceView class]])
+  while (touchHandlerView != nil && ![touchHandlerView isKindOfClass:[RCTSurfaceView class]]) {
     touchHandlerView = touchHandlerView.superview;
+  }
 #else
   UIView *parent = childView;
   while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)])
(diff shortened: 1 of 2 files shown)


ios/RNGestureHandlerManager.mm
```diff
@@ -164,18 +196,25 @@ - (id)handlerWithTag:(NSNumber *)handlerTag
 
 - (void)registerViewWithGestureRecognizerAttachedIfNeeded:(UIView *)childView
 {
+#ifdef RCT_NEW_ARCH_ENABLED
+  UIView *touchHandlerView = childView;
+
+  while (touchHandlerView != nil && ![touchHandlerView isKindOfClass:[RCTSurfaceView class]])
+    touchHandlerView = touchHandlerView.superview;
```
tomekzaw, line 203:
```suggestion
  while (touchHandlerView != nil && ![touchHandlerView isKindOfClass:[RCTSurfaceView class]]) {
    touchHandlerView = touchHandlerView.superview;
  }
```

ios/RNGestureHandlerManager.mm
```diff
@@ -208,7 +248,16 @@ - (void)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
+  // this little trick to get access to it
+  for (UIGestureRecognizer *recognizer in [viewWithTouchHandler gestureRecognizers]) {
+    if ([recognizer isKindOfClass:[RCTSurfaceTouchHandler class]]) {
+      touchHandler = recognizer;
+      break;
+    }
+  }
 #else
   RCTTouchHandler *touchHandler = [viewWithTouchHandler performSelector:@selector(touchHandler)];
 #endif
```
j-piasecki, line 263:
```suggestion
#endif // RCT_NEW_ARCH_ENABLED
```

Issue #2487, opened 2023-05-22
Swipeable doesn't work smoothly when placed on a Modal


zuhasohail18:
### Description

I am using Modal from 'react-native-modal'. I have a FlatList rendering multiple Swipeable from react-native-gesture-handler inside that Modal. Each Swipeable has two requirements for my app:
1. It must swipe to left/right and show required options.
2. If clicked on, it must redirect to required screen.

The problem I'm facing is that both swipe and touch gestures conflict and even when I'm trying to swipe, it detects the gesture as touch and navigates to new screen.
This problem only occurs when I use Swipeable inside a modal. On simple views, it works fine.

I'm using following versions of libraries:
"react-native-modal": "13.0.1",
"react-native-gesture-handler": "2.8.0"

### Steps to reproduce

1. Make a simple Modal on a screen
2. Put a swipeable inside that modal
3. Try to swipe left/right from almost center of swipeable component.

### Snack or a link to a repository

https://snack.expo.dev/BPdkQDF5y

### Gesture Handler version

2.8.0

### React Native version

0.70.6

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

schniddsel:
Same Issue here, fix in 2.10.1 did not solve the Problem #2482 It is the same behaviour

PR #2498, opened 2023-06-01
Correctly attach root view recognizer in modals on iOS

j-piasecki:
## Description

On iOS, the conditions to attach `RootViewGestureRecognizer` were satisfied neither on the new nor the old architecture.
This PR changes this logic:
- unifies extracting the RN's touch handler to simply loop on all recognizers attached to the root view, as it works in every case
- adds a check for `ModalHostViewController` when attaching the root view recognizer and if the view is in the modal the logic is different:
  - on the old arch: [RCTModalHostView](https://github.com/facebook/react-native/blob/b50874cad41f06519a263fa53d15f76c836fd167/packages/react-native/React/Views/RCTModalHostView.m#L95-L100) has a `UIView` as its child, and this one may only have one `RCTView` as its child - this one has the `RCTTouchHandler` recognizer attached
  - on the new arch: [RCTFabricModalHostViewController](https://github.com/facebook/react-native/blob/b50874cad41f06519a263fa53d15f76c836fd167/packages/react-native/React/Fabric/Mounting/ComponentViews/Modal/RCTFabricModalHostViewController.mm#L42) directly attaches `RCTSurfaceTouchHandler` to its direct subview

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2487.

## Test plan

Tested on Example and FabricExample apps and on the code from the issue.

### DIFF
--- a/ios/RNGestureHandlerManager.mm
+++ b/ios/RNGestureHandlerManager.mm
@@ -3,6 +3,7 @@
 #import <React/RCTComponent.h>
 #import <React/RCTEventDispatcher.h>
 #import <React/RCTLog.h>
+#import <React/RCTModalHostViewController.h>
 #import <React/RCTRootContentView.h>
 #import <React/RCTRootView.h>
 #import <React/RCTUIManager.h>
@@ -15,6 +16,7 @@
 #import "RNRootViewGestureRecognizer.h"
 
 #ifdef RCT_NEW_ARCH_ENABLED
+#import <React/RCTFabricModalHostViewController.h>
 #import <React/RCTSurfaceTouchHandler.h>
 #import <React/RCTSurfaceView.h>
 #import <React/RCTViewComponentView.h>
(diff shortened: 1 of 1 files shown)


ios/RNGestureHandlerManager.mm
```diff
@@ -199,15 +201,25 @@ - (void)registerViewWithGestureRecognizerAttachedIfNeeded:(UIView *)childView
...
-  while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)])
-    parent = parent.superview;
+  UIView *touchHandlerView = nil;
+
+  if ([[childView reactViewController] isKindOfClass:[RCTModalHostViewController class]]) {
+    touchHandlerView = [childView reactViewController].view.subviews[0];
+  } else {
+    UIView *parent = childView;
+    while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)])
+      parent = parent.superview;
```
tomekzaw, line 219:
```suggestion
    while (parent != nil && ![parent respondsToSelector:@selector(touchHandler)]) {
      parent = parent.superview;
    }
```

ios/RNGestureHandlerManager.mm
```diff
@@ -247,20 +259,19 @@ - (void)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
 
-  // touchHandler (RCTSurfaceTouchHandler) is private in RCTFabricSurface so we have to do
-  // this little trick to get access to it
+  // this way we can extract the touch handler on both architectures relatively easily
   for (UIGestureRecognizer *recognizer in [viewWithTouchHandler gestureRecognizers]) {
+#ifdef RCT_NEW_ARCH_ENABLED
     if ([recognizer isKindOfClass:[RCTSurfaceTouchHandler class]]) {
+#else
+    if ([recognizer isKindOfClass:[RCTTouchHandler class]]) {
+#endif
```
tomekzaw, line 270:
```suggestion
#endif // RCT_NEW_ARCH_ENABLED
```

tomekzaw (approved):

j-piasecki:
> For managed expo projects running on ios, do we need a GestureHandlerRootView because in v2.9.0 I did not need it. Now im getting an error that I do need it.

Strictly speaking, you don't need `GestureHandlerRootView` _at the moment_ on iOS, although I cannot with certainty say that it will not change in the future. For now, the root view is required on Android, but we decided to keep the error also on iOS since adding `GestureHandlerRootView` doesn't have any negative consequences and it helps guarantee that if your app works on one platform, it will also work on the other one.

> Also if I do need it, putting it very close to the root navigator should be fine right? That way I can put it in one place and use RNGH as I please.

Yes, that's actually the recommended approach - https://docs.swmansion.com/react-native-gesture-handler/docs/installation#js.

Eduard-Hasa:
> Strictly speaking, you don't need `GestureHandlerRootView` _at the moment_ on iOS, although I cannot with certainty say that it will not change in the future. For now, the root view is required on Android, but we decided to keep the error also on iOS since adding `GestureHandlerRootView` doesn't have any negative consequences and it helps guarantee that if your app works on one platform, it will also work on the other one.

Awesome. Tysm for the feedback. For whom it may concern the docs could benefit from mentioning it a little more clearly on which setups need GestureHandlerRootView and which ones dont. It would really help clarify things for new developers getting started. Just a simple chart showing expo managed needing GestureHandlerRootView and web not or whatever it is that you guys decide. I saw the bug reported that web was showing that error incorrectly so I thought for sure ios was a bug too. In the documentation the heading bare react native is right above the setup of GestureHandlerRootView so its easy to think its only for bare react native setups.

Thanks again. Love the library.

Issue #2517, opened 2023-06-15
[Static code analysis] Coverity Issues - Resource Leak


vamsigp:
### Description

Coverity Issue(s) - Static Code analyzer tool

File Name - NativeViewGestureHandler.kt
Method - onCancel()
Issue - resource returned from "android.view.MotionEvent.obtain(time, time, 3, 0f, 0f, 0)". Variable "event" going out of scope leaks the resource it refers to.
Fix - need to recycle the motion event.

FileName - RNGestureHandlerRootHelper.kt
Method - onCancel()
Issue - resource returned from "android.view.MotionEvent.obtain(time, time, 3, 0f, 0f, 0)". Variable "event" going out of scope leaks the resource it refers to.
Fix - need to recycle the motion event.

### Steps to reproduce

Run static code analyzer / coverity

### Snack or a link to a repository

NA

### Gesture Handler version

1.8.0

### React Native version

0.63.3

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Real device

### Acknowledgements

Yes

PR #2523, opened 2023-06-22
Recycle obtained `MotionEvents` on Android

j-piasecki:
## Description

Adds recycling for `MotionEvents` after they've been used.

Closes https://github.com/software-mansion/react-native-gesture-handler/issues/2517

## Test plan

Tested on the Example app


m-bert (approved):
LGTM

Issue #2530, opened 2023-07-06
TouchableOpacity.onPress not invoked when inside a Mapbox map inside PanGestureHandler.


mfazekas:
### Description

It seems that there is an issue with using `react-native-gesture-handler` and `@rnmapbox/maps` together. 

@rnmapbox/map can have an interactive View shown over a map at a specific coordinate. Normally you can put a TouchableOpacity there and onPress will be invoked as it should. Hovewer when this is inside a `PanGestureHandler` then onPress is not invoked inside the MapView, but it does get insivoked outside of MapView. Sometimes I do see the feedback of Touchable opacity but onPress is not called.

https://github.com/software-mansion/react-native-gesture-handler/assets/52435/7cf421da-7191-40b2-93be-58222479543f

In Xcode the view hierarchy looks like this:

<img width="465" alt="image" src="https://github.com/software-mansion/react-native-gesture-handler/assets/52435/7a7efada-80c4-4fc7-aac5-eea9b3685ef8">

### Steps to reproduce

Full repo is available at: https://github.com/mfazekas/rn_gestures_mbox_bug_report, see README.md as some changes are required in your `~/.netrc` as well:

The code is like:

```jsx
const MapComponent = () => {
  return (
    <View style={{flex: 1}}>
      <MapView style={{ flex: 1 }}>
        <Camera centerCoordinate={[-12.393354, 48.130477]} />
        <MarkerView
          coordinate={[-12.393354, 48.130477]}
          anchor={{ x: 0.5, y: 1 }}
        >
          <TouchableOpacity
            onPress={() => console.log('button in map pressed')}
            style={{
              backgroundColor: 'transparent',
              width: 60,
              height: 60,
            }}
          >
            <View style={{backgroundColor:'red', width: 60, height: 60, borderRadius: 10, padding: 10}}>
              <Text>Press me!</Text>
            </View>
          </TouchableOpacity>
        </MarkerView>
      </MapView>
      <TouchableOpacity
            onPress={() => console.log('button outside map pressed')}
            style={{
              backgroundColor: 'transparent',
              width: 60,
              height: 60,
              margin: 20,
            }}
          >
        <View style={{backgroundColor:'red', width: 60, height: 60, borderRadius: 10, padding: 10}}>
          <Text>Press me!</Text>
        </View>
      </TouchableOpacity>
    </View>
  );
};

const MapWPanGestureHandler = () => {
  return (
[...]

PR #2533, opened 2023-07-10
Only cancel JS responder when activating recognizer comes from Gesture Handler

j-piasecki:
## Description

Upon activation of a gesture recognizer, the iOS root view recognizer would cancel JS responders so they wouldn't conflict with each other, however this is currently done even if the activating recognizer doesn't come from Gesture Handler which may result in different behaviors for custom native components when used with or without Gesture Handler.

This PR adds a check to ensure the activating recognizer has an associated gesture handler before canceling JS responders.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2530

## Test plan

Tested on Example app and on the repro from the linked issue


tomekzaw (approved):

paulschreiber:
When will a new release include these changes? They don't appear to be in 2.12.1, which was release after July 11.

Issue #2544, opened 2023-07-14
Pan gesture handler with manualActivation(true) blocks TouchableOpacity inside GestureDetector on iOS.


gardenrabbit:
### Description

Pan gesture handler with manualActivation(true) blocks TouchableOpacity inside GestureDetector.
it works properly on Android.

### Steps to reproduce

1. create a pan gesture with manualActivation(true) for GestureDetector.
2. put TouchableOpacity in GestureDetector.
3. when pressing the TouchableOpacity, UI shows reaction but doesn't call onPress.

### Snack or a link to a repository

https://snack.expo.dev/e6oAaMt2s

### Gesture Handler version

2.12.0 

### React Native version

0.71.11

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

PR #2545, opened 2023-07-17
Make manual activation recognizer not cancel `UIControls`

j-piasecki:
## Description

On iOS, manual activation is implemented in a somewhat hacky way:
- if a gesture is declared with manual activation, an additional MA recognizer is added to a view
- that recognizer activated immediately, and the original recognizer requires it to fail in order to activate
- when we want to activate to original recognizer we simply cancel the MA recognizer

That's because there's no API for delaying or forcing recognizer activation on iOS, but making require-to-fail relations and failing recognizers is pretty straightforward.

The issue with that approach was the fact, that activation of the MA recognizer caused all touches in the view to be canceled, which meant that every UIControl wouldn't work.
This PR adds `cancelsTouchesInView = NO` to the MA recognizer, to prevent it from canceling underlying `UIControls`.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2544

## Test plan

Tested on the Example app and on the code from the issue.

### DIFF
--- a/ios/RNManualActivationRecognizer.m
+++ b/ios/RNManualActivationRecognizer.m
@@ -10,8 +10,9 @@
 {
   if ((self = [super initWithTarget:self action:@selector(handleGesture:)])) {
     _handler = gestureHandler;
-    self.delegate = self;
     _activePointers = 0;
+    self.delegate = self;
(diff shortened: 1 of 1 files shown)


m-bert (approved):
LGTM

Issue #700, opened 2019-07-24
Set cursor to the grabbing hand when gesture is active


EvanBacon:
😎

```css
cursor: grab;
```

Warus15:
Added in [2.6.1](https://github.com/software-mansion/react-native-gesture-handler/releases/tag/2.6.1) 😎

Note that this change is available in new web implementation. To enable it, call `enableExperimentalWebImplementation()` in the root of your project.

TomSwift:
Why is this desirable?   I just encountered the behavior and I don't want it, and not seeing a way to turn it off.

chriscoomber:
@TomSwift I agree, and I've come up with a workaround to turn it off:

- Find the view that your `GestureDetector` wraps and get a `ref` to it.
- In every gesture's `onBegin`, set `viewRef.current.style.cursor = 'auto';`

For example:

```typescript
const viewRef = useRef<Animated.View>(null);

/* ... */

const panGesture = Gesture.Pan()
  .onBegin(() => {
    // Web fix: don't use the grabby hand when drawing.
    if (Platform.OS === 'web' && viewRef.current !== null) {
      (viewRef.current as unknown as HTMLElement).style.cursor = 'auto';
    }
  })
  /* ...more config */;

/* ... */

<GestureDetector gesture={panGesture}>
  <Animated.View ref={viewRef} /* ...more props */>
```

This works because they set the cursor to `'grab'` right after calling `activate()`, and I think that's always followed by the state being transitioned to `BEGIN`.

m-bert:
Hi @TomSwift and @chriscoomber! We understand that this change may not be desirable and not everyone likes the idea. With that in mind I've prepared [this PR](#2550), which adds `activeCursor` prop. This way cursor keeps default appearance and if you would like to change it, you can use new prop to give it any css cursor value you'd like. I hope this will help!

PR #2550, opened 2023-07-19
`activeCursor` prop

m-bert:
## Description

This PR adds `activeCursor` prop which allows users to change cursor appearance upon handler activation. Initially it was added in [2.6.1](https://github.com/software-mansion/react-native-gesture-handler/releases/tag/2.6.1) as a response to [this issue](https://github.com/software-mansion/react-native-gesture-handler/issues/700). However, not everyone liked the idea (which is understandable), so we decided to turn this change into property. This way anyone who was using it still will be able to do so, and those who didn't like the change don't have to remove it on their own.

## Old API example

```jsx
<PanGestureHandler activeCursor="grab">
      <View style={styles.box}></View>
</PanGestureHandler>
```

## New API example

```jsx
  const panGesture = Gesture.Pan().activeCursor('wait');

  return (
      <View style={styles.box}></View>
    </GestureDetector>
  );
```

## Test plan

Tested on example app

### DIFF
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -167,10 +167,18 @@ export default abstract class GestureHandler {
       this.currentState === State.ACTIVE ||
       this.currentState === State.BEGAN
     ) {
-      this.moveToState(State.FAILED, sendIfDisabled);
-      if (this.config.activeCursor && this.config.activeCursor != 'auto') {
+      // Here the order of this if statement and moveToState call is important.
+      // At this point we can use currentState as previuos state, because immediately after changing cursor we call moveToState method.
+      // By checking whether previous state was ACTIVE, we can decide if we should reset the cursor or not.
+      if (
+        this.config.activeCursor &&
+        this.config.activeCursor != 'auto' &&
+        this.currentState === State.ACTIVE
+      ) {
         this.view.style.cursor = 'auto';
       }
+
+      this.moveToState(State.FAILED, sendIfDisabled);
     }
 
     this.resetProgress();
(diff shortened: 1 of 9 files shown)


src/web/handlers/GestureHandler.ts
```diff
@@ -167,8 +167,18 @@ export default abstract class GestureHandler {
       this.currentState === State.ACTIVE ||
       this.currentState === State.BEGAN
     ) {
+      // Here the order of this if statement and moveToState call is important.
+      // At this point we can use currentState as previuos state, because immediately after changing cursor we call moveToState method.
+      // By checking whether previous state was ACTIVE, we can decide if we should reset the cursor or not.
+      if (
+        this.config.activeCursor &&
+        this.config.activeCursor !== 'auto' &&
```
j-piasecki, line 175: I might not be seeing something, but why check if the `activeCursor` is not `auto`?

m-bert: Originally it was introduced to avoid situations where you have two (or more) handlers attached to one view, and one of them has `activeCursor` set to `"auto"`, while other has some custom cursor. In that case the one with `"auto"` would reset cursor.

j-piasecki (approved):
🚀

m-bert:
Hi @hirbod! Which Buttons do you have in mind? I've just checked `RectButton` and `BaseButton` and they both support `activeCursor` prop. Is there something that I've missed?

hirbod:
@m-bert I had to restart the TS Server, sorry about that. It didn't autocomplete the prop for me in my tests.

hirbod:
@m-bert I tried to replicate inside [this snack](https://snack.expo.dev/@hirbod/rngh-button-nativewind?platform=web) but it seems like it still turns the cursor into a grabbing hand (try the 2nd button)

Will need to test outside of Snack but it should work, even though its not the recommended version for snack.

m-bert:
@hirbod, have you managed to test that outside of Snack? While I do see this problem while using expo snack, it works fine outside of it.

hirbod:
@m-bert I can confirm that it works as expected outside of Snack.

PR #2551, opened 2023-07-24
Improve multi-pointer behavior (with ScrollView) on Android

j-piasecki:
## Description

This PR solves two problems:
- `ScrollView` wrapped with `NativeViewGestureHandler` wasn't canceling the root view handler, which in turn meant that [`shouldIntercept` flag was never set](https://github.com/software-mansion/react-native-gesture-handler/blob/355a82fd3cf39b1bd4a57af958f040b563a1823e/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt#L71) and event was [dispatched to both, the handler and the view](https://github.com/software-mansion/react-native-gesture-handler/blob/355a82fd3cf39b1bd4a57af958f040b563a1823e/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootView.kt#L35-L37).
- After a handler was activated and the root handler was canceled (setting `shouldIntercept` to true), and a new pointer was added to the screen, the root handler would begin and set `shouldIntercept` back to false. Again, this meant dispatching events both to the handler and the view.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/1679

## Test plan

Tested on the Example app and [reproduction from the issue](https://github.com/software-mansion/react-native-gesture-handler/issues/1679#issuecomment-1471576215)

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
@@ -58,6 +58,8 @@ class RNGestureHandlerRootHelper(private val context: ReactContext, wrappedView:
   private inner class RootViewGestureHandler : GestureHandler<RootViewGestureHandler>() {
     override fun onHandle(event: MotionEvent, sourceEvent: MotionEvent) {
       val currentState = state
+      // we shouldn't stop intercepting events when there is an active handler already, which could happen when
+      // adding a new pointer to the screen after a handler activates
       if (currentState == STATE_UNDETERMINED && (!shouldIntercept || orchestrator?.isAnyHandlerActive() != true)) {
         begin()
         shouldIntercept = false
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
```diff
@@ -59,8 +59,10 @@ class RNGestureHandlerRootHelper(private val context: ReactContext, wrappedView:
     override fun onHandle(event: MotionEvent, sourceEvent: MotionEvent) {
       val currentState = state
       if (currentState == STATE_UNDETERMINED) {
-        begin()
-        shouldIntercept = false
+        if (!shouldIntercept || orchestrator?.isAnyHandlerActive() != true) {
+          begin()
+          shouldIntercept = false
+        }
```
m-bert, line 65: I would probably combine this if with the outer one using `and`, but this also looks ok.

m-bert (approved):
LGTM

LuannguyenAxonActive (approved):

ajanuar:
@j-piasecki any plan to merge this PR? thanks

RodolfoGS:
Here @ dmregister mentioned that the patch with these changes worked, the errors are gone, and no side effects were observed https://github.com/software-mansion/react-native-gesture-handler/issues/1679#issuecomment-1716324209

I'll apply the changes too

PR #2553, opened 2023-07-25
Documentation: link leading to a 404 fix

orpos:
## Description
I found a link that is not working due to changes of location


j-piasecki (approved):
Thanks ❤️!

PR #2556, opened 2023-07-30
fix: export Touchable* component props

G07cha:
## Description

Touchable* components contain types that aren't compatible with React Native's it might be necessary for consumer to import and use them instead.

## Test plan

Compiled my branch and tested imports with `lib` output.

### DIFF
--- a/src/components/touchables/TouchableHighlight.tsx
+++ b/src/components/touchables/TouchableHighlight.tsx
@@ -7,7 +7,7 @@ import GenericTouchable, {
 import {
   StyleSheet,
   View,
-  TouchableHighlightProps,
+  TouchableHighlightProps as RNTouchableHighlightProps,
   ColorValue,
   ViewProps,
 } from 'react-native';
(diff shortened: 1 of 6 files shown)


src/components/touchables/TouchableHighlight.tsx
```diff
@@ -35,7 +38,7 @@ export default class TouchableHighlight extends Component<
     underlayColor: 'black',
   };
 
-  constructor(props: TouchableHighlightProps & GenericTouchableProps) {
+  constructor(props: RNTouchableHighlightProps & GenericTouchableProps) {
```
j-piasecki, line 41:
```suggestion
  constructor(props: TouchableHighlightProps) {
```

Is there a reason to leave an explicit union here instead of using the new type?

G07cha: Not at all, forgot to update. Thanks for pointing it out!

j-piasecki (approved):
🚀

PR #2558, opened 2023-07-31
Remove error with `GestureHandlerRootView` on web

### LINKED ISSUE #2557: [Web] Gesture Handler Root View error throws (unnecessarily)
### Description

Screenshot 2023-07-30 at 5 20 02 PM

This error is not necessary on Web, right? Websites are unlike apps, in that they shouldn't necessarily have a view wrapping them, as this can mess with scroll styles. If there is a provider that is going on inside of `GestureHandlerRootView`, then it should be split out as `GestureHandlerRootProvider`. If this context doesn't provide any actual benefit on Web, then this error should only be called on native.

### Steps to reproduce

Add any gesture view in a web app, without wrapping it in GestureHandlerRootView.

Let me know if this is needed.

### Gesture Handler version

2.12.0

### React Native version

0.72.3

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

Gestures on web don't work, unless they are wrapped inside `GestureHandlerRootView`. This PR removes the error, so that root view is no longer necessary.

Fixes #2557 

## Test plan

Tested on example app.

### DIFF
--- a/src/handlers/createHandler.tsx
+++ b/src/handlers/createHandler.tsx
@@ -402,7 +402,7 @@ export default function createHandler<
     }
 
     render() {
-      if (__DEV__ && !this.context && !isJestEnv()) {
+      if (__DEV__ && !this.context && !isJestEnv() && Platform.OS !== 'web') {
         throw new Error(
           name +
             ' must be used as a descendant of GestureHandlerRootView. Otherwise the gestures will not be recognized. See https://docs.swmansion.com/react-native-gesture-handler/docs/installation for more details.'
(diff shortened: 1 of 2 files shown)


Eduard-Hasa:
@j-piasecki 

Should this fix be needed for ios as well or does ios expo managed now require GestureHandlerRootView ?

j-piasecki:
[Answered here](https://github.com/software-mansion/react-native-gesture-handler/pull/2498#issuecomment-1702310654)

PR #2561, opened 2023-08-01
fix: export SwipeableProps

G07cha:
## Description

Allows library consumer to import them.

## Test plan

Generated a new build and tested with types from `lib/typescript`.


j-piasecki (approved):
🚀

PR #2563, opened 2023-08-04
Continuously check for pointer capture on web

j-piasecki:
## Description

Read the comment in the code.

## Test plan

Tested on the example app.


m-bert (approved):
While using `gotPointerCapture` seems like a more natural solution for me, I agree that it seems pointless to implement it right now after we have done it the other way. 

Great commentary!

PR #2564, opened 2023-08-08
Add empty example

m-bert:
## Description

This PR adds empty example to our example app. It can be helpful when it comes to testing reproductions from issues or just some new features/ideas. You can find similar idea in `react-native-reanimated` and it is very useful.

## Test plan

Tested (?) on example app


j-piasecki (approved):
Huh, I didn't even realize I needed that 🚀

PR #2575, opened 2023-08-24
Don't create a `RippleDrawable` when the color is transparent

j-piasecki:
## Description

When the `GestureHandlerButton` was disabled immediately after clicking it, the ripple animation would show even if its color was set to transparent. This PR updates the background creation logic not to create the `RippleDrawable` when its color would be transparent, which prevents the issue from happening.

It also removes some unnecessary SDK version checks, since RN 0.64 is the last supported version starting with Gesture Handler 2.10.0

## Test plan

Tested on the snippet from https://github.com/software-mansion/react-native-gesture-handler/pull/2418#issuecomment-1685239555


m-bert (approved):
LGTM

georgiemathews:
When will this be released, this is currently fixing build issues when compiling with android SDK 31

PR #2577, opened 2023-08-30
Remove cursor changes left after hover

j-piasecki:
## Description

Removes cursor changes from Pan that were introduced by the same PR as `Hover`. It was most likely left there after a conflict resolution instead of being removed then.


m-bert (approved):

j-piasecki:
Seems like it was included in another PR

PR #2579, opened 2023-08-31
Decouple the JS implementation from the web APIs

j-piasecki:
## Description

- adds a `GestureHandlerDelegate` interface to encapsulate web-specific calls from the js implementation
- makes `EventManager` generic so it doesn't rely on web types
- in some places adds `Platform.OS` checks, when there is the need to manipulate the underlying `HTMLElement`

It could still be beneficial to pass the ref to the js implementation instead of the node handle (which is the same thing on the web, but not on other platforms).

## Test plan

Tested on the example app.

### DIFF
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -34,7 +34,7 @@ export default abstract class GestureHandler {
   protected config: Config = { enabled: false };
   protected view!: HTMLElement;
 
-  protected eventManagers: EventManager[] = [];
+  protected eventManagers: EventManager<unknown>[] = [];
   protected tracker: PointerTracker = new PointerTracker();
 
   // Orchestrator properties
@@ -84,7 +84,7 @@ export default abstract class GestureHandler {
     }
   }
 
-  private addEventManager(manager: EventManager): void {
+  private addEventManager(manager: EventManager<unknown>): void {
     manager.setOnPointerDown(this.onPointerDown.bind(this));
     manager.setOnPointerAdd(this.onPointerAdd.bind(this));
     manager.setOnPointerUp(this.onPointerUp.bind(this));
@@ -113,7 +113,7 @@ export default abstract class GestureHandler {
     this.tracker.resetTracker();
     this.onReset();
     this.resetProgress();
-    this.eventManagers.forEach((manager: EventManager) =>
+    this.eventManagers.forEach((manager: EventManager<unknown>) =>
       manager.resetManager()
     );
     this.currentState = State.UNDETERMINED;
@@ -816,7 +816,7 @@ export default abstract class GestureHandler {
     return this.view;
   }
 
-  public getEventManagers(): EventManager[] {
+  public getEventManagers(): EventManager<unknown>[] {
     return this.eventManagers;
   }
 
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -30,7 +30,6 @@ export default abstract class GestureHandler {
   private handlerTag!: number;
   protected config: Config = { enabled: false };
 
-  protected eventManagers: EventManager<unknown>[] = [];
   protected tracker: PointerTracker = new PointerTracker();
 
   // Orchestrator properties
@@ -59,7 +58,7 @@ export default abstract class GestureHandler {
     this.delegate.init(viewRef, this);
   }
 
-  public addEventManager(manager: EventManager<unknown>): void {
+  public attachEventManager(manager: EventManager<unknown>): void {
     manager.setOnPointerDown(this.onPointerDown.bind(this));
     manager.setOnPointerAdd(this.onPointerAdd.bind(this));
     manager.setOnPointerUp(this.onPointerUp.bind(this));
@@ -72,8 +71,6 @@ export default abstract class GestureHandler {
     manager.setOnPointerMoveOver(this.onPointerMoveOver.bind(this));
     manager.setOnPointerMoveOut(this.onPointerMoveOut.bind(this));
     manager.setListeners();
-
-    this.eventManagers.push(manager);
   }
 
   //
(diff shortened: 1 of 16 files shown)


src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -0,0 +1,127 @@
...
+  onEnd(): void {
+    const config = this.gestureHandler.getConfig();
+
+    if (
+      config.activeCursor &&
+      config.activeCursor !== 'auto' &&
+      this.gestureHandler.getState() === State.ACTIVE
+    ) {
+      this.view.style.cursor = 'auto';
+    }
```
m-bert, line 101: Since this part is repeated 3 times in `onEnd`, `onCancel` and `onFail` methods, I would extract it to a function. Also I would do the same to conditions in `if` statement, something like:

```jsx
shouldResetCursor(){
    const config = this.gestureHandler.getConfig();

    return config.activeCursor && config.activeCursor !== 'auto' && this.gestureHandler.getState() === State.ACTIVE
}

tryResetCursor(){
    if(this.shouldResetCursor()){
    }
}
```

and then:

```suggestion
    this.tryResetCursor()
```

src/web/handlers/GestureHandler.ts
```diff
@@ -623,8 +569,16 @@ export default abstract class GestureHandler {
...
 
-  protected transformNativeEvent() {
-    return {};
+  protected transformNativeEvent(): Record<string, unknown> {
+    // those properties are shared by most handlers and if not this method will be overriden
+    const rect = this.delegate.measureView();
+
+    return {
+      x: this.tracker.getLastAvgX() - rect.pageX,
+      y: this.tracker.getLastAvgY() - rect.pageY,
```
m-bert, line 578: Why do we switch to `pageX | pageY` from `left | top`?

src/web/handlers/GestureHandler.ts
```diff
@@ -44,47 +39,26 @@ export default abstract class GestureHandler {
   protected shouldResetProgress = false;
   protected pointerType: PointerType = PointerType.NONE;
 
-  public constructor() {
-    this.hasCustomActivationCriteria = false;
+  protected delegate: GestureHandlerDelegate<unknown>;
```
m-bert, line 42: Why can't we do

```suggestion
  protected delegate: GestureHandlerDelegate<HTMLElement>;
```

src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -0,0 +1,127 @@
...
+export class GestureHandlerWebDelegate
+  implements GestureHandlerDelegate<HTMLElement>
+{
+  private _view!: HTMLElement;
+  private gestureHandler!: GestureHandler;
+  private eventManagers: EventManager<unknown>[] = [];
+
+  get view(): HTMLElement {
+    return this._view;
+  }
```
m-bert, line 22: This may be a matter of preference, but I wouldn't mix those two conventions and I would stick to 

```suggestion
  public getView(): HTMLElement {
    return this.view;
  }
```

m-bert (changes requested):
This looks good, left some comments though.

 I guess platform checks are there because we are using `GestureHandlerWebDelegate` also inside Windows and MacOS modules - in that case wouldn't it be better to create empty delegate as a placeholder? Then we wouldn't have to check whether platform is `web` inside files that are designed for web.

src/web/handlers/LongPressGestureHandler.ts
```diff
@@ -23,17 +24,14 @@ export default class LongPressGestureHandler extends GestureHandler {
   public init(ref: number, propsRef: React.RefObject<unknown>) {
     super.init(ref, propsRef);
 
-    this.view.oncontextmenu = () => false;
+    if (Platform.OS === 'web') {
+      (this.delegate.view as HTMLElement).oncontextmenu = () => false;
+    }
```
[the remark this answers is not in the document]
j-piasecki: Yes, assuming we want to use the JS implementation on different platforms than web 👀.

src/web/handlers/GestureHandler.ts
j-piasecki: I changed `getBoundingClientRect` to use `measure` from the delegate: https://github.com/software-mansion/react-native-gesture-handler/pull/2579/files#diff-c183a9c3b38cc57a48e82ed291800d840b640919c24da2d057c24be0878695d9R60-R69. It still uses `left` and `right`, but I changed the names to be `pageX` and `pageY` as it's more explicit about what those numbers refer to (they change during scroll).

src/web/handlers/GestureHandler.ts
j-piasecki: Same as above - we can, but shouldn't if we want to use the js implementation on different platforms.

src/web/handlers/NativeViewGestureHandler.ts
```diff
@@ -20,15 +21,19 @@ export default class NativeViewGestureHandler extends GestureHandler {
 
     this.setShouldCancelWhenOutside(true);
 
-    this.view.style['touchAction'] = 'auto';
+    if (Platform.OS === 'web') {
```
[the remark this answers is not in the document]
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/2579/commits/2e8155dc0d6ac0b59844046b63ebcc65cfded6c3

src/web/tools/GestureHandlerWebDelegate.ts
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/2579/commits/fd36dd004a06f8a9bd83cf22bf216b988d7e9d2e. I kept it as one method for now though.

src/web/tools/GestureHandlerWebDelegate.ts
j-piasecki: Makes sense - https://github.com/software-mansion/react-native-gesture-handler/pull/2579/commits/48e70b7148bf2bd53cd954592d749940709f8979

src/web/handlers/GestureHandler.ts
m-bert: Oh, okay. I missed the fact that after all it still uses `left` and `top` (somehow I thought it uses `pageX` and `pageY` now, even though they're part of `MouseEvent`, not `DOMRect`)

m-bert (approved):

PR #2589, opened 2023-09-16
New onSwipeableOpenDragStart and onSwipeableCloseDragStart events

### LINKED ISSUE #935: Wrong fire onSwipeableWillOpen

rampazzo1989:
## Description

Fixes #935 
The event onSwipeableWillOpen only starts right before the onSwipeableOpen event, not when the drag to open / close starts. This PR includes onSwipeableOpenStartDrag (direction: 'left' | 'right') and onSwipeableCloseStartDrag (direction: 'left' | 'right'), allowing execute actions when the animation to open/close starts (when drag starts).

## Test plan

Tested in the example app.


matheusinfo:
That would be great, i already had this problem on my app and had to fix it manually with patch

src/components/Swipeable.tsx
```diff
@@ -334,6 +344,32 @@ export default class Swipeable extends Component<
     if (ev.nativeEvent.oldState === State.ACTIVE) {
       this.handleRelease(ev);
     }
+    if (ev.nativeEvent.state === State.ACTIVE) {
+      const { velocityX, translationX: dragX } = ev.nativeEvent;
+
+      const { rowState } = this.state;
+
+      const { friction } = this.props;
+      const translationX = (dragX + DRAG_TOSS * velocityX) / friction!;
```
m-bert, line 353:
```suggestion
      const { velocityX, translationX: dragX } = ev.nativeEvent;
      const { rowState } = this.state;
      const { friction } = this.props;
      
      const translationX = (dragX + DRAG_TOSS * velocityX) / friction!;
```

src/components/Swipeable.tsx
```diff
@@ -334,6 +344,32 @@ export default class Swipeable extends Component<
...
+
+      let direction: 'left' | 'right' | undefined = undefined;
+
+      if (rowState === 0) {
+        direction = translationX > 0 ? 'left' : 'right';
+      } else if (rowState === 1) {
+        direction = 'left';
+      } else if (rowState === -1) {
+        direction = 'right';
+      }
```
m-bert, line 363:
```suggestion
      const direction =
        rowState === -1
          ? 'right'
          : rowState === 1
          ? 'left'
          : translationX > 0
          ? 'left'
          : 'right';
```

src/components/Swipeable.tsx
```diff
@@ -334,6 +344,32 @@ export default class Swipeable extends Component<
...
+
+      if (direction) {
+        if (rowState === 0) {
+          this.props.onSwipeableOpenStartDrag?.(direction);
+        } else {
+          this.props.onSwipeableCloseStartDrag?.(direction);
+        }
+      }
+    }
```
m-bert, line 372:
```suggestion
      if (rowState === 0) {
        this.props.onSwipeableOpenStartDrag?.(direction);
      } else {
        this.props.onSwipeableCloseStartDrag?.(direction);
      }
```

m-bert (changes requested):
Hi @rampazzo1989, thanks for this PR! It looks ok, but I'd like you to make some cosmetic changes, so I've left some comments.

rampazzo1989:
> Hi @rampazzo1989, thanks for this PR! It looks ok, but I'd like you to make some cosmetic changes, so I've left some comments.

Thank you for the review @m-bert! I've done the changes, really got better!

src/components/Swipeable.tsx
```diff
@@ -334,6 +344,28 @@ export default class Swipeable extends Component<
     if (ev.nativeEvent.oldState === State.ACTIVE) {
       this.handleRelease(ev);
     }
+    if (ev.nativeEvent.state === State.ACTIVE) {
```
m-bert, line 347:
```suggestion

    if (ev.nativeEvent.state === State.ACTIVE) {
```

m-bert (approved):

m-bert:
Now it looks much better, thank you once again! :white_check_mark:

latekvo:
Hey @rampazzo1989,

when working on `ReanimatedSwipeable`, i encountered this part of your code:

```js
const direction =
  rowState === -1
    ? 'right'
    : rowState === 1
    ? 'left'
    : translationX > 0
    ? 'left'
    : 'right';
```

When swiping to the left when the `left` panel is open, it returns `left`, but when swiping to the left when no panels are open, it returns `right`.

So when returning `left` and `right` to the callback, was your intention to return the swipe direction or the side of the panel that would be opened?

On another note, do you think the callbacks should always receive the swiping direction like they do when no panels are open, or should it stay how it is, with swiping direction returned always being `right` when `right` panel is open, and `left` when `left` one is?

Issue #2571, opened 2023-08-16
Add AGP8 support by adding buildConfig = true to buildFeatures


pvegh:
### Description

Upgrading to Android Gradle Plugin 8.1 and trying not to use the global flag
android.defaults.buildfeatures.buildconfig=true
which leads to slower builds than defining this only in needed modules
I'm getting the error:

```
* What went wrong:
A problem occurred configuring project ':react-native-gesture-handler'.
> defaultConfig contains custom BuildConfig fields, but the feature is disabled.
  To enable the feature, add the following to your module-level build.gradle:
  `android.buildFeatures.buildConfig true`
```

Fix: please add the flag
```
buildFeatures {
        buildConfig = true
    }
```
Other projects do something like this:
```
def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
    if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
        namespace "com.horcrux.svg"
    }
```
to build.gradle for AGP >= 8.1

### Steps to reproduce

1. Upgrade to AGP 8.1
2. add android.defaults.buildfeatures.buildconfig=true to gradle.properties (done by AGP upgrade wizard in AS)
3. ./gradlew assDebug

### Snack or a link to a repository

none

### Gesture Handler version

2.12.1

### React Native version

0.72.2

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

pvegh:
Can't do a repro sorry, I'm not into the RN game that much.

pvegh:
Can you please take a look at this? I know there's no repro here but it should be straightforward anyway

pvegh:
RN Thread about this: https://github.com/react-native-community/discussions-and-proposals/issues/671

pvegh:
In my opinion it doesn't need a repro, it's a build simple feature with full description what is missing

pvegh:
Created a PR, probably need some help since I don't usually do this. https://github.com/software-mansion/react-native-gesture-handler/pull/2599

pvegh:
Would a patch version with this be possible?

j-piasecki:
[Yes](https://github.com/software-mansion/react-native-gesture-handler/releases/tag/2.13.2) 😃

NitinMittal08:
FAILURE: Build failed with an exception.

* What went wrong:
[...]

PR #2599, opened 2023-09-21
add buildconfig flag for AGP8+

pvegh:
## Description
Add support for AGP8 by adding the buildConfig Flag indicating that this library uses custom buildconfig fields and needs it's BuildConfig class generated.
Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2571

## Test plan

No idea, need some help here since I'm not a RN person. I'm just trying to push support for AGP8 so we don't have to enable generating BuildConfig classes generally in our project, which has a performance impact.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -106,6 +106,12 @@ android {
         namespace "com.swmansion.gesturehandler"
     }
 
+    if (agpVersion.tokenize('.')[0].toInteger() >= 8) {
+        buildFeatures {
+            buildConfig = true
+        }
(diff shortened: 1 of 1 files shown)


m-bert (changes requested):
Hi @pvegh! Sorry for late response.

This PR looks fine, could you please merge `main` branch and add to description that it fixes #2571? Then we can merge it.

pvegh:
Done

m-bert (approved):

Issue #2620, opened 2023-10-06
Differences in onStart Callback with activateAfterLongPress Property for Pan Gesture between iOS and Android


huuthinh95:
### Description

When using the [Pan gesture](https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/pan-gesture) and adding the properties [activateAfterLongPress,](https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/pan-gesture#activateafterlongpressduration-number) the callback [onStart](https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/pan-gesture#onstartcallback) behaves differently on two platforms, iOS and Android (iOS is called twice, while Android is called once). I believe there is a problem on iOS because onStart should only be called once, just like on Android.

### Steps to reproduce

1. Open link Snack
2. Long-press the view for 500ms and then perform the Pan gesture.
3. Check log

### Snack or a link to a repository

https://snack.expo.dev/@huuthinhbk123/pangesture

### Gesture Handler version

2.13.1

### React Native version

0.72.3

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

codegabk:
Which version of gesture handler can i use to get this fix ?

PR #2628, opened 2023-10-10
Fix double start when `activateAfterLongPress` is used

j-piasecki:
## Description

It seems like native iOS recognizers (or at least Pan) don't respect when their state is changed from outside, which caused a problem with `.activateAfterLongPress` - before sending event in `Active` state, the state of the native recognizer is updated to `UIGestureRecognizerStateBegan`. Despite this, after moving the finger it tries to change the state to `UIGestureRecognizerStateBegan` once more, causing weird `Active` -> `Began`, `Began` -> `Active` chain.

This PR adds a simple check to detect the first event and stop processing in case it happens.

~~One thing that *could* be related is the fact that when the gesture fails before activation, the gesture stays in the `Began` state until the finger is lifted. This still needs to be investigated.~~ I think this is a separate issue.

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/2620

## Test plan

Tested on the example app.


apple/RNGestureHandler.m
```diff
@@ -240,6 +240,15 @@ - (void)sendEventsInState:(RNGestureHandlerState)state
       return;
     }
 
+    // When activating a handler under custom conditions (like activateAfterLongPress) recognizers don't respect
+    // manually changing their state, if we send a custom event in state ACTIVE and the recognizer will later update its
+    // state, we will end up sending ACTIVE->BEGAN and BEGAN->ACTIVE chain. To prevent this, we simply detect the first
+    // weird state change and stop it (then we don't update _lastState), so the second call ends up without state change
+    // and is fine.
```
m-bert, line 247: I would try to rewrite the first sentence - it's pretty long and hard to read.
```suggestion
    // Recognizers don't respect manually changing their state (that happens when we are activating handler
    // under custom conditions). If we send a custom event in state ACTIVE and the recognizer will later update its
    // state, we will end up sending ACTIVE->BEGAN and BEGAN->ACTIVE chain. To prevent this, we simply detect the first
    // weird state change and stop it (then we don't update _lastState), so the second call ends up without state change
    // and is fine.
```

m-bert (approved):

apple/RNGestureHandler.m
j-piasecki: > Recognizers don't respect manually changing their state

But they do, via `StateManager` in touch events

PR #2631, opened 2023-10-12
Fix scroll interactions

### LINKED ISSUE #2617: `FlatList` component does not scroll vertically on touch in mobile responsive mode
### Description

Hello, I'm experiencing an issue with the `FlatList` component from `react-native-gesture-handler` when it's used in conjunction with the `Swipeable` component. Specifically, the `FlatList` component does not scroll vertically on touch in mobile responsive mode.

### Steps to reproduce

1. Clone the branch with the development from the public repository: https://github.com/echarbeto/tamagui_test/tree/TEST/flatlist_with_swipeable_elements
2. 
3. Install the necessary libraries. Please note that this is an Expo SDK 49 project.
4. 
5. Run the project using the command npm run start.
6. 
7. Access the project from your browser at localhost:8081.
8. 
9. If accessing from a desktop, switch to mobile view.
10. 
11. Try to scroll vertically.

The issue occurs when trying to scroll vertically in mobile view. The FlatList component does not respond to vertical scrolling actions.
Screenshot_20231002_235335

https://github.com/echarbeto/tamagui_test/tree/TEST/flatlist_with_swipeable_elements

### Gesture Handler version

2.13.1

### React Native version

0.72.5

### Platforms

Web

### JavaScript runtime

None

### Workflow

Expo managed workflow

### Architecture

None

### Build type

Debug mode

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

During development of web version we initially decided to set `touchAction: none` to all gesture handlers' views in order to avoid scrolling while gestures like pan are active. Now it turns out that because of this decision, components like `FlatList` or `ScrollView` may become useless when combined with some handlers (or, like in the issue mentioned below - `Swipeable` components).

This PR aims to fix that problem by replacing `touchAction: none` with `e.preventDefault()`. 

Fixes #2617 

## Test plan

Tested on:
- Draggable example
- Transformation example (with `ScrollView` and `LoremIpsum`)
<details>
<summary> Modified code from issue </summary>

``` jsx
import React from 'react';
import { FlatList, GestureHandlerRootView } from 'react-native-gesture-handler';
import Swipeable from 'react-native-gesture-handler/Swipeable';

export default function Home() {
  type ItemProps = {
    item: {
      text: string;
    };
  };

  const data = Array.from({ length: 50 }, (_, i) => ({
    id: i,
    text: `Item ${i}`,
  }));

  const Item = ({ item }: ItemProps) => {
    return (
      <View style={{ margin: 10 }}>
        <Swipeable
          renderRightActions={() => (
            <View
              style={{
                justifyContent: 'center',
              }}>
              <Text style={{ color: 'white', textAlign: 'center' }}>
                Delete
              </Text>
            </View>
          )}>
          <View
            style={{
              height: 50,
              backgroundColor: 'white',
              justifyContent: 'center',
            }}>
            <Text>{item.text}</Text>
          </View>
        </Swipeable>
      </View>
    );
  };

  return (
    <GestureHandlerRootView>
      <FlatList
        data={data}
        keyExtractor={(item) => item.id.toString()}
        renderItem={({ item }) => <Item item={item} />}
        style={{ maxHeight: 400 }}
      />
    </GestureHandlerRootView>
  );
}

```
</details>


src/web/tools/InteractionManager.ts
```diff
@@ -97,10 +99,14 @@ export default class InteractionManager {
 
   public shouldHandlerBeCancelledBy(
     _handler: GestureHandler,
-    _otherHandler: GestureHandler
+    otherHandler: GestureHandler
   ): boolean {
-    //TODO: Implement logic
-    return false;
+    const shouldBeCancelledByScrollView =
```
j-piasecki, line 104: There's no actual check whether the `NativeViewGestureHandler` is attached to a scroll view. Either change the name or include the check, otherwise it may be confusing.

src/web/handlers/PanGestureHandler.ts
```diff
@@ -246,8 +246,14 @@ export default class PanGestureHandler extends GestureHandler {
...
 
-  protected onPointerUp(event: AdaptedEvent): void {
+  protected onPointerUp(
+    event: AdaptedEvent,
+    ev: TouchEvent | PointerEvent
+  ): void {
     super.onPointerUp(event);
+    if (ev) {
+      ev.preventDefault();
+    }
```
j-piasecki, line 256:
```suggestion
    ev?.preventDefault();
```

src/web/handlers/PanGestureHandler.ts
```diff
@@ -263,7 +269,10 @@ export default class PanGestureHandler extends GestureHandler {
       this.fail();
     }
   }
-  protected onPointerRemove(event: AdaptedEvent): void {
+  protected onPointerRemove(event: AdaptedEvent, ev: TouchEvent): void {
+    if (ev) {
+      ev.preventDefault();
+    }
```
j-piasecki, line 275: Same

src/web/handlers/PanGestureHandler.ts
```diff
@@ -286,9 +295,13 @@ export default class PanGestureHandler extends GestureHandler {
     }
   }
 
-  protected onPointerMove(event: AdaptedEvent): void {
+  protected onPointerMove(event: AdaptedEvent, ev: TouchEvent): void {
     this.tracker.track(event);
 
+    if (ev) {
+      ev.preventDefault();
+    }
```
j-piasecki, line 303: Same

src/web/handlers/PanGestureHandler.ts
```diff
@@ -299,11 +312,15 @@ export default class PanGestureHandler extends GestureHandler {
...
 
-  protected onPointerOutOfBounds(event: AdaptedEvent): void {
+  protected onPointerOutOfBounds(event: AdaptedEvent, ev: TouchEvent): void {
     if (this.getShouldCancelWhenOutside()) {
       return;
     }
 
+    if (ev) {
+      ev.preventDefault();
+    }
```
j-piasecki, line 322: Same

src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -287,10 +287,7 @@ export default class GestureHandlerOrchestrator {
       return false;
     }
 
-    if (
-      handler !== otherHandler &&
-      (handler.isAwaiting() || handler.getState() === State.ACTIVE)
-    ) {
+    if (handler.isAwaiting() || handler.getState() === State.ACTIVE) {
```
j-piasecki, line 290: This one seems a bit weird. Why do we want a handler to be cancellable by itself?

m-bert: It won't be cancelled by itself. The thing is, I've written it exactly how it's done in Android implementation, but if you look at [this line](https://github.com/software-mansion/react-native-gesture-handler/blob/a0506c026fd4a73238b491a1bb0a6b6c2b973ad3/src/web/tools/GestureHandlerOrchestrator.ts#L276C11-L276C11) you'll notice that this check is not necessary.

src/web/handlers/GestureHandler.ts
```diff
@@ -272,42 +272,60 @@ export default abstract class GestureHandler {
   // Event actions
   //
 
-  protected onPointerDown(event: AdaptedEvent): void {
+  protected onPointerDown(
+    adaptedEvent: AdaptedEvent,
+    _originalEvent?: TouchEvent | PointerEvent
```
j-piasecki, line 277: On Android handler methods also accept two events, and the untransformed one is called `sourceEvent`, I think we should stay consistent with names.

src/web/handlers/PanGestureHandler.ts
```diff
@@ -246,9 +246,16 @@ export default class PanGestureHandler extends GestureHandler {
     }
   }
 
-  protected onPointerUp(event: AdaptedEvent): void {
+  protected onPointerUp(
+    event: AdaptedEvent,
+    ev: TouchEvent | PointerEvent
```
j-piasecki, line 251: In `GestureHandler` the second argument is optional, shouldn't it also be the case here?

src/web/handlers/PanGestureHandler.ts
```diff
@@ -263,7 +270,11 @@ export default class PanGestureHandler extends GestureHandler {
       this.fail();
     }
   }
-  protected onPointerRemove(event: AdaptedEvent): void {
+  protected onPointerRemove(event: AdaptedEvent, ev: TouchEvent): void {
```
j-piasecki, line 273: Same here. (why `TouchEvent` instead of `TouchEvent | PointerEvent`?)

src/web/handlers/PanGestureHandler.ts
```diff
@@ -286,7 +297,11 @@ export default class PanGestureHandler extends GestureHandler {
     }
   }
 
-  protected onPointerMove(event: AdaptedEvent): void {
+  protected onPointerMove(event: AdaptedEvent, ev: TouchEvent): void {
```
j-piasecki, line 300: Same as above

src/web/handlers/PanGestureHandler.ts
```diff
@@ -299,11 +314,15 @@ export default class PanGestureHandler extends GestureHandler {
     super.onPointerMove(event);
   }
 
-  protected onPointerOutOfBounds(event: AdaptedEvent): void {
+  protected onPointerOutOfBounds(event: AdaptedEvent, ev: TouchEvent): void {
```
j-piasecki, line 317: Same as above

src/web/handlers/GestureHandler.ts
m-bert: Done in [5c26a86](https://github.com/software-mansion/react-native-gesture-handler/pull/2631/commits/5c26a86fd4ccfd2a17e2d54e49d54a50c5b2843a)

src/web/handlers/PanGestureHandler.ts
m-bert: Marking it as optional in `GestureHandler` allowed us to call `super(event)` without passing `sourceEvent`. Here we need it to be present. However it may be confusing, (especially when calling `sourceEvent?.cancelable`), so it's good idea to also mark it as optional.

Done in [5c26a86](https://github.com/software-mansion/react-native-gesture-handler/pull/2631/commits/5c26a86fd4ccfd2a17e2d54e49d54a50c5b2843a)

src/web/handlers/PanGestureHandler.ts
m-bert: Done in [5c26a86](https://github.com/software-mansion/react-native-gesture-handler/pull/2631/commits/5c26a86fd4ccfd2a17e2d54e49d54a50c5b2843a) and [cf5e1cc](https://github.com/software-mansion/react-native-gesture-handler/pull/2631/commits/cf5e1cc89a8febd6e07649c7e3af630d238e1c86)

src/web/handlers/PanGestureHandler.ts
m-bert: Done in [5c26a86](https://github.com/software-mansion/react-native-gesture-handler/pull/2631/commits/5c26a86fd4ccfd2a17e2d54e49d54a50c5b2843a) and [cf5e1cc](https://github.com/software-mansion/react-native-gesture-handler/pull/2631/commits/cf5e1cc89a8febd6e07649c7e3af630d238e1c86)

src/web/tools/InteractionManager.ts
```diff
@@ -97,10 +99,13 @@ export default class InteractionManager {
...
     _handler: GestureHandler,
-    _otherHandler: GestureHandler
+    otherHandler: GestureHandler
   ): boolean {
-    //TODO: Implement logic
-    return false;
+    return (
+      otherHandler instanceof NativeViewGestureHandler &&
+      otherHandler.getLastSentState() === State.ACTIVE &&
+      otherHandler.getState() === State.ACTIVE
```
j-piasecki, line 107: Is there a reason for checking both, `state` and `lastSentState`?

src/web/handlers/PanGestureHandler.ts
```diff
@@ -246,9 +246,16 @@ export default class PanGestureHandler extends GestureHandler {
...
   }
 
-  protected onPointerUp(event: AdaptedEvent): void {
+  protected onPointerUp(
+    event: AdaptedEvent,
+    sourceEvent?: TouchEvent | PointerEvent
+  ): void {
     super.onPointerUp(event);
 
+    if (sourceEvent?.cancelable) {
```
j-piasecki, line 255: Don't other handlers (pinch, rotate) also need something like that?

m-bert: I thought that they will need it, however testing on `Transformations` example showed that it is not necessary.

src/web/tools/InteractionManager.ts
m-bert: After some testing I can tell that there isn't. Initially I used that to check whether scroll was already active, or just got activated, but it seems that it isn't required anymore.

src/web/handlers/PanGestureHandler.ts
j-piasecki: I don't think there's a scroll view in that example

m-bert: There isn't but I modifed it so it is similar to Draggable, i.e:

```jsx
  return (
    <ScrollView style={styles.home}>
      <LoremIpsum />
      <Photo />
      <LoremIpsum />
    </ScrollView>
  );
}
```

j-piasecki:
Looks like there still are some things left to iron out

https://github.com/software-mansion/react-native-gesture-handler/assets/21055725/1cb24eba-cf57-4911-897d-2168d170edf3

https://github.com/software-mansion/react-native-gesture-handler/assets/21055725/a12fb4da-658e-48da-96a3-68c738685bc3

https://github.com/software-mansion/react-native-gesture-handler/assets/21055725/73a84a8d-f182-43d9-9011-f3519335fe37

All captured on Android 13, Chrome 118.0.5993.111

m-bert:
@j-piasecki I agree that there are some problems that should be fixed. However, I've checked the one with `swipeable` on `main` and it seems that it is already there, even without this PR. Could you please check if that's true?

j-piasecki:
> However, I've checked the one with swipeable on main and it seems that it is already there, even without this PR. Could you please check if that's true?

Works fine for me on the current `main` (8d3d5e0b7)

https://github.com/software-mansion/react-native-gesture-handler/assets/21055725/aaa5b43c-f061-4b4b-a91c-34265aa86bb2

GuestInCorle:
@m-bert What's blocking this pull request from getting a review?

j-piasecki:
iOS doesn't seem to work at all yet - all handlers are running simultaneously with scrolling the page and navigation gestures

m-bert:
Hi @GuestInCorle!

Problems that @j-piasecki described seem to affect both platforms, Android and iOS.

I'm not sure about the navigation part and I have yet to check that. However, most of the issues seem to originate from default `minDistance` in `PanGestureHandler` (which right now is set to 15px, see [this line](https://github.com/software-mansion/react-native-gesture-handler/blob/35ec17d636220c1ad4226456aa3b03c846ba16a8/src/web/constants.ts#L1)). By manipulating this value I was able to achieve what I think is the solution that we are looking for.

m-bert:
This PR was superseded by #2788, followed by #2794.

PR #2636, opened 2023-10-16
Circumvent a detox bug

ggilles:
## Description

We are using `react-native-gesture-handler` in our app, and we are also using [Detox](https://github.com/wix/Detox) to run end-to-end tests.

Unfortunately, because of a bug in Detox and the particular usage of `NSObject:performSelector` in `react-native-gesture-handler`, we run into [this issue](https://github.com/wix/Detox/issues/4194) where only the first single tap is detected by `react-native-gesture-handler` when running under Detox.

The root cause of the problem is that Detox doesn't correctly monitor `cancelPreviousPerformRequestsWithTarget`: this result in the selector being called even when it should not.

Clearly, the problem is on the Detox side, but Detox is also quite complex and fixing the bug on the Detox side will take a long time.

This pull request adds a boolean allowing to ignore spurious calls to the selector.

## Test plan

We checked that this change really [solves the bug in the repro app we created](https://github.com/JulienFremyMP/DetoxApp)

### DIFF
--- a/ios/Handlers/RNPanHandler.m
+++ b/ios/Handlers/RNPanHandler.m
@@ -34,6 +34,7 @@
   __weak RNGestureHandler *_gestureHandler;
   NSUInteger _realMinimumNumberOfTouches;
   BOOL _hasCustomActivationCriteria;
+  BOOL _activateAfterLongPressScheduled;
 }
 
 - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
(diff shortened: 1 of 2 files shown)


m-bert:
I agree with @j-piasecki - this should be fixed on detox side.

ggilles:
@j-piasecki @m-bert Thanks for taken the time to consider the change, I understand your take on it. Working on fixing it on the Detox side indeed: https://github.com/wix-incubator/DetoxSync/pull/68. Meanwhile we will use our own fork of RNGH.

PR #2642, opened 2023-10-17
Fix background color when `foreground` is set.

### LINKED ISSUE #2632: Foreground={true} removes background for touchables
### Description

Not sure if this is expected behavior mentioned somewhere in the docs, but setting the `foreground` prop to `true` on any touchable causes any background styles to be discarded in Android.

image

### Steps to reproduce

1. Open the Snack
2. Select Android
3. Run either locally or online

https://snack.expo.dev/@joshuajadaniel/rngh---foreground-property-effect-on-background

### Gesture Handler version

~2.12.0

### React Native version

0.72.3

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

This PR aims to fix background color issue on android. If you set `foreground` prop to `true` on our `Buttons`, background color doesn't update.

Fixes #2632 

## Test plan

Tested on the code from issue.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -234,8 +234,7 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
       if (useDrawableOnForeground && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
         foreground = selectable
         if (_backgroundColor != Color.TRANSPARENT) {
-          setBackgroundColor(_backgroundColor)
-          updateBackgroundColor(_backgroundColor, borderRadius, selectable)
+          updateBackgroundColor(_backgroundColor, borderRadius, null)
         }
       } else if (_backgroundColor == Color.TRANSPARENT && rippleColor == null) {
         background = selectable
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -188,12 +188,21 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
       return false
     }
 
+    private fun updateBackgroundColor(backgroundColor: Int, borderRadius: Float, selectable: Drawable?) {
+      val colorDrawable = PaintDrawable(backgroundColor)
+
+      if (borderRadius != 0f) {
+        colorDrawable.setCornerRadius(borderRadius)
+      }
+
+      val layerDrawable = LayerDrawable(if (selectable != null) arrayOf(colorDrawable, selectable) else arrayOf(colorDrawable))
+      background = layerDrawable
+    }
+
     fun updateBackground() {
       if (!needBackgroundUpdate) {
         return
       }
-      needBackgroundUpdate = false
-
       if (_backgroundColor == Color.TRANSPARENT) {
         // reset background
         background = null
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -188,12 +188,21 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
...
+
+      val layerDrawable = LayerDrawable(if (selectable != null) arrayOf(colorDrawable, selectable) else arrayOf(colorDrawable))
+      background = layerDrawable
+    }
+
     fun updateBackground() {
       if (!needBackgroundUpdate) {
         return
       }
-      needBackgroundUpdate = false
```
j-piasecki, line 195: I don't see a reason for removing this (and if there is one, it would basically always stay `true` so there would be no need for this variable to exist).

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -224,18 +233,12 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
         foreground = selectable
         if (_backgroundColor != Color.TRANSPARENT) {
           setBackgroundColor(_backgroundColor)
```
j-piasecki, line 235:
```suggestion
```

I believe this one is redundant. The intent was probably to call `setBackgroundColor` method from `View`, but the button overrides it, so it ended up doing basically `_backgroundColor = _backgroundColor`.

Changing it to `super.setBackgroundColor()` would solve that, but then we would ignore the border radius logic.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -224,18 +233,12 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
...
+          updateBackgroundColor(_backgroundColor, borderRadius, selectable)
```
j-piasecki, line 236:
```suggestion
          updateBackgroundColor(_backgroundColor, borderRadius, null)
```

Now the ripple effect is applied to both the background and the foreground (you can check by comparing two buttons side-by-side, the one with `foreground={true}` is darker when pressed). Since the foreground is already set to the selectable drawable, there's no need to also add it to the background.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: Good catch, I must have deleted it by accident while playing with some `Log.d`.

Brought back in [9cc93c3](https://github.com/software-mansion/react-native-gesture-handler/pull/2642/commits/9cc93c3baeeef1c80d17176169069e9e14420cd4)

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: I also thought that it won't do anything, but I left it to make sure that the value in that field is correct. But as you said, we can probably ignore that (I don't think it will cause any trouble).

Done in [ebcf766](https://github.com/software-mansion/react-native-gesture-handler/pull/2642/commits/ebcf766d4fbedbfabec8b64f849fa938bc24fb37)

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: Good point, I've just checked that and it indeed was darker. 

Done in [ebcf766](https://github.com/software-mansion/react-native-gesture-handler/pull/2642/commits/ebcf766d4fbedbfabec8b64f849fa938bc24fb37)

j-piasecki (approved):
🚀

PR #2654, opened 2023-10-23
Fix tvOS build

### LINKED ISSUE #2647: v2.13+ not building on tvOS
### Description

Since #2455 was merged and v2.13 was released the project isn't building on an app that uses react-native-tvos. This is because it's trying to load iOS specific code on the tvOS target.

### Steps to reproduce

1. Build package for tvOS
2. Get "unavailable on tvOS" errors for the new hover functions

https://github.com/react-native-tvos/react-native-tvos

### Gesture Handler version

2.13.0

### React Native version

0.72.6

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

tvOS apps are not building since release 2.13.0, because `UIHoverGestureRecognizer` used in `Hover` does not exist on tvOS. This PR changes `#if` inside `RNHoverHandler` in order to avoid using this recognizer on tvOS. It means that functions inside `Hover` can be treated as `no-op`

Fixes #2647 

## Test plan

Tested that tvOS example builds when used with `Hover`

### DIFF
--- a/FabricExample/ios/Podfile.lock
+++ b/FabricExample/ios/Podfile.lock
@@ -1057,7 +1057,7 @@ PODS:
     - React-jsi (= 0.72.0-rc.5)
     - React-logger (= 0.72.0-rc.5)
     - React-perflogger (= 0.72.0-rc.5)
-  - RNGestureHandler (2.13.3):
+  - RNGestureHandler (2.13.4):
     - hermes-engine
     - RCT-Folly (= 2021.07.22.00)
     - RCTRequired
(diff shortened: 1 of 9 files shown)


RuudBurger:
It doesn't build because of this line: https://github.com/software-mansion/react-native-gesture-handler/blob/main/ios/RNGestureHandlerButtonComponentView.mm#L49

Added 
```
#if !TARGET_OS_TV
  _buttonView.exclusiveTouch = newProps.exclusive;
#endif
```
and then your branch builds.

geroale:
Looking forward for the merge

m-bert:
@RuudBurger I've added your suggested change, could you please check if now this PR works?

j-piasecki (approved):

PR #2664, opened 2023-10-31
Raise limit on simultaneous gesture handlers from 20 to 50

vthorsteinsson:
## Description

In `GestureHandlerOrchestrator.kt`, there is a pretty arbitrary limit on the number of simultaneous gesture handlers,
via the constant `SIMULTANEOUS_GESTURE_HANDLER_LIMIT` which is set to 20. A comment with the constant states
that "this limit doesn't necessarily need to exist" but that it is there for simple efficiency reasons.

My company publishes a React Native game app which hits this limit at run-time, so we have been applying a patch to
`react-native-gesture-handler` in our repo to raise this limit to 50. This solves our problem and appears to have
no adverse side effects. Raising the constant from 20 to 50 makes a couple of vectors slightly bigger but the
memory consumption due to this is negligible.

## Test plan

We have been applying this patch to a production app for a couple of months now with no apparent adverse effects.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
@@ -606,9 +606,9 @@ class GestureHandlerOrchestrator(
   }
 
   companion object {
-    // The limit doesn't necessarily need to exists, it was just simpler to implement it that way
+    // The limit doesn't necessarily need to exist, it was just simpler to implement it that way
     // it is also more allocation-wise efficient to have a fixed limit
-    private const val SIMULTANEOUS_GESTURE_HANDLER_LIMIT = 20
+    private const val SIMULTANEOUS_GESTURE_HANDLER_LIMIT = 50
 
     // Be default fully transparent views can receive touch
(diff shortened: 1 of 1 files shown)


m-bert:
Hi @vthorsteinsson! Thanks for this PR! We all agree that upper limit for gesture handlers present in app shouldn't exist. While raising the limit to 50 should be enough, we think it is better to remove it (at least for now). That's why I've prepared [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2672). It would be great if you could check if it works for you and prevents your app from crashing

vthorsteinsson:
Hi @m-bert and many thanks for your efforts! We can of course perform in-house testing of the other PR and will do so. Be aware, though, that this bug was difficult to catch in local testing and mostly occurred for users in the field (unfortunately). But we are optimistic that your approach solves the problem.

vthorsteinsson:
Hi again @m-bert ! FYI we have tested [your PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2672) locally and our app seems to run fine with it.

m-bert:
Hi! That's great to hear! I'll merge it asap :)

m-bert:
Hi! I'm closing this PR since the one that I've mentioned above has just been merged. Thank you one more time for this submission!

PR #2666, opened 2023-11-03
Move REACT_NATIVE_VERSION to native-only code

### LINKED ISSUE #2660: Web-only app breaks
### Description

Screenshot 2023-10-27 at 2 47 15 PM

`react-native` isn't used on Web. This check should be moved to a native-only file.

### Steps to reproduce

Create a web-only project that imports RNGH.

 https://stackblitz.com/edit/nextjs-tcj82d?file=pages%2Findex.tsx,package.json,next.config.js,rn-pkc.json

### Gesture Handler version

2.13.4

### React Native version

N/A

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

chriscoomber:
## Description

Fixes #2660

Depending on this package in a web-only app is causing a compile error. This is due to this package importing `react-native/package.json`, which for web-only builds doesn't exist. This import is only done to obtain the react-native version number, which isn't used for web anyway. So it seems easy enough to move this version number to a native-only file.

I chose to export a **function** so that it's a _logic error_ to use `getReactNativeVersion` on web, and an error is thrown (right now it's only used in one place, behind a check that makes sure it runs on native only). It might have been easier to just export a **const**, with fake values on web, but that would hide the error for a potentially more confusing error down the line.

## Test plan

Not tested, sorry. Would appreciate if someone could help out here.


m-bert (approved):
Hi @chriscoomber! Thanks for submitting this PR! Please make sure that CI passes and we're good to go 🟢

PR #2672, opened 2023-11-09
Remove Gesture Handler limits on Android

m-bert:
## Description

While looking at PRs I've stumbled upon [this one](https://github.com/software-mansion/react-native-gesture-handler/pull/2664). Turns out that `SIMULTANEOUS_GESTURE_HANDLER_LIMIT` does not refer to simultaneous gestures, but to all gesture handlers present in app. The upper limit shouldn't exist - this PR aims to remove it.

## Test plan

Tested on example app.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
@@ -7,6 +7,7 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.EditText
 import java.util.*
+import kotlin.collections.ArrayList
 
 class GestureHandlerOrchestrator(
   private val wrapperView: ViewGroup,
@@ -19,13 +20,10 @@ class GestureHandlerOrchestrator(
    * traversing view hierarchy and looking for gesture handlers.
    */
   var minimumAlphaForTraversal = DEFAULT_MIN_ALPHA_FOR_TRAVERSAL
-
-  private val gestureHandlers = arrayOfNulls<GestureHandler<*>?>(SIMULTANEOUS_GESTURE_HANDLER_LIMIT)
-  private val awaitingHandlers = arrayOfNulls<GestureHandler<*>?>(SIMULTANEOUS_GESTURE_HANDLER_LIMIT)
-  private val preparedHandlers = arrayOfNulls<GestureHandler<*>?>(SIMULTANEOUS_GESTURE_HANDLER_LIMIT)
-  private val handlersToCancel = arrayOfNulls<GestureHandler<*>?>(SIMULTANEOUS_GESTURE_HANDLER_LIMIT)
-  private var gestureHandlersCount = 0
-  private var awaitingHandlersCount = 0
+  private val gestureHandlers = arrayListOf<GestureHandler<*>?>()
+  private val awaitingHandlers = arrayListOf<GestureHandler<*>?>()
+  private val preparedHandlers = arrayListOf<GestureHandler<*>?>()
+  private val handlersToCancel = arrayListOf<GestureHandler<*>?>()
   private var isHandlingTouch = false
   private var handlingChangeSemaphore = 0
   private var finishedHandlersCleanupScheduled = false
@@ -60,9 +58,9 @@ class GestureHandlerOrchestrator(
     }
   }
 
-  private inline fun compactHandlersIf(handlers: Array<GestureHandler<*>?>, count: Int, predicate: (handler: GestureHandler<*>?) -> Boolean): Int {
+  private inline fun compactHandlersIf(handlers: ArrayList<GestureHandler<*>?>, predicate: (handler: GestureHandler<*>?) -> Boolean): Int {
     var out = 0
-    for (i in 0 until count) {
+    for (i in 0 until handlers.size) {
       if (predicate(handlers[i])) {
         handlers[out++] = handlers[i]
       }
@@ -72,7 +70,7 @@ class GestureHandlerOrchestrator(
 
   private fun cleanupFinishedHandlers() {
     var shouldCleanEmptyCells = false
-    for (i in gestureHandlersCount - 1 downTo 0) {
+    for (i in gestureHandlers.size - 1 downTo 0) {
       val handler = gestureHandlers[i]!!
       if (isFinished(handler.state) && !handler.isAwaiting) {
         gestureHandlers[i] = null
@@ -85,18 +83,17 @@ class GestureHandlerOrchestrator(
         }
       }
     }
+
     if (shouldCleanEmptyCells) {
-      gestureHandlersCount = compactHandlersIf(gestureHandlers, gestureHandlersCount) { handler ->
-        handler != null
-      }
+      gestureHandlers.removeAll(listOf(null))
     }
+
     finishedHandlersCleanupScheduled = false
   }
 
   private fun hasOtherHandlerToWaitFor(handler: GestureHandler<*>): Boolean {
-    for (i in 0 until gestureHandlersCount) {
-      val otherHandler = gestureHandlers[i]!!
-      if (!isFinished(otherHandler.state) && shouldHandlerWaitForOther(handler, otherHandler)) {
+    for (otherHandler in gestureHandlers) {
+      if (!isFinished(otherHandler!!.state) && shouldHandlerWaitForOther(handler, otherHandler)) {
         return true
       }
     }
@@ -115,9 +112,11 @@ class GestureHandlerOrchestrator(
   }
 
   private fun cleanupAwaitingHandlers() {
-    awaitingHandlersCount = compactHandlersIf(awaitingHandlers, awaitingHandlersCount) { handler ->
+    val awaitingHandlersCount = compactHandlersIf(awaitingHandlers) { handler ->
       handler!!.isAwaiting
     }
+
+    awaitingHandlers.subList(awaitingHandlersCount, awaitingHandlers.size).clear()
   }
 
   /*package*/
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):
Looks good to me (although I haven't tested it)

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -72,7 +70,7 @@ class GestureHandlerOrchestrator(
 
   private fun cleanupFinishedHandlers() {
     var shouldCleanEmptyCells = false
-    for (i in gestureHandlersCount - 1 downTo 0) {
+    for (i in gestureHandlers.size - 1 downTo 0) {
```
j-piasecki, line 73:
```suggestion
    for (handler in gestureHandlers.reversed()) {
```

Same in other places when iterating in reverse

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -399,27 +397,27 @@ class GestureHandlerOrchestrator(
...
+    awaitingHandlers.add(handler)
     with(handler) {
       isAwaiting = true
       activationIndex = this@GestureHandlerOrchestrator.activationIndex++
     }
   }
 
   private fun recordHandlerIfNotPresent(handler: GestureHandler<*>, view: View) {
-    for (i in 0 until gestureHandlersCount) {
+    for (i in 0 until gestureHandlers.size) {
```
j-piasecki, line 414:
```suggestion
    for (otherHandler in gestureHandlers) {
```

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -60,9 +58,9 @@ class GestureHandlerOrchestrator(
     }
   }
 
-  private inline fun compactHandlersIf(handlers: Array<GestureHandler<*>?>, count: Int, predicate: (handler: GestureHandler<*>?) -> Boolean): Int {
+  private inline fun compactHandlersIf(handlers: ArrayList<GestureHandler<*>?>, predicate: (handler: GestureHandler<*>?) -> Boolean): Int {
```
j-piasecki, line 61:
```suggestion
  private inline fun compactHandlersIf(handlers: List<GestureHandler<*>?>, predicate: (handler: GestureHandler<*>?) -> Boolean): Int {
```

There's no point to depend on a specific list type

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
m-bert: Done in [6957bb1](https://github.com/software-mansion/react-native-gesture-handler/pull/2672/commits/6957bb1387ac38cfad1c24e4c5b5e63333f7c28f)

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -399,27 +397,27 @@ class GestureHandlerOrchestrator(
   }
 
   private fun addAwaitingHandler(handler: GestureHandler<*>) {
-    for (i in 0 until awaitingHandlersCount) {
+    for (i in 0 until awaitingHandlers.size) {
```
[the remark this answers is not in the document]
m-bert: I've removed that loop by using `.contains` method, let me know what you think ([6957bb1](https://github.com/software-mansion/react-native-gesture-handler/pull/2672/commits/6957bb1387ac38cfad1c24e4c5b5e63333f7c28f))

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
m-bert: Same as above ([6957bb1](https://github.com/software-mansion/react-native-gesture-handler/pull/2672/commits/6957bb1387ac38cfad1c24e4c5b5e63333f7c28f))

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -19,13 +20,10 @@ class GestureHandlerOrchestrator(
...
-  private val gestureHandlers = arrayOfNulls<GestureHandler<*>?>(SIMULTANEOUS_GESTURE_HANDLER_LIMIT)
-  private val awaitingHandlers = arrayOfNulls<GestureHandler<*>?>(SIMULTANEOUS_GESTURE_HANDLER_LIMIT)
-  private val preparedHandlers = arrayOfNulls<GestureHandler<*>?>(SIMULTANEOUS_GESTURE_HANDLER_LIMIT)
-  private val handlersToCancel = arrayOfNulls<GestureHandler<*>?>(SIMULTANEOUS_GESTURE_HANDLER_LIMIT)
-  private var gestureHandlersCount = 0
-  private var awaitingHandlersCount = 0
+  private val gestureHandlers = arrayListOf<GestureHandler<*>?>()
+  private val awaitingHandlers = arrayListOf<GestureHandler<*>?>()
+  private val preparedHandlers = arrayListOf<GestureHandler<*>?>()
+  private val handlersToCancel = arrayListOf<GestureHandler<*>?>()
```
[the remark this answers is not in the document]
m-bert: You're right. I left `null` possibility because of how removing handler worked before last commit, but now I think it should be fine ([6957bb1](https://github.com/software-mansion/react-native-gesture-handler/pull/2672/commits/6957bb1387ac38cfad1c24e4c5b5e63333f7c28f))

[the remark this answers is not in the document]
m-bert: But I would be grateful if you could check if that approach is correct (line 74)

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
m-bert: As far as I can see you cannot use index operator `[]` on `List` (and I think that makes sense). Given that, I don't think it is worth to change this methods body just to get rid of `ArrayList` in favor of `List`.

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -115,18 +112,19 @@ class GestureHandlerOrchestrator(
   }
 
   private fun cleanupAwaitingHandlers() {
-    awaitingHandlersCount = compactHandlersIf(awaitingHandlers, awaitingHandlersCount) { handler ->
+    val awaitingHandlersCount = compactHandlersIf(awaitingHandlers) { handler ->
       handler!!.isAwaiting
     }
+
+    awaitingHandlers.subList(awaitingHandlersCount, awaitingHandlers.size).clear()
```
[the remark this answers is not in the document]
m-bert: `removeRange` doesn't work as it gives `Cannot access 'removeRange': it is protected in 'ArrayList'` error.

But the second one should be ok.

piaskowyk (approved):

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -60,23 +57,10 @@ class GestureHandlerOrchestrator(
...
-
   private fun cleanupFinishedHandlers() {
-    var shouldCleanEmptyCells = false
-    for (i in gestureHandlersCount - 1 downTo 0) {
-      val handler = gestureHandlers[i]!!
+    for (handler in gestureHandlers.reversed()) {
       if (isFinished(handler.state) && !handler.isAwaiting) {
-        gestureHandlers[i] = null
-        shouldCleanEmptyCells = true
+        gestureHandlers.remove(handler)
```
j-piasecki, line 63: Can we modify the collection while iterating over it?

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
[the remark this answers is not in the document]
j-piasecki: I'm not sure what you mean, there's a closing bracket on line 74 😅

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
m-bert: I'm pretty sure that calling `Collection.reversed()` returns new collection - in that case it should be fine ([see docs](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/reversed.html) for context)

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -182,21 +159,22 @@ class GestureHandlerOrchestrator(
...
       }
     }
-    for (i in toCancelCount - 1 downTo 0) {
-      handlersToCancel[i]!!.cancel()
+
+    for (otherHandler in handlersToCancel.reversed()) {
+      otherHandler.cancel()
     }
 
+    handlersToCancel.clear()
```
j-piasecki, line 174: Do we need `handlersToCancel` at all here? I think you can iterate over `gestureHandlers.asReversed()` and just cancel relevant handlers inside the loop.

m-bert: I agree - done in [da2c421](https://github.com/software-mansion/react-native-gesture-handler/pull/2672/commits/da2c4211da791d8b347723d90e77a0a7a4db29b6)

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -160,19 +159,12 @@ class GestureHandlerOrchestrator(
       activationIndex = this@GestureHandlerOrchestrator.activationIndex++
     }
 
-    // Cancel all handlers that are required to be cancel upon current handler's activation
-    for (otherHandler in gestureHandlers) {
+    for (otherHandler in gestureHandlers.reversed()) {
```
j-piasecki, line 162:
```suggestion
    for (otherHandler in gestureHandlers.asReversed()) {
```

I don't think the list is being mutated here, so we shouldn't need a copy.

j-piasecki (approved):
🚀

PR #2675, opened 2023-11-16
Fix inputs focus on safari

### LINKED ISSUE #2673: Select Input not working (Refs) on Safari
### Description

When using select components (drop down pickers) the after focusing an element on (Safari web) the element won't release the focus.

261797283-7261afd3-9b12-4585-ad92-1dde85a7d430

https://github.com/react-navigation/react-navigation/issues/11546

We have tracked the problem to an specific commit on RNGH (Enable the new web implementation by default)
https://github.com/software-mansion/react-native-gesture-handler/commit/7799d7d74e832b85a286648451d812f14d2bb5de
### Steps to reproduce

1. install React Native Web 0.19.0
2. have a select element focused on a page/screen inside a drawer navigator* (react-navigation/drawer react-native-picker/picker)
3. Open it on Safari web
4. select the Select Element
5. the element breaks the flow

https://github.com/ArthurDias01/rngh-bug

### Gesture Handler version

2.10.0

### React Native version

0.72.6

### Platforms

Web

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

None

### Device model

Safari Web - macOS 14 Sonoma

### Acknowledgements

Yes

m-bert:
## Description

It seems that since release of new web implementations `select` doesn't work as it should on Safari. The problem is caused by `setPointerCapture` method. Even though it is called on all browsers, input stays focused only on Safari. 

This PR changes method of calling `setPointerCapture` - now we do this only for `Div` elements.

Fixes #2673

## Test plan

Tested on example app

### DIFF
--- a/src/web/tools/PointerEventManager.ts
+++ b/src/web/tools/PointerEventManager.ts
@@ -19,7 +19,10 @@ export default class PointerEventManager extends EventManager<HTMLElement> {
       const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.DOWN);
       const target = event.target as HTMLElement;
 
-      target.setPointerCapture(adaptedEvent.pointerId);
+      if (target instanceof HTMLDivElement) {
+        target.setPointerCapture(adaptedEvent.pointerId);
+      }
+
       this.markAsInBounds(adaptedEvent.pointerId);
       this.trackedPointers.add(adaptedEvent.pointerId);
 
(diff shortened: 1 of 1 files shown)


m-bert:
It could. The other thing that I considered was to avoid calling `setPointerCapture` on `inputs` (for example in case of mentioned issue it would be `select`). Let me know which one do you prefer.

Also I think we can merge this one and keep in mind that there's another approach if something happens.

PR #2676, opened 2023-11-17
Add `mouseButton` prop

m-bert:
## Description

After giving it some thought, we decided to add new modifier to our handlers - `mouseButton`. This way users will be able to choose mouse buttons that handler will respond to.

For now this prop is available on all handlers, but it is a topic for a short discussion.

I think it is also worth to mention, that right now all three mouse buttons work on web, so this prop will allow us to 'disable' all buttons except those provided as arguments.

This PR also adds two small examples (one for buttons and one for context menu) that can be used to test given prop on other platforms (when this functionality will be added)

## Test plan

Tested on example app.


src/web/tools/PointerEventManager.ts
```diff
@@ -190,7 +206,7 @@ export default class PointerEventManager extends EventManager<HTMLElement> {
       pointerId: event.pointerId,
       eventType: eventType,
       pointerType: event.pointerType as PointerType,
-      buttons: event.buttons,
+      button: this.mouseButtonsMapper.get(event.button),
```
j-piasecki, line 209: Do we care if there's more than one button pressed at the same time?

If we do, I'd rename it to `buttons`. Not sure about the format though - I'm not sure if passing a union back translates to a good ux. Maybe an array in that case?

src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -31,6 +31,8 @@ export class GestureHandlerWebDelegate
     this.gestureHandler = handler;
     this.view = findNodeHandle(viewRef) as unknown as HTMLElement;
 
+    this.view.oncontextmenu = () => false;
```
j-piasecki, line 34: Do we want to disable the context menu for every handler? I think we could do it only when right mouse button is expected (and for long press on touch), no?

src/web/interfaces.ts
```diff
@@ -121,22 +122,19 @@ export interface AdaptedEvent {
...
-  SCROLL,
-  SCROLL_LEFT,
-  SCROLL_RIGHT,
-  SCROLL_LEFT_RIGHT,
+export enum MouseButton {
+  LEFT = 1,
+  MIDDLE = 2,
+  RIGHT = 4,
+  BTN4 = 8,
+  BTN5 = 16,
```
j-piasecki, line 137:
```suggestion
  BUTTON_4 = 8,
  BUTTON_5 = 16,
```

I would go for more expressive names, we have autocompletion anyway.

src/web/handlers/GestureHandler.ts
```diff
@@ -736,6 +737,14 @@ export default abstract class GestureHandler {
     return false;
   }
 
+  protected wasPressedButtonCorrect(mouseButton: MouseButton | undefined) {
+    return (
+      !mouseButton ||
```
j-piasecki, line 742: Left mouse button is `0` which is falsy

src/web/handlers/GestureHandler.ts
```diff
@@ -736,6 +737,14 @@ export default abstract class GestureHandler {
     return false;
   }
 
+  protected wasPressedButtonCorrect(mouseButton: MouseButton | undefined) {
```
j-piasecki, line 740: Not sure about the name:
1. I'd change it to the present tense to keep it consistent
2. `correct` sounds strange to me in this context, maybe something along `matches config`

Idk, I can be convinced I'm wrong here.

example/src/App.tsx
```diff
@@ -16,6 +16,7 @@ import { ComboWithGHScroll } from './release_tests/combo';
 import { TouchablesIndex, TouchableExample } from './release_tests/touchables';
 import Rows from './release_tests/rows';
 import Fling from './release_tests/fling';
+import Buttons from './release_tests/mouseButtons';
```
j-piasecki, line 19:
```suggestion
import MouseButtons from './release_tests/mouseButtons';
```

I'd keep the full name, especially given that we support other buttons

example/src/release_tests/mouseButtons/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+  const middleTap = Gesture.Tap()
+    .mouseButton(MouseButton.MIDDLE)
+    .onEnd(() => console.log('Tap with middle'));
+
+  const leftRightTap = Gesture.Tap()
+    .mouseButton(MouseButton.LEFT | MouseButton.RIGHT)
+    .onEnd(() => console.log('Tap with left | right'));
+
+  const allTap = Gesture.Tap()
```
j-piasecki, line 61: These are not all buttons 😈.

Anyway, this made me think - is adding something like `MouseButton.ALL` useful?

src/web/tools/TouchEventManager.ts
```diff
@@ -157,7 +156,6 @@ export default class TouchEventManager extends EventManager<HTMLElement> {
       pointerId: event.changedTouches[index].identifier,
       eventType: eventType,
       pointerType: PointerType.TOUCH,
-      buttons: MouseButtons.NONE,
```
j-piasecki, line 160: I'd keep this in ming regarding more than one button. If we were to pass them here, having the type as possibly `undefined` adds one more step when using the API.

If we don't care about multiple buttons at once, then it doesn't matter I guess.

j-piasecki (changes requested):

src/web/tools/PointerEventManager.ts
m-bert: I don't think we should care about that since it doesn't seem to be well supported (or should I say, it doesn't seem to be supported at all).

If you click on something and then add another click with another button (for example, first you left click and then add right click to it), you won't get second event. Moreover, even if you click with two buttons at once, event seems to contain information only about "first" pointer.

So even if we consider using `event.buttons`, which contains combination of buttons, it would work only inside `pointermove`.

cc @tomekzaw

src/web/tools/GestureHandlerWebDelegate.ts
m-bert: From what I understood we wanted to disable it for every handler, but I don't mind doing it only when we expect right click.

src/web/handlers/GestureHandler.ts
m-bert: At this point left button was already mapped to value `1` in `PointerEventManager`

```tsx
this.mouseButtonsMapper.set(0, MouseButton.LEFT);
```

src/web/tools/TouchEventManager.ts
m-bert: I'm not sure what do you mean by that. We are talking about `TouchEventManager` - it won't get any event from mouse. Why would we want to pass them here?

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -51,6 +51,8 @@ export default class FlingGestureHandler extends GestureHandler {
   }
 
   private tryEndFling(): boolean {
+    console.log(this.tracker.getLastAvgX(), this.startX);
```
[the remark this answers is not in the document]
m-bert: Done in [64ff5d4](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/64ff5d4c385a0a2987bd8ce931fbf340803c32ab)

src/web/interfaces.ts
m-bert: I was used to these abbreviations, but renaming it to full names also sounds good. Done in [193e240](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/193e24009e37266cd1a9ec800c3525a91627c889)

example/src/App.tsx
m-bert: Done in [0cbaebd](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/0cbaebd85a445c9469aa45a6f9994c2d30d06578)

example/src/release_tests/mouseButtons/index.tsx
m-bert: Given that we've assumed the left button as the default one, it sounds like it will come in handy.

src/web/handlers/GestureHandler.ts
m-bert: What about `isButtonInConfig`?

src/web/tools/GestureHandlerWebDelegate.ts
j-piasecki: We can talk about it to make sure it will work consistently across platforms then

src/web/handlers/GestureHandler.ts
j-piasecki: Sounds good

src/web/tools/TouchEventManager.ts
j-piasecki: I just meant that (assuming we support more than one button at once), passing `0` or `[]` here removes the step of checking if the value is `undefined`.

src/web/tools/PointerEventManager.ts
j-piasecki: > If you click on something and then add another click with another button (for example, first you left click and then add right click to it), you won't get second event.

It works if you subscribe to `mousedown` event, but I guess that's another thing we may want to talk about 😅

src/web/handlers/GestureHandler.ts
m-bert: Done in [09e11fd](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/09e11fdf9a0e9ea06446a026ed4a5bf68058b745)

example/src/release_tests/contextMenu/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <View style={[styles.box, styles.grandParent]}>
+          <GestureDetector gesture={p2} enableContextMenu={true}>
+            <View style={[styles.box, styles.parent]}>
+              <GestureDetector gesture={p3} enableContextMenu={false}>
+                <View style={[styles.box, styles.child]} />
+              </GestureDetector>
+            </View>
+          </GestureDetector>
+        </View>
+      </GestureDetector>
```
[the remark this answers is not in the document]
m-bert: The outer one will have context menu disabled since `p1` has `mouseButton(MouseButton.RIGHT)`

example/src/release_tests/mouseButtons/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+        </GestureDetector>
+      </View>
+      <View style={{ flexDirection: 'row' }}>
+        <GestureDetector gesture={gestureHandlers[3]}>
+          <View style={styles.boxLeftRight} />
+        </GestureDetector>
+        <GestureDetector gesture={gestureHandlers[4]}>
+          <View style={styles.boxAll} />
+        </GestureDetector>
+      </View>
```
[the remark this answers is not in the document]
m-bert: Sure, changed to mapping in [60b8487](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/60b84877d793bdc44406903c2a6f13def8d1b8f3)

example/src/release_tests/mouseButtons/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+}
+
+const styles = StyleSheet.create({
+  center: {
+    alignItems: 'center',
+    justifyContent: 'space-around',
+  },
+  boxLeft: {
+    width: 100,
+    height: 100,
```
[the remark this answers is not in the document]
m-bert: Also done in [60b8487](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/60b84877d793bdc44406903c2a6f13def8d1b8f3)

src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -9,13 +9,21 @@ import TouchEventManager from './TouchEventManager';
...
 {
   private view!: HTMLElement;
   private gestureHandler!: GestureHandler;
   private eventManagers: EventManager<unknown>[] = [];
+  private boundDisableContextMenu: (e: MouseEvent) => void;
+  private boundEnableContextMenu: (e: MouseEvent) => void;
+
+  constructor() {
+    this.boundDisableContextMenu = this.disableContextMenu.bind(this);
+    this.boundEnableContextMenu = this.enableContextMenu.bind(this);
```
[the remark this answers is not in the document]
m-bert: Without those bindings you'll get this error:
``` 
error  Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead  @typescript-eslint/unbound-method
```

But looking at it I assume it is better to use `this:void` instead of binding `this`.

Changed in [c75fdb7](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/c75fdb766e6993f48b6a6511cb250e977928b6a2)

src/web/handlers/GestureHandler.ts
```diff
@@ -736,8 +738,22 @@ export default abstract class GestureHandler {
...
+    return (
+      !mouseButton ||
+      (!this.config.mouseButton && mouseButton === MouseButton.LEFT) ||
+      (this.config.mouseButton && mouseButton & this.config.mouseButton)
+    );
+  }
+
   protected resetConfig(): void {}
 
+  public finalize(): void {
```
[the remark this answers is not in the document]
m-bert: Yes, it can be pretty misleading. 

Changed in [1da7bfb](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/1da7bfbb71e81a90ab41800260a5ba91eb5b550e)

src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -86,6 +96,41 @@ export class GestureHandlerWebDelegate
...
+
+  addContextMenuListeners(config: Config): void {
+    if (this.shouldDisableContextMenu(config)) {
+      this.view.addEventListener('contextmenu', this.boundDisableContextMenu);
+    } else if (config.enableContextMenu) {
+      this.view.addEventListener('contextmenu', this.boundEnableContextMenu);
+    }
+  }
+
+  public removeContextMenuListeners(config: Config): void {
```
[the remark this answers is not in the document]
m-bert: Good idea, maybe we will add something more to this method later.

Added in [8279c90](https://github.com/software-mansion/react-native-gesture-handler/pull/2676/commits/8279c908fdd36fc9031b6ca8953f74c165e8614e)

src/web/handlers/GestureHandler.ts
```diff
@@ -749,9 +749,7 @@ export default abstract class GestureHandler {
   protected resetConfig(): void {}
 
   public onDestroy(): void {
-    (this.delegate as GestureHandlerWebDelegate).removeContextMenuListeners(
-      this.config
-    );
+    (this.delegate as GestureHandlerWebDelegate).destroy(this.config);
```
j-piasecki, line 752: Instead of casting you can add this method to the `GestureHandlerDelegate` interface

src/web/handlers/LongPressGestureHandler.ts
```diff
@@ -23,10 +22,6 @@ export default class LongPressGestureHandler extends GestureHandler {
 
   public init(ref: number, propsRef: React.RefObject<unknown>) {
     super.init(ref, propsRef);
-
-    if (Platform.OS === 'web') {
-      (this.delegate.getView() as HTMLElement).oncontextmenu = () => false;
-    }
```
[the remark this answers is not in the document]
j-piasecki: > Changing it in longPressGesture would also require another change for old API

What do you mean? It should be simply ignored when used with the old API if there's no implementation for it, no?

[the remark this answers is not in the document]
m-bert: But you can use `enableContextMenu` on old API. For example you can open _Draggable_ example and see that if you add `enableContextMenu={false}` to `PanGestureHandler`, context menu will be disabled.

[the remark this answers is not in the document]
j-piasecki: Ah, got you. I guess it makes sense to make it platform-specific then.

j-piasecki (approved):

PR #2679, opened 2023-11-20
Remove `handlersToCancel` on web.

m-bert:
## Description

While working on removing [gesture handlers limit on android](https://github.com/software-mansion/react-native-gesture-handler/pull/2672) I've though that we can also remove `handlersToCancel` array on web.

## Test plan

Tested on example app.


j-piasecki (approved):
:shipit:

PR #2680, opened 2023-11-23
Add `mouseButton` implementation on Android

m-bert:
## Description

This PR adds implementation of `mouseButton` prop on android.

Since `actionButton` field is available only in API >= 23, we provide full support only for those versions. The only thing that changes for API < 23 is ignoring `ACTION_BUTTON_*`, because we don't want handlers to react to both type of events.

Note that it requires [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2676) to work.

## Test plan

Tested on `MouseButtons` example.


android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -67,6 +67,10 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
   }
 
   override fun onHandle(event: MotionEvent, sourceEvent: MotionEvent) {
+    if (!shouldActivateWithMouse(sourceEvent)) {
+      return
+    }
```
j-piasecki, line 72: Wouldn't it make sense to move this check to `GestureHandler.handle(...)`? This way the logic would be in one place for all handlers.

src/web/interfaces.ts
```diff
@@ -132,8 +132,8 @@ export interface AdaptedEvent {
 
 export enum MouseButton {
   LEFT = 1,
-  MIDDLE = 2,
-  RIGHT = 4,
+  RIGHT = 2,
+  MIDDLE = 4,
```
j-piasecki, line 136: I assume this doesn't impact web/macos, but to doublecheck: does it?

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -688,6 +695,46 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
...
+        // When we receive ACTION_MOVE, we have to check buttonState field.
+        if (action == MotionEvent.ACTION_MOVE && !isButtonInConfig(buttonState)) {
+          return@shouldActivateWithMouse false
+        }
+      } else if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+        // We do not fully support mouse below API 23, so we will ignore BUTTON events.
+        if (action == MotionEvent.ACTION_BUTTON_PRESS || action == MotionEvent.ACTION_BUTTON_RELEASE) {
+          return@shouldActivateWithMouse false
+        }
+      }
```
j-piasecki, line 732: I think the check for API version can be moved to the top of the method, and we can always return `true` if below `M` - the behavior will stay as it is currently on those platforms, and above API 23 the new logic will run. Additionally, I believe that the second API check (`>= M`) would be unnecessary.

Correct me if I'm wrong though.

m-bert: The main problem is that mouse clicks generate 2 events. This function acts like a filter - it assures that handler will receive only one of them. If we return `true` all the time, handlers will receive both events (like `ACTION_DOWN` and `ACTION_BUTTON_PRESS`).  While I haven't tested if anything breaks in that case, I don't think it's a good idea to leave it that way.

src/handlers/gestureHandlerCommon.ts
```diff
@@ -111,6 +111,8 @@ export type ActiveCursor =
   | 'zoom-in'
   | 'zoom-out';
 
+export type EnableContextMenu = boolean;
+
```
j-piasecki, line 115: This seems unrelated, no?

j-piasecki (approved):
Looks good but please remove the context menu bit 😄

src/handlers/gestureHandlerCommon.ts
m-bert: Done in[9936d77] (https://github.com/software-mansion/react-native-gesture-handler/pull/2680/commits/9936d772b4ff4a5e1183c57f8377863d9a3874d9)

m-bert: Done in [9936d77](https://github.com/software-mansion/react-native-gesture-handler/pull/2680/commits/9936d772b4ff4a5e1183c57f8377863d9a3874d9)

Issue #2658, opened 2023-10-26
RNGestureHandlerModule.mm fatally throws NSGenericException in iOS production due to collection mutating while being enumerated


sh-pq:
### Description

We use GestureDetector in our application inside the contents of an inverted FlatList. Within that context, we're observing GestureDetector throw a fatal exception (stack trace below) that crashes the application. The fatal error originates from [RNGestureHandlerModule.mm line 259](https://github.com/software-mansion/react-native-gesture-handler/blob/f2b115e1aaa8cf05687db4db0f6bb40a333692fc/ios/RNGestureHandlerModule.mm#L259).

- This was reproduced on 2.13.0 and then 2.13.2, the only two versions we've used so far.
- The errors aren't 100% deterministic but we can reproduce it frequently during our tests.
- We only observe the error being thrown on production releases in iOS. (In case it matters, they're testflight releases; we haven't released any version of the application with this dependency to the store yet.) Debug iOS, and debug + production Android, are all fine.
- Errors occur from simply loading the component, through scrolling the list, or through pressing or long pressing within the gesture detector.

#### MVCE notes

I have attached a MVCE but I have some notes to make about it.

- We use [React Native Stack Navigation](https://reactnavigation.org/docs/stack-navigator), and the component containing the FlatList is a screen inside the navigation, but I'm not sure how crucial it is to the MVCE so I've left it out.
- There's one bit missing from the MVCE: the FlatList there doesn't scroll, and I don't remember what I should be doing to get it to scroll — it's been weeks since I last wrote the scrolling FlatList in this area and I'm not sure what bit I'm missing that'd make it scroll.

#### Stack trace

Here `Lambda` is the name of our application.

```
Fatal Exception: NSGenericException

*** Collection <__NSArrayM: 0x283e03b70> was mutated while being enumerated.

Fatal Exception: NSGenericException
0  CoreFoundation                 0x9cb4 __exceptionPreprocess
1  libobjc.A.dylib                0x183d0 objc_exception_throw
2  CoreFoundation                 0x176920 -[__NSSingleObjectEnumerator init]
3  Lambda                         0x6cf580 -[RNGestureHandlerModule didMountComponentsWithRootTag:] + 259 (RNGestureHandlerModule.mm:259)
4  Lambda                         0x6704a0 -[RCTSurfacePresenter mountingManager:didMountComponentsWithRootTag:] + 453 (RCTSurfacePresenter.mm:453)
[...]

PR #2687, opened 2023-12-04
Remove `didMountComponentsWithRootTag` on the new arch

j-piasecki:
## Description

Removes usage of `didMountComponentsWithRootTag` on the new architecture. It was doing exactly the same thing as `flushOperations` but was executed on a different thread, which could cause crashes. Removing this method shouldn't change any behavior as `flushOperations` is scheduled after every operation.

Closes https://github.com/software-mansion/react-native-gesture-handler/issues/2658
 
## Test plan

See https://github.com/software-mansion/react-native-gesture-handler/issues/2658


WoLewicki (approved):
LGTM!

PR #2690, opened 2023-12-07
fix: Fix types for `TouchableWithoutFeedback`

mrousavy:
## Description

Fixes the types for the TouchableWithoutFeedback.

Previously, the TouchableWithoutFeedback exposed itself as it's `TouchableWithoutFeedbackProps`, now it correctly exposes it's props.


j-piasecki (approved):
Thanks!

Issue #2594, opened 2023-09-19
Buttons(iOS | Android): BorderRadius issue with styled components


hirbod:
### Description

We're attempting to port all instances of `Pressable` from react-native to RNGH Buttons (e.g., RectButton). We utilize NativeWind (akin to Tailwind) and convert them into styled components as follows:

```tsx
import { RectButton } from "react-native-gesture-handler";
const StyledRectButton = styled(RectButton);
```

Setting a `borderRadius` is straightforward:

```tsx
<StyledRectButton tw="rounded-full" />
```

This approach works seamlessly for every component. However, it encounters issues with the Buttons exported by RNGH. I've identified the root of the problem: the Buttons incorporate an `Animated.View` as a child within the Button and derive the `borderRadius` from the flattened Stylesheet.

While passing `borderRadius` via the `style` prop to the `StyledComponent` does work, it complicates the process. You can see the relevant code here: 

https://github.com/software-mansion/react-native-gesture-handler/blob/2cdcdf68d5864178ec00022462f8e7b90d450614/src/components/GestureButtons.tsx#L270-L295

Ideally, we'd like to pass the child component (turning it into an AnimatedStyledView) and be able to add `tw/className` props to it, without the need to wrap it in an outer view. At present, migration is unfeasible for us because wrapping the RNGH Button with a view disrupts a significant amount of our styling.

If I recall correctly, we could modify the children's styles in the past. However, this wouldn't be particularly helpful now, as we need the ability to pass our styled props from Tailwind.

### Steps to reproduce

**See the Snack**
https://snack.expo.dev/@hirbod/rngh-button-nativewind?platform=ios

Works fine on web, but does not work on iOS. On iOS, it does apply the borderRadius correctly for the view with the underlayColor (see the red color) but it fails for the main view.

On Android, it fails completely and doesn't even add a color or anything at all. 

### Snack or a link to a repository

https://snack.expo.dev/@hirbod/rngh-button-nativewind?&platform=ios

### Gesture Handler version

2.9.0

### React Native version

0.72.3

### Platforms

Android, iOS

### JavaScript runtime

None

### Workflow

Expo bare workflow

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

None

### Acknowledgements

Yes

PR #2691, opened 2023-12-12
fix: Add support for borderRadii on RectButton

kacperkapusciak:
# Description

https://github.com/software-mansion/react-native-gesture-handler/assets/39658211/4ebc5761-c571-4c9a-a9c0-c2e0b7ac0018

This PR adds support for border radius style props like: `borderTopLeftRadius`, `borderTopRightRadius`, `borderBottomLeftRadius`, `borderBottomRightRadius` to RectButton.

It implements the border radii by wrapping a RawButton with two Views - the outer one curves the native RawButton and the inner acts like a mask cutting the corners of the button. The radius of the inner view had to be adjusted by subtracting the borderWidth from it - [see this tweet](https://twitter.com/lilykonings/status/1567317037126680576).

FIxes https://github.com/software-mansion/react-native-gesture-handler/issues/2594


m-bert (approved):
Logic looks fine, thanks for fixing that! ❤️  

I'll leave review of TS part for @tjzel

tjzel (approved):
TS seems approvable now 🚀

src/components/splitStyleProp.ts
```diff
@@ -0,0 +1,169 @@
...
+    const innerBorderRadiiStyles = shrinkBorderRadiiByBorderWidth(
+      { ...styleGroups.borderRadiiStyles },
+      borderWidth
+    );
+
+    innerStyles = { ...innerStyles, ...innerBorderRadiiStyles };
+  }
+
+  return { outerStyles, innerStyles, restStyles };
+}
```
j-piasecki, line 169: 1. Umm, do we care about `borderTopWidth`, `borderRightWidth`, ...?
2. It's better that now, so should we?

j-piasecki (changes requested):

src/components/splitStyleProp.ts
j-piasecki: I'll assume that this comment being ignored by @kacperkapusciak means something along the lines of `and who is that? mr. grouchy, destroyer of good fun, slayer of children's smiles` and we'll figure it out in the due time.

j-piasecki (approved):
🔥

versten1uk:
Still reproduced in `2.24.0`

PR #2705, opened 2023-12-20
Reset relevant handler when recognizer gets reset.

j-piasecki:
## Description

At the moment `reset` method is only called in `gestureRecognizerShouldBegin`: https://github.com/software-mansion/react-native-gesture-handler/blob/35ec17d636220c1ad4226456aa3b03c846ba16a8/apple/RNGestureHandler.m#L487
However, I've noticed that this method is called after some events are already sent, mainly the `BEGIN` event, which is often sent as soon as `onTouchesBegan` is executed for the first pointer.

According to the [apple documentation](https://developer.apple.com/documentation/uikit/uigesturerecognizer/1620004-reset?language=objc), the `reset` method of recognizer is the place to reset the internal state of recognizer, which in this case in my opinion should also include the state of the handler.

The issue this fixes is the one described in https://github.com/software-mansion/react-native-gesture-handler/pull/2628:
> One thing that could be related is the fact that when the gesture fails before activation, the gesture stays in the Began state until the finger is lifted. This still needs to be investigated.

This was coming from the fact that the handler was failing before it was allowed to 'begin' (iOS begin not RNGH begin), so `gestureRecognizerShouldBegin` was not called, thus not resetting the failing handler. Because of that, it was sending events containing wrong states.

## Test plan

Tested on the Example app.

### DIFF
--- a/apple/Handlers/RNPanHandler.m
+++ b/apple/Handlers/RNPanHandler.m
@@ -123,6 +123,7 @@
 
   if (self.state == UIGestureRecognizerStatePossible && [self shouldFailUnderCustomCriteria]) {
     self.state = UIGestureRecognizerStateFailed;
+    [self triggerAction];
     return;
   }
 
--- a/apple/Handlers/RNPanHandler.m
+++ b/apple/Handlers/RNPanHandler.m
@@ -218,6 +218,7 @@
   [NSObject cancelPreviousPerformRequestsWithTarget:self selector:@selector(activateAfterLongPress) object:nil];
   self.enabled = YES;
   [super reset];
+  [_gestureHandler reset];
 }
 
 - (void)updateHasCustomActivationCriteria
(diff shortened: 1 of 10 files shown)


apple/Handlers/RNPanHandler.m
```diff
@@ -123,6 +123,7 @@ - (void)interactionsMoved:(NSSet *)touches withEvent:(UIEvent *)event
 
   if (self.state == UIGestureRecognizerStatePossible && [self shouldFailUnderCustomCriteria]) {
     self.state = UIGestureRecognizerStateFailed;
+    [self triggerAction];
```
m-bert, line 126: I assume that it should be here, but to clarify - should it?

m-bert (commented):
LGTM

apple/Handlers/RNPanHandler.m
j-piasecki: Yeah, it's loosely related to this - the fail event here wouldn't be sent until `reset` gets called by the OS, which happens after the finger is lifted from the screen - the gesture could fail, but the information about that would be delayed.

m-bert (approved):

PR #2706, opened 2023-12-20
Add rotation and pinch velocity on macOS

### LINKED ISSUE #2695: Add velocity information on macOS
### Description

RIght now, we don't provide any information about velocity of rotation and pinch gestures on macOS, since there is no information about it in [NSRotationGestureRecognizer](https://developer.apple.com/documentation/appkit/nsrotationgesturerecognizer?language=objc). It still seems that we can get this information like on Android.

### Steps to reproduce

-

-

### Gesture Handler version

-

### React Native version

-

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

jfedak:
## Description

Fixes #2695

This PR adds pinch and rotation velocity information on macOS. In order to make it work, we needed to change a way of handling events inside `RNRotationHandler` and `RNPinchHandler` - from now on, it uses `rotateWithEvent` and `magnifyWithEvent` respectively. The action is based on the state of handler. 

Both handlers calculate current velocity in a similar way to [Android handler](https://github.com/software-mansion/react-native-gesture-handler/blob/83d8e7b02b525a8a5d18dc0eb680565470daa92d/android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureHandler.kt#L27-L30).

## Test plan

PR was tested on `Pinch & rotate` example.


apple/Handlers/RNPinchHandler.m
```diff
@@ -67,28 +76,23 @@ - (void)interactionsCancelled:(NSSet *)touches withEvent:(UIEvent *)event
...
+    [self interactionsEnded:[NSSet setWithObject:event] withEvent:event];
+  } else if (self.state == NSGestureRecognizerStateCancelled) {
+    [self interactionsCancelled:[NSSet setWithObject:event] withEvent:event];
+  }
 
-- (void)touchesCancelledWithEvent:(NSEvent *)event
-{
-  [super touchesCancelledWithEvent:event];
-  [self interactionsCancelled:[NSSet setWithObject:event] withEvent:event];
+  _velocity = (self.magnification - prevMagnification) / (event.timestamp - prevTime);
```
j-piasecki, line 93: To double-check, what's the unit of `event.timestamp`? On iOS the unit of timer is device-dependant (https://github.com/software-mansion/react-native-gesture-handler/blob/db51032fcd0f81449736564e2f3fd055f29a28a4/apple/Handlers/RNLongPressHandler.m#L117).

j-piasecki: Then the time delta should be multiplied by `1000` (both here and in the rotation handler), on Android the time unit we're using is milliseconds.

apple/Handlers/RNRotationHandler.m
```diff
@@ -60,28 +70,28 @@ - (void)interactionsCancelled:(NSSet *)touches withEvent:(UIEvent *)event
...
+    case NSGestureRecognizerStateCancelled:
+      [self interactionsCancelled:[NSSet setWithObject:event] withEvent:event];
+      break;
+  }
 
-- (void)touchesCancelledWithEvent:(NSEvent *)event
-{
-  [super touchesCancelledWithEvent:event];
-  [self interactionsCancelled:[NSSet setWithObject:event] withEvent:event];
+  _velocity = (self.rotation - prevRotation) / (event.timestamp - prevTime);
```
m-bert, line 92: This delta should also be multiplied by `1000`, right?

m-bert (changes requested):

m-bert (approved):
🟢

PR #2709, opened 2023-12-22
Fix `onUp` in `Fling`

m-bert:
## Description

Fling gesture ends when pointer was moved farther than minimum distance.  The problem is that `onUp` method calls `endFling` after pointer was removed from tracker. This means that calling `this.tracker.getLastX(this.keyPointer)` (the same for `Y` position) inside `tryEndFling` will return undefined.

It is rather cosmetic change - it doesn't really affect behavior since we call `tryEndFling` inside `onPointerMove`. 

This PR also changes minimum distance the pointer has to travel, since `160` is way to big.

## Test plan

<details>
<summary> Tested on the following code </summary>

```jsx
import React from 'react';

  const f = Gesture.Fling()
    .onEnd((e) => console.log(e))
    .onFinalize((e, s) => console.log(e, s));

  return (
      <GestureDetector gesture={f}>
        <View style={styles.box} />
      </GestureDetector>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    backgroundColor: '#ecf0f1',
    padding: 8,
  },
  box: {
    width: 500,
    height: 500,
    backgroundColor: 'blue',
  },
});

```

</details>

### DIFF
--- a/src/web/handlers/FlingGestureHandler.ts
+++ b/src/web/handlers/FlingGestureHandler.ts
@@ -5,7 +5,7 @@ import { AdaptedEvent, Config } from '../interfaces';
 import GestureHandler from './GestureHandler';
 
 const DEFAULT_MAX_DURATION_MS = 800;
-const DEFAULT_MIN_ACCEPTABLE_DELTA = 160;
+const DEFAULT_MIN_ACCEPTABLE_DELTA = 32;
 const DEFAULT_DIRECTION = Direction.RIGHT;
 const DEFAULT_NUMBER_OF_TOUCHES_REQUIRED = 1;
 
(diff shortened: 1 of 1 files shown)


src/web/handlers/FlingGestureHandler.ts
```diff
@@ -5,7 +5,7 @@ import { AdaptedEvent, Config } from '../interfaces';
 import GestureHandler from './GestureHandler';
 
 const DEFAULT_MAX_DURATION_MS = 800;
-const DEFAULT_MIN_ACCEPTABLE_DELTA = 160;
+const DEFAULT_MIN_ACCEPTABLE_DELTA = 32;
```
m-bert, line 8: Even though it should be fine (and also this functionality can be implemented using `Pan`), I think it would be better if users could change this value. What do you think @j-piasecki?

j-piasecki: Yeah, having it set to 160 is a bit much. We could make it also use velocity to determine whether to activate (if the distance was shorter than the minimum delta, but the gesture was fast). Anyway, that's something for another PR, if we want to do this at all.

PR #2713, opened 2024-01-04
Fix macOS warning message 

jfedak:
## Description

This PR just fix a message in native view handler warning. 

## Test plan

n/a

### DIFF
--- a/apple/Handlers/RNNativeViewHandler.mm
+++ b/apple/Handlers/RNNativeViewHandler.mm
@@ -216,7 +216,7 @@
 
 - (instancetype)initWithTag:(NSNumber *)tag
 {
-  RCTLogWarn(@"Manual gesture handler is not supported on macOS");
+  RCTLogWarn(@"Native view gesture handler is not supported on macOS");
   if ((self = [super initWithTag:tag])) {
     _recognizer = [NSGestureRecognizer alloc];
   }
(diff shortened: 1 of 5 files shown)


apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -216,7 +216,7 @@ @implementation RNNativeViewGestureHandler
 
 - (instancetype)initWithTag:(NSNumber *)tag
 {
-  RCTLogWarn(@"Manual gesture handler is not supported on macOS");
+  RCTLogWarn(@"Native view gesture handler is not supported on macOS");
```
m-bert, line 219: I think this looks better. Let me know what you think.

```suggestion
  RCTLogWarn(@"NativeViewGestureHandler is not supported on macOS");
```

m-bert (approved):
Good catch. Also, I think that using convention like in suggestion below looks a bit better, but since all other warnings use spaces, I'll leave it up to you to decide.

apple/Handlers/RNNativeViewHandler.mm
jfedak: Yup, imo it looks better for long names. When I wrote code for warnings for the first time I followed the message from hover handler on tvOS, but now I see that we do different for force touch (one single word).

j-piasecki: Yeah, it looks nicer. I'm fine with both, merging this one and changing them all in a follow-up, or just changing them in this PR.

j-piasecki (approved):

m-bert (approved):

PR #2715, opened 2024-01-04
fix(worklet): fix workletization of touch event gesture callbacks

computerjazz:
## Description

Workletized touch event gesture callbacks were causing Reanimated errors to be thrown in RN 73 (Expo 50 beta).
Not sure if this is a bug in RNGH or Reanimated, but moving the function definition outside of the object seems to fix the issue (sounds like a bug in the reanimated babel plugin if nested functions should be able to be workiletized?)

Fixes https://github.com/software-mansion/react-native-reanimated/issues/5555

This case was previously broken:

```tsx
function Demo() {

  panGesture.onTouchesMove((evt, mgr) => {
    'worklet'
    console.log('move!!')
  })

  return (
    <GestureHandlerRootView style={{ flex: 1, backgroundColor: 'seashell' }}>
        <Animated.View style={{ flex: 1 }} />
      </GestureDetector>
    </GestureHandlerRootView>
  )
}
```

### DIFF
--- a/src/handlers/gestures/gestureStateManager.ts
+++ b/src/handlers/gestures/gestureStateManager.ts
@@ -18,8 +18,7 @@ const warningMessage = tagMessage(
 const REANIMATED_AVAILABLE = Reanimated?.useSharedValue !== undefined;
 const setGestureState = Reanimated?.setGestureState;
 
-export const GestureStateManager = {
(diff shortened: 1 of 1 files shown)


j-piasecki (approved):
Thanks!

PR #2728, opened 2024-01-21
handle case when setImmediate is not available in the browser

flyingart:
## Description

I found some cases, when setImmediate is not available (f.ex in Coherent browser). This "workaround" seems to solve the issue. I didn't find more places where similar tweak is needed.

## Test plan

Make sure that everything works like before.


m-bert:
Hi @flyingart! Thanks for this PR!

> Coherent browser

Do you mean Coherent GT? I have to admit that I'm not familiar with this one.

> I didn't find more places where similar tweak is needed

This change should be sufficient, as all other calls for `queueMicrotask`/`setImmediate` use `ghQueueMicrotask` wrapper.

j-piasecki:
This file was updated to use also `requestAnimationFrame`, I'll close the PR for now. If this doesn't solve your case, let me know. https://github.com/software-mansion/react-native-gesture-handler/blob/6c16cf2b9cda35b7891f4d7841c0a9884ff664a6/src/ghQueueMicrotask.ts#L3-L8

PR #2741, opened 2024-02-02
Change functions in windows module file to no-ops.

m-bert:
## Description

A while ago we added `RNGestureHandlerModule.windows.ts` file, to allow usage of our library on Windows. The problem is that it was an exact copy of `.web.ts` file, therefore `HTMLElement` is not defined. This PR changes all functions inside this module file into no-ops, so that applications will build correctly. 

## Test plan

Confirmed that app builds correctly on Windows


j-piasecki (approved):
Don't forget to update the title and description 😅

PR #2757, opened 2024-02-12
Pass `{ flex: 1 }` as default style to `GestureHandlerRootView`

kacperkapusciak:
## Description
Adding `{ flex: 1 }` style is [described in the installation docs](https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/installation#js) but I _always_ forget to add it when creating a new app that uses gestures. I think others developers can relate.

This PR adds a `{ flex: 1 }`  fallback style to `GestureHandlerRootView` so users won't see a blank screen when they wrap their app with `GestureHandlerRootView`. 
 
## Test plan

Run an example app

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -161,7 +161,7 @@ const Stack = createStackNavigator<RootStackParamList>();
 
 export default function App() {
   return (
-    <GestureHandlerRootView style={styles.root}>
+    <GestureHandlerRootView>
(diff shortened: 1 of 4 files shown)


j-piasecki (approved):

m-bert (approved):
❤️

m-bert (approved):

PR #2760, opened 2024-02-16
Add `pointerType` to event

m-bert:

### DIFF
--- a/apple/Handlers/RNHoverHandler.m
+++ b/apple/Handlers/RNHoverHandler.m
@@ -155,7 +155,8 @@ API_AVAILABLE(ios(13.4))
 - (RNGestureHandlerEventExtraData *)eventExtraData:(UIGestureRecognizer *)recognizer
 {
   return [RNGestureHandlerEventExtraData forPosition:[recognizer locationInView:recognizer.view]
-                                withAbsolutePosition:[recognizer locationInView:recognizer.view.window]];
+                                withAbsolutePosition:[recognizer locationInView:recognizer.view.window]
+                                     withPointerType:UITouchTypePencil];
 }
 
 @end
--- a/apple/Handlers/RNNativeViewHandler.mm
+++ b/apple/Handlers/RNNativeViewHandler.mm
@@ -39,6 +39,7 @@
 
 - (void)touchesBegan:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
 {
+  [_gestureHandler setCurrentPointerType:event];
   [_gestureHandler.pointerTracker touchesBegan:touches withEvent:event];
 }
 
@@ -148,21 +149,21 @@
 
   [self sendEventsInState:RNGestureHandlerStateActive
            forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES]];
+            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES withPointerType:_pointerType]];
 }
 
 - (void)handleTouchUpOutside:(UIView *)sender forEvent:(UIEvent *)event
 {
   [self sendEventsInState:RNGestureHandlerStateEnd
            forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:NO]];
+            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:NO withPointerType:_pointerType]];
 }
 
 - (void)handleTouchUpInside:(UIView *)sender forEvent:(UIEvent *)event
 {
   [self sendEventsInState:RNGestureHandlerStateEnd
            forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES]];
+            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES withPointerType:_pointerType]];
 }
 
 - (void)handleDragExit:(UIView *)sender forEvent:(UIEvent *)event
@@ -173,11 +174,11 @@
     [control cancelTrackingWithEvent:event];
     [self sendEventsInState:RNGestureHandlerStateEnd
              forViewWithTag:sender.reactTag
-              withExtraData:[RNGestureHandlerEventExtraData forPointerInside:NO]];
+              withExtraData:[RNGestureHandlerEventExtraData forPointerInside:NO withPointerType:_pointerType]];
   } else {
     [self sendEventsInState:RNGestureHandlerStateActive
              forViewWithTag:sender.reactTag
-              withExtraData:[RNGestureHandlerEventExtraData forPointerInside:NO]];
+              withExtraData:[RNGestureHandlerEventExtraData forPointerInside:NO withPointerType:_pointerType]];
   }
 }
 
@@ -185,14 +186,14 @@
 {
   [self sendEventsInState:RNGestureHandlerStateActive
            forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES]];
+            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES withPointerType:_pointerType]];
 }
 
 - (void)handleTouchCancel:(UIView *)sender forEvent:(UIEvent *)event
 {
   [self sendEventsInState:RNGestureHandlerStateCancelled
            forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:NO]];
+            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:NO withPointerType:_pointerType]];
 }
 
 @end
--- a/apple/Handlers/RNPanHandler.m
+++ b/apple/Handlers/RNPanHandler.m
@@ -187,6 +187,7 @@
 
 - (void)touchesBegan:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
 {
+  [_gestureHandler setCurrentPointerType:event];
   // super call was moved to interactionsBegan method to keep the
   // original order of calls
   [self interactionsBegan:touches withEvent:event];
@@ -409,7 +410,8 @@
                            withAbsolutePosition:[recognizer locationInView:recognizer.view.window]
                                 withTranslation:[recognizer translationInView:recognizer.view.window]
                                    withVelocity:[recognizer velocityInView:recognizer.view.window]
-                            withNumberOfTouches:recognizer.numberOfTouches];
+                            withNumberOfTouches:recognizer.numberOfTouches
+                                withPointerType:_pointerType];
 }
 #endif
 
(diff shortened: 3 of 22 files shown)


apple/Handlers/RNHoverHandler.m
```diff
@@ -155,7 +155,8 @@ - (void)configure:(NSDictionary *)config
 - (RNGestureHandlerEventExtraData *)eventExtraData:(UIGestureRecognizer *)recognizer
 {
   return [RNGestureHandlerEventExtraData forPosition:[recognizer locationInView:recognizer.view]
-                                withAbsolutePosition:[recognizer locationInView:recognizer.view.window]];
+                                withAbsolutePosition:[recognizer locationInView:recognizer.view.window]
+                                     withPointerType:UITouchTypePencil];
```
j-piasecki, line 159: Hover also works with mouse

apple/RNGestureHandler.h
```diff
@@ -87,4 +89,10 @@
 - (void)sendEvent:(nonnull RNGestureHandlerStateChange *)event;
 - (void)sendTouchEventInState:(RNGestureHandlerState)state forViewWithTag:(nonnull NSNumber *)reactTag;
 
+#if !TARGET_OS_OSX
+- (void)setCurrentPointerType:(nonnull UIEvent *)event;
+#else
+- (void)setCurrentPointerType;
```
j-piasecki, line 95: Can we use a different name for macos that somehow suggests that it will be mouse?

src/web/handlers/GestureHandler.ts
```diff
@@ -38,7 +40,7 @@ export default abstract class GestureHandler {
   protected awaiting = false;
   protected active = false;
   protected shouldResetProgress = false;
-  protected pointerType: PointerType = PointerType.NONE;
+  protected pointerType: WebPointerType = WebPointerType.NONE;
```
j-piasecki, line 43: Wouldn't it make more sense to just use `PointerType` and remove `WebPointerType` entirely at this point? You could just map it when building an intermediate event instead of having two separate enums doing exactly the same thing.

.github/workflows/ios-build.yml
```diff
@@ -29,6 +29,10 @@ jobs:
         with:
           node-version: 18
           cache: 'yarn'
+      - name: setup-cocoapods
+        uses: maxim-lobanov/setup-cocoapods@v1
+        with:
+          version: 1.15.2
```
j-piasecki, line 35: This looks like something that should be in a separate PR.

m-bert: Sure, [here it is](https://github.com/software-mansion/react-native-gesture-handler/pull/2765). I will remove these lines after we merge it.

src/web/handlers/GestureHandler.ts
m-bert: Done in [407bbbe](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/407bbbe216a9c60db9fa1460c5e3216dd9ebfd4b)

apple/Handlers/RNPanHandler.m
```diff
@@ -9,6 +9,8 @@
 #import "RNPanHandler.h"
 
 #if TARGET_OS_OSX
+#import "RNGestureHandlerPointerType.h"
```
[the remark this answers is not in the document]
m-bert: Done in [846cfbf](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/846cfbff68ee1a63c2d6f37ce7f067b1ce9c87bd)

apple/RNGestureHandler.h
m-bert: Done in [adbdf3c](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/adbdf3c613917eccbecff1b5bb3257b8bb0d0d6d)

apple/Handlers/RNHoverHandler.m
m-bert: Currently we don't have option to distinguish between `mouse` and `apple pen` on iOS. I suggest leaving is as it is and describing this behavior in docs.

apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -148,21 +149,21 @@ - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
 
   [self sendEventsInState:RNGestureHandlerStateActive
            forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES]];
+            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES withPointerType:_pointerType]];
```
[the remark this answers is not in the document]
m-bert: Set in [711becb](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/711becb19782f076bc4e2c147e0780e0609c38f5)

NiuGuohui:
Can you support other point data? For example, pressure.

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -378,6 +380,11 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
     lastAbsolutePositionY = GestureUtils.getLastPointerY(adaptedTransformedEvent, true)
     lastEventOffsetX = adaptedTransformedEvent.rawX - adaptedTransformedEvent.x
     lastEventOffsetY = adaptedTransformedEvent.rawY - adaptedTransformedEvent.y
+
+    if (sourceEvent.action == MotionEvent.ACTION_DOWN || sourceEvent.action == MotionEvent.ACTION_HOVER_ENTER || sourceEvent.action == MotionEvent.ACTION_HOVER_MOVE) {
```
jfedak, line 384: (just for clarification) I assume there is no need to set type of pointer on `MotionEvent.ACTION_HOVER_EXIT` right?

apple/RNGestureHandler.m
```diff
@@ -176,6 +176,38 @@ - (void)setEnabled:(BOOL)enabled
...
 }
 
+#if !TARGET_OS_OSX
+- (void)setCurrentPointerType:(UIEvent *)event
+{
+  UITouch *touch = [[event allTouches] anyObject];
+
+  switch (touch.type) {
+    case UITouchTypeDirect:
+      _pointerType = RNGestureHandlerTouch;
```
jfedak, line 186: There is a small difference between naming on Android and iOS. On Android we call this type of pointer `POINTER_TYPE_FINGER` and here it's called `RNGestureHandlerTouch`. Do we want to rename to something like `POINTER_TYPE_TOUCH` or `RNGestureHandlerFinder`?

src/web/tools/PointerEventManager.ts
```diff
@@ -1,13 +1,13 @@
...
-  PointerType,
-} from '../interfaces';
+import { AdaptedEvent, EventTypes, MouseButton } from '../interfaces';
 import EventManager from './EventManager';
-import { isPointerInBounds } from '../utils';
+import { PointerTypeMapping, isPointerInBounds } from '../utils';
+import { PointerType } from '../../PointerType';
 
 const POINTER_CAPTURE_EXCLUDE_LIST = new Set<string>(['SELECT', 'INPUT']);
+const PointerTypes = {
```
jfedak, line 7: Basically same thing here, sometimes we use names like `touch/pen` while in other cases we use names like `finger/stylus`.

jfedak (commented):
Some minor comments, as everything else seems to be allright.

jfedak:
Also one more thing (according to our last week talk). It seems that it's possible to have other pointer types than just mouse on macOS, but we didn't manage to get any of them in our case. Maybe it would be possible if we implement methods like `tabletPoint` or `tabletProximity` in our handlers, but it's more related to all macOS stuff and shouldn't be considered here.

apple/RNGestureHandler.m
m-bert: I think it would make sense to stick to `touch`.

src/web/tools/PointerEventManager.ts
m-bert: Here `pen` is used because this is what we get inside `PointerEvents`. I could change it to `PointerTypes.Stylus` to match our convention. What do you think?

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
m-bert: Yes, I don't think we need to set that on exit. The main idea was to set this variable when gesture receives first event (inside `onPointerDown` on web, `touchesBegan` on iOS, etc.). 

Now I see that in `NativeViewGestureHandler` on iOS i call `setCurrentPointerType` also on `TouchUp` methods. I will check if it is necessary.

apple/RNGestureHandler.m
m-bert: Android done in [025f549](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/025f5491ece30c8cae814d737d8589c466eb61e8).
Web done in [f74ce39](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/f74ce39c515e028d889eee9f27a1de348364624a).

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
m-bert: As I said, I've tested that and it seems that we don't need most of those calls. I removed them in [ab677e2](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/ab677e27049ef754cb2c14a1f4fbdbe63dd18990). I've left one in `handleDragEnter` since I'm not sure when this method is called. If you know how to test it let me know, I'll be grateful, cc @j-piasecki.

src/web/tools/PointerEventManager.ts
m-bert: As we discussed earlier, I've changed that to `Stylus` in order to stay consistent with our inner convention. Done in [6a92b89](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/6a92b89b170d91af9e09f5ae4f7d5abbb5c434d9).

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
j-piasecki: `handleDragEnter` should be called when pointer moves into the view boundary without touching down. I'm not sure whether it requires to start drag inside the view (touch the view -> drag outside -> drag back inside) or not (touch outside the view -> drag into the view)

j-piasecki (approved):
LGTM, if all of @jfedak comments are addressed

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
m-bert: Okay, I've just checked that and it does require to start drag inside of view. Given that, I removed the line that sets `pointerType` inside that function in [77ab5ce](https://github.com/software-mansion/react-native-gesture-handler/pull/2760/commits/77ab5cecef9433dc6b08c4ca76a17cc9e3f76f7a).

PR #2779, opened 2024-02-21
Fix wrong admonition color on initial render

kacperkapusciak:
### Before

https://github.com/software-mansion/react-native-gesture-handler/assets/39658211/04f2494a-ba8f-40dc-a850-f53e6df03d9e

### After

https://github.com/software-mansion/react-native-gesture-handler/assets/39658211/a618b26a-9af7-4a5f-9805-bb4c7b21e894


patrycjakalinska (approved):
Approved 🚀

PR #2785, opened 2024-02-28
Adjust Android build.gradle to to run project with Gradle 8

norman-kapschefsky:
## Description

To build the `android` project with gradle 8+, the added configs are required.

## Test plan

Tested in Android project with ...
- gradle 8.2
- gradle 7.6

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -155,6 +155,10 @@ android {
         targetCompatibility JavaVersion.VERSION_11
     }
 
+    kotlinOptions {
+        jvmTarget = '11'
+    }
+
     if (isNewArchitectureEnabled()) {
         externalNativeBuild {
             cmake {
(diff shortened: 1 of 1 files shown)


j-piasecki:
We already set `namespace` and `buildConfig` depending on AGP version: https://github.com/software-mansion/react-native-gesture-handler/blob/d55d8de21bce537aaef2c8f30a4a50a0ce4cc152/android/build.gradle#L102-L110

If that's not enough in your case, can you share more details about your setup?

norman-kapschefsky:
> We already set `namespace` and `buildConfig` depending on AGP version:
> 
> https://github.com/software-mansion/react-native-gesture-handler/blob/d55d8de21bce537aaef2c8f30a4a50a0ce4cc152/android/build.gradle#L102-L110
> 
> If that's not enough in your case, can you share more details about your setup?

Thx for your respond @j-piasecki 
Ah my bad. Due to the fact, that we use an older version of "react-native-gesture-handler", I thought its missing. Sorry.

Good that you have it already 👍. Will close the PR.

PR #2766, opened 2024-02-20
feat: add necessary changes for new arch on RN 0.74

WoLewicki:
## Description

PR with necessary changes for new arch bridgeless support on RN 0.74

PR #2786, opened 2024-03-01
Move `customDirectEventTypes` to separate files.

m-bert:
## Description

`customDirectEventTypes` import inside `createHandler` introduced in #2766 makes web unable to compile since `react-native-web` doesn't provide this object. This PR moves mentioned import into separate file and splits it into `.ts` and `.web.ts` parts 

## Test plan

Tested on web and FabricExample app.


j-piasecki (approved):

WoLewicki (commented):
LGTM

PR #2787, opened 2024-03-04
Add `touchAction` prop.

m-bert:
## Description

This PR adds new prop to gesture handlers - `touchAction`. 

While working on [fix scroll interactions PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2631) we've found it difficult to remove `touch-action: none;` and ensure correct behavior of handlers with scroll, especially with swipeable components. What made it even harder were differences between platforms. 

We decided to kill two birds with one stone - give our users possibility to manipulate `touchAction`, which will also make it easier for us to properly handle scroll.

## Test plan

Tested on example app.

### DIFF
--- a/src/handlers/gestures/GestureDetector.tsx
+++ b/src/handlers/gestures/GestureDetector.tsx
@@ -19,6 +19,7 @@ import {
   HandlerStateChangeEvent,
   scheduleFlushOperations,
   UserSelect,
+  TouchAction,
 } from '../gestureHandlerCommon';
 import {
   GestureStateManager,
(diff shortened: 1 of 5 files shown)


src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -48,6 +44,10 @@ export class GestureHandlerWebDelegate
       this.view.style['userSelect'] = config.userSelect;
     }
 
+    this.view.style['touchAction'] = config.touchAction ?? 'none';
+    //@ts-ignore This one disables default events on Safari
+    this.view.style['WebkitTouchCallout'] = 'none';
```
m-bert, line 49: Quote from [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout):

>When a target is touched and held on iOS, Safari displays a callout information about the link. This property allows disabling that behavior.

So I think we can keep it as `none`.

src/handlers/gestures/GestureDetector.tsx
```diff
@@ -644,6 +655,10 @@ export const GestureDetector = (props: GestureDetectorProps) => {
     applyEnableContextMenuProp(props.enableContextMenu, gestureConfig);
   }
 
+  if (props.touchAction !== undefined) {
+    applyTouchActionProp(props.touchAction, gestureConfig);
+  }
```
j-piasecki, line 660: We may want to figure out a better approach to do this, that's outside of the scope of this PR though 😄

j-piasecki (approved):

Issue #477, opened 2019-02-26
Border does not appear in styling RectButton


yodaheis:
@pcooney10 I'm facing the issue of border not appearing on Android. It does work well on iOS though.
style props like **borderRadius works**, **but borderWidth, borderColor does not work on Android**

_Originally posted by @yodaheis in https://github.com/kmagiera/react-native-gesture-handler/issues/225#issuecomment-467333032_

I would like to bring to your attention border styling related issues on Android.
Border doesn't appear on Android while it works well on iOS.

*Environment*:
expo: 32.0.0

vlkosinov:
Also there is no ability to apply `borderTopStartRadius`, `borderTopEndRadius`, `borderBottomStartRadius`, `borderBottomEndRadius` properties to RectButton

anarkafkas:
I found that, for IOS, borderWidth works as expected but borderBottomWidth and borderTopWidth do not work. For Android, none of the three works.

ribbit-rota:
Any update on this one?

erickjth:
Same here

guitexa:
not working on Android

jakub-gonet:
`<RectButton />` is based on native `RNGestureHandlerButton` component which does implement `borderRadius` in a way that prevents using individual corners ([here's more info about that](https://github.com/software-mansion/react-native-gesture-handler/blob/1d23e205975e010d393bc164962ebcd1fd037031/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.java#L126-L132)).

It also doesn't support setting `borderWidth`, nor `borderWidth`.

iOS implementation subclasses standard `RCTViewManager` class and supports every property that `<View />` supports.


Hard to say if it will be fixed in the near future, but I'll keep it in my issues-that-should-be-fixed-sometime list 🙂

RBioZ:
same here

leandrosouzaa:
> same here

try add the rect button inside a view.

pauloels:
same here

Gitarcitano:
Any update? Having the same problem here!

euthribeiro:
This is because RectButton preserves the native shape of each platform. To solve this you can wrap your RectButton inside a View and style the View

a-eid:
I'm wondering if there is any update on this. @jakub-gonet

DenisDov:
i do that way:

```tsx
<RectButton
  hitSlop={16}
  style={{
[...]

PR #2792, opened 2024-03-04
Allows to apply different border radius to RectButton component

camilossantos2809:
## Description
This PR is intended to allow the use of different border radius values on the RectButton component. As mentioned in comments on #477 , it doesn’t work on Android.
I tried to follow the same implementation pattern from `borderRadius`, but using `setCornerRadii` method when informed at least one prop from specific corner.
Also, it is possible to merge the values from `borderRadius` or from specific corner.

## Test plan

I added a component (RectButtonBorders) on example app with some variations of values to test the changes.


j-piasecki (approved):
Thanks for taking the time for it ❤️!

I modified it a bit to also apply the custom corner radii to the ripple drawable if it's there.

PR #2796, opened 2024-03-08
Make FlingHandler use velocity as the activation metric.

LatekVo:
## Description
Fixed fling component handler to activate based on fling velocity instead of fling distance.
The fling also no longer has to be performed from start to finish entirely within the component's bounds.
This feature was tested on both emulated and live android phones.
## Test plan
- clone this branch
- run `yarn`
- go to `example`
- copy fling example from the docs into the `example/src/empty/EmptyExample.tsx` file

### Web
- run `yarn start`

### Android
- go to `example/android`
- run `yarn start`

### IOS
- IOS remains unchanged, it already implements this mechanism in a different way.

## Reference
https://github.com/software-mansion/react-native-gesture-handler/assets/74246391/6868ff68-4e6e-41fb-93a7-0cd35e85cd6e


m-bert (changes requested):
Good job! 😄

There are some things that we should change. Since both implementations are identical, I've left comments with explanations only in `.ts` files. 

> The fling also no longer has to be performed from start to finish entirely within the component's bounds.

I'm not sure if that's what we actually want 😅 Handlers have property called `shouldCancelWhenOutside`, so behavior of moving out of the component's bounds should match this property, cc @j-piasecki.

Also please change name of this PR and in `Web` section you can change `yarn start` to `yarn web`

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -83,13 +75,19 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
...
+    )
+
     velocityTracker.recycle()
 
-    // hypot may be overkill for this simple function, dot product may be sufficient
     val totalVelocity = hypot(velocityVector.x, velocityVector.y)
 
-    val isAligned = compareSimilarity(directionVector, velocityVector) > 0.75
-    val isFast = totalVelocity > this.minAcceptableDelta
+    val isAligned = alignmentList.reduce { any, element -> any or element }
```
m-bert, line 89:
```suggestion
    val isAligned = alignmentList.any { it }
```

Same as in `.ts`, just different function.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -4,17 +4,15 @@ import android.os.Handler
...
 
 class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
   var numberOfPointersRequired = DEFAULT_NUMBER_OF_TOUCHES_REQUIRED
   var direction = DEFAULT_DIRECTION
 
   private val maxDurationMs = DEFAULT_MAX_DURATION_MS
-  private val minAcceptableDelta = DEFAULT_MIN_ACCEPTABLE_DELTA
-  private var startX = 0f
-  private var startY = 0f
+  private val minVelocity = DEFAULT_MIN_ACCEPTABLE_DELTA
```
m-bert: I think we can also change `DEFAULT_MIN_ACCEPTABLE_DELTA` into `DEFAULT_MIN_VELOCITY` in companion object.

m-bert (changes requested):

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -120,7 +162,7 @@ export default class FlingGestureHandler extends GestureHandler {
     }
   }
 
-  protected onPointerMove(event: AdaptedEvent): void {
+  protected pointerMove(event: AdaptedEvent): void {
```
m-bert, line 165:
```suggestion
  private pointerMove(event: AdaptedEvent): void {
```

I think this can be `private` instead of `protected`

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -35,26 +34,74 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
...
+      compareAlignment(velocityUnitVector, SimpleVector(1.0, 0.0), DIRECTION_RIGHT),
+      compareAlignment(velocityUnitVector, SimpleVector(0.0, -1.0), DIRECTION_UP),
+      compareAlignment(velocityUnitVector, SimpleVector(0.0, 1.0), DIRECTION_DOWN)
+    )
+
+    velocityTracker.recycle()
+
+    val totalVelocity = hypot(velocityVector.x, velocityVector.y)
+
+    val isAligned = alignmentList.any { it }
```
j-piasecki, line 90: I feel like this should be its own class (minus the velocity stuff at the end). And then `Simple` can be dropped from the name, since it will no longer be that simple 😄.

Instead of creating directional vectors for every check, they can be static fields of the class (which are achieved using [`companion object` in Kotlin](https://kotlinlang.org/docs/object-declarations.html#companion-objects)).

And I'd move it outside the `tryEndFling` so it's inner class of Fling or move it to a separate file altogether.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -35,26 +34,74 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
...
-  ) {
-    handler!!.removeCallbacksAndMessages(null)
-    activate()
-    true
-  } else {
-    false
-  }
+    }
+
+    fun compareSimilarity(
```
j-piasecki, line 51:
```suggestion
    fun calculateSimilarity(
```
Given that `compareAlignment` returns `Boolean` and this one returns `Double` I'd say that names should reflect that.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -35,26 +34,74 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
...
+        toSafeNumber(vec.y / magnitude)
       )
-  ) {
-    handler!!.removeCallbacksAndMessages(null)
-    activate()
-    true
-  } else {
-    false
-  }
+    }
```
j-piasecki, line 49: Not a strong opinion, but I'd prefer an early return here instead of calling helper (not sure about the threshold here, or maybe even simple comparison to `0.0` would work?), i.e.:
```kt
      val magnitude = hypot(vec.x, vec.y)
      
      if (magnitude < 0.001) {
        return SimpleVector(0.0, 0.0)
      }

      return SimpleVector(vec.x / magnitude, vec.y / magnitude)
```

Possibly even make a zero vector a static field, so it would return `SimpleVector.ZERO`.

LMK what you think

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -35,26 +34,74 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
...
+    )
+
+    val velocityUnitVector = toUnitVector(velocityVector)
 
+    val alignmentList = arrayOf(
+      compareAlignment(velocityUnitVector, SimpleVector(-1.0, 0.0), DIRECTION_LEFT),
+      compareAlignment(velocityUnitVector, SimpleVector(1.0, 0.0), DIRECTION_RIGHT),
+      compareAlignment(velocityUnitVector, SimpleVector(0.0, -1.0), DIRECTION_UP),
+      compareAlignment(velocityUnitVector, SimpleVector(0.0, 1.0), DIRECTION_DOWN)
+    )
```
j-piasecki, line 84: Maybe I'm dumb, but I don't see checking if the alignment is consistent with what the user passed in `direction` prop. It seems like we're always checking all four directions and activating on any of them.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -3,15 +3,16 @@ package com.swmansion.gesturehandler.core
...
 class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
   var numberOfPointersRequired = DEFAULT_NUMBER_OF_TOUCHES_REQUIRED
   var direction = DEFAULT_DIRECTION
 
   private val maxDurationMs = DEFAULT_MAX_DURATION_MS
-  private val minAcceptableDelta = DEFAULT_MIN_ACCEPTABLE_DELTA
-  private var startX = 0f
-  private var startY = 0f
+  private val minVelocity = DEFAULT_MIN_VELOCITY
+  private val minDirectionAlignment = DEFAULT_MIN_DIRECTION_ALIGNMENT
```
j-piasecki, line 15:
```suggestion
  private val minDirectionalAlignment = DEFAULT_MIN_DIRECTIONAL_ALIGNMENT
```

Dunno, the adjective looks nicer to me. I can be talked out of it though.

j-piasecki (commented):
Soon...
image

Most of what I wrote also applies to the web version, but I didn't want to copy&paste it.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
LatekVo: Hello, thank you for the whole review :)
Consistence with user input is checked inside `compareAlignment()`.
We use this array because the user may request more than 1 directions to be 'activatory', as `direction` is a bitmask. 
As you suggested, I will separate this code into a separate class, which will resolve most of the issues present here.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
LatekVo: Applied.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
LatekVo: Applied.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
LatekVo: I've separated most of the logic into a separate Vector class, please let me know if this is what You had in mind, and if i should place it in a separate file, or is it fine to keep it where it is :)

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -51,21 +91,32 @@ export default class FlingGestureHandler extends GestureHandler {
...
+      Direction.LEFT,
+      Direction.RIGHT,
+      Direction.UP,
+      Direction.DOWN,
+    ].map(
+      (direction) =>
+        velocityVector.computeSimilarity(
+          new Vector().fromDirection(direction)
+        ) > this.minDirectionalAlignment && direction & this.direction
+    );
```
m-bert, line 110:
```suggestion
    const alignmentList = Object.values(Direction).map((direction) => {
      const similarity = velocityVector.computeSimilarity(
        new Vector().fromDirection(direction)
      );

      return (
        similarity > this.minDirectionalAlignment && direction & this.direction
      );
    });
```

I would create separate array to hold directions or use `Object.values()`. Also the condition after `=>` is hard to read.

Also I've noticed that we use `Direction` from `constants.ts`, but we have a file called `Directions.ts`, we could change that if you want to :)

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -132,6 +183,14 @@ export default class FlingGestureHandler extends GestureHandler {
     super.onPointerMove(event);
   }
 
+  protected onPointerMove(event: AdaptedEvent): void {
+    this.pointerMove(event);
+  }
+
+  protected onPointerOutOfBounds(event: AdaptedEvent): void {
+    this.pointerMove(event);
```
m-bert, line 191: In `pointerMove` you call `super.onPointerMove()`, but I believe in that case it should be `super.onPointerOutOfBounds()`. I'd remove calling `super` methods in `pointerMove` and call them inside overriden functions.

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -1,30 +1,73 @@
...
+      this.uX = this.uY = 0;
+    }
+
+    this.uX = this.x / magnitude;
+    this.uY = this.y / magnitude;
+
+    return this;
+  }
+
+  computeSimilarity(vector: any) {
```
m-bert, line 46:
```suggestion
  computeSimilarity(vector: Vector) {
```

I know there are places in our codebase where we still use `any`, but here we can simply use `Vector`.

m-bert (commented):
This time I won't leave comments on android side because implementations are identical.

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -1,30 +1,73 @@
...
+  }
+
+  fromVelocity(tracker: PointerTracker, pointerId: number) {
+    this.x = tracker.getVelocityX(pointerId);
+    this.y = tracker.getVelocityY(pointerId);
+
+    const magnitude = Math.hypot(this.x, this.y);
+    if (magnitude < 0.001) {
+      this.uX = this.uY = 0;
+    }
```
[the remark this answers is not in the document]
j-piasecki: I think something along the lines of `epsilon` would work.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -35,26 +81,45 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
...
-  }
+  private fun tryEndFling(event: MotionEvent): Boolean {
+
+    fun compareAlignment(
+      vector: Vector,
+      direction: Int,
+      directionVec: Vector = Vector().fromDirection(direction),
+    ): Boolean =
+      vector.computeSimilarity(directionVec) > minDirectionalAlignment &&
+        (this.direction and direction != 0)
```
j-piasecki, line 92:
```suggestion
    fun compareAlignment(
      vector: Vector,
      direction: Int,
    ): Boolean =
      (this.direction and direction != 0) &&
        vector.computeSimilarity(Vector().fromDirection(direction)) > minDirectionalAlignment
```

The third argument seems unnecessary here since it's never used.

Bitwise operations will be faster to check than operations on vector, so changing order will save us some processor cycles that we will never notice anyway.

Also, can this also be moved to the `Vector` class? The first argument becomes `this` this way.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -3,15 +3,63 @@ package com.swmansion.gesturehandler.core
...
+
+    val magnitude = hypot(this.x, this.y)
+    if (magnitude < 0.001) {
+      this.uX = 0.0
+      this.uY = 0.0
+    }
+
+    this.uX = this.x / magnitude
+    this.uY = this.y / magnitude
+  }
```
j-piasecki, line 45: Those seem like they should be static and creating a new vector:
```
val vec = Vector.fromDirection(...)
```
instead of 
```
val vec = Vector().fromDirection(...)
```

That would also simplify it a bit, since you can move the normalization code to the constructor, so it's not repeated in both methods.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -3,15 +3,63 @@ package com.swmansion.gesturehandler.core
...
+
+  fun computeSimilarity(vector: Vector): Double {
+    return this.uX * vector.uX + this.uY * vector.uY
+  }
+
+  fun computeMagnitude(): Double {
+    return hypot(this.x, this.y)
+  }
```
j-piasecki, line 53: We can make this into a read-only property
```suggestion
  val magnitude: Double
    get() = hypot(this.x, this.y)
```

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -3,15 +3,63 @@ package com.swmansion.gesturehandler.core
...
 import android.view.MotionEvent
+import android.view.VelocityTracker
+import com.swmansion.gesturehandler.core.GestureHandler.Companion
+import kotlin.math.hypot
+
+class Vector {
+  var x: Double = 0.0
+  var y: Double = 0.0
+  private var uX: Double = 0.0
+  private var uY: Double = 0.0
```
j-piasecki, line 14: We can use more letters 😄 
```suggestion
  private var unitX: Double = 0.0
  private var unitY: Double = 0.0
```

Or we can go even further:
```suggestion
  private var unit: Vector
```
Though, it may be overcomplicating at this point since we only need those for the dot product.

m-bert: I agree that it looks better with "more letters" (at first I wasn't sure what exactly is `uX`), but creating unit vector inside is a bit too much for me 😅

src/web/tools/Vector.ts
```diff
@@ -0,0 +1,58 @@
...
+export default class Vector {
+  x: number = 0;
+  y: number = 0;
+  unitX: number = 0;
+  unitY: number = 0;
+
+  fromDirection(direction: Directions) {
+    [this.x, this.y] = [this.unitX, this.unitY] = DirectionTypeMappings.get(
+      direction
+    ) ?? [0, 0];
```
m-bert, line 13:
```suggestion
[this.x, this.y] = [this.unitX, this.unitY] =
      DirectionTypeMappings.get(direction)!;
```

Assuming I haven't missed something, this value won't be undefined so we can use non-null assertion (correct me if I'm wrong).

src/web/tools/Vector.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    return this;
+  }
+
+  fromVelocity(tracker: PointerTracker, pointerId: number) {
+    this.x = tracker.getVelocityX(pointerId);
+    this.y = tracker.getVelocityY(pointerId);
+
+    const magnitude = Math.hypot(this.x, this.y);
+    if (magnitude < 1) {
```
m-bert, line 23: Even though `1` looks less confusing than `0.001`, I'd still like to give it a name (cc @j-piasecki).

We could also do something like

```js
    this.unitX = magnitude < 1 ? 0 : this.x / magnitude
    this.unitY = magnitude < 1 ? 0 : this.y / magnitude
```

But I don't have strong preferences here so it's up to you to decide which one you like more 😅

src/web/tools/Vector.ts
```diff
@@ -0,0 +1,58 @@
...
+  isSimilar(vector: Vector, threshold: number) {
+    return this.computeSimilarity(vector) > threshold;
+  }
+
+  get magnitude() {
+    return Math.hypot(this.x, this.y);
+  }
+}
+
+const DirectionTypeMappings = new Map<Directions, Array<number>>([
```
m-bert, line 46:
```suggestion
const DirectionMappings = new Map<Directions, number[]>([
```

We tend to use `T[]` to denote array instead of `Array<T>`. Also I don't think we need `Type` infix in this name.

src/web/tools/Vector.ts
```diff
@@ -0,0 +1,58 @@
...
+  [Directions.LEFT, [-1, 0]],
+  [Directions.RIGHT, [1, 0]],
+  [Directions.UP, [0, -1]],
+  [Directions.DOWN, [0, 1]],
+]);
+
+export const DirectionTypeVectorMappings = new Map<Directions, Vector>(
```
m-bert, line 53:
```suggestion
export const DirectionToVectorMappings = new Map<Directions, Vector>(
```

Same here, I don't think we need `Type` in that name and `To` sounds more natural, as it reflects `Direction -> Vector` relationship.

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
LatekVo: Please let me know if the current implementation is what you've had in mind.

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -120,18 +127,26 @@ export default class FlingGestureHandler extends GestureHandler {
     }
   }
 
-  protected onPointerMove(event: AdaptedEvent): void {
+  private pointerMove(event: AdaptedEvent): void {
```
m-bert, line 130:
```suggestion
  private pointerMoveAction(event: AdaptedEvent): void {
```
I still find this name a bit confusing 😅

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
LatekVo: Current implementation kept `directionVec` but it is now just a reference to a static instance of `Vector`, 
let me know if that is fine, or whether it should still be removed.

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -51,21 +48,31 @@ export default class FlingGestureHandler extends GestureHandler {
...
+      this.keyPointer
+    );
+
+    const getAlignment = (direction: Directions) => {
+      const directionVector = DirectionTypeVectorMappings.get(direction);
+      return (
+        directionVector &&
+        direction & this.direction &&
+        velocityVector.isSimilar(directionVector, this.minDirectionalAlignment)
+      );
```
m-bert, line 62:
```suggestion
      const directionVector = DirectionTypeVectorMappings.get(direction)!;
      return (
        direction & this.direction &&
        velocityVector.isSimilar(directionVector, this.minDirectionalAlignment)
      );
```

Similar to the situation in `Vector.ts`, as far as I can see `directionVector` will never be undefined, so we can use non-null assertion (correct me if I'm wrong).

m-bert (commented):
Looks good! Left some comments, this time only cosmetic things 😄

src/web/tools/Vector.ts
LatekVo: You are correct

src/web/tools/Vector.ts
```diff
@@ -0,0 +1,54 @@
...
+    [this.x, this.y] = [this.unitX, this.unitY] = DirectionMappings.get(direction)!;
+
+    return this;
+  }
+
+  fromVelocity(tracker: PointerTracker, pointerId: number) {
+    this.x = tracker.getVelocityX(pointerId);
+    this.y = tracker.getVelocityY(pointerId);
+
+    const minimalVelocity = 1;
```
m-bert, line 20: I have mixed feelings about this one. Name is fine and the condition below looks better, but on the other hand I think this type of constants should be declared at least outside of the function (like at the top of the file, or maybe even in `constants.ts` file) cc @j-piasecki.

m-bert (commented):
It looks much better now! Please change name of this PR and make sure that CI passes 😄

android/src/main/java/com/swmansion/gesturehandler/core/Vector.kt
```diff
@@ -0,0 +1,60 @@
+package com.swmansion.gesturehandler.core
+
+import android.view.VelocityTracker
+import com.swmansion.gesturehandler.core.GestureHandler.Companion.DIRECTION_DOWN
+import com.swmansion.gesturehandler.core.GestureHandler.Companion.DIRECTION_LEFT
+import com.swmansion.gesturehandler.core.GestureHandler.Companion.DIRECTION_RIGHT
+import com.swmansion.gesturehandler.core.GestureHandler.Companion.DIRECTION_UP
+import kotlin.math.hypot
+
+class Vector(var x: Double, var y: Double) {
```
j-piasecki, line 10:
```suggestion
class Vector(val x: Double, val y: Double) {
```

android/src/main/java/com/swmansion/gesturehandler/core/Vector.kt
```diff
@@ -0,0 +1,60 @@
...
+    private var unitX: Double = x
+    private var unitY: Double = y
```
j-piasecki, line 12: I *think* those can be `vals` as well since they are now initialized in the constructor, I'm not sure though.
```suggestion
    private val unitX: Double
    private val unitY: Double
```

android/src/main/java/com/swmansion/gesturehandler/core/Vector.kt
```diff
@@ -0,0 +1,60 @@
...
+    private fun computeSimilarity(vector: Vector): Double {
+        return unitX * vector.unitX + unitY * vector.unitY
+    }
+
+    fun isSimilar(vector: Vector, threshold: Double): Boolean {
+        return computeSimilarity(vector) > threshold
+    }
+
+    val magnitude
+        get() = hypot(x, y)
```
j-piasecki, line 31: This also *should* work since `x` and `y` are passed in the constructor
```suggestion
    val magnitude = hypot(x, y)
```

android/src/main/java/com/swmansion/gesturehandler/core/Vector.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.swmansion.gesturehandler.core.GestureHandler.Companion.DIRECTION_RIGHT
+import com.swmansion.gesturehandler.core.GestureHandler.Companion.DIRECTION_UP
+import kotlin.math.hypot
+
+class Vector(var x: Double, var y: Double) {
+    private var unitX: Double = x
+    private var unitY: Double = y
+
+    init {
+        val magnitude = hypot(x, y)
```
j-piasecki, line 15: ... and then this shouldn't be necessary
```suggestion
```

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -35,26 +33,44 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
...
-      )
-  ) {
-    handler!!.removeCallbacksAndMessages(null)
-    activate()
-    true
-  } else {
-    false
-  }
+  private fun tryEndFling(event: MotionEvent): Boolean {
+    val velocityTracker = VelocityTracker.obtain()
```
j-piasecki, line 37: The velocity tracker should be initialized when gesture begins, now it's calculating velocity only based on the last event...

android/src/main/java/com/swmansion/gesturehandler/core/FlingGestureHandler.kt
```diff
@@ -35,26 +33,44 @@ class FlingGestureHandler : GestureHandler<FlingGestureHandler>() {
...
+    addVelocityMovement(velocityTracker, event)
+
+    val velocityVector = Vector.fromVelocity(velocityTracker)
+
+    velocityTracker.recycle()
```
j-piasecki, line 42: ... and recycled when gesture is activated or failed/cancelled.

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -51,21 +48,30 @@ export default class FlingGestureHandler extends GestureHandler {
...
+      return (
+        direction & this.direction &&
+        velocityVector.isSimilar(directionVector, this.minDirectionalAlignment)
+      );
+    };
+
+    // list of alignments to all activated directions
+    const alignmentList = Object.values(Directions).map(getAlignment);
+
+    const isAligned = alignmentList.some((element) => element);
```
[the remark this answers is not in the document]
LatekVo: Yes, looks much better. I didn't know that Booleans were constructible objects in JS.

LatekVo:
Give me one minute, i forgot to rewrite the JS vector as well.

j-piasecki (approved):
Looks good, besides the little thing with `VelocityTracker`!

I have one more thing: right now the the vectors are considered aligned when the angle between them is less than ~41 degrees. It works great, but when setting the direction of `UP | RIGHT` (or any other 2 adjacent directions) it leaves a cone of ~18 degrees where the gesture doesn't activate. I think we can change the `DEFAULT_MIN_DIRECTION_ALIGNMENT` to `0.707` which would mean considering two vectors aligned when the angle between them is less than `45.008` degrees, so there would not be a no-gesture-recognized cone in-between. What do you think?

Ideally, we would want to reduce the cone for a particular direction to like 60-45 deg and expand them when 2 adjacent directions are set. We may also consider going in the direction of iOS, which does a weird thing of rotating the cone (I think? cc. @m-bert) when adjacent directions are set. That's something up for discussion though, and definitely for another PR.

m-bert:
I don't think that trying to mimic `iOS` behavior is a good idea. We will have some cross-platform inconsistency, but still we will achieve something what I believe can be called "expected behavior" on `web` and `android`.

> I think we can change the DEFAULT_MIN_DIRECTION_ALIGNMENT to 0.707 which would mean considering two vectors aligned when the angle between them is less than 45.008 degrees, so there would not be a no-gesture-recognized cone in-between.

I agree with that 👍 

>Ideally, we would want to reduce the cone for a particular direction to like 60-45 deg and expand them when 2 adjacent directions are set.

If we consider expanding the cone, `45 deg` sounds a bit too much for me. In that case I would set it to `~30 deg` for one direction and expand it when we set adjacent directions. We can discuss it next time though.

LatekVo:
Alright, I'm increasing the constant to 0.86 (30 deg), and adding angle expansion functionality.

src/web/tools/Vector.ts
```diff
@@ -3,28 +3,32 @@ import { MINIMAL_FLING_VELOCITY } from '../constants';
...
 export default class Vector {
-  x: number = 0;
-  y: number = 0;
-  unitX: number = 0;
-  unitY: number = 0;
+  x = 0;
+  y = 0;
+  unitX = 0;
+  unitY = 0;
+  magnitude = 0;
```
m-bert, line 10:
```suggestion
  private readonly x;
  private readonly y;
  private readonly unitX;
  private readonly unitY;
  public readonly magnitude;
```

Since all of those are assigned only in constructor You can make them `readonly` . Also all of them (except `magnitude`) could be private, which makes me think about adding a getter.

Earlier you used `get magnitude` syntax which was fine (however now we would have to add something like `_magnitude` since it is used in constructor):

```jsx
  get magnitude(){
    return this._magnitude;
  }
```
Or we could simply add java style getter:

```jsx
  getMagnitude(){
    return this.magnitude;
  }
```

m-bert (approved):
LGTM 😄 I've left one last comment, but overall it looks fine. Thanks for your hard work ❤️  

@j-piasecki, I'd like to know your opinion about magnitude getter before we merge it 😄

src/web/tools/Vector.ts
j-piasecki: What do we get from a getter here? It would be `readonly` anyway.

m-bert: I believe this is more matter of preference, for me it just looks better. That's why I asked about your opinion 😄

m-bert:
:shipit::shipit::shipit:

PR #2802, opened 2024-03-12
fix(web): After swiping closed, the Swipeable component cannot be swiped open again

yatessss:
## Description

After swiping closed, the component did not reset `this.currentState` to `State.UNDETERMINED` internally, which prevents the execution of the swipe open method.


m-bert:
Hi! Thanks for submitting this PR 😄 Before we merge it, could you please provide a reproduction? I have not noticed those problems with swipeables.

yatessss:
Hello, I have conducted extensive testing and found that the issue may lie with `react-native-web` Registering DOM nodes using `React` works fine, but registering DOM nodes through react-native-web causes problems (e.g., swiping left to open becomes ineffective once). By commenting out the two sections of code separately in the `main.tsx` file, the issue can be reproduced.

https://stackblitz.com/edit/vitejs-react-native-gesture-handler-k8wghi-sgju9p?file=src%2Fmain.tsx

yatessss:
> Hi @yatessss! Thanks for submitting this PR 😄 Before we merge it, could you please provide a reproduction? I have not noticed those problems with swipeables.

Hello, I have conducted extensive testing and found that the issue may lie with `react-native-web`. Registering DOM nodes using `React` works fine, but registering DOM nodes through `react-native-web` causes problems (e.g., swiping left to open becomes ineffective once). By commenting out the two sections of code separately in the `main.tsx` file, the issue can be reproduced.

https://stackblitz.com/edit/vitejs-react-native-gesture-handler-k8wghi-sgju9p?file=src%2Fmain.tsx

m-bert:
Maybe I've missed something, but your example seems to work:


https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/7839bafc-a8f9-415c-a4a6-390b1cedc2b1

yatessss:
> Maybe I've missed something, but your example seems to work:
> 
>  Nagranie.z.ekranu.2024-03-13.o.15.50.50.mov

If you swipe from right to left, you will encounter the issue.😂😂

m-bert:
Okay, now I see what you're talking about. I think it can be done in a different way since `resetProgress` method is used, as it's name suggests, to reset handler progress (for example set `scale` in `Pinch` to `1`).

i don't want to close your PR since you're the one who found this problem and it would be great if you could fix it 😄 What I suggest is changing 

```jsx
    if (!this.enabled && this.isFinished()) {
      this.currentState = State.UNDETERMINED;
    }
```

to

```jsx
    if (this.isFinished()) {
      this.currentState = State.UNDETERMINED;
    }
```
in `moveToState` method inside `GestureHandler.ts` file.

You can check if ti solves your problem and if it does we will merge it :smile:

yatessss:
ok, thanks for your suggestion. i will make the changes to this PR and commit changes after confirming that it runs smoothly.

yatessss:
Although the issue has been resolved, I would still like to understand the specific reasons behind it. Why does registering with react-native-web cause this problem?

m-bert (approved):
Thanks for fixing that ❤️

m-bert:
@yatessss Sadly, right now I can't answer this question 😞 If I find some time to investigate it, I will try to dive deeper into this problem.

PR #2807, opened 2024-03-15
Add corner area detection to Fling gesture.

LatekVo:
## Description

This PR implements a requested feature to activate fling on corners of two adjacent activated directions.

## Test plan

- In project root run `yarn`
- Go to `example/` and run `yarn`
- Paste Fling example from Gesture Handler docs into EmptyExample.tsx
- Run `yarn web` for Web or `yarn start` for IOS and Android

### DIFF
--- a/src/Directions.ts
+++ b/src/Directions.ts
@@ -13,7 +13,7 @@ export const CornerDirections = {
 } as const;
 
 /* eslint-disable @typescript-eslint/no-redeclare -- backward compatibility; it can be used as a type and as a value */
-export type Directions = typeof Directions[keyof typeof Directions];
+export type Directions = (typeof Directions)[keyof typeof Directions];
+// prettier-ignore
 export type CornerDirections = typeof CornerDirections[keyof typeof CornerDirections];
 /* eslint-enable @typescript-eslint/no-redeclare */
-
--- a/src/Directions.ts
+++ b/src/Directions.ts
@@ -5,5 +5,15 @@ export const Directions = {
   DOWN: 8,
 } as const;
 
-// eslint-disable-next-line @typescript-eslint/no-redeclare -- backward compatibility; it can be used as a type and as a value
+export const CornerDirections = {
+  UP_RIGHT: Directions.UP | Directions.RIGHT,
+  DOWN_RIGHT: Directions.DOWN | Directions.RIGHT,
+  UP_LEFT: Directions.UP | Directions.LEFT,
+  DOWN_LEFT: Directions.DOWN | Directions.LEFT,
+} as const;
+
+/* eslint-disable @typescript-eslint/no-redeclare -- backward compatibility; it can be used as a type and as a value */
 export type Directions = typeof Directions[keyof typeof Directions];
+export type CornerDirections = typeof CornerDirections[keyof typeof CornerDirections];
+/* eslint-enable @typescript-eslint/no-redeclare */
+
--- a/src/Directions.ts
+++ b/src/Directions.ts
@@ -1,19 +1,18 @@
+const RIGHT = 1;
+const LEFT = 2;
+const UP = 4;
+const DOWN = 8;
+
 export const Directions = {
-  RIGHT: 1,
-  LEFT: 2,
-  UP: 4,
-  DOWN: 8,
+  RIGHT: RIGHT,
+  LEFT: LEFT,
+  UP: UP,
+  DOWN: DOWN,
+  UP_RIGHT: UP | RIGHT,
+  DOWN_RIGHT: DOWN | RIGHT,
+  UP_LEFT: UP | LEFT,
+  DOWN_LEFT: DOWN | LEFT,
 } as const;
 
-export const CompositeDirections = {
-  UP_RIGHT: Directions.UP | Directions.RIGHT,
-  DOWN_RIGHT: Directions.DOWN | Directions.RIGHT,
-  UP_LEFT: Directions.UP | Directions.LEFT,
-  DOWN_LEFT: Directions.DOWN | Directions.LEFT,
-} as const;
-
-/* eslint-disable @typescript-eslint/no-redeclare -- backward compatibility; it can be used as a type and as a value */
+// eslint-disable-one-line @typescript-eslint/no-redeclare -- backward compatibility; it can be used as a type and as a value
 export type Directions = typeof Directions[keyof typeof Directions];
-export type CompositeDirections =
-  typeof CompositeDirections[keyof typeof CompositeDirections];
-/* eslint-enable @typescript-eslint/no-redeclare */
--- a/src/Directions.ts
+++ b/src/Directions.ts
@@ -14,6 +14,6 @@ export const CornerDirections = {
 
 /* eslint-disable @typescript-eslint/no-redeclare -- backward compatibility; it can be used as a type and as a value */
 export type Directions = (typeof Directions)[keyof typeof Directions];
-// prettier-ignore
-export type CornerDirections = typeof CornerDirections[keyof typeof CornerDirections];
+export type CornerDirections =
+  (typeof CornerDirections)[keyof typeof CornerDirections];
 /* eslint-enable @typescript-eslint/no-redeclare */
--- a/src/Directions.ts
+++ b/src/Directions.ts
@@ -1,5 +1,3 @@
-import { lastArrayValues } from './web/utils';
-
 const RIGHT = 1;
 const LEFT = 2;
 const UP = 4;
@@ -14,26 +12,15 @@ export const Directions = {
 } as const;
 
 // internal interface
-export const CompositeDirections = {
-  RIGHT: RIGHT,
-  LEFT: LEFT,
-  UP: UP,
-  DOWN: DOWN,
+export const DiagonalDirections = {
   UP_RIGHT: UP | RIGHT,
   DOWN_RIGHT: DOWN | RIGHT,
   UP_LEFT: UP | LEFT,
   DOWN_LEFT: DOWN | LEFT,
 } as const;
 
-export const axialDirectionsList = Object.values(Directions);
-export const diagonalDirectionsList = lastArrayValues(
-  Object.values(CompositeDirections),
-  4
-);
-
 // eslint-disable-next-line @typescript-eslint/no-redeclare -- backward compatibility; it can be used as a type and as a value
 export type Directions = typeof Directions[keyof typeof Directions];
-
 // eslint-disable-next-line @typescript-eslint/no-redeclare
-export type CompositeDirections =
-  typeof CompositeDirections[keyof typeof CompositeDirections];
+export type DiagonalDirections =
+  typeof DiagonalDirections[keyof typeof DiagonalDirections];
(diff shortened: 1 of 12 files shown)


LatekVo:
I am still polishing up some details, but if you have any questions or suggestions, feel free to ask.

I actually want to ask for your opinion, current `cornerFillAlignmentList` initialisation prevents corners from being calculated if `enableCornerActivation` is `false`, but since that will probably never happen, i feel like including that condition only makes the code less readable. 

So instead, i could move this condition a few lines lower to `isAligned` like this: 
```ts
const alignmentList = Object.values(Directions).map((direction) =>
  getAlignment(direction, this.minimalAlignmentCosine)
);
const cornerFillAlignmentList = Object.values(CornerDirections).map((direction) =>
  getAlignment(direction, this.minimalCornerAlignmentCosine)
);

const isAligned = alignmentList.some(Boolean) || (cornerFillAlignmentList.some(Boolean) && this.enableCornerActivation);
```

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -18,13 +19,25 @@ export default class FlingGestureHandler extends GestureHandler {
   private maxDurationMs = DEFAULT_MAX_DURATION_MS;
   private minVelocity = DEFAULT_MIN_VELOCITY;
   private minDirectionalAlignment = DEFAULT_MIN_DIRECTION_ALIGNMENT;
+  private enableCornerActivation = DEFAULT_ENABLE_CORNER_ACTIVATION;
```
m-bert, line 22: As far as I can see this value is always `true`. Making a prop that sets it doesn't make a lot of sense if you ask me - if I were a user, I'd assume that passing for example `Directions.UP | Directions.LEFT` is enough. 

Is there any situation that we want to disable this behavior? (cc @j-piasecki)

LatekVo: Removed all instances of this flag.

src/Directions.ts
```diff
@@ -5,5 +5,15 @@ export const Directions = {
   DOWN: 8,
 } as const;
 
-// eslint-disable-next-line @typescript-eslint/no-redeclare -- backward compatibility; it can be used as a type and as a value
+export const CornerDirections = {
```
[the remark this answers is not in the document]
m-bert: Now when I think of it, it could be slightly harder to implement `Vector.fromDirection`.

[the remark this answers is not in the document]
LatekVo: The real issue i see is that minimal detection angle would be capped at >=30 percent, as previously Corners had a different, adaptive detection angle to fill the missing uncovered space, and after combining both classes, it no longer works that way.
I also think exposing directions like UP_RIGHT etc. to the user looks unclean.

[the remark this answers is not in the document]
LatekVo: Nevermind, since current minimal angle is 40 degrees, and it is an unchangeable constant, i will apply all your suggestions.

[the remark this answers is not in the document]
m-bert: > That's a good call. I don't think it would be harder though, as fromDirection looks for exact matches, and i don't see edge-cases where it could return the wrong thing.

It does, but in case of `ALL` we don't have exact match 😄 

> But it makes Directions implementation less readable, it is not possible to reference this. nor Directions. when inside, thus we have to rely on normal numbers or a separate object.

We could make a structure like `BaseDirections`, `DiagonalDirections` and then combine them into `Directions`, though it could be too much.

> I also think exposing directions like UP_RIGHT etc. to the user looks unclean.

I agree, that's why I want to hear what @j-piasecki has to say in that case, as exposing things such as `UP_RIGHT` may or may not make sense (I think it would if we didn't have `|` operator to combine directions).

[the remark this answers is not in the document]
LatekVo: > It does, but in case of ALL we don't have exact match 😄

We have an exact match for every single possible mask, because `fromDirection` is called for every `Directions` member, and evaluates it's individual activation. For example, if both `UP` and `RIGHT` are active, then all 3 `UP`, `RIGHT` and `UP_RIGHT` will trigger the `.some(Boolean)` individually. If we were to add an `ALL`, then while it wouldn't match by itself with anything, it would activate each of the other directions individually, guaranteeing `.some(Boolean)` to be activated.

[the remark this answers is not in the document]
LatekVo: Ok but since we agree on:

> I also think exposing directions like UP_RIGHT etc. to the user looks unclean.

I think I can separate the exposed and the used api into `Directions`, `CompositeDirections` and `CombinedDirections`, of which only `Directions` will be exposed to the user.
Is that okay by you?

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -17,14 +17,23 @@ export default class FlingGestureHandler extends GestureHandler {
...
 
   private maxNumberOfPointersSimultaneously = 0;
   private keyPointer = NaN;
 
+  private minimalAlignmentCosine = 0;
+
   public init(ref: number, propsRef: React.RefObject<unknown>): void {
     super.init(ref, propsRef);
+
+    const degToRad = (degrees: number) => (degrees * Math.PI) / 180;
```
j-piasecki, line 31: Move this to `utils` please

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -7,7 +7,7 @@ import Vector from '../tools/Vector';
 
 const DEFAULT_MAX_DURATION_MS = 800;
 const DEFAULT_MIN_VELOCITY = 700;
-const DEFAULT_MIN_DIRECTION_ALIGNMENT = 0.75;
+const DEFAULT_DIRECTIONAL_ALIGNMENT_CONE = 40;
```
j-piasecki, line 10: Shouldn't this be 45 if we want to cover the full 90 degrees when adjacent directions are set? I think now it leaves two 5-degree cones where the gesture won't be recognized.

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -17,14 +17,23 @@ export default class FlingGestureHandler extends GestureHandler {
...
+
   public init(ref: number, propsRef: React.RefObject<unknown>): void {
     super.init(ref, propsRef);
+
+    const degToRad = (degrees: number) => (degrees * Math.PI) / 180;
+
+    const alignmentRadians = degToRad(this.directionalAlignmentCone);
+    const maxDirectionalDeviation = alignmentRadians / 2;
+
+    this.minimalAlignmentCosine = Math.cos(maxDirectionalDeviation);
```
j-piasecki, line 36: This is constant for every gesture, no? If so, we can move the calculations outside and assign the result to a constant.

src/web/handlers/FlingGestureHandler.ts
LatekVo: Since we implemented the diagonal fill, this value can go as low as 30 degrees without leaving blank spots, as 2 cones protrude axially, and then a third one is added between them. Additionally if we decide to leave the division between Diagonal and Axial directions, we already have a function ready to adapt to the missing unfilled space.
In summary, in this particular commit there is a 10 degree cone left blank by the Axial directions, and filled by the Diagonal directions.

LatekVo:
I will apply all relevant changes to the kotlin side in the next commit.

src/Directions.ts
```diff
@@ -1,18 +1,43 @@
...
   UP_LEFT: UP | LEFT,
   DOWN_LEFT: DOWN | LEFT,
 } as const;
 
+export const axialDirectionsList = firstArrayValues(
+  Object.values(CompositeDirections),
+  4
+);
+
+export const diagnalDirectionsList = lastArrayValues(
```
j-piasecki, line 33: This is unused

src/Directions.ts
```diff
@@ -1,18 +1,43 @@
...
+  LEFT: LEFT,
+  UP: UP,
+  DOWN: DOWN,
   UP_RIGHT: UP | RIGHT,
   DOWN_RIGHT: DOWN | RIGHT,
   UP_LEFT: UP | LEFT,
   DOWN_LEFT: DOWN | LEFT,
 } as const;
 
+export const axialDirectionsList = firstArrayValues(
```
j-piasecki, line 28: This is just `Object.values(Directions)`

src/web/utils.ts
```diff
@@ -15,3 +15,14 @@ export const PointerTypeMapping = new Map<string, PointerType>([
...
+export const degToRad = (degrees: number) => (degrees * Math.PI) / 180;
+
+export const coneToDeviation = (degrees: number) =>
+  Math.cos(degToRad(degrees / 2));
+
+export const firstArrayValues = (array: Array<any>, amount: number) =>
+  array.filter((_, index) => index < amount);
+
+export const lastArrayValues = (array: Array<any>, amount: number) =>
+  array.filter((_, index, object) => index > object.length - amount - 1);
```
j-piasecki, line 28: And then those aren't needed

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -73,11 +71,18 @@ export default class FlingGestureHandler extends GestureHandler {
...
+    );
+
+    const diagonalAlignmentList = axialDirectionsList.map((direction) =>
+      getAlignment(direction, DIAGONAL_DEVIATION_COSINE)
     );
 
-    const isAligned = alignmentList.some(Boolean);
+    const isAligned = axialAlignmentList
+      .concat(diagonalAlignmentList)
+      .some(Boolean);
```
j-piasecki, line 84:
```suggestion
    const isAligned = axialAlignmentList.some(Boolean) || diagonalAlignmentList.some(Boolean);
```

`concat` creates a new array each time with values from both sources copied

src/Directions.ts
LatekVo: This is an error, `diagnalDirectionsList` should be used, but i accidently used `axialDirectionsList` in it's place.

src/web/handlers/FlingGestureHandler.ts
LatekVo: Please tell me if this current implementation is what you've had in mind.

src/web/handlers/FlingGestureHandler.ts
LatekVo: I ended up using the adaptive method i mentioned.

m-bert (commented):
Looks good, but I'd like to discuss something about `CompositeDirections`.

src/Directions.ts
```diff
@@ -1,9 +1,39 @@
...
+// internal interface
+export const CompositeDirections = {
+  RIGHT: RIGHT,
+  LEFT: LEFT,
+  UP: UP,
+  DOWN: DOWN,
+  UP_RIGHT: UP | RIGHT,
+  DOWN_RIGHT: DOWN | RIGHT,
+  UP_LEFT: UP | LEFT,
+  DOWN_LEFT: DOWN | LEFT,
```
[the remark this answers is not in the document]
LatekVo: We decided to remove `Composite` directions all together since it wasn't neither more readable nor more efficient

m-bert (approved):
LGTM

PR #2759, opened 2024-02-14
Fix nested taps activation

m-bert:
## Description

While working on #2739 we discovered that our orchestration process doesn't work as expected when we have nested `Tap` gestures . This PR introduces few changes that make those gestures work as intended. 

Fixes #2739 

## Test plan

Tested on example app and on modified code from issue.

<details>
<summary> Test code </summary>

```jsx
import React from 'react';
import { StyleSheet, View } from 'react-native';
import { Gesture, GestureDetector } from 'react-native-gesture-handler';
import Animated from 'react-native-reanimated';

const log = (e: any, handler: string) =>
  console.log(`[${e.handlerTag}][${handler}]: ${e.state}`);

const orangeTapGesture = Gesture.Tap()
  .runOnJS(true)
  .maxDeltaX(5)
  .onBegin((e) => log(e, 'Orange tap'))
  .onStart((e) => log(e, 'Orange tap'))
  .onFinalize((e, success) => {
    log(e, 'Orange tap');
    if (success) {
      console.log('---> Orange Tap <---');
    }
  });

const orangeDoubleTapGesture = Gesture.Tap()
  .numberOfTaps(2)
  .runOnJS(true)
  .onBegin((e) => log(e, 'Orange Double Tap'))
  .onStart((e) => log(e, 'Orange Double Tap'))
  .onFinalize((e, success) => {
    log(e, 'Orange Double Tap');
    if (success) {
      console.log('---> Orange Double Tap <---');
    }
  });

const blueTapGesture = Gesture.Tap()
  .runOnJS(true)
  .requireExternalGestureToFail(orangeTapGesture, orangeDoubleTapGesture)
  .onBegin((e) => log(e, 'Blue Tap'))
  .onStart((e) => log(e, 'Blue Tap'))
  .onFinalize((e, success) => {
    log(e, 'Blue Tap');
    if (success) {
      console.log('---> Blue Tap <---');
    }
  });

const blueDoubleTapGesture = Gesture.Tap()
  .numberOfTaps(2)
  .runOnJS(true)
  .onBegin((e) => log(e, 'Blue Double Tap'))
  .onStart((e) => log(e, 'Blue Double Tap'))
  .onFinalize((e, success) => {
    log(e, 'Blue Double Tap');
    if (success) {
      console.log('---> Blue Double Tap <---');
    }
  });

const redLongPressGesture = Gesture.LongPress()
  .runOnJS(true)
  .onBegin((e) => log(e, 'Red Longpress'))
[...]

PR #2819, opened 2024-03-21
Web `Orchestrator` refactor.

m-bert:
## Description

While working on #2759 I've decided that it will be good idea to also introduce some changes into web version of `orchestrator`. 

While this PR is mostly a refactor, it also brings other changes, for example it adds `awaitingHandlersTags` `set` which is used on android (current implementation was unsafe).

Unfortunately change made in `GestureHandler.ts` reverts #2802, but it is necessary for orchestration process to work. We will find better way to solve this problem in follow-up PR.

## Test plan

<details>
<summary> Tested on the following code: </summary>

```jsx
import React from 'react';

const log = (e: any, handler: string) =>
  console.log(`[${e.handlerTag}][${handler}]: ${e.state}`);

const orangeTapGesture = Gesture.Tap()
  .runOnJS(true)
  .maxDeltaX(5)
  .onBegin((e) => log(e, 'Orange tap'))
  .onStart((e) => log(e, 'Orange tap'))
  .onFinalize((e, success) => {
    log(e, 'Orange tap');
    if (success) {
      console.log('---> Orange Tap <---');
    }
  });

const orangeDoubleTapGesture = Gesture.Tap()
  .numberOfTaps(2)
  .runOnJS(true)
  .onBegin((e) => log(e, 'Orange Double Tap'))
  .onStart((e) => log(e, 'Orange Double Tap'))
  .onFinalize((e, success) => {
    log(e, 'Orange Double Tap');
    if (success) {
      console.log('---> Orange Double Tap <---');
    }
  });

const blueTapGesture = Gesture.Tap()
  .runOnJS(true)
  .requireExternalGestureToFail(orangeTapGesture, orangeDoubleTapGesture)
  .onBegin((e) => log(e, 'Blue Tap'))
  .onStart((e) => log(e, 'Blue Tap'))
  .onFinalize((e, success) => {
    log(e, 'Blue Tap');
    if (success) {
      console.log('---> Blue Tap <---');
    }
  });

const blueDoubleTapGesture = Gesture.Tap()
  .numberOfTaps(2)
  .runOnJS(true)
  .onBegin((e) => log(e, 'Blue Double Tap'))
  .onStart((e) => log(e, 'Blue Double Tap'))
  .onFinalize((e, success) => {
    log(e, 'Blue Double Tap');
    if (success) {
      console.log('---> Blue Double Tap <---');
    }
  });

const redLongPressGesture = Gesture.LongPress()
  .runOnJS(true)
  .onBegin((e) => log(e, 'Red Longpress'))
  .onStart((e) => log(e, 'Red Longpress'))
  .onFinalize((e, success) => {
    log(e, 'Red Longpress');
    if (success) {
      console.log('---> Red Longpress <---');
    }
  });

const redDoubleTapGesture = Gesture.Tap()
  .numberOfTaps(2)
  .runOnJS(true)
  .onBegin((e) => log(e, 'Red Double Tap'))
  .onStart((e) => log(e, 'Red Double Tap'))
  .onFinalize((e, success) => {
    log(e, 'Red Double Tap');
    if (success) {
      console.log('---> Red Double Tap <---');
    }
  });

const Showcase = () => {
  const gesture = Gesture.Race(redDoubleTapGesture, redLongPressGesture);

  return (
    <View style={[styles.center, { flex: 1, backgroundColor: 'white' }]}>
        <Animated.View style={[styles.outer, styles.center]}>
          <Blue />
        </Animated.View>
      </GestureDetector>
    </View>
  );
};

export default Showcase;

const Blue = () => {
  const composedGesture = Gesture.Exclusive(
    blueDoubleTapGesture,
    blueTapGesture
  );

  return (
    <GestureDetector gesture={composedGesture}>
      <Animated.View style={[styles.blue, styles.center]}>
        <Orange />
      </Animated.View>
    </GestureDetector>
  );
};

const Orange = () => {
  const composedGesture = Gesture.Exclusive(
    orangeDoubleTapGesture,
    orangeTapGesture
  );

  return (
    <GestureDetector gesture={composedGesture}>
      <Animated.View style={styles.orange} />
    </GestureDetector>
  );
};

  center: {
    display: 'flex',
    alignItems: 'center',
  },

  outer: {
    backgroundColor: 'red',
    width: 400,
    height: 400,
    borderRadius: 200,
  },
  blue: {
    width: 250,
    height: 250,
    backgroundColor: 'blue',

    borderRadius: 175,
  },
  orange: {
    width: 100,
    height: 100,
    backgroundColor: 'orange',
    borderRadius: 50,
  },
});

```

</details>

### DIFF
--- a/src/web/tools/GestureHandlerOrchestrator.ts
+++ b/src/web/tools/GestureHandlerOrchestrator.ts
@@ -143,23 +143,28 @@ export default class GestureHandlerOrchestrator {
     this.handlingChangeSemaphore += 1;
 
     if (this.isFinished(newState)) {
-      this.awaitingHandlers.forEach((otherHandler) => {
-        if (this.shouldHandlerWaitForOther(otherHandler, handler)) {
-          if (newState === State.END) {
-            otherHandler?.cancel();
-            if (otherHandler.getState() === State.END) {
-              // Handle edge case, where discrete gestures end immediately after activation thus
-              // their state is set to END and when the gesture they are waiting for activates they
-              // should be cancelled, however `cancel` was never sent as gestures were already in the END state.
-              // Send synthetic BEGAN -> CANCELLED to properly handle JS logic
-              otherHandler.sendEvent(State.CANCELLED, State.BEGAN);
-            }
-            otherHandler?.setAwaiting(false);
-          } else {
-            this.tryActivate(otherHandler);
-          }
+      for (const otherHandler of this.awaitingHandlers) {
+        if (!this.shouldHandlerWaitForOther(otherHandler, handler)) {
+          continue;
         }
-      });
+
+        if (newState !== State.END) {
+          this.tryActivate(otherHandler);
+          continue;
+        }
+
+        otherHandler.cancel();
+
+        if (otherHandler.getState() === State.END) {
+          // Handle edge case, where discrete gestures end immediately after activation thus
+          // their state is set to END and when the gesture they are waiting for activates they
+          // should be cancelled, however `cancel` was never sent as gestures were already in the END state.
+          // Send synthetic BEGAN -> CANCELLED to properly handle JS logic
+          otherHandler.sendEvent(State.CANCELLED, State.BEGAN);
+        }
+
+        otherHandler.setAwaiting(false);
+      }
     }
 
     if (newState === State.ACTIVE) {
@@ -184,7 +189,7 @@ export default class GestureHandlerOrchestrator {
 
     this.scheduleFinishedHandlersCleanup();
 
-    if (this.awaitingHandlers.indexOf(handler) < 0) {
+    if (!this.awaitingHandlers.includes(handler)) {
       this.cleanupAwaitingHandlers(handler);
     }
   }
--- a/src/web/tools/GestureHandlerOrchestrator.ts
+++ b/src/web/tools/GestureHandlerOrchestrator.ts
@@ -133,8 +133,8 @@ export default class GestureHandlerOrchestrator {
       }
     }
 
-    this.awaitingHandlers = this.awaitingHandlers.filter(
-      (otherHandler) => !shouldWait(otherHandler)
+    this.awaitingHandlers = this.awaitingHandlers.filter((otherHandler) =>
+      this.awaitingHandlersTags.has(otherHandler.getTag())
     );
   }
 
--- a/src/web/tools/GestureHandlerOrchestrator.ts
+++ b/src/web/tools/GestureHandlerOrchestrator.ts
@@ -207,12 +207,11 @@ export default class GestureHandlerOrchestrator {
       }
     }
 
-    this.awaitingHandlers.forEach((otherHandler) => {
+    for (const otherHandler of this.awaitingHandlers) {
       if (this.shouldHandlerBeCancelledBy(otherHandler, handler)) {
-        otherHandler?.cancel();
-        otherHandler?.setAwaiting(true);
+        otherHandler.setAwaiting(true);
       }
-    });
+    }
 
     handler.sendEvent(State.ACTIVE, State.BEGAN);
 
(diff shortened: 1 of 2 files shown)


src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -33,47 +34,48 @@ export default class GestureHandlerOrchestrator {
...
-    return hasToWait;
+    return this.gestureHandlers.some(hasToWaitFor);
   }
 
   private shouldBeCancelledByFinishedHandler(
     handler: IGestureHandler
   ): boolean {
-    return this.gestureHandlers.some(
-      (otherHandler) =>
+    const shouldBeCancelled = (otherHandler: IGestureHandler) => {
```
j-piasecki, line 71:
```suggestion
    const shouldBeCancelledBy = (otherHandler: IGestureHandler) => {
```

src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -84,44 +86,56 @@ export default class GestureHandlerOrchestrator {
...
+
+    for (const otherHandler of this.awaitingHandlers) {
+      if (shouldWait(otherHandler)) {
+        this.cleanHandler(otherHandler);
+        this.awaitingHandlersTags.delete(otherHandler.getTag());
       }
     }
+
+    this.awaitingHandlers = this.awaitingHandlers.filter(
+      (otherHandler) => !shouldWait(otherHandler)
```
j-piasecki, line 137: It looks like `awaitingHandlersTags` and `awaitingHandlers` should be in sync, so cannot we check if the handler tag is in `awaitingHandlersTags` instead or repeating the checks again?

src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -196,12 +218,11 @@ export default class GestureHandlerOrchestrator {
...
 
-    this.awaitingHandlers.forEach((otherHandler) => {
+    for (const otherHandler of this.awaitingHandlers) {
       if (this.shouldHandlerBeCancelledBy(otherHandler, handler)) {
-        otherHandler?.cancel();
-        otherHandler?.setAwaiting(true);
+        otherHandler.setAwaiting(true);
       }
-    });
+    }
```
j-piasecki, line 225: This loop doesn't make a lot of sense, we go through the awaiting handlers and mark them as awaiting. On Android, we mark them as *not* awaiting and then clean up the awaiting handlers list, which makes more sense since `shouldHandlerBeCancelledBy` is the condition. Because of that, I'm also not sure about removing the call to `.cancel()`, but the logic on web here is not 1:1 to Android.

src/web/tools/GestureHandlerOrchestrator.ts
m-bert: Done in [145e9f7](https://github.com/software-mansion/react-native-gesture-handler/pull/2819/commits/145e9f710b298a064f7ff16ef34195826e3cb807).

src/web/tools/GestureHandlerOrchestrator.ts
m-bert: Good catch! 

Done in [98e2637](https://github.com/software-mansion/react-native-gesture-handler/pull/2819/commits/98e2637671b376864d780138902e5f95e74094df).

src/web/tools/GestureHandlerOrchestrator.ts
m-bert: >This loop doesn't make a lot of sense, we go through the awaiting handlers and mark them as awaiting.

I agree. It was there for a very long time, probably we overlooked it during development of new web implementation. I've changed that to `false` and it seems to work fine. 

>Because of that, I'm also not sure about removing the call to .cancel(), but the logic on web here is not 1:1 to Android.

I think that removing `cancel` is fine, since this is the same condition as in the loop above and we agreed, that if handler is in `gestureHandlers` then it has to be in `awaitingHandlers` (that's why we also removed this `cancel` on android).

Done in [cca5d5f](https://github.com/software-mansion/react-native-gesture-handler/pull/2819/commits/cca5d5fa776cccc43e3135fac5e0b05071b81d3c).

j-piasecki (approved):

Issue #2816, opened 2024-03-20
Get wrong, negative value from e.absoluteY on Android


doublelam:
### Description

<img width="637" alt="image" src="https://github.com/software-mansion/react-native-gesture-handler/assets/12025501/d4cbd5b1-b910-4e38-a4f8-cb5e9062b496">

Different platforms behave differently. I got a negative value when I moved the pointer to the top of the screen on Android but got 0 for iOS. I need this value to calculate if the pointer is in a specified element area, so when I use this value to compare with a pageY value, it gives a wrong result.

PS: I found that normally the value varies on ios and Android makes sense because StatusBar occupied some height on Android. But when StatusBar is set to translucent, the value is not correct enough anymore.

### Steps to reproduce

Just get an `absoluteY` value from `onUpdate` function

### Snack or a link to a repository

null

### Gesture Handler version

2.13.1

### React Native version

0.72.6

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

PR #2826, opened 2024-03-26
Use correct origin point for calculating the absolute position on Android

j-piasecki:
## Description

In https://github.com/software-mansion/react-native-gesture-handler/pull/1812 I changed how absolute position is calculated to be relative to the window instead of the screen, but there were two things wrong with it:
- It used [getWindowVisibleDisplayFrame](https://developer.android.com/reference/android/view/View#getWindowVisibleDisplayFrame(android.graphics.Rect)) which `tells you the available area where content can be placed and remain visible to users`. This was always taking status bar height into account, even if the app could be drawn behind it.
- It was calling this method on the `decorView`, which is drawn under system bars anyway

This PR changes it to use `contentView` of the current activity and [getLocationOnScreen](https://developer.android.com/reference/android/view/View#getLocationOnScreen(int[])) which `gets the coordinates of this view in the coordinate space of the device screen, irrespective of system decorations and whether the system is in multi-window mode`.

In a nutshell, after this PR the absolute position will take into account whether the app is displayed behind the status bar.

## Test plan

Tested on the example app:

https://github.com/software-mansion/react-native-gesture-handler/assets/21055725/c3230a2d-7bba-483b-b160-1c9fb32b0da1

https://github.com/software-mansion/react-native-gesture-handler/assets/21055725/701d9fa0-0c9f-4126-af58-f80c6a0a6526

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -4,14 +4,13 @@ import android.app.Activity
 import android.content.Context
 import android.content.ContextWrapper
 import android.graphics.PointF
-import android.graphics.Rect
 import android.os.Build
 import android.view.MotionEvent
 import android.view.MotionEvent.PointerCoords
 import android.view.MotionEvent.PointerProperties
 import android.view.View
-import android.view.Window
 import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.ReactContext
 import com.facebook.react.bridge.UiThreadUtil
 import com.facebook.react.bridge.WritableArray
 import com.facebook.react.uimanager.PixelUtil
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -192,10 +188,11 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
...
   protected open fun onPrepare() {}
 
-  private fun getWindow(context: Context?): Window? {
+  private fun getActivity(context: Context?): Activity? {
     if (context == null) return null
-    if (context is Activity) return context.window
-    if (context is ContextWrapper) return getWindow(context.baseContext)
+    if (context is ReactContext) return context.currentActivity
+    if (context is Activity) return context
+    if (context is ContextWrapper) return getActivity(context.baseContext)
```
m-bert, line 195:
```suggestion
    return when(context){
      is ReactContext -> context.currentActivity
      is Activity -> context
      is ContextWrapper -> getActivity(context.baseContext)
      else -> null
    }
```

Wouldn't this work? I think it looks cleaner.

m-bert (approved):
LGTM

m-bert:
Also you can add that it fixes #2816

android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/2826/commits/422aa713724fc9dfad11757af2259c5cad7b2989

PR #2830, opened 2024-03-26
Replace `mach_absolute_time` with `CACurrentMediaTime`

j-piasecki:
## Description

Replaces `mach_absolute_time` with `CACurrentMediaTime` which doesn't need to be specified in the privacy manifest.

## Test plan

Tested on the Example app.


m-bert (approved):
LGTM

PR #2831, opened 2024-03-26
Add check when removing handlers from orchestrator

m-bert:
## Description

#2819 removed changes made in #2802. After further investigation we found out that `removeHandlerFromOrchestrator` didn't check if handlers were present in orchestrator, therefore in some situation calling this function ended up in removing wrong handler. Calling `indexOf` returned `-1` and, as you can read in [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice):

> Negative index counts back from the end of the array

So instead of ignoring that, we simply removed wrong handler.

## Test plan

Tested on example app (mostly on swipeable example) and example code from #2819

### DIFF
--- a/src/web/tools/GestureHandlerOrchestrator.ts
+++ b/src/web/tools/GestureHandlerOrchestrator.ts
@@ -32,9 +32,14 @@ export default class GestureHandlerOrchestrator {
   }
 
   public removeHandlerFromOrchestrator(handler: IGestureHandler): void {
-    this.gestureHandlers.splice(this.gestureHandlers.indexOf(handler), 1);
-    this.awaitingHandlers.splice(this.awaitingHandlers.indexOf(handler), 1);
-    this.awaitingHandlersTags.delete(handler.getTag());
+    if (this.gestureHandlers.includes(handler)) {
+      this.gestureHandlers.splice(this.gestureHandlers.indexOf(handler), 1);
+    }
+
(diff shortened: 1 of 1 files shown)


src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -32,9 +32,14 @@ export default class GestureHandlerOrchestrator {
   }
 
   public removeHandlerFromOrchestrator(handler: IGestureHandler): void {
-    this.gestureHandlers.splice(this.gestureHandlers.indexOf(handler), 1);
-    this.awaitingHandlers.splice(this.awaitingHandlers.indexOf(handler), 1);
-    this.awaitingHandlersTags.delete(handler.getTag());
+    if (this.gestureHandlers.includes(handler)) {
+      this.gestureHandlers.splice(this.gestureHandlers.indexOf(handler), 1);
```
j-piasecki, line 36: You can save the index to a variable and check if it's `-1` so you don't look through the array twice (same below).

m-bert: Done in [5f00c22](https://github.com/software-mansion/react-native-gesture-handler/pull/2831/commits/5f00c2221ffab64ead24c8b841cc065566eccbc6).

j-piasecki (approved):

PR #2835, opened 2024-03-27
Change `.web` files to default and add `.native`

m-bert:
## Description

This PR moves content of `.web` files into default ones, which are replaced by `.native`. We do this because `metro` is good at resolving those types of files, but other bundlers are not.

Change in `tsconfig` assures, that types are correctly resolved.

## Test plan

Built and played a little bit with example app on:

- Web
- iOS
- Android


piaskowyk (approved):
👏

j-piasecki (approved):

Zoxive:
After getting latest with this change, Full types for FlatList, TextInput, etc are all missing.

(Locally I'm on typescript 5.4.5)

```typescript

import { FlatList, TextInput } from 'react-native-gesture-handler';

export function NotificationsPage() {
  const notificationIds = ['one', 'two']
  const textInputRef = useRef<TextInput>(null); // ERROR 'TextInput' refers to a value, but is being used as a type here
  
  return (
	  <FlatList
		  data={notificationIds}
		  renderItem={({ item }) => <NotificationBarItem notificationId={item} />}
		  keyExtractor={(item) => item} // ERROR item is unknown 
	  />
  </View>
);
}
```

Im not sure what the correct fix is. I first thought i could simply export the types in https://github.com/software-mansion/react-native-gesture-handler/blob/main/src/index.ts but the types seem to work fine inside the repository directly. (Because ts resolves the .native.tsx file)

Im guessing that the types were lost when the NPM package was bundled?

hirbod:
@m-bert i think this PR should be reverted or fixed asap.

It breaks types on a wide range of projects. Some more insight here https://github.com/software-mansion/react-native-gesture-handler/issues/2889#issuecomment-2090010196

PR #2843, opened 2024-04-02
Disable flipper on CI

m-bert:
## Description

Currently our iOS build fails on Flipper build. This PR adds `NO_FLIPPER` flag since it is no longer required.

## Test plan

Check that CI passes.

### DIFF
--- a/.github/workflows/ios-build.yml
+++ b/.github/workflows/ios-build.yml
@@ -21,6 +21,8 @@ jobs:
     concurrency:
       group: ios-${{ matrix.working-directory }}-${{ github.ref }}
       cancel-in-progress: true
+    env:
+      NO_FLIPPER: 1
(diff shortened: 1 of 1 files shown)


.github/workflows/ios-build.yml
```diff
@@ -21,6 +21,8 @@ jobs:
     concurrency:
       group: ios-${{ matrix.working-directory }}-${{ github.ref }}
       cancel-in-progress: true
+    env:
+      NO_FLIPPER: 1
```
tomekzaw, line 25: We can set that environmental variable just for "Install pods" step, e.g.
```sh
bundle install && NO_FLIPPER=1 bundle exec pod install
```

tomekzaw (approved):
LGTM!

PR #2848, opened 2024-04-04
Bump `typescript` version in example

m-bert:
## Description

Our latest fails on CI were caused by outdated `typescript` in example app.  #2835 added `moduleSuffixes` into `compilerOptions` in `tsconfig.json`, but this option was added in version 4.7 of `typescript` and our example app currently has version 4.5.5.

This PR updates `typescript` version in our example app to much version in the root.

## Test plan

Check that CI passes.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -55,7 +55,7 @@
     "metro-react-native-babel-preset": "0.76.5",
     "prettier": "^2.2.1",
     "react-test-renderer": "18.2.0",
-    "typescript": "^4.5.5",
+    "typescript": "5.0.4",
     "webpack-dev-server": "3.11.3"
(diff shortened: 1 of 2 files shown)


example/package.json
```diff
@@ -55,7 +55,7 @@
     "metro-react-native-babel-preset": "0.76.5",
     "prettier": "^2.2.1",
     "react-test-renderer": "18.2.0",
-    "typescript": "^4.5.5",
+    "typescript": "5.0.4",
```
tjzel, line 58:
```suggestion
    "typescript": "5.0.4",
```

tjzel (approved):

PR #2855, opened 2024-04-08
Add `Next` example app

m-bert:
## Description

This PR adds `Next.js` example to our repository.

## Test plan

Well... run example app I guess?

1.  `cd next-example`
2.  `yarn dev`
3.  Visit `localhost:3000`


m-bert:
> [!CAUTION]
> We've decided to drop this project for now as we couldn't find good way to deal with typescript issues. 
> It is possible that we will get back to it in the future.

PR #2874, opened 2024-04-23
Add `TypeDoc` to new API

m-bert:
## Description

I thought that adding TypeDocs might be a good idea and it could enhance developer experience (I've come up with this idea when I discovered that we have `withRef` modifier). This PR adds TypeDocs to new API (but we can also think of filling missing docs in old API)

## Test plan

Hover functions/components and read docs.

### DIFF
--- a/src/handlers/gestures/flingGesture.ts
+++ b/src/handlers/gestures/flingGesture.ts
@@ -13,11 +13,21 @@ export class FlingGesture extends BaseGesture<FlingGestureHandlerEventPayload> {
     this.handlerName = 'FlingGestureHandler';
   }
 
+  /**
+   * Determine exact number of points required to handle the fling gesture.
+   * @param pointers
+   */
   numberOfPointers(pointers: number) {
     this.config.numberOfPointers = pointers;
     return this;
   }
 
+  /**
+   * Expressed allowed direction of movement.
+   * Expected values are exported as constants in the Directions object.
+   * @param direction
+   * @see https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/fling-gesture/#directionvalue-directions
+   */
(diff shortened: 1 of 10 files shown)


src/handlers/gestures/flingGesture.ts
```diff
@@ -13,11 +13,21 @@ export class FlingGesture extends BaseGesture<FlingGestureHandlerEventPayload> {
...
+   * @param pointers
+   */
   numberOfPointers(pointers: number) {
     this.config.numberOfPointers = pointers;
     return this;
   }
 
+  /**
+   * Expressed allowed direction of movement.
+   * Expected values are exported as constants in the Directions object.
```
j-piasecki, line 27: Would be nice to mention that multiple directions at once can be recognized.

j-piasecki (approved):

src/handlers/gestures/gesture.ts
```diff
@@ -202,6 +224,10 @@ export abstract class BaseGesture<
     return this;
   }
 
+  /**
+   * Set the `onTouchesDown` callback which is called every time a finger is placed on the screen.
```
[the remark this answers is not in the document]
m-bert: Done in [ef3e648](https://github.com/software-mansion/react-native-gesture-handler/pull/2874/commits/ef3e6482945ac4ce63f69eb167013398e8c88cc3)

src/handlers/gestures/flingGesture.ts
m-bert: Done in [4de93ca](https://github.com/software-mansion/react-native-gesture-handler/pull/2874/commits/4de93ca0ef7f3048ba5c7ae1aba35ff3d4f9ec92)

PR #2881, opened 2024-04-26
Fix old Gesture Handler API integration with Reaniamted

piaskowyk:
## Description

This PR resolves the recognition issue of the Reanimated Event handler for the previous Gesture Handler API. This adjustment was necessary due to a modification in the internal property name that was utilized to identify the Reanimated event handler (as seen here: https://github.com/software-mansion/react-native-reanimated/pull/5743).

## Test plan

Run this example - https://github.com/software-mansion/react-native-reanimated/blob/main/app/src/examples/DragAndSnapExample.tsx

### DIFF
--- a/src/handlers/createHandler.tsx
+++ b/src/handlers/createHandler.tsx
@@ -343,9 +343,9 @@ export default function createHandler<
         const actionType = (() => {
           if (
             (this.props?.onGestureEvent &&
-              'current' in this.props.onGestureEvent) ||
+              'workletEventHandler' in this.props.onGestureEvent) ||
             (this.props?.onHandlerStateChange &&
-              'current' in this.props.onHandlerStateChange)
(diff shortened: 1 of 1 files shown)


src/handlers/createHandler.tsx
```diff
@@ -343,9 +343,9 @@ export default function createHandler<
         const actionType = (() => {
           if (
             (this.props?.onGestureEvent &&
-              'current' in this.props.onGestureEvent) ||
+              'workletEventHandler' in this.props.onGestureEvent) ||
             (this.props?.onHandlerStateChange &&
-              'current' in this.props.onHandlerStateChange)
+              'workletEventHandler' in this.props.onHandlerStateChange)
```
j-piasecki, line 348: Wouldn't that break for Rea < 3.9.0, or whatever version the `useEvent` got changed?

j-piasecki (approved):

PR #2882, opened 2024-04-29
Update yarn setup in the repository

j-piasecki:
## Description

Sets the package manager used in the example apps and in the root of the repository to `yarn@1.22.22`.

## Test plan

Install dependencies & build apps


m-bert:
Shouldn't we also update `Podfile.lock` and `yarn.lock` in `example` app?

j-piasecki:
There are no changes there

m-bert (approved):
:shipit:

PR #2884, opened 2024-04-29
[Web] Add `minVelocity` to `Pan` custom activation criteria

m-bert:
## Description

Recently I've discovered that using `.minVelocity()` modifier doesn't affect `Pan` activation on web. After further investigation I've found out that list with custom activation criteria had `minVelocityX` and `minVelocityY`, but `minVelocity` was missing. This PR adds mentioned criterion.

## Test plan

Tested on example app (just add `.minVelocity()` in any example with `Pan`


j-piasecki (approved):
TIL: criterion is a singular form of criteria

PR #2901, opened 2024-05-09
Refactor `GestureDetector`

j-piasecki:
## Description

This PR refactors GestureDetector file:
- splits it into smaller files
- renames some variables
- removes redundant logic
- extracts common logic to utils

## Test plan

Test on the Example app


src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -0,0 +1,159 @@
...
+      console.error(
+        tagMessage(
+          `Some of the callbacks in the gesture are worklets and some are not. Either make sure that all calbacks are marked as 'worklet' if you wish to run them on the UI thread or use '.runOnJS(true)' modifier on the gesture explicitly to run all callbacks on the JS thread.`
+        )
+      );
+    }
+
+    const areAllNotWorklets = !areSomeWorklets && areSomeNotWorklets;
+    // if none of the callbacks are worklets and the gesture is not explicitly marked with
+    // `.runOnJS(true)` show an error
```
m-bert, line 88:
```suggestion
    // `.runOnJS(true)` show a warning
```

src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  return {
+    waitFor: requireToFail,
+    simultaneousHandlers: simultaneousWith,
+    blocksHandlers: blocksHandlers,
+  };
+}
+
+export function checkGestureCallbacksForWorklets(gesture: GestureType) {
+  if (__DEV__) {
```
m-bert, line 66:
```suggestion
  if (!__DEV__) {
    return;
  }
```

src/handlers/gestures/GestureDetector/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+  useEffect(() => {
+    if (!state.firstRender) {
+      onHandlersUpdate();
+    } else {
+      state.firstRender = false;
+    }
+  }, [props]);
+
+  const refFunction = (ref: unknown) => {
+    if (ref !== null) {
```
m-bert, line 233:
```suggestion
    if (ref === null) {
      return;
    }
```

src/handlers/gestures/GestureDetector/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+      // if it's the first render, also set the previousViewTag to prevent reattaching gestures when not needed
+      if (state.previousViewTag === -1) {
+        state.previousViewTag = findNodeHandle(state.viewRef) as number;
+      }
+
+      // pass true as `skipConfigUpdate`, here we only want to trigger the eventual reattaching of handlers
+      // in case the view has changed, while config update would be handled be the `useEffect` above
+      onHandlersUpdate(true);
+
+      if (__DEV__ && isFabric() && global.isFormsStackingContext) {
```
m-bert, line 246:
```suggestion
      if (!__DEV__ || !isFabric() || !global.isFormsStackingContext) {
        return;
      }
```

m-bert (commented):
Looks good!

src/handlers/gestures/GestureDetector/attachHandlers.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export function attachHandlers({
+  preparedGesture,
+  gestureConfig,
+  gestures,
+  viewTag,
+  webEventHandlersRef,
+  mountedRef,
+}: AttachHandlersConfig) {
+  gestureConfig.initialize();
```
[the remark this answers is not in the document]
j-piasecki: This was done in a weird way. The check you're mentioning was actually negated: https://github.com/software-mansion/react-native-gesture-handler/blob/81a7df0001774088f940b6ed45f2cbc9f2250de5/src/handlers/gestures/GestureDetector.tsx#L152-L156

so it would run every time but the first, then in the detector itself it was run only the first time:
https://github.com/software-mansion/react-native-gesture-handler/blob/81a7df0001774088f940b6ed45f2cbc9f2250de5/src/handlers/gestures/GestureDetector.tsx#L779-L781

Now the effect _should_ be the same, without the weird conditional gymnastics

src/handlers/gestures/GestureDetector/useAnimatedGesture.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    const currentCallback = sharedHandlersCallbacks.value;
+    if (!currentCallback) {
+      return;
+    }
+
+    for (let i = 0; i < currentCallback.length; i++) {
+      const gesture = currentCallback[i];
+
+      if (event.handlerTag === gesture.handlerTag) {
```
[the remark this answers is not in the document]
j-piasecki: At least for now, I wouldn't touch the state machine - this has the most potential to break stuff when changed and with other changes, it might be hard to pinpoint the problem exactly.

src/handlers/gestures/GestureDetector/index.tsx
j-piasecki: This one feels like too much of a microoptimization 😅. For my small brain, it's much easier to parse a conjunction than negated alternatives.

m-bert: I get it, let's leave that as it is now.

src/handlers/gestures/GestureDetector/useAnimatedGesture.ts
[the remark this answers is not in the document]
m-bert: Got it. I think we could at least get rid of this one by doing

```suggestion
      if (event.handlerTag !== gesture.handlerTag) {
        continue;
      }
```

I know it may look like an obsession right now, but I believe that reducing those nested braces increases readability of the code. What do you think? 😅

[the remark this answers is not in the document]
j-piasecki: Done in [bab8167](https://github.com/software-mansion/react-native-gesture-handler/pull/2901/commits/bab8167a43bba77ce5c844616fb6a20dbb4eef6c)

src/handlers/gestures/GestureDetector/useViewRefHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+  state: GestureDetectorState,
+  updateAttachedGestures: (skipConfigUpdate?: boolean) => void
+) {
+  const refHandler = useCallback(
+    (ref: React.Component | null) => {
+      if (ref === null) {
+        return;
+      }
+
+      validateDetectorChildren(ref);
```
m-bert, line 27: Is it necessary? As far as I can see it is extracted from [this](https://github.com/software-mansion/react-native-gesture-handler/blob/50d2dce0ef1e70c8baf956815bb156c7f08613f3/src/handlers/gestures/GestureDetector.tsx#L819), where there is no validation performed.

m-bert (approved):
I'm happy that this file got refactored 😄

src/handlers/gestures/GestureDetector/useViewRefHandler.ts
j-piasecki: You're right, it's not. The validation is already done in `updateAttachedHandlers`.

PR #2902, opened 2024-05-10
Change navbar and footer to a reusable component

patrycjakalinska:
In order to use navbar and footer theme components from `@swmansion/t-rex-ui`, we need to rewrite current navbar and footer to be more simplified. 

To achieve that we moved landing background from Navbar and footer svg `<Moon />` from Footer to main page component.

image

image


piaskowyk (approved):
🫡

PR #2903, opened 2024-05-10
Add `ref` property to `Buttons`

### LINKED ISSUE #2894: `RectButton`, `BorderlessButton` and `BaseButton` don't have support for View ref methods
### Description

When passing refs to any of the aforementioned components, none of `react-native`'s `View` methods are available, which means it is not possible to call `measure` or `measureLayout` for these components.

### Steps to reproduce

1. Pass `ref` down to `RectButton`, `BorderlessButton` or `BaseButton`
2. Try to call `ref.current.measure()`

https://snack.expo.dev/@tonihm96/react-native-gesture-handler-example

### Gesture Handler version

2.14.0

### React Native version

0.73.6

### Platforms

Android, iOS, Web, MacOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

Pixel 2 XL UpsideDownCake

### Acknowledgements

Yes

m-bert:
## Description

Currently our components (`BaseButton`, `RectButton` and `BorderlessButton`) don't support `refs`, so it is impossible to use methods like `measure`.

This PR adds wrapper to these components, so that they are now exported as `ForwardRef`.

Fixes #2894

## Test plan

<details>
<summary> Tested on slightly modified code from issue </summary>

```jsx
import { Text, StyleSheet } from 'react-native';

import {
  BaseButton,
  BorderlessButton,
  GestureHandlerRootView,
  RectButton,

  const rectButtonRef = useRef(null);
  const borderlessButtonRef = useRef(null);
  const baseButtonRef = useRef(null);

  const handlePress = () => {
    try {
      baseButtonRef.current?.measure?.((x, y, width, height) => {
        console.log('baseButtonRef', x, y, width, height);
      });

      rectButtonRef.current?.measure?.((x, y) => {
        console.log('rectButtonRef', x, y);
      });

      borderlessButtonRef.current?.measure?.((x, y) => {
        console.log('borderlessButtonRef', x, y);
      });
    } catch (e) {
      console.error(e);
    }
  };

  return (
      <RectButton onPress={handlePress} style={styles.button}>
        <Text style={styles.text}>Press me</Text>
      </RectButton>

      <BaseButton ref={baseButtonRef} style={styles.button}>
        <Text style={styles.text}>Test</Text>
      </BaseButton>
      <BorderlessButton ref={borderlessButtonRef} style={styles.button}>
        <Text style={styles.text}>Test</Text>
      </BorderlessButton>
      <RectButton ref={rectButtonRef} style={styles.button}>
        <Text style={styles.text}>Test</Text>
      </RectButton>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    gap: 20,
  },
  button: {
    justifyContent: 'center',
    alignItems: 'center',
    borderRadius: 5,
    backgroundColor: 'grey',
    paddingHorizontal: 20,
    paddingVertical: 10,
  },
  text: {
    color: 'white',
  },
});
```

</details>


src/components/GestureButtons.tsx
```diff
@@ -92,6 +92,11 @@ export interface BaseButtonProps extends RawButtonProps {
    * Defaults to 600.
    */
   delayLongPress?: number;
+
+  /**
+   * Used to pass ref from parent into `BaseButton`
+   */
+  innerRef?: React.ForwardedRef<unknown>;
```
j-piasecki, line 99: Why not `React.ForwardedRef<React.ComponentType<any>>` since it's being cast anyway?

Or, if that's not possible, simply `React.ForwardedRef<any>`, since there is `any` in the casted type anyway?

j-piasecki, line 99: Also, you mentioned not wanting `innerRef` to be accessible, I think we are exporting this type publicly.

m-bert: Makes sense, done in [fec6943](https://github.com/software-mansion/react-native-gesture-handler/pull/2903/commits/fec69434e1643e874c85b2122948fe480991e0c9)

m-bert: Yup, I've missed that. I've changed that in [cf193d1](https://github.com/software-mansion/react-native-gesture-handler/pull/2903/commits/cf193d1a528060fed55418c9c06b8fdac20cb567), though I'm not sure if this is the best way to do this. Maybe there are some TypeScript tricks that we could use to make it look better, cc @tjzel

tjzel: I can't really think of a solution here that wouldn't need any extensive changes sadly :c

j-piasecki (approved):
Assuming that this doesn't break types, it looks ok to me.

PR #2911, opened 2024-05-14
docs: add @swmansion/t-rex-ui

patrycjakalinska:
## Description

Because of a ramping up maintainance cost across our suite of tools - Reanimated, Gesture Handler & React Native IDE we've decided to create a UI library for our reusable docusaurus components.

The [`@swmansion/t-rex-ui`](https://github.com/software-mansion-labs/t-rex-ui) package aims to unify the components across our documentations. This will make the maintainance easier in the long term by code reuse. We hope that with a single source of truth will avoid the need of fixing a bug in 3-4 places instead of just one.

## Test plan

```
cd docs/
yarn

# to test development build
yarn start

# to test production build
yarn build
yarn serve
```

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -29,6 +29,7 @@
     "@emotion/styled": "^11.10.6",
     "@mdx-js/react": "^1.6.22",
     "@mui/material": "^5.12.0",
+    "@swmansion/t-rex-ui": "^0.0.3",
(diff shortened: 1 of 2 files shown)


kacperkapusciak (approved):
https://www.youtube.com/watch?v=FGBhQbmPwH8

PR #2914, opened 2024-05-16
Fix Detector creating and attaching all gestures twice on first mount

j-piasecki:
## Description

After refactoring, the Detector would create and attach all gestures twice on the first mount. This made gestures effectively not work on Android, since two gestures would be attached to the view. JS side would listen to the events of only the second one, while on the native side, the second one would always be canceled by the second one.

Note that the gestures don't work when the children of the detector change, which is probably caused by the Reanimated upgrade with the new example app.

## Test plan

Check any of the new API examples on Android

### DIFF
--- a/src/handlers/gestures/GestureDetector/index.tsx
+++ b/src/handlers/gestures/GestureDetector/index.tsx
@@ -164,10 +164,10 @@ export const GestureDetector = (props: GestureDetectorProps) => {
   }, []);
 
   useEffect(() => {
-    if (!state.firstRender) {
-      updateAttachedGestures();
-    } else {
+    if (state.firstRender) {
       state.firstRender = false;
+    } else {
+      updateAttachedGestures();
     }
   }, [props]);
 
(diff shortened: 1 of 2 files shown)


src/handlers/gestures/GestureDetector/index.tsx
```diff
@@ -164,10 +164,10 @@ export const GestureDetector = (props: GestureDetectorProps) => {
   }, []);
 
   useEffect(() => {
-    if (!state.firstRender) {
-      updateAttachedGestures();
-    } else {
+    if (state.firstRender) {
       state.firstRender = false;
+    } else {
+      updateAttachedGestures();
```
m-bert, line 170: I'm a bit confused, isn't it the same as before? 😅

j-piasecki: Yes, but the logic isn't inverted, so it should be (at least is for me 😅) easier to understand at a glance.

PR #2916, opened 2024-05-17
Remove lodash

### LINKED ISSUE #2613: Remove lodash
### Description

As far as I can tell lodash is used [once](https://github.com/software-mansion/react-native-gesture-handler/blob/a9272cf2471bd869a3c81e13fa114ecfea3cc835/src/handlers/createHandler.tsx#L391C13-L391C14) anywhere in the repo and could either be replaced with [lodash.isEqual](https://www.npmjs.com/package/lodash.isequal) or native JS.

### Steps to reproduce

N/A

https://github.com/software-mansion/react-native-gesture-handler/blob/a9272cf2471bd869a3c81e13fa114ecfea3cc835/src/handlers/createHandler.tsx#L391C13-L391C14

### Gesture Handler version

2.13.1

### React Native version

N/A

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

Since `lodash` is used only in one place in our codebase, we can simply replace it with our own implementation on `deepEqual`.

Closes #2613 

## Test plan

Run example apps.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -60,6 +60,15 @@ export function isRemoteDebuggingEnabled(): boolean {
   );
 }
 
+/**
+ * Recursively compares two objects for deep equality.
+ *
+ * **Note:** This function does not support cyclic references.
+ *
+ * @param obj1 - The first object to compare.
+ * @param obj2 - The second object to compare.
+ * @returns `true` if the objects are deeply equal, `false` otherwise.
+ */
 export function deepEqual(obj1: any, obj2: any) {
   if (obj1 === obj2) {
     return true;
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -61,7 +61,9 @@ export function isRemoteDebuggingEnabled(): boolean {
 }
 
 export function deepEqual(obj1: any, obj2: any) {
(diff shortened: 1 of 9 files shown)


src/utils.ts
```diff
@@ -59,3 +59,33 @@ export function isRemoteDebuggingEnabled(): boolean {
...
+    return true;
+  }
+
+  if (
+    typeof obj1 !== 'object' ||
+    typeof obj2 !== 'object' ||
+    obj1 === null ||
+    obj2 === null
+  ) {
+    return obj1 === obj2;
```
j-piasecki, line 74: This branch always returns `false`, no?

```suggestion
    return false;
```

m-bert: I'm pretty sure you're right.

j-piasecki (approved):
I'm pretty sure this won't work with cyclical references. We don't really need that at the moment since this is only used to compare gesture configs, but it would be nice to add a comment about that in case it's used somewhere else in the future.

PR #2919, opened 2024-05-17
Update examples

j-piasecki:
## Description

- moves `New Api` section to the top
- unifies white-ish backgrounds across examples
- unifies `Velocity Test` with `Simple Reanimated example`
- unifies `Overlap Parent` and `OverlapSiblings`
- updates `Transformations`
- updates `Camera` since the camera was missing there 
- updates `Pan Responder` to not use `setNativeProps` which was removed on RN web

## Test plan

Check the updated examples


example/App.tsx
```diff
@@ -68,6 +66,7 @@ import VelocityTest from './src/new_api/velocityTest';
 import EmptyExample from './src/empty/EmptyExample';
 import RectButtonBorders from './src/release_tests/rectButton';
 import { ListWithHeader } from './src/ListWithHeader';
+import { COLORS } from 'src/common';
```
m-bert, line 69:
```suggestion
import { COLORS } from './src/common';
```

example/src/ListWithHeader/Header.tsx
```diff
@@ -11,6 +11,7 @@ import Animated, {
   useSharedValue,
   withTiming,
 } from 'react-native-reanimated';
+import { COLORS } from 'src/common';
```
m-bert, line 14: Why not

```suggestion
import { COLORS } from '../common';
```

?

m-bert (changes requested):
Good job! I really like new look of our app :heart:

I've left some suggestions, let me know what do you think about them.

Also there's one problem - `transformations` example no longer works on web. Could you please check that?

m-bert:
Also, one more detail

> - updates Pan Responder to not use setNativeProps which was removed on RN web

I think that combo example [still uses `setNativeProps`](https://github.com/software-mansion/react-native-gesture-handler/blob/921a9b2af4579021c5ca4ff40108e208683435b5/example/src/release_tests/combo/index.tsx#L70). Do we plan to remove that?

j-piasecki:
Most likely yes.

example/src/ListWithHeader/Header.tsx
j-piasecki: VSCode does weird things when auto-importing

example/src/new_api/camera/AnimatedCameraView.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const animatedProps = useAnimatedProps(() => {
+      return {
+        zoom: props.zoom.value - 1,
+      };
+    });
+
+    if (!permission) {
+      return <View />;
+    }
```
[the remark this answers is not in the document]
j-piasecki: That's the same approach as in the docs: https://docs.expo.dev/versions/latest/sdk/camera/#usage. Looks like it returns null only when the permission isn't yet loaded.

example/src/new_api/velocityTest/index.tsx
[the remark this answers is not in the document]
j-piasecki: Done in [e3919c2](https://github.com/software-mansion/react-native-gesture-handler/pull/2919/commits/e3919c24adf8fcce8c209ac22a26f66f41111e0a), thanks for the snippet

j-piasecki:
> Also there's one problem - transformations example no longer works on web. Could you please check that?

It seems like passing the matrix as transform directly didn't work on web. Now it applies styles, but they are not right. It looks like on web `focal` and `anchor` points (for pinch and rotation gestures respectively) aren't in the coordinate space of the view like on Android and iOS. Could you check that? And should I open an issue for this or do you want to?

m-bert:
> And should I open an issue for this or do you want to?

You can open an issue. I will look into it later 😅

j-piasecki:
I've moved the complex transform logic to the separate PR so this stops being blocked by wrong anchor/focal points on web - https://github.com/software-mansion/react-native-gesture-handler/pull/3124

example/app.json
```diff
@@ -28,6 +28,14 @@
       "favicon": "./assets/favicon.png"
     },
     "plugins": [
+      [
+        "expo-camera",
+        {
+          "cameraPermission": "Allow RNGH example to access your camera",
+          "microphonePermission": "Allow RNGH example to access your microphone",
+          "recordAudioAndroid": true
```
m-bert, line 36: Which example needs audio access? 😅

m-bert (commented):
> I've moved the complex transform logic to the separate PR so this stops being blocked by wrong anchor/focal points on web - https://github.com/software-mansion/react-native-gesture-handler/pull/3124

So I think we can remove matrices from description:

> updates Transformations to use matrices and update the look

example/app.json
j-piasecki: I think it's for the camera

m-bert: As far as I can see, camera works without those two 😅

j-piasecki: I probably copied it from the docs: https://docs.expo.dev/versions/latest/sdk/camera/#example-appjson-with-config-plugin.

I assume the app doesn't ask for microphone permissions?

m-bert: Yup, it doesn't. Also I've removed those permissions from `app.json` and it works fine (though I have not looked into other examples, I just assumed that they don't need that 😅).

m-bert (approved):

PR #2921, opened 2024-05-21
Fix GestureDetector not working when its children change

j-piasecki:
## Description

On web `findNodeHandle` was returning the ref it received as an argument, which in the case of GestureDetector was a reference to the Wrap component. This worked fine until the children of the wrap changed (but the wrap itself didn't, which caused the handlers not to be reattached), in which case the new view didn't have event listeners added and gestures didn't work.

This PR changes the used method to always call `findNodeHandle` which returns the reference to the underlying DOM element, and view change detection works correctly in this case.

## Test plan

<details>
<summary>Tested on the example app and on the following code:</summary>

```jsx
import { Button, StyleSheet, View } from 'react-native';
import { GestureDetector, Gesture } from 'react-native-gesture-handler';
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withSpring,

function Ball(props) {
  const isPressed = useSharedValue(false);
  const offset = useSharedValue({ x: 0, y: 0 });

  const animatedStyles = useAnimatedStyle(() => {
    return {
      transform: [
        { translateX: offset.value.x },
        { translateY: offset.value.y },
        { scale: withSpring(isPressed.value ? 1.2 : 1) },
      ],
      backgroundColor: isPressed.value ? 'yellow' : 'blue',
    };
  });

  const gesture = Gesture.Pan()
    .onBegin(() => {
      'worklet';
      isPressed.value = true;
    })
    .onChange((e) => {
      'worklet';
      offset.value = {
        x: e.changeX + offset.value.x,
        y: e.changeY + offset.value.y,
      };
    })
    .onFinalize(() => {
      'worklet';
      isPressed.value = false;
    });

  return (
      <Animated.View style={[styles.ball, animatedStyles]} key={props.counter} />
    </GestureDetector>
  );
}

  const [counter, setCounter] = useState(0);

  return (
      <Button title="Remount" onPress={() => setCounter((prev) => prev + 1)} />
      <Ball counter={counter} />
    </View>
  );
}

  container: {
    flex: 1,
  },
  ball: {
    width: 100,
    height: 100,
    borderRadius: 100,
    backgroundColor: 'blue',
    alignSelf: 'center',
  },
});
```
</details>

### DIFF
--- a/src/handlers/gestures/GestureDetector/index.tsx
+++ b/src/handlers/gestures/GestureDetector/index.tsx
@@ -1,14 +1,10 @@
 /* eslint-disable react/no-unused-prop-types */
 import React, { useContext, useEffect, useMemo, useRef } from 'react';
+import { Platform, findNodeHandle } from 'react-native';
 import { GestureType } from '../gesture';
-import {
-  findNodeHandle,
-  UserSelect,
-  TouchAction,
-} from '../../gestureHandlerCommon';
+import { UserSelect, TouchAction } from '../../gestureHandlerCommon';
 import { ComposedGesture } from '../gestureComposition';
 import { isJestEnv } from '../../../utils';
-import { Platform } from 'react-native';
 
 import GestureHandlerRootViewContext from '../../../GestureHandlerRootViewContext';
 import { AttachedGestureState, GestureDetectorState } from './types';
--- a/src/handlers/gestures/GestureDetector/useDetectorUpdater.ts
+++ b/src/handlers/gestures/GestureDetector/useDetectorUpdater.ts
@@ -1,6 +1,6 @@
 import React, { useCallback } from 'react';
+import { findNodeHandle } from 'react-native';
 import { GestureType } from '../gesture';
-import { findNodeHandle } from '../../gestureHandlerCommon';
 import { ComposedGesture } from '../gestureComposition';
 
 import {
--- a/src/handlers/gestures/GestureDetector/useViewRefHandler.ts
+++ b/src/handlers/gestures/GestureDetector/useViewRefHandler.ts
@@ -1,10 +1,9 @@
-import { findNodeHandle } from '../../gestureHandlerCommon';
-
 import { isFabric, tagMessage } from '../../../utils';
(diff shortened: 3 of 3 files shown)


m-bert (commented):
I'm not so happy about using `findNodeHandle` - afaik it uses `findDOMNode` under the hood, which is marked as deprecated and [it is meant to be removed in future major version of React](https://react.dev/reference/react-dom/findDOMNode). We already started removing it on web in `Reanimated`.

However, I think we can merge this now and later on we can think of how we can deal with this problem, if returning `node` is not enough.

Also, shouldn't we replace all occurrences of `findNodeHandle`? Right now we still have places where we use the one from `gestureHandlerCommon`, is it necessary?

j-piasecki:
I agree about removing it, though it's fixing the problem and we're already using it anyway (in the delegate). 

> Also, shouldn't we replace all occurrences of findNodeHandle? Right now we still have places where we use the one from gestureHandlerCommon, is it necessary?

I think all references in the code related to the new API is updated. I'd rather not touch the old API as to not break anything accidentally, especially at the point where I don't know why the custom `findNodeHandle` was working like that in the first place.

m-bert:
>I think all references in the code related to the new API is updated.

I think [this one](https://github.com/software-mansion/react-native-gesture-handler/blob/0a10f2206274ffff012b463bb0392854105936f0/src/handlers/gestures/GestureDetector/index.tsx#L5) was not updated.

> I don't know why the custom `findNodeHandle` was working like that in the first place.

Well, I'm not sure why it was implemented this way, but I do agree that on web `findNodeHandle` acts almost like an identity function (though I've seen some cases where it doesn't, as far as I remember the problem was with `FlatList`, but I'm not sure).

j-piasecki:
> I think [this one](https://github.com/software-mansion/react-native-gesture-handler/blob/0a10f2206274ffff012b463bb0392854105936f0/src/handlers/gestures/GestureDetector/index.tsx#L5) was not updated.

Fixed

m-bert (approved):
LGTM

PR #2923, opened 2024-05-22
Unify touch events callbacks on `web` with respect to `Android`

### LINKED ISSUE #2907: WEB/NATIVE inconsistent behaviour of pan gesture callbacks with manual activation.
### Description

In the `onTouchesDown` callback, `stateManager.fail()` is called. It works as expected in native, but in the web, callbacks `onTouchesMove`, `onTouchesUp` etc continue to trigger, which should not happen.

Run the code below, swipe your finger across the screen, and watch the console.

```
import { Gesture, GestureDetector, GestureHandlerRootView } from 'react-native-gesture-handler';

    .manualActivation(true)
    .onTouchesDown((e, stateManager) => {
      stateManager.fail();
      console.log('onTouchesDown, fail')
    })
    .onTouchesUp(() => console.log('onTouchesUp'))
    .onTouchesMove(() => console.log('onTouchesMove'))
    .onBegin(() => console.log('onBegin'))
    .onChange(() => console.log('onChange'))
    .onTouchesCancelled(() => console.log('onTouchesCancelled'))
    .onFinalize(() => console.log('onFinalize'))
    .onStart(() => console.log('onStart'))
    .onUpdate(() => console.log('onUpdate'))
    .onEnd(() => console.log('onEnd'));
	
  return (
    <GestureHandlerRootView style={{ flex: 1 }}>
        <View style={{ flex: 1, backgroundColor: 'yellow' }} />
      </GestureDetector>
    </GestureHandlerRootView>
  )
}
```

In the web, you will see the following output:
```
onTouchesDown, fail
onBegin
onTouchesMove
onTouchesUp
onFinalize
```
In the native, you will see the following output:
```
onBegin
onTouchesCancelled
onFinalize
onTouchesDown, fail

```

It seems like this is an [issue](https://github.com/software-mansion/react-native-gesture-handler/issues/2868)  that hasn't been fully resolved #2869.

@m-bert what do you think?

### Steps to reproduce

try it on Snack

https://snack.expo.dev/DPDgCkZ0hbai5s51p3AcM

### Gesture Handler version

2.16.2

### React Native version

0.71.9

### Platforms

Android, iOS, Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

As stated in #2907, web shows inconsistent behavior with native platforms when it comes to touch event callbacks. This PR unifies its behavior with `Android` by calling `sendTouchEvent` after handler goes to `BEGAN` state. Previously it wouldn't fail, because transition from `UNDETERMINED` into `FAIL` is not possible on web.

Fixes #2907.

### `onBegin` and `onFinalize`

We agreed with @j-piasecki that `onFinalize` should be called only when `onBegin` was called. However, in that specific case (i.e. snippet from issue) we are not sure if those callbacks should be triggered and it is still a thing for a discussion.

## Test plan

<details>
<summary> Tested on the following code from issue </summary>

```jsx
import { Gesture, GestureDetector, GestureHandlerRootView } from 'react-native-gesture-handler';

    .manualActivation(true)
    .onTouchesDown((e, stateManager) => {
      stateManager.fail();
      console.log('onTouchesDown, fail')
    })
    .onTouchesUp(() => console.log('onTouchesUp'))
    .onTouchesMove(() => console.log('onTouchesMove'))
    .onBegin(() => console.log('onBegin'))
    .onChange(() => console.log('onChange'))
    .onTouchesCancelled(() => console.log('onTouchesCancelled'))
    .onFinalize(() => console.log('onFinalize'))
    .onStart(() => console.log('onStart'))
    .onUpdate(() => console.log('onUpdate'))
    .onEnd(() => console.log('onEnd'));
	
  return (
    <GestureHandlerRootView style={{ flex: 1 }}>
        <View style={{ flex: 1, backgroundColor: 'yellow' }} />
      </GestureDetector>
    </GestureHandlerRootView>
  )
}
```

</details>


src/web/handlers/GestureHandler.ts
```diff
@@ -288,33 +288,20 @@ export default abstract class GestureHandler implements IGestureHandler {
     if (this.pointerType === PointerType.TOUCH) {
       GestureHandlerOrchestrator.getInstance().cancelMouseAndPenGestures(this);
     }
-
-    if (this.config.needsPointerData) {
-      this.sendTouchEvent(event);
-    }
```
j-piasecki, line 294: Why aren't we sending events in onPointerDown anymore?

src/web/handlers/GestureHandler.ts
```diff
@@ -329,43 +316,42 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+  private tryToSendMoveEvent(out: boolean, event: AdaptedEvent): void {
     if (
       this.enabled &&
       this.active &&
       (!out || (out && !this.shouldCancelWhenOutside))
     ) {
       this.sendEvent(this.currentState, this.currentState);
     }
+
+    this.tryToSendTouchEvent(event);
```
j-piasecki, line 348: Shouldn't this be handled by `onPointerMove`?

src/web/handlers/FlingGestureHandler.ts
```diff
@@ -114,6 +114,8 @@ export default class FlingGestureHandler extends GestureHandler {
 
     super.onPointerDown(event);
     this.newPointerAction();
+
+    this.tryToSendTouchEvent(event);
```
j-piasecki, line 118: Why not `super.onPointerDown(event)`?

src/web/handlers/GestureHandler.ts
m-bert: What I wanted to achieve was to send touch event after gesture had opportunity to begin. All `super.onPointerDown` are called at the beginning of overriden methods because parent class does some additional steps (like adding handler to `orchestrator`). The easiest way to get this flow was to remove sending touch event in parent class and call it at the end of child's `onPointerDown`.

src/web/handlers/GestureHandler.ts
m-bert: I think it is fine to leave it here. This method is called `tryToSendMoveEvent`, so sending `touchmove` seems ok. Also note that we have two cases - `onPointerMove` and `onPointerOutOfBounds`, so instead of calling `tryToSendTouchEvent` twice, we can leave it here and it will cover both cases.

m-bert: If there's something that could be changed here, I would do something like this:

```jsx

  private tryToSendMoveEvent(out: boolean, event: AdaptedEvent): void {
    if (!out || (out && !this.shouldCancelWhenOutside)) {
      if (this.enabled && this.active) {
        this.sendEvent(this.currentState, this.currentState);
      }

      this.tryToSendTouchEvent(event);
    }
  }

```

also if we simplify the first condition to make an early return we get something that is much easier to read:

```jsx

...
    if (out && this.shouldCancelWhenOutside) {
      return;
    }
...

```

This way we could send touch events only when handler config allows it.

src/web/handlers/FlingGestureHandler.ts
m-bert: Because we don't send touch events in parent class anymore 😅

m-bert: Also, there is `super.onPointerDown(event)` visible on that suggestion (line 115) 😅

src/web/handlers/GestureHandler.ts
j-piasecki: In this case, it may be worth to consider refactoring it so that `onPointerDown` in the base `GestureHandler` class looks something like this:

```
onPointerDown(event) {
  // do the common stuff

  this.handleDownEvent(event);
  this.tryToSendTouchEvent(event);
}
```

and move the handler specific stuff to `handleDownEvent` method.

It's outside the scope of this PR though, but if you agree a TODO in code would be great so it's not lost to time 😅.

src/web/handlers/GestureHandler.ts
j-piasecki: Actually, I think that `this.enabled` should also be moved to the check doing early return - we shouldn't be sending any events do disabled gestures.

m-bert: Yes, I agree. Done in [9f44240](https://github.com/software-mansion/react-native-gesture-handler/pull/2923/commits/9f44240383864e8296ff367d871c49ff3b475f6a)

src/web/handlers/GestureHandler.ts
m-bert: Sounds good, left a `TODO` in [f2ac147](https://github.com/software-mansion/react-native-gesture-handler/pull/2923/commits/f2ac1471b9c9c25396246d8db0debbb4787358c7)

j-piasecki (approved):

PR #2928, opened 2024-06-01
chore(types): add missing pointerType to GestureTouchEvent

mgcrea:
## Description

Using `Gesture.Tap().onTouchesDown((event) => {})` event is a `GestureTouchEvent` that is missing the `pointerType` key while it does exist in the actual object.


j-piasecki:
I don't think that `pointerType` is implemented on Android. I'll try to come back to this soon.

j-piasecki (approved):
Thanks for bringing it up!

Issue #2929, opened 2024-06-03
Focal and anchor points are not in the coordinate space of the view on web


j-piasecki:
### Description

On Android and iOS the focal (for pinch) and anchor (for rotation) points are always in the coordinate space of the view, while on web they are not. It looks like they are in the window coordinate space, though I didn't investigate it that thoroughly.

https://github.com/software-mansion/react-native-gesture-handler/assets/21055725/78632cfc-24f2-43e1-8f8e-afd8a0e4b0ae

### Steps to reproduce

```jsx
import React from 'react';
import { StyleSheet, View } from 'react-native';
import { Gesture, GestureDetector } from 'react-native-gesture-handler';
import Animated, {
  useAnimatedStyle,
  useSharedValue,
} from 'react-native-reanimated';

function Pointer(props: { x: number; y: number }) {
  return (
    <View
      style={{
        position: 'absolute',
        left: props.x,
        top: props.y,
        width: 16,
        height: 16,
        borderRadius: 8,
        backgroundColor: 'red',
        transform: [{ translateX: -8 }, { translateY: -8 }],
      }}
    />
  );
}

export default function EmptyExample() {
  const translation = useSharedValue({ x: 0, y: 0 });

  const style = useAnimatedStyle(() => {
    return {
      transform: [
        { translateX: translation.value.x },
        { translateY: translation.value.y },
      ],
    };
  });

  const [pointerPos, setPointerPos] = React.useState({ x: 100, y: 100 });
  const [pointerVisible, setPointerVisible] = React.useState(false);

  const pan = Gesture.Pan()
    .averageTouches(true)
    .onChange((e) => {
      translation.value = {
        x: translation.value.x + e.changeX,
        y: translation.value.y + e.changeY,
      };
    });

  const pinch = Gesture.Pinch()
    .onStart((e) => {
      setPointerVisible(true);
      setPointerPos({ x: e.focalX, y: e.focalY });
    })
    .onEnd(() => {
      setPointerVisible(false);
    })
    .runOnJS(true);

  const rotation = Gesture.Rotation()
    .onStart((e) => {
      setPointerVisible(true);
      setPointerPos({ x: e.anchorX, y: e.anchorY });
    })
    .onEnd(() => {
      setPointerVisible(false);
    })
    .runOnJS(true);

  return (
    <View style={styles.container}>
      <GestureDetector gesture={Gesture.Simultaneous(pan, rotation)}>
        <Animated.View
[...]

PR #2931, opened 2024-06-03
`PointerTracker` refactor.

m-bert:
## Description

Fixing #2929 required adding additional information to `TrackerElement`. If we were to leave current structure of `PointerTracker`, we would have to add few more methods. After refactoring, there are only 2 that were not present before:

- `getLastViewRelativeCoords`
- `getViewRelativeCoordsSum`

I'd like to make similar changes to our handlers in the future (i.e. use `coords` object instead of `lastX` and `lastY` variables).

## Test plan

Tested on example app.

### DIFF
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -405,10 +405,9 @@ export default abstract class GestureHandler implements IGestureHandler {
       nativeEvent: {
         numberOfPointers: this.tracker.getTrackedPointersCount(),
         state: newState,
-        pointerInside: this.delegate.isPointerInBounds({
-          x: this.tracker.getLastAvgX(),
-          y: this.tracker.getLastAvgY(),
-        }),
+        pointerInside: this.delegate.isPointerInBounds(
+          this.tracker.getAbsoluteCoordsAverage()
+        ),
         ...this.transformNativeEvent(),
         handlerTag: this.handlerTag,
         target: this.viewRef,
@@ -442,10 +441,10 @@ export default abstract class GestureHandler implements IGestureHandler {
 
       all.push({
         id: id,
-        x: element.lastX - rect.pageX,
-        y: element.lastY - rect.pageY,
-        absoluteX: element.lastX,
-        absoluteY: element.lastY,
+        x: element.abosoluteCoords.x - rect.pageX,
+        y: element.abosoluteCoords.y - rect.pageY,
+        absoluteX: element.abosoluteCoords.x,
+        absoluteY: element.abosoluteCoords.y,
       });
     });
 
@@ -465,10 +464,10 @@ export default abstract class GestureHandler implements IGestureHandler {
 
         changed.push({
           id: id,
-          x: element.lastX - rect.pageX,
-          y: element.lastY - rect.pageY,
-          absoluteX: element.lastX,
-          absoluteY: element.lastY,
+          x: element.abosoluteCoords.x - rect.pageX,
+          y: element.abosoluteCoords.y - rect.pageY,
+          absoluteX: element.abosoluteCoords.x,
+          absoluteY: element.abosoluteCoords.y,
         });
       });
     }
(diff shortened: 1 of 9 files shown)


src/web/handlers/GestureHandler.ts
```diff
@@ -534,18 +533,18 @@ export default abstract class GestureHandler implements IGestureHandler {
...
       });
 
       changed.push({
         id: id,
-        x: element.lastX - rect.pageX,
-        y: element.lastY - rect.pageY,
-        absoluteX: element.lastX,
-        absoluteY: element.lastY,
+        x: element.abosoluteCoords.x - rect.pageX,
+        y: element.abosoluteCoords.y - rect.pageY,
```
j-piasecki, line 545: Same

src/web/handlers/GestureHandler.ts
```diff
@@ -534,18 +533,18 @@ export default abstract class GestureHandler implements IGestureHandler {
 
       all.push({
         id: id,
-        x: element.lastX - rect.pageX,
-        y: element.lastY - rect.pageY,
-        absoluteX: element.lastX,
-        absoluteY: element.lastY,
+        x: element.abosoluteCoords.x - rect.pageX,
+        y: element.abosoluteCoords.y - rect.pageY,
```
j-piasecki, line 537: Same

src/web/handlers/GestureHandler.ts
```diff
@@ -465,10 +464,10 @@ export default abstract class GestureHandler implements IGestureHandler {
 
         changed.push({
           id: id,
-          x: element.lastX - rect.pageX,
-          y: element.lastY - rect.pageY,
-          absoluteX: element.lastX,
-          absoluteY: element.lastY,
+          x: element.abosoluteCoords.x - rect.pageX,
+          y: element.abosoluteCoords.y - rect.pageY,
```
j-piasecki, line 468: Same

src/web/handlers/GestureHandler.ts
```diff
@@ -442,10 +441,10 @@ export default abstract class GestureHandler implements IGestureHandler {
 
       all.push({
         id: id,
-        x: element.lastX - rect.pageX,
-        y: element.lastY - rect.pageY,
-        absoluteX: element.lastX,
-        absoluteY: element.lastY,
+        x: element.abosoluteCoords.x - rect.pageX,
+        y: element.abosoluteCoords.y - rect.pageY,
```
j-piasecki, line 445: Same

src/web/tools/PointerTracker.ts
```diff
@@ -1,12 +1,10 @@
...
 import VelocityTracker from './VelocityTracker';
 
 export interface TrackerElement {
-  lastX: number;
-  lastY: number;
-
-  timeStamp: number;
-
+  abosoluteCoords: Point;
+  viewRelativeCoords: Point;
```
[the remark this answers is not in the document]
m-bert: I agree 👍 

Done in [010f18d](https://github.com/software-mansion/react-native-gesture-handler/pull/2931/commits/010f18d5d7c59f312f7312f5146fd0664d436129)

src/web/tools/PointerTracker.ts
```diff
@@ -121,92 +110,93 @@ export default class PointerTracker {
...
-  /**
-   *
-   * @param pointerId
-   * Returns Y coordinate of given pointer
-   */
-  // eslint-disable-next-line @typescript-eslint/unified-signatures
-  public getLastY(pointerId: number): number;
-
-  public getLastY(pointerId?: number): number {
+  public getLastViewRelativeCoords(pointerId?: number) {
```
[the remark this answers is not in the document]
m-bert: Done in [010f18d](https://github.com/software-mansion/react-native-gesture-handler/pull/2931/commits/010f18d5d7c59f312f7312f5146fd0664d436129)

src/web/tools/PointerTracker.ts
```diff
@@ -121,92 +110,93 @@ export default class PointerTracker {
...
       }
     });
 
-    return sumX;
+    return sum;
   }
-  public getSumY(ignoredPointer?: number): number {
-    let sumY = 0;
+
+  public getViewRelativeCoordsSum(ignoredPointer?: number) {
```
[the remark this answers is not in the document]
m-bert: Done in [010f18d](https://github.com/software-mansion/react-native-gesture-handler/pull/2931/commits/010f18d5d7c59f312f7312f5146fd0664d436129)

m-bert:
@j-piasecki it's true that `lastCoords.x - rect.pageX` looks like `relativeCoords`. In most cases it is. The only problem is that when sending events from `onPointerUp` method those values will differ, because `tracker` doesn't have information about pointer anymore. I've managed to make it work, but it requires few changes in handlers' logic and I think it is out of scope of this PR. 

Given that, I think we can merge this and then look into changing those calculations into `relativeCoords` from `tracker`. What do you think?

j-piasecki:
> Given that, I think we can merge this and then look into changing those calculations into relativeCoords from tracker. What do you think?

Sounds good.

j-piasecki (approved):

PR #2934, opened 2024-06-05
Add swipeable example rewritten to new API

latekvo:
This PR reimplements the `Swipeable` component using `Reanimated 2` instead of `Animated`.


latekvo:
```ts
/**
 * A function called upon animation completion. If the animation is cancelled, the callback will receive `false` as the argument; otherwise, it will receive `true`.
 */
export type AnimationCallback = (finished?: boolean, current?: AnimatableValue) => void;
```
what's up with this function, why does it say `called upon animation completion` but then there are both `finished` and `current` arguments available, suggesting it is called every frame?

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,606 @@
...
+  openRight: () => void;
+  reset: () => void;
+}
+
+const Swipeable = forwardRef<
+  SwipeableProps & React.RefAttributes<ExposedFunctions>
+>(function Swipeable(
+  props: SwipeableProps,
+  ref: ForwardedRef<SwipeableProps & RefAttributes<ExposedFunctions>>
+) {
```
j-piasecki, line 234:
```suggestion
const Swipeable = forwardRef<ExposedFunctions, SwipeableProps>(
  function Swipeable(
    props: SwipeableProps,
    ref: ForwardedRef<ExposedFunctions>
  ) {
```

Ref methods and props are separate from each other. `forwardRef` allows you to specify both - the first generic type is for ref, the second one is for props.

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,606 @@
...
+
+  useEffect(() => {
+    if (!props.friction) props.friction = defaultProps.friction;
+    if (!props.overshootFriction)
+      props.overshootFriction = defaultProps.overshootFriction;
+    if (!props.useNativeAnimations)
+      props.useNativeAnimations = defaultProps.useNativeAnimations;
+
+    updateAnimatedEvent();
+  }, []);
```
j-piasecki, line 260: I don't think this is the right to do. Reassigning prop values is an anti-pattern, instead, you can use something like this:
```js
const { foo = 'default' } = props;
```

and either destructure the props at the top level and use those values in the code, or move default values to global constants and use them where necessary. Personally I'd lean more towards the first approach.

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,606 @@
...
+    }
+    return false;
+  };
+  */
+
+  const transX = useSharedValue(0); // only IV
+  const showLeftAction = useSharedValue(0); // can AV
+  const leftActionTranslate = useSharedValue(0); // only IV
+  const showRightAction = useSharedValue(0); // can AV;
+  const rightActionTranslate = useSharedValue(0); // only IV
```
j-piasecki, line 282: Please don't use acronyms in the comments.

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,606 @@
...
+    transX.value = withSpring(
+      toValue,
+      {
+        duration: 1000,
+        dampingRatio: 1.2,
+        stiffness: 500,
+        overshootClamping: true,
+        restDisplacementThreshold: 0.01,
+        restSpeedThreshold: 0.01,
+        reduceMotion: ReduceMotion.System,
```
j-piasecki, line 389: Do we actually need to set that? Seems like it could be the default value.

j-piasecki (commented):
Nice job so far. Here are a few comments:

example/src/new_api/swipeable/Swipeable.tsx
latekvo: Thank you, this fixed a lot of other related issues.

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,573 @@
...
+      props.friction !== props.friction ||
+      props.overshootLeft !== props.overshootLeft ||
+      props.overshootRight !== props.overshootRight ||
+      props.overshootFriction !== props.overshootFriction
+    ) {
+      updateAnimatedEvent(props);
+      return true;
+    }
+    return false;
+  };
```
latekvo, line 246: I think it would be good to deprecate this function in favour of `onUpdate` callback which would be provided by the user. This would be more aligned with how other react components work. Would that be alright? @j-piasecki

example/src/new_api/swipeable/Swipeable.tsx
latekvo: removed those all together, was only there to keep track of what happened in the old API

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,606 @@
...
+      <View
+        onLayout={({ nativeEvent }) =>
+          (rightOffset.value = nativeEvent.layout.x)
+        }
+      />
+    </Animated.View>
+  );
+
+  const panGesture = Gesture.Pan();
+  const tapGesture = Gesture.Tap();
```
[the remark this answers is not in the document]
latekvo: the browser is throwing warnings that the gesture callbacks should be explicitly workletized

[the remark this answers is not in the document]
latekvo: i dropped the 'worklet' directives, but are you sure about chaining the entire thing? I think it makes the code less readable

example/src/new_api/swipeable/Swipeable.tsx
latekvo: this was an internal function all along, my mistake

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,525 @@
...
+          duration: 1000,
+          dampingRatio: 1.2,
+          stiffness: 500,
+          velocity: velocityX,
+          ...props.animationOptions,
+        },
+        (isFinished) => {
+          if (isFinished) {
+            if (toValue > 0) {
+              runOnJS(() => props.onSwipeableOpen?.('left', swipeableMethods));
```
latekvo, line 324: I have to ask you whether you (c.c. @j-piasecki) think these frequent uses of `runOnJS` are ok, or if you'd prefer to have them replaced with something else.
I've had to place them to avoid crashes on android, which otherwise required every callback to be workletized.

latekvo:
## Current state of this component
https://github.com/software-mansion/react-native-gesture-handler/assets/74246391/bf762aac-308c-4164-a78b-45d8d260a04c

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,558 @@
...
+}
+
+const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
+  function Swipeable(
+    props: SwipeableProps,
+    ref: ForwardedRef<SwipeableMethods>
+  ) {
+    const rowState = useSharedValue<number>(0);
+
+    // this block has to be reduced, most of these values can be easily removed
```
j-piasecki, line 200: Not sure if this is up-to-date

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,558 @@
...
+        runOnJS(props.onSwipeableWillOpen)('left');
+      } else if (toValue < 0 && props.onSwipeableWillOpen) {
+        runOnJS(props.onSwipeableWillOpen)('right');
+      } else if (props.onSwipeableWillClose) {
+        const closingDirection = fromValue > 0 ? 'left' : 'right';
+        runOnJS(props.onSwipeableWillClose)(closingDirection);
+      }
+    };
+
+    const handleRelease = (
```
j-piasecki, line 346: This is used only in one place and can be moved much closer to where it's called (or even be inlined)

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,558 @@
...
+    const tapGesture = Gesture.Tap().onStart(() => {
+      if (rowState.value !== 0) {
+        close();
+      }
+    });
+
+    const panGesture = Gesture.Pan()
+      .onUpdate((event: GestureUpdateEvent<PanGestureHandlerEventPayload>) => {
+        // fixme: apply progress atop of drag
+        // fixme: apply offset from cursor to translation
```
j-piasecki, line 483: No sure if up-to-date

example/src/new_api/swipeable/Swipeable.tsx
latekvo: it's not anymore, i removed those that could have been

j-piasecki: Ok, the comment can be removed then

example/src/new_api/swipeable/Swipeable.tsx
```diff
@@ -0,0 +1,565 @@
...
+          animateRow(calculateCurrentOffset(), -rightWidth.value);
+        },
+        reset() {
+          'worklet';
+          userDrag.value = 0;
+          appliedTranslation.value = 0;
+          rowState.value = 0;
+        },
+      }),
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
j-piasecki, line 384: Is this necessary when you have dependencies below?

j-piasecki (approved):
🚀

example/src/new_api/swipeable/Swipeable.tsx
latekvo: not all are there, adding all of them would result in updating the useMemo every frame

PR #2936, opened 2024-06-10
Create a separate component out of the new Swipeable row

latekvo:
This PR takes configures `ReanimatedSwipeable` to be a part of the `gesture handler` package.


latekvo:
https://github.com/software-mansion/react-native-gesture-handler/assets/74246391/7d6a7a5a-47f0-4ea2-9ced-324e7a7ecb6f

latekvo:
This change has to be displayed in the docs, as for most people the reanimated component is likely preferable. Should we do it in this PR (c.c. @kacperkapusciak ), or push this one and open a new one for that?

latekvo:
found a new issue with `forwardRef`'d callbacks i didn't notice before, converting to draft

example/src/new_api/swipeable/AppleStyleSwipeableRow.tsx
```diff
@@ -8,7 +8,9 @@ import Animated, {
   interpolate,
   useAnimatedStyle,
 } from 'react-native-reanimated';
-import Swipeable, { SwipeableMethods } from 'src/new_api/swipeable/Swipeable';
+import Swipeable, {
+  SwipeableMethods,
+} from '../../../../src/components/ReanimatedSwipeable';
```
j-piasecki, line 13:
```suggestion
} from 'react-native-gesture-handler/ReanimatedSwipeable';
```

example/src/new_api/swipeable/GmailStyleSwipeableRow.tsx
```diff
@@ -8,7 +8,9 @@ import Animated, {
   interpolate,
   useAnimatedStyle,
 } from 'react-native-reanimated';
-import Swipeable, { SwipeableMethods } from 'src/new_api/swipeable/Swipeable';
+import Swipeable, {
+  SwipeableMethods,
+} from '../../../../src/components/ReanimatedSwipeable';
```
j-piasecki, line 13:
```suggestion
} from 'react-native-gesture-handler/ReanimatedSwipeable';
```

example/src/new_api/swipeable/AppleStyleSwipeableRow.tsx
```diff
@@ -135,16 +137,16 @@ export default function AppleStyleSwipeableRow({
...
-      renderRightActions={(_, progress) =>
+      renderRightActions={(_: any, progress: SharedValue<number>) =>
         renderRightActions(_, progress, swipeableRow)
       }
-      onSwipeableWillOpen={(direction) => {
+      onSwipeableWillOpen={(direction: any) => {
         console.log(`Opening swipeable from the ${direction}`);
       }}
-      onSwipeableWillClose={(direction) => {
+      onSwipeableWillClose={(direction: any) => {
```
j-piasecki, line 149: Why set the types explicitly here, shouldn't they be inferred?

And if they are necessary here, why `any`?

example/src/new_api/swipeable/GmailStyleSwipeableRow.tsx
```diff
@@ -82,10 +84,10 @@ export default function GmailStyleSwipeableRow({
       leftThreshold={80}
       enableTrackpadTwoFingerGesture
       rightThreshold={40}
-      renderLeftActions={(_, progress) =>
+      renderLeftActions={(_: any, progress: SharedValue<number>) =>
         renderLeftActions(_, progress, swipeableRow)
       }
-      renderRightActions={(_, progress) =>
+      renderRightActions={(_: any, progress: SharedValue<number>) =>
```
j-piasecki, line 90: Same

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,41 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        }
+      },
+      [props.onSwipeableClose, props.onSwipeableOpen]
+    );
+
     const animateRow = useCallback(
       (fromValue: number, toValue: number, velocityX?: number) => {
         'worklet';
-
+        console.log('animatin');
```
j-piasecki, line 353:
```suggestion
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,41 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       );
     };
 
+    const dispatchImmidiateEvents = useCallback(
+      (fromValue: number, toValue: number) => {
+        if (toValue > 0 && props.onSwipeableWillOpen) {
+          runOnJS(props.onSwipeableWillOpen)('left');
```
j-piasecki, line 322:
```suggestion
         props.onSwipeableWillOpen('left');
```

This function is not a worklet so `runOnJS` isn't needed here.

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,41 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        } else if (toValue < 0 && props.onSwipeableWillOpen) {
+          runOnJS(props.onSwipeableWillOpen)('right');
```
j-piasecki, line 324:
```suggestion
          props.onSwipeableWillOpen('right');
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,41 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        } else if (props.onSwipeableWillClose) {
+          const closingDirection = fromValue > 0 ? 'left' : 'right';
+          runOnJS(props.onSwipeableWillClose)(closingDirection);
```
j-piasecki, line 327:
```suggestion
          props.onSwipeableWillClose(closingDirection);
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,41 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        }
+      },
+      [props.onSwipeableWillClose, props.onSwipeableWillOpen]
+    );
+
+    const dispatchEndEvents = useCallback(
+      (fromValue: number, toValue: number) => {
+        if (toValue > 0 && props.onSwipeableOpen) {
+          runOnJS(props.onSwipeableOpen)('left', swipeableMethods.current);
```
j-piasecki, line 336:
```suggestion
          props.onSwipeableOpen('left', swipeableMethods.current);
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,41 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        } else if (toValue < 0 && props.onSwipeableOpen) {
+          runOnJS(props.onSwipeableOpen)('right', swipeableMethods.current);
```
j-piasecki, line 338:
```suggestion
          props.onSwipeableOpen('right', swipeableMethods.current);
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,41 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        } else if (props.onSwipeableClose) {
+          const closingDirection = fromValue > 0 ? 'left' : 'right';
+          runOnJS(props.onSwipeableClose)(
```
j-piasecki, line 341:
```suggestion
          props.onSwipeableClose(
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -334,37 +364,20 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
-                  swipeableMethods.current
-                );
-              } else if (props.onSwipeableClose) {
-                const closingDirection = fromValue > 0 ? 'left' : 'right';
-                runOnJS(props.onSwipeableClose)(
-                  closingDirection,
-                  swipeableMethods.current
-                );
-              }
+              dispatchEndEvents(fromValue, toValue);
```
j-piasecki, line 367:
```suggestion
              runOnJS(dispatchEndEvents)(fromValue, toValue);
```

Since `dispatchEndEvents` isn't a worklet (and we probably don't want it to be), you need `runOnJS` here.

src/components/ReanimatedSwipeable.tsx
```diff
@@ -334,37 +364,20 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
 
-        if (toValue > 0 && props.onSwipeableWillOpen) {
-          runOnJS(props.onSwipeableWillOpen)('left');
-        } else if (toValue < 0 && props.onSwipeableWillOpen) {
-          runOnJS(props.onSwipeableWillOpen)('right');
-        } else if (props.onSwipeableWillClose) {
-          const closingDirection = fromValue > 0 ? 'left' : 'right';
-          runOnJS(props.onSwipeableWillClose)(closingDirection);
-        }
+        dispatchImmidiateEvents(fromValue, toValue);
```
j-piasecki, line 372:
```suggestion
        runOnJS(dispatchImmidiateEvents)(fromValue, toValue);
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -379,47 +392,39 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
-      rightOffset,
-      rightWidth,
-      rowState,
-      rowWidth,
-      userDrag,
-    ]);
+    swipeableMethods.current = {
+      close() {
+        'worklet';
+        console.log('somethin');
```
j-piasecki, line 398:
```suggestion
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,41 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       );
     };
 
+    const dispatchImmidiateEvents = useCallback(
```
j-piasecki, line 319:
```suggestion
    const dispatchImmediateEvents = useCallback(
```

example/src/new_api/swipeable/AppleStyleSwipeableRow.tsx
latekvo: these new eslint rules flag inferring as a bad practice.
`any` is used in cases we don't care about the type of said variable, in our case it's only used when the variable either isn't used, or is printed

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,10 +316,37 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        if (toValue > 0 && props.onSwipeableWillOpen) {
+          props.onSwipeableWillOpen('left');
+        } else if (toValue < 0 && props.onSwipeableWillOpen) {
+          props.onSwipeableWillOpen('right');
+        } else if (props.onSwipeableWillClose) {
+          const closingDirection = fromValue > 0 ? 'left' : 'right';
+          props.onSwipeableWillClose(closingDirection);
+        }
+      },
+      [props.onSwipeableWillClose, props.onSwipeableWillOpen, swipeableMethods]
```
latekvo: I have to ask about this line.
My linter marks this as:

> unnecessary `prop.onSwipeableWillClose`, `props.onSwipeableWillOpen` deps

But isn't the current setup accommodating changes to those particular fields on the props?
These fields could change without triggering the broad `prop` dep as far as I understand.

latekvo: (in the latest commit, the `props` dep is not missing in the quoted code)

latekvo:
I just found out that I haven't achieved full independence from reanimated after all, I was just using incorrect imports in the examples.

latekvo:
Alright, after some further testing, when exporting `react-native-gesture-handler` the lib works as intended.
`react-native-reanimated` is NOT installed, until user explicitly installs it, even after add `ReanimatedSwipeable` to the project.
But what I don't think is normal nor expected, is `react-native-reanimated` gets installed to the root `node_modules` when running `yarn` within this repo.

src/components/ReanimatedSwipeable.tsx
j-piasecki: I don't get any warnings locally, though I see that the code got changed in the meantime. It could've just lagged behind a bit and a reload would have solved it.

tsconfig.json
```diff
@@ -21,5 +21,10 @@
     "noUnusedParameters": true,
     "noUnusedLocals": true
   },
-  "include": ["src/**/*.ts", "src/**/*.tsx", "jestSetup.js"],
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx",
+    "jestSetup.js",
+    "Swipeable/Swipeable.tsx"
```
j-piasecki, line 28: I don't think this file exists

src/components/ReanimatedSwipeable.tsx
```diff
@@ -17,7 +16,7 @@ import {
   GestureUpdateEvent,
   PanGestureHandlerEventPayload,
   PanGestureHandlerProps,
-} from 'react-native-gesture-handler';
+} from '../.';
```
j-piasecki, line 19: Similarly to the other PR, I don't think it should be imported from `index`. While in this case, it doesn't introduce any cycles (this component is not imported in `index`), it may turn out to be problematic in the future if someone forgets about it.

j-piasecki:
I'm getting the following error when trying to use the component on an app with RN `0.74.1` and Reanimated `3.12.0`:
Screenshot 2024-06-14 at 09 25 04

I think it may be caused by using `props` directly in methods called on (and from via `runOnJS`) the UI thread. Reanimated will [freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) everything in the worklets closure, so it may be necessary to only capture what's actually used.

I.e. there's a difference between
```jsx
useAnimatedStyle(() => {
  return { width: props.width }
});
```
and 
```jsx
const width = props.width;
useAnimatedStyle(() => {
  return { width: width }
});
```

src/components/ReanimatedSwipeable.tsx
latekvo: After rebuilding i unfortunately still get the following error:
> React Hook useCallback has unnecessary dependencies: 'props.onSwipeableWillClose', 'props.onSwipeableWillOpen', and 'swipeableMethods'. Either exclude them or remove the dependency array.

I think I'll just have to `eslint-disable-next-line` it, because I am quite certain that these are in fact necessary.

latekvo:
Besides the built-in examples issue, when using this library as usual, everything is now tested and working. Couldn't find any discrepancies so far when testing side by side with the original, but I still have to check the exact values returned by the callback and rendering functions, as they are now computed by a completely different mechanism.

latekvo:
i believe all the issues i could find are now fixed, besides a slight difference in spring config, but as far as i looked, there isn't a simple way to convert the legacy spring config 1:1 to a new one

j-piasecki:
There's an issue with layout in the new example (on all platforms):

Screenshot 2024-06-18 at 08 40 46

latekvo:
### How did you achieve that? I replicate your issue.

https://github.com/software-mansion/react-native-gesture-handler/assets/74246391/f37ab86d-d0f6-4e37-b4b2-112bd83b68b8


Uploading Screen Recording 2024-06-18 at 09.15.58.mov…

example/src/release_tests/swipeableReanimation/index.tsx
```diff
@@ -0,0 +1,144 @@
+import React from 'react';
+import { Text, Animated as Laminated, StyleSheet } from 'react-native';
```
j-piasecki, line 2: Is there a story behind `Laminated` 😄?

Anyway, If you need to use both Animated and Reanimated in the same file, I'd stick to keeping Animated as Animated, and import Reanimated as Reanimated.

src/components/ReanimatedSwipeable.tsx
```diff
@@ -584,4 +633,10 @@ const styles = StyleSheet.create({
     ...StyleSheet.absoluteFillObject,
     flexDirection: I18nManager.isRTL ? 'row' : 'row-reverse',
   },
+  childrenContainer: {
+    flex: 1,
+    alignItems: 'center',
+    width: '100%',
+    height: '100%',
+  },
```
j-piasecki, line 641: I don't think this is needed here (and it was causing the issue from the screen above)

src/components/ReanimatedSwipeable.tsx
```diff
@@ -544,23 +580,36 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
-        style={[styles.container, props.containerStyle]}>
+        style={[styles.container, containerStyle]}>
         {leftElement}
         {rightElement}
         <GestureDetector
           gesture={composedGesture}
           touchAction="pan-y"
-          {...props}>
-          <Animated.View style={[animatedStyle, props.childrenContainerStyle]}>
+          {...remainingProps}>
```
j-piasecki, line 606: This way you cannot 'swipe back' the swipeable. I think that pan should be attached to the outer `AnimatedView`, and tap to the inner one.

example/src/release_tests/swipeableReanimation/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+import {
+  GestureHandlerRootView,
+  RectButton,
+} from 'react-native-gesture-handler';
+
+import { Swipeable } from 'react-native-gesture-handler';
+import { default as ReanimatedSwipeable } from 'react-native-gesture-handler/ReanimatedSwipeable';
```
j-piasecki, line 10:
```suggestion
import ReanimatedSwipeable from 'react-native-gesture-handler/ReanimatedSwipeable';
```

Doesn't this work?

example/src/release_tests/swipeableReanimation/index.tsx
latekvo: I'm sorry, i never thought i'd upload that example and forgot to change the names 🫢

src/components/ReanimatedSwipeable.tsx
latekvo: It ensures the element is clickable and draggable across it's entire area, 
but I think you're right suggest to removing it, as the wasn't clickable across the entire area.

example/src/release_tests/swipeableReanimation/index.tsx
latekvo: In the file, it's exported as `Swipeable`, to be a drop-in replacement.
I can change it's name to `ReanimatedSwipeable` if you think that would work better.

j-piasecki: It's a default export, you don't have to import it under the same name it was exported (and you don't need to use `as` to rename).

What you did would be necessary if it was a named export:
```jsx
export { ReanimatedSwipeable }
```
vs
```jsx
export default ReanimatedSwipeable
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -544,27 +580,37 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
-          </Animated.View>
-        </GestureDetector>
-      </Animated.View>
+      <GestureDetector gesture={panGesture} touchAction="pan-y">
+        <Animated.View
+          onLayout={onRowLayout}
+          style={[styles.container, containerStyle]}>
+          {leftElement}
+          {rightElement}
+          <GestureDetector gesture={tapGesture} {...remainingProps}>
```
j-piasecki, line 602:
```suggestion
          <GestureDetector gesture={tapGesture}>
```

That's not necessary, it basically wouldn't do anything since gestures are configured in the builder in the new API.

src/components/ReanimatedSwipeable.tsx
```diff
@@ -544,27 +580,37 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+      () => ({
+        transform: [{ translateX: appliedTranslation.value }],
+        pointerEvents: rowState.value === 0 ? 'auto' : 'box-only',
+      }),
+      [appliedTranslation, rowState]
+    );
+
+    const containerStyle = props.containerStyle;
+    const childrenContainerStyle = props.childrenContainerStyle;
+    const remainingProps = { ...props };
```
j-piasecki, line 593:
```suggestion
```

example/src/release_tests/swipeableReanimation/index.tsx
latekvo: I see, didn't understand at first. I had no idea that's how default exports work.

latekvo:
left swipe threshold is twice as large as it should be on both the new and the legacy version, will double check, but looks like it

example/src/release_tests/swipeableReanimation/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+  RectButton,
+} from 'react-native-gesture-handler';
+
+import { Swipeable } from 'react-native-gesture-handler';
+import ReanimatedSwipeable from 'react-native-gesture-handler/ReanimatedSwipeable';
+import {
+  default as Reanimated,
+  SharedValue,
+  useAnimatedStyle,
+} from 'react-native-reanimated';
```
j-piasecki, line 15:
```suggestion
import Reanimated, {
  SharedValue,
  useAnimatedStyle,
} from 'react-native-reanimated';
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -523,16 +557,30 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+
+    panGesture.config = { ...remainingProps } as Extract<
+      SwipeableProps,
+      typeof panGesture.config
+    >;
+
+    tapGesture.config = { ...remainingProps } as Extract<
+      SwipeableProps,
+      typeof tapGesture.config
+    >;
```
j-piasecki, line 583:
```suggestion
```

That's a bad idea - you're effectively removing the config set above to assign props to it. The only prop that it would make sense for is `enableTrackpadTwoFingerGesture` and you can just set it on the gesture directly:
```jsx
panGesture.enableTrackpadTwoFingerGesture(props.enableTrackpadTwoFingerGesture)
```

Also you can remove `useNativeAnimations` from `SwipeableProps` since all Reanimated animations are native.

latekvo: I think it wasn't so bad, it automatically recognized all suitable props, making room for change in the future as well, and applied them before any of them were explicitly set. It's also quite readable.
But I replaced it as per your recommendation.

latekvo:
Just found some weird behaviour on `IOS` devices :/
`Taps` aren't detect and `swipe-back` doesn't register.

src/components/ReanimatedSwipeable.tsx
```diff
@@ -544,27 +584,31 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
-          </Animated.View>
-        </GestureDetector>
-      </Animated.View>
+      <GestureDetector gesture={panGesture} touchAction="pan-y">
+        <Animated.View
+          onLayout={onRowLayout}
+          style={[styles.container, containerStyle]}>
+          {leftElement}
+          {rightElement}
+          <GestureDetector gesture={tapGesture}>
```
j-piasecki, line 605:
```suggestion
          <GestureDetector gesture={tapGesture} touchAction="pan-y">
```

j-piasecki (approved):
Looks good 🚀.

I left one comment (fixes scrolling using touch on web), please wait for @m-bert's review before merging.

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,54 +317,87 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        if (toValue > 0 && props.onSwipeableWillOpen) {
+          props.onSwipeableWillOpen('left');
+        } else if (toValue < 0 && props.onSwipeableWillOpen) {
+          props.onSwipeableWillOpen('right');
+        } else if (props.onSwipeableWillClose) {
+          const closingDirection = fromValue > 0 ? 'left' : 'right';
+          props.onSwipeableWillClose(closingDirection);
+        }
+      },
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
m-bert, line 331: According to our conversation, this should be removed

```suggestion
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -317,54 +317,87 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        if (toValue > 0 && props.onSwipeableOpen) {
+          props.onSwipeableOpen('left', swipeableMethods.current);
+        } else if (toValue < 0 && props.onSwipeableOpen) {
+          props.onSwipeableOpen('right', swipeableMethods.current);
+        } else if (props.onSwipeableClose) {
+          const closingDirection = fromValue > 0 ? 'left' : 'right';
+          props.onSwipeableClose(closingDirection, swipeableMethods.current);
+        }
+      },
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
m-bert, line 351: Same as above
```suggestion
```

example/src/release_tests/swipeableReanimation/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+    console.log('[L] showLeftProgress:', value.value);
+  });
+  drag.addListener((value: any) => {
+    console.log('[L] appliedTranslation:', value.value);
+  });
+
+  const trans = Animated.subtract(drag, 50);
+
+  return (
+    <RectButton>
```
m-bert, line 61: It would be nice to have some feedback if we use `RectButton` (could be for example `console.log`). Also, why legacy version uses `button` if the one based on `Reanimated` doesn't?

m-bert (changes requested):
Imho this PR has too many linter changes that are out of scope (like `tsconfig.json` inside `MacOSExample`). It would be better to do another PR that unifies all indentations etc. (cc @j-piasecki). I don't consider it as a blocker though.

What I do consider as a blocker is lack of `touchAction: "pan-y"` in `GestureDetector` with `tapGesture`. Without it, `View` will have `touch-action: none;` which will effectively disable scrolling (you can check that on code from [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2788)).

Also, please remove `eslint-disable` (as we discussed it shouldn't be necessary here).

Other than that it looks solid, good job! 😄

example/src/new_api/swipeable/index.tsx
```diff
@@ -42,11 +42,13 @@ const SwipeableRow = ({ item, index }: { item: DataRow; index: number }) => {
...
 };
 
+const Separator = () => <View style={styles.separator} />;
+
 export default function App() {
   return (
     <FlatList
       data={DATA}
-      ItemSeparatorComponent={() => <View style={styles.separator} />}
+      ItemSeparatorComponent={Separator}
```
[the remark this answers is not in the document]
latekvo: Linter marks inline components like this as a bad practice.

m-bert (approved):
Thanks for adding `touchAction` 😄

I still believe that it would be nice to have some feedback from `RectButtons` in new example (using them as `View` doesn't make much sense for me), so please either remove them or add a callback before merging 😅

latekvo:
> I still believe that it would be nice to have some feedback from `RectButtons` in new example (using them as `View` doesn't make much sense for me), so please either remove them or add a callback before merging 😅

I removed those completely, they were used in the apple/gmail swipeable showcase, and I must have copied them accidentally.

jackstudd:
Hi @latekvo !🙋‍♂️

Thanks for this! It works really great and with better performances! However, I noticed there's no longer the ability to disable the swipe. I'm guessing this could be done with a props and passed to the `Gesture.Pan.enabled()`.

I tried to just remove the `renderActions` based on my disabled condition, but we can still swipe, we just no longer see the action.

I would gladly do the PR myself if needed!

jackstudd:
Hey @latekvo !

Just tried it out, it works perfectly, many thanks!

PR #2939, opened 2024-06-11
Add `relativeCoords` average to `PointerTracker`

m-bert:
## Description

This PR adds missing `relativeCoords` average in `PointerTracker`.

## Test plan

Add `console.log(this.tracker.getRelativeCoordsAverage());` in `onPointerMove` callback inside `PanGestureHandler` and check that logs contain average position.

### DIFF
--- a/src/web/tools/PointerTracker.ts
+++ b/src/web/tools/PointerTracker.ts
@@ -22,7 +22,8 @@ export default class PointerTracker {
 
   private lastMovedPointerId: number;
 
-  private cachedAverages: { x: number; y: number } = { x: 0, y: 0 };
+  private cachedAbsoluteAverages: { x: number; y: number } = { x: 0, y: 0 };
+  private cachedRelativeAverages: { x: number; y: number } = { x: 0, y: 0 };
 
   public constructor() {
     this.lastMovedPointerId = NaN;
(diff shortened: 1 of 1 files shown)


src/web/tools/PointerTracker.ts
```diff
@@ -160,8 +163,22 @@ export default class PointerTracker {
...
-      x: isNaN(avgX) ? this.cachedAverages.x : avgX,
-      y: isNaN(avgY) ? this.cachedAverages.y : avgY,
+      x: isNaN(avgX) ? this.cachedAbsoluteAverages.x : avgX,
+      y: isNaN(avgY) ? this.cachedAbsoluteAverages.y : avgY,
+    };
+
+    return averages;
+  }
+
+  public getRelativeCoordsAverage() {
```
m-bert, line 173: Here I see two options:

- We can leave 2 methods, one for `absolute` and one for `relative` coords
- We can merge them into one method

The second option could look like this:

```jsx
public getAverageCoords(relative = false) {
  // calculate average
}
```
It can be done in different ways (default value, optional argument, required argument), but you get the point. Let me know what do you think. cc @j-piasecki, @latekvo

latekvo: I think it would be better to follow the convention, right below there are `getAbsoluteCoordsSum()` and `getRelativeCoordsSum()` functions.

And while this creates a bit more duplicate code, I also think it's much more readable than having a single function with a `boolean` argument.
Perhaps the second option would work fine with an `enum`, but at that point it'd be making things too complicated.

But of course it's best to wait for j-piasecki's opinion 👍

j-piasecki: Yeah, I think keeping them separate makes sense. They are not that long and, unless I'm missing something, the only part that would apply to both cases would be calculating the average so we wouldn't get that much from merging the two.

j-piasecki (approved):
🪨

PR #2941, opened 2024-06-12
Add touchType to touchEvent

latekvo:
This PR adds pointerType to touchEvent on android and web.
This propet is already present on ios but it's missing from web and android.


latekvo:
Both android and web implementations are now working.
On a side note, jobs keep on getting randomly cancelled, and have to be manually re-run

j-piasecki (approved):

Issue #1221, opened 2020-10-29
Pressable API ?


TomasSestak:
## Motivation

Is it coming? Is there plan to bring ti to gesture handler?

nandorojo:
This would be awesome!

PR #2942, opened 2024-06-12
Create gesturized pressable component

latekvo:
This PR adds our own `Pressable` component

Closes https://github.com/software-mansion/react-native-gesture-handler/issues/1221


Pressable/package.json
```diff
@@ -0,0 +1,6 @@
+{
```
j-piasecki, line 1: There's no need to have a separate "inner" package for pressable

package.json
```diff
@@ -39,6 +39,7 @@
     "android/noreanimated/src/main/java/",
     "apple/",
     "Swipeable/",
+    "Pressable/",
```
j-piasecki, line 42:
```suggestion
```

src/components/Pressable.tsx
```diff
@@ -0,0 +1,281 @@
...
+}
+
+export interface PressableAndroidRippleConfig {
+  color?: null | ColorValue | undefined;
+  borderless?: null | boolean | undefined;
+  radius?: null | number | undefined;
+  foreground?: null | boolean | undefined;
+}
+
+export interface PressableProps
```
j-piasecki, line 40: I think it would make sense to split this component into multiple files, type definitions can be moved out

src/components/Pressable.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+import {
+  Gesture,
+  GestureDetector,
+  GestureHandlerRootView,
+  GestureStateChangeEvent,
+  GestureTouchEvent,
+  LongPressGestureHandlerEventPayload,
+  TouchData,
+} from '../.';
```
j-piasecki, line 23: Now that I think about it, it shouldn't be imported this way as it creates a cycle

src/components/Pressable.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  // onBlur and onFocus don't exist in the docs
+
+  touch.hitSlop(props.hitSlop);
+  press.hitSlop(props.hitSlop);
+  hover.hitSlop(props.hitSlop);
+
+  // add props.pressRetentionOffset, according to docs, they're relative to pressable, not hitSlop
+
+  touch.enabled(!(props.disabled ?? false));
```
j-piasecki, line 251:
```suggestion
  touch.enabled(props.disabled !== true);
```

src/components/Pressable.tsx
```diff
@@ -0,0 +1,281 @@
...
+          bottom: props.pressRetentionOffset,
+          right: props.pressRetentionOffset,
+        }
+      : props.pressRetentionOffset;
+
+  const touch = Gesture.Manual()
+    .onTouchesDown((event) => {
+      // note: hitslop checking support is built in
+      props.onPressIn?.(event);
+      previousTouchData.current = event.allTouches;
```
j-piasecki, line 199: If `onPressOut` is called only for the last pointer, should `onPressIn` be called for every pointer?

Have you checked how `Pressable` from React Native behaves in this case?

src/components/Pressable.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  const pressRetentionOffset: Insets | null | undefined =
+    typeof props.pressRetentionOffset === 'number'
+      ? {
+          top: props.pressRetentionOffset,
+          left: props.pressRetentionOffset,
+          bottom: props.pressRetentionOffset,
+          right: props.pressRetentionOffset,
+        }
+      : props.pressRetentionOffset;
```
j-piasecki, line 193: `pressRetentionOffset` is relative to the view size - setting it to `20` will expand the offset by that amount in every direction.

I think you end up comparing a rect of something like `[20, 20, 20, 20]` to the absolute coordinates of the touch.

This should be doable by setting a different `hitSlop` on the gesture depending on whether the component is already pressed.

src/components/Pressable.tsx
```diff
@@ -0,0 +1,281 @@
...
+      if (
+        previousTouchData.current?.find((touch) =>
+          touchWithinBounds(touch, pressRetentionOffset)
+        )
+      ) {
+        props.onPressOut?.(event);
+      }
+    });
+
+  const press = Gesture.LongPress().onEnd((event, success) => {
```
j-piasecki, line 221:
```suggestion
  const longPressGesture = Gesture.LongPress().onEnd((event, success) => {
```

src/components/Pressable.tsx
```diff
@@ -0,0 +1,281 @@
...
+    if (success) {
+      props.onLongPress?.(event);
+    }
+  });
+
+  const hover = Gesture.Hover()
```
j-piasecki, line 227:
```suggestion
  const hoverGesture = Gesture.Hover()
```

src/components/Pressable.tsx
```diff
@@ -0,0 +1,281 @@
...
+    typeof props.pressRetentionOffset === 'number'
+      ? {
+          top: props.pressRetentionOffset,
+          left: props.pressRetentionOffset,
+          bottom: props.pressRetentionOffset,
+          right: props.pressRetentionOffset,
+        }
+      : props.pressRetentionOffset;
+
+  const touch = Gesture.Manual()
```
j-piasecki, line 195:
```suggestion
  const touchGesture = Gesture.Manual()
```

j-piasecki, line 195: I think all of the gestures should be run on the JS thread (`.runOnJS(true)`).

j-piasecki, line 195: It's very possible, that we should be using `Gesture.Native` here combined with `GestureHandlerButton` for tracking touches (I think there may be differences when it comes to handling hitslop), though that needs to be checked

src/components/Pressable.tsx
latekvo: i think `!== false` would work as intended, since we want all `true`, `undefined` and `null` to be truthy as well

src/components/Pressable.tsx
latekvo: I'm in the progress of working on it, current setup is completely invalid

latekvo:
Original `onBlur` and `onFocus`:
When in `blurred` state, the first click on the pressable calls `onFocus`.
When in `focused` state, when the user clicks outside of `Pressable`, `onBlur` is called and `blurred` state is set.

src/components/Pressable.tsx
latekvo: You're right, I checked the original and `onPressIn` is triggered on the first press only. Same with `onPressOut`

src/components/Pressable.tsx
latekvo: It is now working, although I'll see if i can optimise it further because i really dislike how it looks now.

src/components/PressableProps.tsx
```diff
@@ -0,0 +1,160 @@
...
+  onPressOut?: null | ((event: GestureTouchEvent) => void);
+
+  /**
+   * Called when a long-tap gesture is detected.
+   */
+  onLongPress?:
+    | null
+    | ((
+        event: GestureStateChangeEvent<LongPressGestureHandlerEventPayload>
+      ) => void);
```
j-piasecki, line 72: I think that here we would like to match the event types from the RN's implementation.

src/components/Pressable.tsx
latekvo: You're right

latekvo:
there are a lot of `any`, `unknown` and other bad practices in the latest commit.
I wanted to commit changes without uploading otherwise completely erranous version.
Will remove them soon.

src/components/Pressable.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const touchGesture = Gesture.Native()
+    .onTouchesDown((event) => {
+      // check if all touching fingers were lifted up on the previous event
+      if (
+        !previousTouchData.current ||
+        !previousChangeData.current ||
+        previousTouchData.current?.length === previousChangeData.current?.length
+      ) {
+        props.onPressIn?.(event);
```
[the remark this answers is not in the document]
latekvo: Good call, I think `onEnd` may work well for detecting the pointer being out of bounds while accounting for `pressRetentionOffset` and `shouldCancelWhenOutside`.

latekvo:
Last blocker: setting `hitSlop` doesn't have an effect on the pressable area.

latekvo:
Could someone please confirm there is a crash on android with the standalone `new_api` example?
It appears `fontWeight` style is causing it, but I still haven't ruled out that my particular android installation is somehow corrupted, which happened to me before :P

latekvo:
### State of things
- on `android`, we're only lacking the feature to activate onPressIn, when a pressed mouse/finger slides over the component
- on `ios`, both `ManualGesture`'s and `NativeGesture`'s `hitSlop` setting behaves differently than on android, only allowing to extend a mouse click that has already been placed, but doesn't extend the clickable area itself.
- on `web`, I won't be implementing either `hitSlop` or `retentionOffset`.

latekvo:
I see no further issues, and all bahaviours align with the original.
I'll work on further testing, reduction of code, and polishing the adapted event, but as a whole, it's ready for review.

latekvo:
for the web, we cannot use `RNButton`, and have to replace it with `RectButton` instead, but that doesn't infer on any functionality as far as I see.

latekvo:
I am aware that the tests have strange styling on `web`, but I will take care of that after I finish writing them.

latekvo:
Screenshot 2024-06-21 at 15 09 16

This is up for investigation, first of all, kotlin/native side interface doesn't align with the react typing, secondly the number I am providing doesn't seem to align with any kind of rgb standard, and is likely something else. 
Converting from hex `0xRRGGBB` to `int` won't get the desired result, neither will `0xRGB`.
The random number that is present there is just a one that works at all, to showcase what is happening.

example/src/release_tests/gesturizedPressable/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+    alignSelf: 'flex-start',
+    marginLeft: 20,
+    marginTop: 0,
+    marginBottom: 5,
+    color: '#555',
+  },
+  scrollable: {},
+  separator: {
+    width: '90%',
+    // ios bug - different border styling doesn't work on just 1 side
```
j-piasecki, line 96: What do you mean?

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
...
+  bottom: (a.bottom ?? 0) + (b.bottom ?? 0),
+});
+
+const touchToPressEvent = (data: TouchData, timestamp: number): PressEvent => ({
+  identifier: data.id,
+  locationX: data.x,
+  locationY: data.y,
+  pageX: data.absoluteX,
+  pageY: data.absoluteY,
+  target: 0, // fixme if possible, set to correct target ID
```
j-piasecki, line 38: What is the correct target ID in this case?

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
...
+  id: 0, // fixme if possible, set to correct pointer ID
+  x: event.x,
+  y: event.y,
+  absoluteX: event.absoluteX,
+  absoluteY: event.absoluteY,
+});
+
+const isTouchWithinInset = (
+  // all of these are guaranteed to be their type,
+  // but we cannot use '!' anymore
```
j-piasecki, line 58: What do you mean?

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
...
+  touch?: TouchData,
+  inset?: Insets,
+  dimensions?: { width: number; height: number }
+) =>
+  // ^ to right, ^ to bottom
```
j-piasecki, line 63: Same?

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+  return {
+    nativeEvent: {
+      touches: [pressEvent],
+      changedTouches: [pressEvent],
+      identifier: event.handlerTag,
+      locationX: event.x ?? -1,
+      locationY: event.y ?? -1,
+      pageX: event.absoluteX ?? -1,
+      pageY: event.absoluteY ?? -1,
```
j-piasecki, line 88: Shouldn't those be defined on the event?

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+  return {
+    nativeEvent: {
+      touches: nativeTouches,
+      changedTouches: nativeChangedTouches,
+      identifier: event.handlerTag,
+      locationX: event.allTouches.at(0)?.x ?? -1,
+      locationY: event.allTouches.at(0)?.y ?? -1,
+      pageX: event.allTouches.at(0)?.absoluteX ?? -1,
+      pageY: event.allTouches.at(0)?.absoluteY ?? -1,
```
j-piasecki, line 114: Same

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
...
+      setStyleProp(props.style({ pressed: isPressed }));
+    } else {
+      setStyleProp(props.style);
+    }
+    if (typeof props.children === 'function') {
+      setChildrenProp(props.children({ pressed: isPressed }));
+    } else {
+      setChildrenProp(props.children);
+    }
+  };
```
j-piasecki, line 164: I don't think we should be keeping styles and children in the state. IMO it would be better to keep `pressedDown` as state, and move function checks (and execution) to the render function, [like in RN](https://github.com/facebook/react-native/blob/f7fe68861a52a0428698a9352d39a6e39afefe42/packages/react-native/Libraries/Components/Pressable/Pressable.js#L350-L352).

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
...
+    typeof props.pressRetentionOffset === 'number'
+      ? numberAsInset(props.pressRetentionOffset)
+      : props.pressRetentionOffset ?? {};
+
+  const pressGesture = Gesture.LongPress().onStart((event) => {
+    if (isPressedDown.current) {
+      props.onLongPress?.(adaptStateChangeEvent(event));
+      isPressCallbackEnabled.current = false;
+    }
+  });
```
j-piasecki, line 150: It may be worthwhile to consider wrapping gesture objects with `useMemo` so that they are not recreated on every render.

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
+import React, { useEffect, useRef, useState } from 'react';
```
j-piasecki, line 1: You can make a new directory called `Pressable` and move the file with props along with this one there. Then you can add an `index` file exporting what's necessary.

src/components/Pressable.tsx
```diff
@@ -0,0 +1,277 @@
...
+const DEFAULT_HOVER_DELAY = 0;
+
+const numberAsInset = (value: number): Insets => ({
+  left: value,
+  right: value,
+  top: value,
+  bottom: value,
+});
+
+const addInsets = (a: Insets, b: Insets): Insets => ({
```
j-piasecki, line 25: This, along with other helpers, can be moved to a separate `utils` file IMO.

src/components/Pressable.tsx
```diff
@@ -0,0 +1,286 @@
...
+  touches: [], // intentionally empty
+  changedTouches: [], // intentionally empty
+});
+
+const changeToTouchData = (
+  event: GestureStateChangeEvent<
+    HoverGestureHandlerEventPayload | LongPressGestureHandlerEventPayload
+  >
+): TouchData => ({
+  id: 0, // fixme if possible, set to correct pointer ID
```
j-piasecki, line 49: Bump, to not forget about it

src/components/Pressable.tsx
```diff
@@ -0,0 +1,286 @@
...
+  }, []);
+
+  const defaultRippleColor = props.android_ripple ? undefined : 'transparent';
+
+  return (
+    <GestureDetector gesture={gesture}>
+      <RNButton
+        testID={props.testID}
+        ref={pressableRef}
+        // this hitSlop block is required by ios
```
j-piasecki, line 275: Is this still necessary after adding `rippleGesture`?

example/src/release_tests/gesturizedPressable/androidRipple.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { StyleSheet, View } from 'react-native';
+import TestingBase from './testingBase';
+
+export function RippleExample() {
+  return (
+    <View style={styles.container}>
+      <TestingBase
+        style={styles.pressable}
+        android_ripple={{
+          // @ts-ignore sources say string, android actually wants a number
```
j-piasecki, line 11:
```suggestion
```

src/components/Pressable.tsx
```diff
@@ -0,0 +1,286 @@
...
+      // original triggers onPressOut on cancel, but not onPress
+      props.onPressOut?.(adaptTouchEvent(event));
+
+      setPressedState(false);
+    });
+
+  const rippleGesture = Gesture.Native();
+
+  pressGesture.minDuration(props.delayLongPress ?? DEFAULT_LONG_PRESS_DURATION);
+  pressGesture.minDuration(props.delayLongPress ?? DEFAULT_LONG_PRESS_DURATION);
```
j-piasecki, line 232: Why do it twice?

src/components/Pressable.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+  useEffect(() => {
+    setPressedState(props.testOnly_pressed ?? false);
+  }, []);
+
+  const defaultRippleColor = props.android_ripple ? undefined : 'transparent';
+
+  return (
+    <GestureDetector gesture={gesture}>
+      <RNButton
```
j-piasecki, line 272: I think it should also receive `enabled` prop

src/components/Pressable.tsx
```diff
@@ -0,0 +1,286 @@
...
+  pressGesture.shouldCancelWhenOutside(true);
+  hoverGesture.shouldCancelWhenOutside(true);
+
+  touchGesture.enabled(props.disabled !== true);
+  pressGesture.enabled(props.disabled !== true);
+  hoverGesture.enabled(props.disabled !== true);
+
+  touchGesture.runOnJS(true);
+  pressGesture.runOnJS(true);
+  hoverGesture.runOnJS(true);
```
j-piasecki, line 255: Shouldn't this also be done for `rippleGesture`?

src/components/Pressable.tsx
```diff
@@ -0,0 +1,286 @@
...
+  const gesture = Gesture.Simultaneous(
+    hoverGesture,
+    pressGesture,
+    touchGesture,
+    rippleGesture
+  );
+
+  useEffect(() => {
+    setPressedState(props.testOnly_pressed ?? false);
+  }, []);
```
j-piasecki, line 266: Does this need to be an effect, or can it be set as the initial state?

src/components/PressableProps.tsx
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Additional distance outside of this view in which a touch is considered a
+   * press before `onPressOut` is triggered.
+   */
+  pressRetentionOffset?: null | Insets | number;
+
+  /**
+   * If true, doesn't play system sound on touch.
+   */
+  android_disableSound?: null | boolean;
```
j-piasecki, line 121: I don't see this implemented

src/components/Pressable.tsx
```diff
@@ -0,0 +1,286 @@
+import React, { useEffect, useRef, useState } from 'react';
+import { GestureObjects as Gesture } from '../handlers/gestures/gestureObjects';
```
j-piasecki, line 2: And just so it's not lost to time, I think the default config for ripple differs between this implementation and RN's one - here it's enabled by default.

src/components/PressableProps.tsx
```diff
@@ -0,0 +1,145 @@
...
+   * the component is currently pressed and returns view styles.
+   */
+  style?:
+    | StyleProp<ViewStyle>
+    | ((state: PressableStateCallbackType) => StyleProp<ViewStyle>);
+
+  /**
+   * Duration (in milliseconds) to wait after press down before calling onPressIn.
+   */
+  unstable_pressDelay?: number;
```
j-piasecki, line 144: Is this one implemented?

example/src/release_tests/gesturizedPressable/index.tsx
latekvo: Screenshot 2024-06-24 at 10 21 52

IOS doesn't allow for different styles for different borders.

src/components/Pressable.tsx
latekvo: My bad, I though Eslint prohibited usage of `!`, but it doesn't seem like a problem here.

src/components/Pressable.tsx
latekvo: I see how it's unclear, I will remove it.
It meant to show these values increase from left to right, and from top to bottom.

src/components/Pressable.tsx
latekvo: I'm not sure 100% if it's defined on every platform, will have to double check it.

src/components/Pressable.tsx
latekvo: I disabled this in the latest commit, although I'm not sure if setting `rippleColor` to `transparent` is a good way of doing so 😅 

There don't seem to be any dedicated methods for disabling it on `RNButton`, and disabling the `Native` gesture had no effect either.

src/components/PressableProps.tsx
latekvo: It is, but I still have to work on making it work well.

On the original pressable, it converts the onPressIn into a long press, as in you have to hold it down for `unstable_pressDelay` milliseconds before it registers.

src/components/Pressable.tsx
latekvo: It's not necessary anymore.

src/components/Pressable.tsx
latekvo: mistake when i was testing `unstable_pressDelay`

src/components/Pressable.tsx
latekvo: image

j-piasecki: Why mark them as nullable (`?`) though? It's called in one place and 2 of these 3 arguments seem to be always defined.

src/components/PressableProps.tsx
j-piasecki: Isn't it `unstable_pressDelay` for `onPressIn` and `unstable_pressDelay + delayLongPress` for `onLongPress`?

latekvo: yes it is, why do you ask?

latekvo: ok i double checked now.
on the original, `unstable_pressDelay` does not set the minimum finger holding distance.
If the finger is pulled up before `unstable_pressDelay` passes, `onPressIn` will be dispatched either way.
But if the `unstable_pressDelay` passes, and the finger is still holding down on the `pressable`, `onPressIn` will automatically dispatch.

latekvo: It's implemented as of the latest commit.

src/components/Pressable.tsx
latekvo: Because the Inset type explicitly has every of it's child members as `number | undefined`, and so I cannot just use them as it is. 

Same goes for the event types, which while they are certain to have a `number` type, the events themselves are `event | undefined`.

And for the `touch?: TouchData` object, it's just a choice between having it non-null checked either here in `utils`, or in `Pressable`, where the passed argument is read from an array, which implicitly makes it `| undefined`

example/src/release_tests/gesturizedPressable/index.tsx
latekvo: image
It's hard to explain

latekvo: image

latekvo: The key here is `borderStyle`, it's the only thing causing issue on IOS.

src/components/Pressable.tsx
latekvo: That'd be the `RNButton`'s `Native ID`. I know it's there i just have to find a way to access it.

example/src/release_tests/gesturizedPressable/index.tsx
latekvo: And it works perfectly on web, here is how it is supposed to look :P

image

src/components/Pressable.tsx
j-piasecki: Sounds fairly reasonable to me 😄

src/components/Pressable.tsx
latekvo: Moved

src/components/Pressable.tsx
latekvo: Done

src/components/Pressable.tsx
latekvo: Done

src/components/Pressable.tsx
latekvo: Let me know if the current code is what you had in mind.

src/components/Pressable.tsx
latekvo: Addressed this problem in depth in a different comment further up.

src/components/PressableProps.tsx
latekvo: Implemented in the latest commit.

src/components/Pressable.tsx
latekvo: Removed this useEffect as well as the useStates you mentioned in other comment.

src/components/Pressable.tsx
j-piasecki: > The Inset type explicitly has every of it's child members as number | undefined, and so I cannot just use them as it is.

But `inset` itself seems to always be defined (it's always `normalizedHitslop`), no?

> And for the touch?: TouchData object, it's just a choice between having it non-null checked either here in utils, or in Pressable, where the passed argument is read from an array, which implicitly makes it | undefined.

You're right.

src/components/Pressable.tsx
latekvo: Done

src/components/Pressable.tsx
latekvo: Done

example/src/release_tests/gesturizedPressable/androidRipple.tsx
latekvo: removed

latekvo:
I still need to fix styling for examples on `web`, and fix `pressDelay` on web, it's not being cancelled on touch up.

src/components/Pressable.tsx
j-piasecki: I'm almost sure events are not bubbling. I think target is present in the events on iOS, not sure if it's present on other platforms so it may not be useful.

If you cannot figure out a way to get it, it may make sense to implement it on Android & Web.

src/components/Pressable.tsx
j-piasecki: I don't really see that, did you push the commit?

src/components/Pressable.tsx
latekvo: Could not find a conventional way of getting `targetId`, and `pointerId` in those places that they are missing.

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+  const styleProp =
+    typeof props.style === 'function'
+      ? props.style({ pressed: pressedState })
+      : props.style;
+
+  const childrenProp =
+    typeof props.children === 'function'
+      ? props.children({ pressed: pressedState })
+      : props.children;
```
j-piasecki, line 40: Can this be moved closer to the `return` statement?

src/components/Pressable.tsx
j-piasecki: Yes, just left one more comment regarding it but otherwise it's pretty good.

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,240 @@
...
+          setTimeout(
+            () => props.onHoverOut?.(adaptStateChangeEvent(event)),
+            props.delayHoverOut ?? DEFAULT_HOVER_DELAY
+          );
+        }),
+    [props.onHoverIn, props.onHoverOut, props.delayHoverIn, props.delayHoverOut]
+  );
+
+  const pressDelayTimeoutRef = useRef<number | null>(null);
+  const onPressInRoutine = useCallback((event: GestureTouchEvent) => {
```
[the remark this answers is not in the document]
latekvo: sure, done

src/components/Pressable.tsx
latekvo: sorry, i took a look and misunderstood it as this line:
> rippleGesture.hitSlop(normalizedHitSlop);

I'm not sure if it's needed to duplicate `enabled` and `shouldCancelWhenOutside`, as it's a blank gesture that doesn't really do anything besides being there for `RNButton`.

I think it'd be better to separate it from these blocks and label it with a comment as to why it is there.
I think it's both cluttering and confusing why this gesture is there to anyone who isn't familiar with nuances of `ripple`.

src/components/Pressable.tsx
latekvo: > But inset itself seems to always be defined (it's always normalizedHitslop), no?

You're right, I'm correcting this mistake.

src/components/Pressable.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+        if (
+          event.allTouches.find((touch) =>
+            touchWithinBounds(touch, pressRetentionOffset, pressableDimensions)
+          )
+        ) {
+          props.onPress?.(event);
+          props.onPressOut?.(event);
+        }
+      });
```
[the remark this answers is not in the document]
latekvo: Resolved, this method didn't work but the current implementation deals with it well.

src/components/Pressable.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const gesture = Gesture.Simultaneous(
+    hoverGesture,
+    pressGesture,
+    touchGesture
+  );
+
+  return (
+    <GestureHandlerRootView>
+      <GestureDetector gesture={gesture}>
+        <View
```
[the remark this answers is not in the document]
latekvo: Done

src/components/Pressable.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  touchGesture.runOnJS(true);
+  pressGesture.runOnJS(true);
+  hoverGesture.runOnJS(true);
+
+  const gesture = Gesture.Simultaneous(
+    hoverGesture,
+    pressGesture,
+    touchGesture
+  );
```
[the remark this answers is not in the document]
latekvo: Fixed, iirc this was an issue with which state i was watching on the long press gesture

latekvo:
Regarding the border example, do we want to nest another `View` inside the `RNButton` to allow for some broken styles to be more aligned with the original?

Screenshot 2024-06-26 at 09 31 41
Screenshot 2024-06-26 at 09 32 52

j-piasecki:
> Regarding the border example, do we want to nest another View inside the RNButton to allow for some broken styles to be more aligned with the original? (c.c. @j-piasecki )

Possibly, though It's something I'd like to avoid, given it's similar to Touchables from GH and there were many issues regarding that.

example/src/release_tests/gesturizedPressable/delayedPressExample.tsx
```diff
@@ -0,0 +1,80 @@
...
+export function DelayedPressExample() {
+  const startColor = '#fff';
+  const pressColor = '#ff0';
+  const longPressColor = '#f0f';
+  const animatedColor = useSharedValue(startColor);
+
+  const pressDelay = 1000;
+
+  // original adds this value to the base longPressDuration
+  // new implementation uses this value as the longPressDuration
```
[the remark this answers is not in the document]
latekvo: This comment is old and wrong, current delayed press implementation is identical to the original one, I'll remove it.

example/src/release_tests/gesturizedPressable/delayedPressExample.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const longPressDelay = 1000;
+
+  const onPressIn = () => {
+    console.log('Pressed with delay');
+    animatedColor.value = withSequence(
+      withSpring(pressColor, signallerConfig),
```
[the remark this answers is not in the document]
latekvo: Fixed

example/src/release_tests/gesturizedPressable/testingBase.tsx
```diff
@@ -0,0 +1,54 @@
...
+    </GesturizedPressable>
+    <Pressable {...props}>
+      <View style={styles.textWrapper}>
+        <Text style={styles.text}>Legacy pressable!</Text>
+      </View>
+    </Pressable>
+  </>
+);
+
+const BG_COLOR = '#F5FCFF';
```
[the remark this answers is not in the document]
latekvo: Done

example/src/release_tests/gesturizedPressable/hitSlopExample.tsx
```diff
@@ -0,0 +1,100 @@
...
+          onHoverOut={() => hoverOut()}
+          onLongPress={() => longPress()}
+        />
+      </View>
+    </>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
```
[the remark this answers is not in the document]
latekvo: created 2 constants in the outer scope of both the component and the styles

j-piasecki:
I think we can try the approach with an additional view as a parent of the button. You would probably style the view and simply pass `StyleSheet.absoluteFill` as the style to the button. You may need to split the other props in some wacky way.

latekvo:
>I think we can try the approach with an additional view as a parent of the button. You would probably style the view and simply pass StyleSheet.absoluteFill as the style to the button. You may need to split the other props in some wacky way.

Hey just to clarify, in the attached screenshot, `View` was the inner most child, and it inherited all the styles, removing them from the button entirely. Everything thing else adjusted perfectly as well.

Is there a reason why you'd prefer the `View` to be the topmost parent, or is it of no importance?

For now I'll commit the component with the additional `View` as it's child, but please let me know if this is what you wanted.

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,247 @@
...
+  touchGesture.runOnJS(true);
+  pressGesture.runOnJS(true);
+  hoverGesture.runOnJS(true);
+  rippleGesture.runOnJS(true);
+
+  // this hitSlop block is required by android
+  touchGesture.hitSlop(appliedHitSlop);
+  pressGesture.hitSlop(appliedHitSlop);
+  hoverGesture.hitSlop(appliedHitSlop);
+  rippleGesture.hitSlop(normalizedHitSlop);
```
m-bert, line 201: Why does `rippleGesture` use different `hitSlop`?

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,247 @@
...
+  const appliedHitSlop = addInsets(
+    normalizedHitSlop,
+    normalizedPressRetentionOffset
+  );
+
+  const isPressableEnabled = props.disabled !== true;
+
+  touchGesture.enabled(isPressableEnabled);
+  pressGesture.enabled(isPressableEnabled);
+  hoverGesture.enabled(isPressableEnabled);
```
m-bert, line 186: Shouldn't it be also set on `rippleGesture`?

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  touchGesture.shouldCancelWhenOutside(true);
+  pressGesture.shouldCancelWhenOutside(true);
+  hoverGesture.shouldCancelWhenOutside(true);
```
m-bert, line 190: Should it be set to `true`?

m-bert (changes requested):
Good job! There are few things that I'd like to point out

1. If you perform a subtle click, `Pressable` will stay in `pressed in` state. Here is how it looks like:

https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/c7fb9c78-816e-48a6-a0e7-c6edc5eddbf5

> [!CAUTION]
> I've recorded it on android, but it happens on all platforms.

2. On `web` if you click on `Pressable` and drag out of its boundaries, you'll trigger `press out` callback. On native platforms it doesn't happen:

https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/9e906875-ec1b-4bf1-9428-d4684cfb6cf0

> [!TIP]
> It is caused by `shouldCancelWhenOutside(true)` inside `Pressable.tsx`. I'm not sure if changing it to `false` will break something

3. We can think about passing `userSelect`/`touchAction` into `GestureDetector` to allow users to customize some of web behaviors (for example, `Touchable` [does support userSelect](https://github.com/software-mansion/react-native-gesture-handler/blob/b1ed6ffcde0d456b38a7e37a50bedde2c78f3ac8/src/components/touchables/GenericTouchable.tsx#L291))

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,247 @@
...
+    props.testOnly_pressed ?? false
+  );
+
+  const pressableRef = useRef<View>(null);
+
+  // disabled when onLongPress has been called
+  const isPressCallbackEnabled = useRef<boolean>(true);
+  const isPressedDown = useRef<boolean>(false);
+
+  const normalizedHitSlop: Insets =
```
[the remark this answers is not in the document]
j-piasecki: I can see your point, but also they kind of are transformed (from a number to a rect). No strong opinion from me, though we can change that if you have something better in mind.

src/components/Pressable/Pressable.tsx
j-piasecki: So that ripple doesn't activate when you tap on the press retention area

example/src/release_tests/gesturizedPressable/bordersExample.tsx
```diff
@@ -0,0 +1,39 @@
...
+  },
+  square: {
+    width: 100,
+    height: 100,
+    borderRadius: 10,
+    borderWidth: 5,
+    borderBottomColor: '#0f0',
+    borderRightColor: '#f00',
+    borderLeftWidth: 1,
+    borderStyle: 'dashed',
```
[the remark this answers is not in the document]
latekvo: Hey just to address your issues.

> If you perform a subtle click, Pressable will stay in pressed in state. Here is how it looks like:

I'll fix that right away.

>On web if you click on Pressable and drag out of its boundaries, you'll trigger press out callback. On native platforms it doesn't happen:

I'll fix that as well.

>Let's change that to solid since iOS doesn't support dashed border as shown on the image below

The point of this example is to show how `Pressable` behaves with dashed/dotted border which is differently supported on android, web and IOS, all of which handle borders differently when you mix different styles.

Previously, `Pressable` had issues with being consistant with the legacy version, and so i added this example.
I fixed this error since, so I think I can remove the borders example all together, as all the styles are now aligned with the legacy pressable.

For more details on the borders issue, there are screens documenting these issues further up in this PR.

src/components/Pressable/Pressable.tsx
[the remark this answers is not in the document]
latekvo: `normalized` in this context means they are transformed into an universal form. 
There is a chance they'll be left without and transforms, but that means they were in the correct format in the first place. 
Most of the times they are changed from `number` to `Insets`.

src/components/Pressable/Pressable.tsx
latekvo: Yes, setting it to `false` wouldn't cancel the gestures when they leave `pressRetentionOffset` area, which is how the legacy pressable behaves.
And there isn't a prop to set it to `false`.

src/components/Pressable/Pressable.tsx
latekvo: It's not necessary, it's dependant on `RNButton`, and it's getting disabled correctly, but I'll add it for clarity.

src/components/Pressable/Pressable.tsx
latekvo: I'll do that.

src/components/Pressable/Pressable.tsx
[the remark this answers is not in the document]
m-bert: I get what it means in that context, I was just pointing out that first thing that comes to my mind when I see `normalized` is something like dividing vector by its magnitude 😅 

I don't say that we have to change that, nor do I have a better name for it. I would probably use something like `transformedHitSlop`, but it doesn't sound good either.

So in conclusion, let's leave that as it is 😅

src/components/Pressable/PressableProps.tsx
```diff
@@ -0,0 +1,147 @@
...
+   * Whether a press gesture can be interrupted by a parent gesture such as a
+   * scroll event. Defaults to true.
+   */
+  cancelable?: null | boolean;
+
+  /**
+   * Duration to wait after hover in before calling `onHoverIn`.
+   * @platform macos windows
+   * | NOTE: not present in RN docs
+   */
```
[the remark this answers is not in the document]
latekvo: These props were present on the legacy `Pressable` but since they weren't documented anywhere, and I couldn't test them, I left that comment.

I'll remove them all together, I don't think we have to go that far with compatibility as we've already removed `onBlur` and `onFocus` due to lack of tools to implement those. (c.c. @j-piasecki )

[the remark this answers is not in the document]
latekvo: Sorry, I see that I actually implement those, but then couldn't test them :P
I think we can leave them in, I'll just have to find a way to build and test this on my computer.

src/components/GestureButtons.tsx
```diff
@@ -63,6 +63,8 @@ export interface RawButtonProps extends NativeViewGestureHandlerProps {
    * Set this to true if you don't want the system to play sound when the button is pressed.
    */
   touchSoundDisabled?: boolean;
+
+  style?: StyleProp<ViewStyle>;
```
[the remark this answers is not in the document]
latekvo: I'll do that

src/components/Pressable/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+  (touch?.y ?? 0) > -(inset.top ?? 0);
+
+const adaptStateChangeEvent = (
+  event: GestureStateChangeEvent<
+    HoverGestureHandlerEventPayload | LongPressGestureHandlerEventPayload
+  >
+): PressableEvent => {
+  const timestamp = Date.now();
+
+  // as far as i see, there isn't a conventional way of getting targetId with the data we get
```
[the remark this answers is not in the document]
latekvo: I'll change that

src/components/Pressable/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+      timestamp: timestamp,
+      force: undefined,
+    },
+  };
+};
+
+const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
+  const timestamp = Date.now();
+
+  // as far as i see, there isn't a conventional way of getting targetId with the data we get
```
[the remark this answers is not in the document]
latekvo: I'll change that

src/components/Pressable/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+      touches: [pressEvent],
+      changedTouches: [pressEvent],
+      identifier: pressEvent.identifier,
+      locationX: event.x,
+      locationY: event.y,
+      pageX: event.absoluteX,
+      pageY: event.absoluteY,
+      target: targetId,
+      timestamp: timestamp,
+      force: undefined,
```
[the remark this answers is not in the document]
latekvo: It could be attained when using a special force touch pencil, trackpad or a screen, but I don't think I have any of those, and so I have no way to test it.

src/components/Pressable/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+  absoluteY: event.absoluteY,
+});
+
+const isTouchWithinInset = (
+  dimensions: { width: number; height: number },
+  inset: Insets,
+  touch?: TouchData
+) =>
+  (touch?.x ?? 0) < (inset.right ?? 0) + dimensions.width &&
+  (touch?.y ?? 0) < (inset.bottom ?? 0) + dimensions.height &&
```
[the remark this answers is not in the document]
latekvo: It's an object which contains the dimensions of a `ReactNode`, and so i think the name `dimensions` is quite fitting.

It's quite a small function, with type specified 1 line above, and alternatives like `size` also partially used by react - `ScaledSize`.

So I don't think changing it will make it any clearer.

latekvo:
>Even if borderStyle causes the issue in Pressable (and I've mentioned it in bordersExample.tsx), I don't understand why do we need this comment here. In the separator styles work fine (though they look different when comparing platforms)

This comment refers to `borderWidth`, which is set to an arbitrary number.
`IOS` does allow for `dotted/slashed` borders, just refuses to render the entire object, when any 1-side styles are set to any of the sides of the border, which is also `dotted/slashed` on any of the sides.

Because of this, we have to set the border width to a really small number, so that all the borders are squashed into a single line, while also making them thick enough, so that the slashed effect does not look like little dots, as it's look depends on border's width.

src/components/Pressable/PressableProps.tsx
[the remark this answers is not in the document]
m-bert: Still, I don't think it's a good idea to point out that supported platforms are desktops. Currently we have very limited support on `macOS` and we do not test things on `windows`. Also it sounds a bit weird that native library has desktop-only props, cc @j-piasecki.

src/components/Pressable/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+): PressEvent => ({
+  identifier: data.id,
+  locationX: data.x,
+  locationY: data.y,
+  pageX: data.absoluteX,
+  pageY: data.absoluteY,
+  target: targetId,
+  timestamp: timestamp,
+  touches: [], // intentionally empty
+  changedTouches: [], // intentionally empty
```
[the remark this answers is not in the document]
latekvo: I changed those to `always empty`.

This strange structure is caused by the legacy version of this object using the same base class for both the parent and the child, even though they weren't recursively referencing themselves.

Consequently, while the parent's `touches` and `changedTouches` were always populated, the child's were always empty.

m-bert:
Okay, after I've stumbled upon

> The key here is borderStyle, it's the only thing causing issue on IOS.

I assumed that the problem was just `borderStyle` and `borderWidth` has nothing to do with it.

src/components/Pressable/utils.ts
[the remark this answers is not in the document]
m-bert: Well, we do have one of those 😉

src/components/Pressable/utils.ts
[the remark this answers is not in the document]
m-bert: `size` sounds more generic so I think it would be fine. But it's up to you, `dimensions` is also ok 😅

src/components/Pressable/PressableProps.tsx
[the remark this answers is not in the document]
latekvo: Alright, I think it's absolutely fair to remove those the same way `onBlur` and `onFocus` were removed.
I'll just wait for j-piasecki's opinion.

latekvo:
>If you perform a subtle click, Pressable will stay in pressed in state. Here is how it looks like.

Together with @j-piasecki we've determined this is an inner issue with gesture handler, and so I don't think it's worth fixing it here, as it should be addressed at it's root cause.

src/components/Pressable/PressableProps.tsx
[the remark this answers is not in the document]
latekvo: They work on web, but I think we already talked about it - to implement them, we would have to track the mouse outside of the component, which doesn't align with gesture handlers capabilities.
And on native, it doesn't work even with mouse hooked up, and I tested this on a real device.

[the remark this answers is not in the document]
m-bert: > (...) we would have to track mouse clicks outside of the component, which doesn't align with gesture handlers capabilities

Those two props are called `delayHoverIn` and `delayHoverOut`, so correct me if I'm wrong, but doesn't it just start `hover` after delay? If so, why would we need to track mouse clicks outside of the component (also, in case of `hover` there's no click, unless I'm missing something)?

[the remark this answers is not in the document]
latekvo: Yes, hover tracking is currently fully functional, I was talking about `onBlur` and `onFocus` which activate on mouse click outside and inside of the component.

Back to the `delayHoverIn`, as far as I see in the code, it should be working, but I'll recheck on Monday.

src/components/Pressable/PressableProps.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Whether a press gesture can be interrupted by a parent gesture such as a
+   * scroll event. Defaults to true.
+   */
+  cancelable?: null | boolean;
+
+  /**
+   * Duration to wait after hover in before calling `onHoverIn`.
+   * @platform web macos windows
```
m-bert, line 85: I still don't like the idea of putting `windows` here cc @j-piasecki

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,269 @@
...
+    normalizedHitSlop,
+    normalizedPressRetentionOffset
+  );
+
+  const isPressableEnabled = props.disabled !== true;
+
+  touchGesture.enabled(isPressableEnabled);
+  pressGesture.enabled(isPressableEnabled);
+  hoverGesture.enabled(isPressableEnabled);
+  rippleGesture.enabled(isPressableEnabled);
```
m-bert, line 199: If it was only enabled it would be fine, but we also have other properties set in that way. Wouldn't it be better to use loop?

```tsx
  const gestures = [touchGesture, pressGesture, hoverGesture, rippleGesture];

  for (const gesture of gestures) {
    gesture.runOnJS(true);

    if (Platform.OS !== 'web') {
      gesture.shouldCancelWhenOutside(true);
    }
  }

  rippleGesture.hitSlop(normalizedHitSlop);
```

m-bert (commented):
I've found one more bug (happens on `iOS` and `android`). If you click on `Pressable` while scrolling, press delay stops working (long press delay still works)

https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/33185fae-0d70-47ed-87d2-d0555489894e

src/components/Pressable/PressableProps.tsx
latekvo: alright

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,269 @@
...
+  );
+
+  const hoverGesture = useMemo(
+    () =>
+      Gesture.Hover()
+        .onBegin((event) => {
+          setTimeout(
+            () => props.onHoverIn?.(adaptStateChangeEvent(event)),
+            props.delayHoverIn ?? DEFAULT_HOVER_DELAY
+          );
```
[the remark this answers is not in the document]
latekvo: I agree, changing

src/components/Pressable/Pressable.tsx
latekvo: Agreed, at first I though wrapping the config into a loop would make it less readable, but now i see this is not the case.

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,269 @@
...
+import {
+  Insets,
+  Platform,
+  StyleProp,
+  View,
+  ViewStyle,
+  processColor,
+  StyleSheet,
+} from 'react-native';
+import RNButton from '../GestureHandlerButton';
```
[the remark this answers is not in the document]
latekvo: Maybe `NativeButton`?

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,269 @@
...
+  touchGesture.shouldCancelWhenOutside(true);
+  pressGesture.shouldCancelWhenOutside(true);
+  hoverGesture.shouldCancelWhenOutside(true);
+  rippleGesture.shouldCancelWhenOutside(true);
+
+  if (Platform.OS === 'web') {
+    touchGesture.shouldCancelWhenOutside(false);
+    pressGesture.shouldCancelWhenOutside(false);
+    hoverGesture.shouldCancelWhenOutside(false);
+  }
```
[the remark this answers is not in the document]
m-bert: > While it is default on browsers, do we want it to be?

Android also [does this by default](https://github.com/software-mansion/react-native-gesture-handler/blob/31a62e96a006ec655175aa8259160415e0f96988/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt#L64). Same for [iOS](https://github.com/software-mansion/react-native-gesture-handler/blob/31a62e96a006ec655175aa8259160415e0f96988/apple/RNGestureHandler.m#L101). The thing is, when it is not set to `false`, handlers will cancel when you move outside of the underlying view. It means that, for example, if user will be panning fast, gesture may be cancelled if he leaves view's boundaries. We don't want that behavior 😅

src/components/Pressable/Pressable.tsx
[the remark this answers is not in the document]
m-bert: I like that 👍

j-piasecki (commented):
When delay hover in/out is set, shouldn't it be canceled when moving outside of the view (cc @m-bert)?

Though I'm not exactly sure how to make it work when delay in is set but delay out is not and moving quickly over the pressable.

latekvo:
> When delay hover in/out is set, shouldn't it be canceled when moving outside of the view?

That's a good point. I have no idea how hover delay behaves on the legacy component, but I think that suggestion makes a bit more sense than the current implementation. I'll do that, shouldn't be a problem at all.

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,262 @@
...
+      ? props.children({ pressed: pressedState })
+      : props.children;
+
+  return (
+    <View style={styleProp}>
+      <GestureDetector gesture={gesture}>
+        <RNButton
+          ref={pressableRef}
+          testID={props.testID}
+          enabled={isPressableEnabled}
```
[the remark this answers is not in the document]
latekvo: I have no idea why I removed `hitSlop` from that button, I believe it was present there at one point.
Adding it back.

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,262 @@
...
+      props.onPress?.(adaptTouchEvent(event));
+    }
+
+    isPressedDown.current = false;
+    setPressedState(false);
+  }, []);
+
+  // fix for: touch out is called before touch in due to async .measure() in onTouchesDown()
+  const handlingOnTouchesDown = useRef<boolean>(false);
+  const onEndHandlingTouchesDown = useRef<() => void>(() => null);
```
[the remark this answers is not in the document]
latekvo: done

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,262 @@
...
+            if (props.unstable_pressDelay) {
+              pressDelayTimeoutRef.current = setTimeout(() => {
+                pressInHandler(event);
+              }, props.unstable_pressDelay);
+            } else {
+              pressInHandler(event);
+            }
+
+            onEndHandlingTouchesDown.current();
+            handlingOnTouchesDown.current = false;
```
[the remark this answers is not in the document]
latekvo: I see the issue, fixed it.

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+            if (props.unstable_pressDelay) {
+              pressDelayTimeoutRef.current = setTimeout(() => {
+                pressInHandler(event);
+              }, props.unstable_pressDelay);
+            } else {
+              pressInHandler(event);
+            }
+
+            onEndHandlingTouchesDown.current();
```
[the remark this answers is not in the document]
latekvo: did `onEndHandlingTouchesDown.current?.();`

m-bert:
> When delay hover in/out is set, shouldn't it be canceled when moving outside of the view?

I think that if you hover on `Pressable` that has `hoverDelayIn` and then you move outside it before callback will be triggered, then it shouldn't be called at all (i.e. it should be cancelled).

src/components/Pressable/Pressable.tsx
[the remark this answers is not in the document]
latekvo: Ok but as far as I see, the original `Pressable` has `shouldCancelWhenOutside` set to `true`, and we want the new `Pressable` to behave like the old one. I'm genuinely not sure if I'm missing something here.

https://github.com/software-mansion/react-native-gesture-handler/assets/74246391/0771cdf5-6ada-4068-99f1-fc92fc1b918c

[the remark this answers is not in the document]
latekvo: And to be clear, removing the `shouldCancelWhenOutside` block as you requested, so that the default value is used looks like this:


https://github.com/software-mansion/react-native-gesture-handler/assets/74246391/5d26ab46-c614-43ad-b6aa-d9cbe2e29e16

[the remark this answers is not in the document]
m-bert: The videos are captured on `iOS` and its implementation was fine. I was talking about web behavior. If you leave `shouldCancelWhenOutside` set to `false` on web, both `Pressables` act in the same way and this is what we want to achieve.

Of course I may be wrong, so If you see that something is not right with this approach feel free to tell me 😅 


https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/2d918e7a-d920-47d8-823a-3b4e94d0a250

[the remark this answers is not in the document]
latekvo: Alight i now understand what you meant all this time. I believe this feature is implemented as of now, sorry for the fuss 😄

src/components/Pressable/Pressable.tsx
```diff
@@ -56,12 +56,19 @@ export default function Pressable(props: PressableProps) {
     [isPressCallbackEnabled, props.onLongPress, isPressedDown]
   );
 
+  // clear timeout on exit,
```
m-bert, line 59:
```suggestion
  // clear timeout on exit
```

But I think that "on exit" doesn't fit well here 😅 Especially that `hoverOutTimeout` is cleared on "enter"

latekvo: updated to: `// cancel timeout when the other one is started`. I'm not sure if maybe it'd be better to delete it all together? What do you think?

m-bert: Well, I don't think this comment is necessary. On the other hand it would be nice to have it (cc @j-piasecki).

```
  // cancel timeout when the other one is started
```

I think it is more of a side effect.

<details>
<summary> If you take a look at the code  </summary>

```tsx
const hoverInTimeout = useRef<number | null>(null);
const hoverOutTimeout = useRef<number | null>(null);

const hoverGesture = useMemo(
  () =>
    Gesture.Hover()
      .onBegin((event) => {
        if (hoverOutTimeout.current) {
          clearTimeout(hoverOutTimeout.current);
        }
        if (props.delayHoverIn) {
          hoverInTimeout.current = setTimeout(
            () => props.onHoverIn?.(adaptStateChangeEvent(event)),
            props.delayHoverIn
          );
          return;
        }
        props.onHoverIn?.(adaptStateChangeEvent(event));
      })
```

</details>

`hoverOutTimeout` is cleared inside `onBegin` callback and there's no check if `hoverInTimeout` is started, so that comment is a bit misleading (at least for me).

latekvo: I agree, I think unlike the other comments which clear up non-oblivious mechanics, this one is just unnecessary.
I'll remove it for now, with possibility of adding it back later.

m-bert:
I've found 2 more problems:

### Functional styling (all platforms)

If you drag out of `Pressable` and return to it, next press doesn't work 

https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/16fe1fb3-05fb-4902-84e3-a28d1aec11ae

### Delay on press (web only)

Sometimes press stops working

https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/b91046c1-bbef-4e2f-852c-7ec57466ad5f

### Conclusion

I think we've done a lot in this PR and further issues may be addressed in following PRs (preferably with `tag` `Area: Pressable`. What's your stance on that, @j-piasecki?

m-bert:
Also I've noticed that if you use `stylus` and you click on "delay hover" example, when you lift stylus you will get `Hover out with delay registered`, without hover in being logged.

And last thing for now, we still have problems with borders on iOS:

image


> (...) I fixed this error since, so I think I can remove the borders example all together, as all the styles are now aligned with the legacy pressable.

Please, either remove the example or change borders to `solid`

latekvo:
> Functional styling (all platforms)

fixed, it was an issue with touch cancelling

> Delay on press (web only)

couldn't replicate it, could you tell me which browser you're using?

> Problems with borders on iOS

removed them

(c.c. @m-bert )

latekvo:
Regarding the `hover in/out` bug on web browser with stylus, I think what you described is normal behaviour.

When just clicking `(press in immediately followed by press out)`, well then before hoverIn's timer finishes `(press in)`, hoverOut's is started `(press out)`, thus only the latter will be displayed.

If you were to hold down the `hover` element, you'd also see `Hovered in with delay`.

This is in line with what you requested yesterday:

> I think that if you hover on Pressable that has hoverDelayIn and then you move outside it before callback will be triggered, then it shouldn't be called at all (i.e. it should be cancelled). - @m-bert 



I hope this explains the issue you've found.

m-bert:
> couldn't replicate it, could you tell me which browser you're using?

I'm using Google Chrome (126.0.6478.63).

> I hope this explains the issue you've found.

Makes sense 😅

m-bert:
I think this is yet another bug 😅 But as I said earlier, I'm not sure if we want to fix it in this PR

https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/e6bfbf42-6558-467b-a050-39e2627ef3cd

j-piasecki:
> I think we've done a lot in this PR and further issues may be addressed in the following PRs.

We can do that. Could you open issues for the problems you've found in that case?

m-bert:
> Could you open issues for the problems you've found in that case?

I will.

src/components/Pressable/Pressable.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const pressOutHandler = useCallback((event: GestureTouchEvent) => {
+    if (
+      !isPressedDown.current ||
+      event.allTouches.length > event.changedTouches.length
+    ) {
+      return;
+    }
+
+    if (props.unstable_pressDelay && pressDelayTimeoutRef.current !== null) {
+      // if pressDelay is set, we want to call onPressIn on touch up
```
m-bert, line 110: It would be nice if you could also explain in that comment why we want to do this 😅

example/src/release_tests/gesturizedPressable/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+    marginBottom: 5,
+    color: '#555',
+  },
+  testSandbox: {
+    marginTop: 5,
+    display: 'flex',
+    alignItems: 'center',
+  },
+  separator: {
+    // ios bug - different border styling doesn't work on just 1 side
```
m-bert, line 120:
```suggestion
```

I don't think it is necessary anymore

m-bert (approved):
Good job! There are few things that we have to polish though - let's do this in follow ups 😅

src/components/Pressable/Pressable.tsx
latekvo: This is a comment regarding how the legacy version behaved.
I changed it to:
```
      // legacy Pressable behaviour - if pressDelay is set, we want to call onPressIn on touch up
```

m-bert: What I meant was to explain why we want to call `onPressIn` on touch up, because it is not that obvious at first sight. We believe that the reason is because if you click on `Pressable` that has delay and then you lift the finger, you want to receive `pressIn` immediately (cc @j-piasecki). Adding that this is legacy behavior doesn't change much if you ask me.

latekvo: Replaced it with this:

image

I hope that plays a good role of explaining the launching of `pressInHandler` inside `pressOutHandler`

m-bert: It's fine. It is only a comment after all 😅

PR #2943, opened 2024-06-13
Unify scaled coordinates between `web` and `native` side.

m-bert:
## Description

On `android` and `iOS` if you scale view and get relative coords, they will act like they were not scaled at all. Let's say that we have a square with size `100x100`. If you click in the middle, `x` and `y` fields in event will both be `50`. Now, if you scale this view by 2 (i.e. new size is `200x200`) and click in the middle, you'll still get values of 50.

Web takes scale into account and on the above situation, second output would be `100`. To unify this behavior we now divide coords by scale factors.

## Test plan

Add `console.log(e.x, e.y)` in any example with scale, for example in `transformations`

### DIFF
--- a/src/web/tools/PointerEventManager.ts
+++ b/src/web/tools/PointerEventManager.ts
@@ -1,7 +1,11 @@
 import EventManager from './EventManager';
 import { MouseButton } from '../../handlers/gestureHandlerCommon';
 import { AdaptedEvent, EventTypes, Point } from '../interfaces';
-import { PointerTypeMapping, isPointerInBounds } from '../utils';
+import {
+  PointerTypeMapping,
+  calculateViewScale,
+  isPointerInBounds,
+} from '../utils';
 import { PointerType } from '../../PointerType';
 
 const POINTER_CAPTURE_EXCLUDE_LIST = new Set<string>(['SELECT', 'INPUT']);
@@ -234,11 +238,13 @@ export default class PointerEventManager extends EventManager<HTMLElement> {
   }
 
   protected mapEvent(event: PointerEvent, eventType: EventTypes): AdaptedEvent {
+    const { scaleX, scaleY } = calculateViewScale(this.view);
+
     return {
       x: event.clientX,
       y: event.clientY,
-      offsetX: event.offsetX,
-      offsetY: event.offsetY,
+      offsetX: event.offsetX / scaleX,
+      offsetY: event.offsetY / scaleY,
       pointerId: event.pointerId,
       eventType: eventType,
       pointerType:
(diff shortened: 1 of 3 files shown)


src/web/utils.ts
```diff
@@ -18,3 +18,36 @@ export const degToRad = (degrees: number) => (degrees * Math.PI) / 180;
...
+
+  const scales = styles.scale.split(' ');
+
+  if (scales[0] !== 'none') {
+    resultScales.scaleX *= parseFloat(scales[0]);
+  }
+
+  if (scales[1]) {
+    resultScales.scaleY *= parseFloat(scales[1]);
+  }
```
latekvo, line 38: Why is the `0` element checked with `!== 'none'` while the `1` element is checked with a boolean conversion?

Also, if you're multiplying numbers by 1, wouldn't it be better to set these values instead of multiplying them? 
`resultScales.scaleX *= parseFloat(scales[0])` should be equivalent to `resultScales.scaleX = parseFloat(scales[0])`, since as far as i see, `resultScales.scaleX` is always `1`

m-bert: Good catch! Done in [0474578](https://github.com/software-mansion/react-native-gesture-handler/pull/2943/commits/0474578f94034f04e14f7fc9c88b2223de989674).

latekvo (approved):
lgtm 🪨

src/web/tools/PointerEventManager.ts
```diff
@@ -234,11 +238,13 @@ export default class PointerEventManager extends EventManager<HTMLElement> {
...
   protected mapEvent(event: PointerEvent, eventType: EventTypes): AdaptedEvent {
+    const { scaleX, scaleY } = calculateViewScale(this.view);
+
     return {
       x: event.clientX,
       y: event.clientY,
-      offsetX: event.offsetX,
-      offsetY: event.offsetY,
+      offsetX: event.offsetX / scaleX,
+      offsetY: event.offsetY / scaleY,
```
j-piasecki, line 247: Wouldn't this also affect other places? I'm not sure in which cases relative coords are used on web.

m-bert: Right now it is used only in `PointerTracker` to get relative coords. Since relative coords have just been introduced, this change shouldn't affect anything.

j-piasecki (approved):
We also need something like that for rotation, don't we?

m-bert:
@j-piasecki Probably we do, I've missed that so thanks for reminding me. Also, I think it will be better to do it in other PR, I hope you don't mind.

PR #2957, opened 2024-06-24
Remove shared value read on the JS thread during detector update

j-piasecki:
## Description

In https://github.com/software-mansion/react-native-gesture-handler/pull/2037 support for memoizing gestures was added, which relied on reading the current handlers from the shared value and comparing them to the new ones.

This worked well with Reanimated 2, where shared values kept a copy of their value on JS thread, and the assumption about Reanimated 3, was that even though the read is synchronous, it's quick enough not to cause problems. This assumption was wrong since it can result in JS thread being blocked until the UI thread finishes its current work to allow for SV read.

This PR eliminates the necessity of reading from the shared value by relying on information already available on the JS thread to decide whether the shared value should be updated.

## Test plan

Test on the Example app to make sure nothing breaks, and make sure that memorized gestures don't result in updating the shared value.

### DIFF
--- a/src/handlers/gestures/GestureDetector/updateHandlers.ts
+++ b/src/handlers/gestures/GestureDetector/updateHandlers.ts
@@ -40,8 +40,22 @@ export function updateHandlers(
     if (!preparedGesture.isMounted) {
       return;
     }
+
+    // if amount of gesture configs changes, we need to update the callbacks in shared value
+    let shouldUpdateSharedValueIfUsed =
+      preparedGesture.attachedGestures.length !== newGestures.length;
+
     for (let i = 0; i < newGestures.length; i++) {
       const handler = preparedGesture.attachedGestures[i];
+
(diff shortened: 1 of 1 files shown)


src/handlers/gestures/GestureDetector/updateHandlers.ts
```diff
@@ -59,32 +73,13 @@ export function updateHandlers(
...
-            newHandlersValue[i].gestureId !== previousHandlersValue[i].gestureId
-          ) {
-            shouldUpdateSharedValue = true;
-            break;
-          }
-        }
-      }
-
-      if (shouldUpdateSharedValue) {
+      if (shouldUpdateSharedValueIfUsed) {
```
AndreiCalazans, line 82: Nit you could move the newHandlersValue to within the `if (shouldUpdateSharedValueIfUsed)` to avoid that variable declaration when not needed.

AndreiCalazans (approved):
Thanks, this works well.

PR #2970, opened 2024-06-28
Resolve circular dependencies on JS side

latekvo:
This PR removes circular dependencies found via:

`yarn madge --extensions js,ts,tsx --circular src`


latekvo:
I among the 30+ files changed so far, i accidentally committed some, that were automatically fixed by eslint that don't really contribute to the goal of this PR.
And now that they are already here, I don't see why they should be removed.

src/RNGestureHandlerModule.web.ts
```diff
@@ -7,6 +7,7 @@ import type { Config } from './web/interfaces';
 import InteractionManager from './web/tools/InteractionManager';
 import NodeManager from './web/tools/NodeManager';
 import * as HammerNodeManager from './web_hammer/NodeManager';
+import * as createGestureHandler from './web_hammer/modifyGestureHandlerRegistry';
```
m-bert, line 10: I'd call it something like `HandlerRegistryManager`, or use `{}` syntax 

```tsx
import { createGestureHandler, ... } from './web_hammer/modifyGestureHandlerRegistry';
```

m-bert (commented):
Looks good, let's hear what @j-piasecki has to say 😅

src/handlers/gestureCommonUtils.ts
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react';
```
j-piasecki, line 1: I think just utils.ts would be fine

src/handlers/handlerTag.ts
```diff
@@ -0,0 +1,5 @@
+let handlerTag = 1;
```
j-piasecki, line 1: I'd stick to naming one-function-files with the name of the function

src/RNGestureHandlerModule.web.ts
```diff
@@ -48,7 +49,7 @@ export default {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       const GestureClass = HammerGestures[handlerName];
       // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-      HammerNodeManager.createGestureHandler(handlerTag, new GestureClass());
+      createGestureHandler.createGestureHandler(handlerTag, new GestureClass());
```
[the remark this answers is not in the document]
m-bert: While I agree that it would be better not to split `NodeManager`, this one only refers to `Hammer` implementation which we don't actively support right now. It may create a bit of a mess, but I believe it is worth it.

Have you tested old implementation @latekvo?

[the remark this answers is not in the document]
latekvo: I rolled back this splitting, as we fixed this particular circular dependency elsewhere.

src/web_hammer/NodeManager.ts
```diff
@@ -1,10 +1,4 @@
-import { ValueOf } from '../typeUtils';
-import { HammerGestures } from '../web/Gestures';
-
-const gestures: Record<
-  number,
-  InstanceType<ValueOf<typeof HammerGestures>>
-> = {};
+export const gestures: Record<number, any> = {};
```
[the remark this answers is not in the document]
latekvo: you're right, i rolled back this change

src/handlers/FlingGestureHandler.ts
```diff
@@ -1,3 +1,4 @@
+import { FlingGestureHandlerEventPayload } from './GestureHandlerEventPayload';
```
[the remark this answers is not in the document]
latekvo: i think i applied this requested change to all instances of imported files i committed in this PR

src/handlers/gestureCommonUtils.ts
latekvo: renamed

src/handlers/handlerTag.ts
latekvo: renamed

src/components/touchables/GenericTouchableProps.tsx
```diff
@@ -0,0 +1,26 @@
+import {
```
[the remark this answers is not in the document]
latekvo: i think i applied this suggestion to all relevant files in this PR

src/web_hammer/NodeManager.ts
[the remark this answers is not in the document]
j-piasecki: Is using `GestureHandler` instead of any adding another cycle?

j-piasecki:
@m-bert I think you mentioned adding a CI workflow dedicated to making sure we don't introduce another cycle by accident, is that happening?

src/web_hammer/NodeManager.ts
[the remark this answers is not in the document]
latekvo: Yes

m-bert:
@j-piasecki Yes, I did mention adding CI workflow 😅  In order for it to make sense we have to get rid of all circular dependencies. At the beginning we didn't care that much about the one connected with `Hammer`, so it was postponed. Now that I see that we did remove all circular dependencies, I think adding CI workflow would be good idea. 

Also, we could add it to pre-commit hook, let me know what do you think.

@latekvo could you add this workflow?

m-bert:
I think it would be nice to add it in this PR 😅

latekvo:
### New marge workflow

image

.github/workflows/static-root-checks.yml
```diff
@@ -29,3 +29,5 @@ jobs:
         run: yarn tsc --noEmit
       - name: Lint
         run: yarn lint:js-root
+      - name: Check for circular dependencies
+        run: yarn madge --extensions js,ts,tsx --circular src
```
m-bert, line 33: Would it be possible to add script in our `package.json` and then use it here? Like `yarn circular-dependency-check`. This way it would be easier to run this in our repo (we wouldn't have to remember all those arguments and so on).

latekvo: ok, will do

latekvo: don't we want a script for both the JS and the native side in this case? 
or is this a bit far fetched for now?

m-bert: Is `madge` capable of doing such things?

Also keep in mind that in order to add this script we would have to have 0 circular dependencies on native side (I don't say we have them, that's something that we have to check).

So in conclusion, it would be great to have such script for native, but not in this PR.

package.json
```diff
@@ -15,7 +15,8 @@
     "lint:js": "eslint --ext '.js,.ts,.tsx' src/ example/src FabricExample/src MacOSExample/src && yarn prettier --check './{src,example,FabricExample,MacOSExample}/**/*.{js,jsx,ts,tsx}'",
     "lint:js-root": "eslint --ext '.js,.ts,.tsx' src/ && yarn prettier --check './src/**/*.{js,jsx,ts,tsx}'",
     "lint:android": "./android/gradlew -p android spotlessCheck -q",
-    "checkIntegrity": "(cd ./FabricExample/android && ./gradlew generateCodegenArtifactsFromSchema -PskipCodegenCopyTask) && (cd ./android && ./gradlew checkIntegrityBetweenArchitectures)"
+    "checkIntegrity": "(cd ./FabricExample/android && ./gradlew generateCodegenArtifactsFromSchema -PskipCodegenCopyTask) && (cd ./android && ./gradlew checkIntegrityBetweenArchitectures)",
+    "checkForCircularDependencies": "yarn madge --extensions js,ts,tsx --circular src"
```
m-bert, line 19:
```suggestion
    "circularDependencyCheck": "yarn madge --extensions js,ts,tsx --circular src"
```

I think it is enough, we don't have to overcomplicate that name. Also I think using `-` as separator is more readable than camelCase, but since we have `checkIntegrity` script we can leave that (or we can change both, cc @j-piasecki)

.github/workflows/static-root-checks.yml
latekvo: Although `madge` likely won't work, for `kotlin` there is a tool called `Classycle` which could be made into a workflow.

But of course, i don't think doing this in this PR would be a good idea either 😄 
I was just asking if it was something you'd like to be done.

package.json
latekvo: Yep, i agree but preferred to follow the convention instead of having both cases in a single filename next to eachother. 

Same for the name, but since you prefer the latter one, I'll change it.

m-bert: Okay, I get what you mean by "same for the name" 😅 Currently we have only one script that starts with `check` so I think it is a bit different compared to `-` and camelCase. Starting with `check` prefix isn't that bad, but having many scripts starting with the same prefix would make `tab` autocomplete a bit frustrating 😅

This is a topic for discussion and I think that following `Reanimated` convention would be better, but let's hear @j-piasecki opinion on that.

j-piasecki: I think we can merge it as is, but create a follow-up that will take care of this, given that we already have a `ts-check` script (to use dashes instead of camelCase).

.github/workflows/static-root-checks.yml
j-piasecki: I think it could be done, though first, it would be great to check how many of them are in the codebase to have an idea of the effort it would take.

j-piasecki (approved):

m-bert (approved):

package.json
latekvo: https://github.com/software-mansion/react-native-gesture-handler/pull/2973
@j-piasecki

PR #2972, opened 2024-07-02
fix: bind scope to null to prevent issues with inline requires

EvanBacon:
## Description

- I've identified an issue with Metro's inlineRequires + experimentalImportSupport while building tree shaking support for Expo CLI. This issue causes an invalid invocation to be thrown because the scope of the build-in functions is not correct.
- More info can be found here https://github.com/EvanBacon/inline-requires-issue

## Test plan

- In my [repro](https://github.com/EvanBacon/inline-requires-issue) the bindings no longer throw. I've raised the issue with Metro team so we can create a system-wide fix, but this should prevent any issues in the meantime.

### DIFF
--- a/src/ghQueueMicrotask.ts
+++ b/src/ghQueueMicrotask.ts
@@ -1,6 +1,6 @@
 // We check for typeof requestAnimationFrame because of SSR
 export const ghQueueMicrotask =
-  typeof requestAnimationFrame === "function"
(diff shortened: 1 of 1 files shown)


m-bert (approved):
Hi @EvanBacon! Thanks for this PR. I took the liberty of fixing prettier since I didn't want to bother you 😅

PR #2973, opened 2024-07-02
Change naming scheme in package.json from camelCase to dash-case

latekvo:
changes JS script naming scheme to be more uniform

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint:js-root": "eslint --ext '.js,.ts,.tsx' src/ && yarn prettier --check './src/**/*.{js,jsx,ts,tsx}'",
     "lint:android": "./android/gradlew -p android spotlessCheck -q",
     "check-integrity": "(cd ./FabricExample/android && ./gradlew generateCodegenArtifactsFromSchema -PskipCodegenCopyTask) && (cd ./android && ./gradlew checkIntegrityBetweenArchitectures)",
-    "circularDependencyCheck": "yarn madge --extensions js,ts,tsx --circular src"
+    "circular-dependency-check": "yarn madge --extensions js,ts,tsx --circular src"
   },
   "react-native": "src/index.ts",
   "main": "lib/commonjs/index.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -15,8 +15,8 @@
     "lint:js": "eslint --ext '.js,.ts,.tsx' src/ example/src FabricExample/src MacOSExample/src && yarn prettier --check './{src,example,FabricExample,MacOSExample}/**/*.{js,jsx,ts,tsx}'",
     "lint:js-root": "eslint --ext '.js,.ts,.tsx' src/ && yarn prettier --check './src/**/*.{js,jsx,ts,tsx}'",
     "lint:android": "./android/gradlew -p android spotlessCheck -q",
-    "checkIntegrity": "(cd ./FabricExample/android && ./gradlew generateCodegenArtifactsFromSchema -PskipCodegenCopyTask) && (cd ./android && ./gradlew checkIntegrityBetweenArchitectures)",
-    "circularDependencyCheck": "yarn madge --extensions js,ts,tsx --circular src"
+    "check-integrity": "(cd ./FabricExample/android && ./gradlew generateCodegenArtifactsFromSchema -PskipCodegenCopyTask) && (cd ./android && ./gradlew checkIntegrityBetweenArchitectures)",
```
m-bert, line 18:
```suggestion
    "integrity-check": "(cd ./FabricExample/android && ./gradlew generateCodegenArtifactsFromSchema -PskipCodegenCopyTask) && (cd ./android && ./gradlew checkIntegrityBetweenArchitectures)",
```

We do have `ts-check` and `circular-dependency-check`, so I think we could name it `integrity-check` (cc @j-piasecki)

latekvo: i agree

j-piasecki: I'd go even further and change it to `cross-arch-integrity-check` - it must have slipped past the original review, as the name `checkIntegrity ` doesn't really say much. (Integrity of what?)

latekvo: Sure, especially since `integrity-check` is used by yarn already: `yarn install ... --skip-integrity-check`

m-bert (approved):

Issue #2979, opened 2024-07-04
[Pressable] Styles are not applied 


m-bert:
### Description

Some of CSS styles do not work with `Pressables`. It happens because `style` property is [applied to outer `View`](https://github.com/software-mansion/react-native-gesture-handler/blob/4ce89e3a36234a8c59dc007f4667821c5e9cec5f/src/components/Pressable/Pressable.tsx#L258) and not the button itself 

<img width="2496" alt="image" src="https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/d7c7a9af-ca36-4909-aac8-64f4326271e6">

### Steps to reproduce

<details>
<summary> Test on the following code </summary>

```tsx
import React from 'react';
import { StyleSheet, Text, View } from 'react-native';
import { Pressable } from 'react-native-gesture-handler';

const onPressCallback = (type: string) =>
  console.log(`${type} pressable onPress`);

const onPressInCallback = (type: string) =>
  console.log(`${type} pressable onPressIn`);

const onPressOutCallback = (type: string) =>
  console.log(`${type} pressable onPressOut`);

export default function App() {
  return (
    <View style={styles.container}>
      <Pressable
        onPress={() => onPressCallback('outer')}
        onPressIn={() => onPressInCallback('outer')}
        onPressOut={() => onPressOutCallback('outer')}
        style={styles.outerPressable}>
        <View style={styles.box} />
      </Pressable>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },

  outerPressable: {
    width: 150,
    height: 150,
    backgroundColor: 'lightpink',

    display: 'flex',
    justifyContent: 'center',
    alignItems: 'center',
  },

  box: {
    width: 100,
    height: 100,
    backgroundColor: 'plum',
  },
});

```

</details> 

### Snack or a link to a repository

Irrelevant

### Gesture Handler version

2.17.1

### React Native version

0.74.1

### Platforms

Android, iOS, Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

Issue #2980, opened 2024-07-04
[Pressable] Nested `Pressables` don't behave as the ones from RN


m-bert:
### Description

When you nest `Pressables` the inner one should activate. Right now you can see logs form both - inner and outer one.

https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/6e72e451-b014-44bf-8291-1c26792a1a84

### Steps to reproduce

<details>
<summary> Use the following code to test this issue: </summary>

```tsx

import React from 'react';
import { StyleSheet, View, Text } from 'react-native';
import { Pressable as GHPressable } from 'react-native-gesture-handler';
import { Pressable as RNPRessable } from 'react-native';

const onPressCallback = (type: string) =>
  console.log(`${type} pressable onPress`);

const onPressInCallback = (type: string) =>
  console.log(`${type} pressable onPressIn`);

const onPressOutCallback = (type: string) =>
  console.log(`${type} pressable onPressOut`);

export default function App() {
  return (
    <View style={styles.container}>
      <View style={styles.innerContainer}>
        <Text style={styles.text}> Gesture Handler Pressable </Text>
        <GHPressable
          onPress={() => onPressCallback('outer')}
          onPressIn={() => onPressInCallback('outer')}
          onPressOut={() => onPressOutCallback('outer')}
          style={styles.outerPressable}>
          <GHPressable
            onPress={() => onPressCallback('inner')}
            onPressIn={() => onPressInCallback('inner')}
            onPressOut={() => onPressOutCallback('inner')}
            style={styles.innerPressable}
          />
        </GHPressable>
      </View>

      <View style={styles.innerContainer}>
        <Text style={styles.text}> React Native Pressable </Text>

        <RNPRessable
          onPress={() => onPressCallback('outer')}
          onPressIn={() => onPressInCallback('outer')}
          onPressOut={() => onPressOutCallback('outer')}
          style={styles.outerPressable}>
          <RNPRessable
            onPress={() => onPressCallback('inner')}
            onPressIn={() => onPressInCallback('inner')}
            onPressOut={() => onPressOutCallback('inner')}
            style={styles.innerPressable}
          />
        </RNPRessable>
      </View>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    flexDirection: 'row',
    justifyContent: 'space-evenly',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },

  innerContainer: {
[...]

PR #2982, opened 2024-07-04
Fix inner styling of Pressable

latekvo:
### References
closes: #2979
mentioned in: #2980

## Description
All styles given to Pressable are applied exclusively to the outer View component.

## Test plan
Pull commit [734c429](https://github.com/software-mansion/react-native-gesture-handler/pull/2982/commits/734c42965f28bac1bcec1ef8ff12309e6337b8ef) to see the issue.
Pull commit [36f1a24](https://github.com/software-mansion/react-native-gesture-handler/pull/2982/commits/36f1a247e469146be038d89271e889c7d6a335e8) to see the same example with this issue resolved. 

Both commits contain examples reproducing this issue under: 
`Example app > Release tests > Pressable Gesturization > Flex styling`.

In the commit [without a fix applied](https://github.com/software-mansion/react-native-gesture-handler/pull/2982/commits/734c42965f28bac1bcec1ef8ff12309e6337b8ef), you can see the square red shape used for presentation is misaligned with it's parent, which is caused by inner box-model styles being assigned to the outer `View` container.

This behaviour does on occur on the [fixed commit](https://github.com/software-mansion/react-native-gesture-handler/pull/2982/commits/36f1a247e469146be038d89271e889c7d6a335e8).


m-bert:
I don't think we need example for this. Also calling it `flexExample` may not be the best - other props may not work as well (check `padding` for example).

It happens because styles passed to `Pressables` are [applied to the outer](https://github.com/software-mansion/react-native-gesture-handler/blob/4ce89e3a36234a8c59dc007f4667821c5e9cec5f/src/components/Pressable/Pressable.tsx#L258) `View` and not the button itself.

m-bert:
I've also updated title and description of #2979 since they suggested that problem lies only in `flex`

latekvo:
>I don't think we need example for this

I agree, that's just a work-in-progress kind of thing to see if the issue is resolved as I am switching between branches and would prefer not to use shelves which tend to be problematic.

I'll make sure to remove it before making the branch ready for review.

m-bert:
> I agree, that's just a work-in-progress kind of thing to see if the issue is resolved as I am switching between branches and would prefer not to use shelves which tend to be problematic.

Ok, It was a bit too early to mention that, sorry for bothering you 😅

j-piasecki:
Not sure if it helps but Kacper was doing something kinda similar with border styling - https://github.com/software-mansion/react-native-gesture-handler/pull/2691

latekvo:
> Not sure if it helps but Kacper was doing something kinda similar with border styling

This is extremely helpful, I see Kacper used `StyleSheet.flatten` to reduce `StyleProp<ViewStyle>` into `ViewStyle` which is the exact thing I was looking for and couldn't find.

src/components/Pressable/Pressable.tsx
```diff
@@ -254,8 +257,14 @@ export default function Pressable(props: PressableProps) {
       ? props.children({ pressed: pressedState })
       : props.children;
 
+  // StyleProp<ViewStyle> is a broad umbrella type for objects, recursive arrays and numbers
```
m-bert, line 260: I'm not sure if we need this one 😅

src/components/Pressable/utils.ts
```diff
@@ -119,12 +119,80 @@ const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
...
+  'paddingStart',
+  'paddingTop',
+  'paddingVertical',
+  'start',
+  'end',
+
+  /**
+   * @platform ios
+   */
+  'direction',
```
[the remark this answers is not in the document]
latekvo: I agree, changed to the suggested comment. Still waiting for j-piasecki's opinion.

src/components/Pressable/Pressable.tsx
latekvo: Removed.

src/components/Pressable/utils.ts
```diff
@@ -119,12 +119,80 @@ const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
   };
 };
 
+type StylePropKeys = (keyof ViewStyle)[];
+
+const innerStyleKeys = [
```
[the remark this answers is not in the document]
latekvo: I added the following:
```
// Source:
// - From ViewStyle extracted FlexStyle sub-interface which contains all of the box-model manipulating props.
// - From FlexStyle handpicked those styles, which act on the inner part of the box-model.
```
As the comment says, these keys were handpicked out of a mixed bag of box-model props acting on both the inner and the outer behaviour.

src/components/Pressable/utils.ts
```diff
@@ -119,12 +119,80 @@ const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
...
+const excludeStyles = (from: ViewStyle, keys: StylePropKeys): ViewStyle => {
+  if (!from) {
+    return {} as ViewStyle;
+  }
+
+  const exclusiveData = { ...from };
+
+  for (const key of keys) {
+    if (exclusiveData[key] !== undefined) {
+      exclusiveData[key] = undefined;
```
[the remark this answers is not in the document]
latekvo: Unfortunately I couldn't use `delete` even though it would clear up debugging, due to eslint rule which forbids deleting dynamically computed keys:
image

src/components/Pressable/utils.ts
```diff
@@ -119,12 +119,80 @@ const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
...
+
+  const extractedData = {} as Record<string, unknown>;
+
+  for (const key of keys) {
+    if (from[key] !== undefined) {
+      extractedData[key] = from[key];
+    }
+  }
+
+  return extractedData as ViewStyle;
```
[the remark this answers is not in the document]
latekvo: I converted both of the functions to use a more funcitonal approach, please let me know if you'd like something changed:

```js
const extractStyles = (from: ViewStyle, keys: StylePropKeys): ViewStyle => {
  if (!from) {
    return {} as ViewStyle;
  }

  return Object.fromEntries(
    keys
      .map((key) => [key, from[key]])
      .filter(([_key, value]) => value !== undefined)
  ) as ViewStyle;
};

const excludeStyles = (from: ViewStyle, keys: StylePropKeys): ViewStyle => {
  if (!from) {
    return {} as ViewStyle;
  }

  const exclusiveData = { ...from };

  keys
    .filter((key) => exclusiveData[key] !== undefined)
    .forEach((key) => (exclusiveData[key] = undefined));

  return exclusiveData;
};
```

src/components/Pressable/utils.ts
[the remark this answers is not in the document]
m-bert: In that case I'd consider adding `eslint-disable`. I'm not a fan of assigning `undefined` to CSS styles as it happens now (unless I've missed something) cc @j-piasecki

src/components/Pressable/utils.ts
```diff
@@ -119,12 +119,72 @@ const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
...
+
+const extractStyles = (from: ViewStyle, keys: StylePropKeys): ViewStyle => {
+  if (!from) {
+    return {} as ViewStyle;
+  }
+
+  return Object.fromEntries(
+    keys
+      .map((key) => [key, from[key]])
+      .filter(([_key, value]) => value !== undefined)
```
j-piasecki, line 160:
```suggestion
    keys
      .filter((key) => key in from)
      .map((key) => [key, from[key]])
```

I am not sure about whether the `in` syntax is better, but the order should be swapped.

src/components/Pressable/utils.ts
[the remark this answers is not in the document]
j-piasecki: I don't think this is relevant anymore as it's changed to the functional approach.

src/components/Pressable/utils.ts
```diff
@@ -119,12 +119,72 @@ const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
...
+
+const excludeStyles = (from: ViewStyle, keys: StylePropKeys): ViewStyle => {
+  if (!from) {
+    return {} as ViewStyle;
+  }
+
+  const exclusiveData = { ...from };
+
+  keys
+    .filter((key) => exclusiveData[key] !== undefined)
```
j-piasecki, line 172:
```suggestion
    .filter((key) => key in exclusiveData)
```

src/components/Pressable/Pressable.tsx
```diff
@@ -254,8 +257,13 @@ export default function Pressable(props: PressableProps) {
       ? props.children({ pressed: pressedState })
       : props.children;
 
+  const flattenedStyles = StyleSheet.flatten((styleProp ?? {}) as ViewStyle);
+
+  const innerStyles = extractStyles(flattenedStyles, innerStyleKeys);
+  const outerStyles = excludeStyles(flattenedStyles, innerStyleKeys);
```
j-piasecki, line 263: Since styles are getting divided into two exclusive groups, wouldn't it be faster to implement one method that would divide them in a single pass? Something like this:

```jsx
const [innerStyles, outerStyles] = dividePressableStyle(flattenedStyles)
```

For now, since it's not used anywhere else, I'd keep `innerStyleKeys` internally in utils and reference directly.

WDYT?

src/components/Pressable/utils.ts
```diff
@@ -119,12 +119,72 @@ const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
   };
 };
 
+type StylePropKeys = (keyof ViewStyle)[];
+
+// Source:
+// - From ViewStyle extracted FlexStyle sub-interface which contains all of the box-model manipulating props.
+// - From FlexStyle handpicked those styles, which act on the inner part of the box-model.
+const innerStyleKeys = [
```
j-piasecki, line 127: Then, if doing a single method to divide styles, this could be a set instead of an array for O(1) access time.

src/components/Pressable/utils.ts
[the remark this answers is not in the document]
m-bert: Partially, we still have assigning `undefined` in `outerStyles`.

The thing is, I'm not 100% sure if passing `undefined` into styles will break something or not. I did get an error once, but it mentioned `translationX` so it was probably due to something else (though I got it in `Pressable` example). If we are sure that it doesn't break anything then I'm fine with that 😅

src/components/Pressable/utils.ts
latekvo: swapped the order.

src/components/Pressable/utils.ts
latekvo: replaced `!==` with `in`

src/components/Pressable/Pressable.tsx
latekvo: Yep, I combined both methods in [5595026](https://github.com/software-mansion/react-native-gesture-handler/pull/2982/commits/55950265f13d9d3bff235719fc5889dd20b219c8), but I'm still yet to optimize it to create both parts in a single pass instead of doing it in 2 passes.

latekvo: As of [e348d8f](https://github.com/software-mansion/react-native-gesture-handler/pull/2982/commits/e348d8fcd1eacb14e9c36cc4bd7e51ad44e602ea), both `inner` and `outer` style separation is combined into a single `forEach` statement.

src/components/Pressable/utils.ts
latekvo: Are you sure this is something we want?
We never access any members of this array besides iterating over the entire thing, 
at which point I don't think it can be optimised further than to `O(n)`, which I believe it currently is.

src/components/Pressable/utils.ts
```diff
@@ -119,6 +119,56 @@ const adaptTouchEvent = (event: GestureTouchEvent): PressableEvent => {
...
+  const outerStyles = { ...from };
+
+  const innerStyles = Object.fromEntries(
+    innerStyleKeys
+      .filter((key) => key in from)
+      .map((key) => {
+        outerStyles[key] = undefined;
+        return [key, from[key]];
+      })
+  ) as ViewStyle;
```
j-piasecki, line 167: I had something more like this in mind:

```jsx
const outerStyles = {}
const innerStyles = {}

for (const key in from) {
  if (innerStyleKeys.has(key)) {
    innerStyles[key] = from[key]
  } else {
    outerStyles[key] = from[key]
  }
}
```

I think it's a bit more concise and avoids the dirty side-effect inside `map`. But it requires `innerStyleKeys` to have O(1) time access, hence set in the other comment.

src/components/Pressable/utils.ts
j-piasecki: See https://github.com/software-mansion/react-native-gesture-handler/pull/2982#discussion_r1668703763

src/components/Pressable/utils.ts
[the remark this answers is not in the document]
latekvo: The error you're talking about originated from a bad argument being passed to an `Animated` function which we don't use anywhere in `Pressable`, I think it was likely caused by a different component.

src/components/Pressable/Pressable.tsx
```diff
@@ -131,6 +144,7 @@ export default function Pressable(props: PressableProps) {
   const touchGesture = useMemo(
     () =>
       Gesture.Manual()
+        .cancelsTouchesInView(true)
```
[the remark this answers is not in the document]
latekvo: My bad, missed it when applying a dated stash.

src/components/Pressable/utils.ts
latekvo: Done, working in: [75d4109](https://github.com/software-mansion/react-native-gesture-handler/pull/2982/commits/75d41093958d014b2a2c7f96cad9a1067b7904fb)

src/components/Pressable/Pressable.tsx
```diff
@@ -256,8 +266,12 @@ export default function Pressable(props: PressableProps) {
       ? props.children({ pressed: pressedState })
       : props.children;
 
+  const flattenedStyles = StyleSheet.flatten((styleProp ?? {}) as ViewStyle);
```
j-piasecki, line 269: Is this cast needed here?

latekvo: removed in [4846fd7](https://github.com/software-mansion/react-native-gesture-handler/pull/2982/commits/4846fd7832f7144eecdfb52ca10058641b0aca68) along with one other unnecessary type cast

latekvo:
Received a go-ahead from m-bert. Merging.

PR #2984, opened 2024-07-06
Unify comments and simplify some conditions.

m-bert:
## Description

This PR changes 3 things:

1. Unifies comments convention across `ts`/`tsx` files.
2. Simplifies one `return` statement in `GestureHandler` on `web`
3. Simplifies `userSelect` assignment in `GestureHandlerWebDelegate`

1. leads to many changed files, so if you think that 2. and 3. deserve their own PRs I'll open them 😅 

### Comments

Changed things:

- start with capital letter
- use `//` instead of `///`
- insert space between `//` and actual comment 

It also adds `ESlint` rule `spaced-comment`, which enforces spaces between `//` and comment content.

### `return`

I've removed `if` condition and moved it right to `return` statement

<details>
<summary> Before </summary>

```tsx
if (
  offsetX >= left &&
  offsetX <= right &&
  offsetY >= top &&
  offsetY <= bottom
) {
  return true;
}
return false;
```

</details>

<details>
<summary> After </summary>

```tsx
return (
  offsetX >= left && offsetX <= right && offsetY >= top && offsetY <= bottom
);
```

</details>

### `userSelect` assignment

I've changed `userSelect` assignment in `GestureHandlerWebDelegate` to use `??` operator

<details>
<summary> Before </summary>

```tsx
if (!config.userSelect) {
  this.view.style['webkitUserSelect'] = 'none';
  this.view.style['userSelect'] = 'none';
} else {
  this.view.style['webkitUserSelect'] = config.userSelect;
  this.view.style['userSelect'] = config.userSelect;
}
```

</details>

<details>
<summary> After </summary>

```tsx
this.view.style['userSelect'] = config.userSelect ?? 'none';
this.view.style['webkitUserSelect'] = config.userSelect ?? 'none';
```

</details>

## Test plan

Verified that example app works as it used to.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -64,6 +64,7 @@
         "ts-expect-error": "allow-with-description"
       }
     ],
-    "curly": "error"
+    "curly": "error",
+    "spaced-comment": "error"
   }
 }
--- a/src/Directions.ts
+++ b/src/Directions.ts
@@ -3,7 +3,7 @@ const LEFT = 2;
 const UP = 4;
 const DOWN = 8;
 
-// public interface
+// Public interface
 export const Directions = {
   RIGHT: RIGHT,
   LEFT: LEFT,
@@ -11,7 +11,7 @@ export const Directions = {
   DOWN: DOWN,
 } as const;
 
-// internal interface
+// Internal interface
 export const DiagonalDirections = {
   UP_RIGHT: UP | RIGHT,
   DOWN_RIGHT: DOWN | RIGHT,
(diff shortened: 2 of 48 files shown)


j-piasecki (approved):
We can make a CI out of this

m-bert:
> We can make a CI out of this

I've partially did this. I don't think adding CI for capital letters makes sense (for example we have some `eslint-disable` comments).

For the key part I've added new `ESlint` rule called `spaced-comment` - it enforces spaces after `//`. It will also catch `///` stuff and similar problems. Unfortunately it is not run with `lint-staged` so it won't be caught in pre-commit, but it will be checked on CI.

Issue #1263, opened 2020-12-10
ScrollView on iOS don't call onHandlerStateChange


mixail-novikov:
## Description
SUBJ

But I see that after the 1.9.0 version there are some changes in the repo, which may solve the problem. I run examples in the repo, they show the same result - the bug is persisted. Probably native changes weren't pushed to cocoapods. When the next version of react-native-gesture-handler planned to be published.

### Expected behavior

onHandlerStateChange runs the same on iOS and android

### Actual behavior

onHandlerStateChange runs on android, but not on iOS

## Snack or minimal code example
https://snack.expo.io/@mixail/scrollview-bug

## Package versions
The same behavior on 1.9.0, tested locally without expo, bare react-native
The version in snack ~1.7.0

- React:
- React Native:
- React Native Gesture Handler:

tomekzaw:
Hello @mixail-novikov! Thanks for reporting. Did you check if it works correctly on a newer version? The issue is quite outdated now so let me just close it with a suggestion to upgrade to react-native-gesture-handler v2.3.2.

j-piasecki:
That's still a thing on GH 2.4 😞. I'll see what can be done about it.

PR #2985, opened 2024-07-10
Fix scroll view handler state change events

coado:
## Description
  
Regarding this [issue](https://github.com/software-mansion/react-native-gesture-handler/issues/1263), this PR fixes the events that are sent from iOS whenever the state change occurs in a RCTScrollView.  

## Test plan
Tested on simulator and real device.

### DIFF
--- a/apple/Handlers/RNNativeViewHandler.mm
+++ b/apple/Handlers/RNNativeViewHandler.mm
@@ -85,7 +85,11 @@
 
 - (BOOL)isScrollView
 {
+#ifdef RCT_NEW_ARCH_ENABLED
+  return [self.view isKindOfClass:[RCTScrollViewComponentView class]];
+#else
   return [self.view isKindOfClass:[RCTScrollView class]];
+#endif
 }
 
 @end
--- a/apple/Handlers/RNNativeViewHandler.mm
+++ b/apple/Handlers/RNNativeViewHandler.mm
@@ -56,7 +56,7 @@
 
   // For now, we are handling only the scroll view case.
   // If more views need special treatment, then we can switch to a delegate pattern
-  if ([_gestureHandler retrieveScrollViewFromRecognizer:self] == nil) {
+  if ([_gestureHandler retrieveScrollView:self.view] == nil) {
     [self reset];
   }
 }
@@ -77,7 +77,7 @@
 
 - (void)updateStateIfScrollView
 {
-  UIScrollView *scrollView = [_gestureHandler retrieveScrollViewFromRecognizer:self];
+  UIScrollView *scrollView = [_gestureHandler retrieveScrollView:self.view];
   if (!scrollView)
     return;
   for (UIGestureRecognizer *scrollViewGestureRecognizer in scrollView.gestureRecognizers) {
@@ -134,20 +134,8 @@
   // We can restore default scrollview behaviour to delay touches to scrollview's children
   // because gesture handler system can handle cancellation of scroll recognizer when JS responder
   // is set
-#ifdef RCT_NEW_ARCH_ENABLED
-  if ([view isKindOfClass:[RCTScrollViewComponentView class]]) {
-    UIScrollView *scrollView = ((RCTScrollViewComponentView *)view).scrollView;
-    scrollView.delaysContentTouches = YES;
-  }
-#else
-  if ([view isKindOfClass:[RCTScrollView class]]) {
-    // This part of the code is coupled with RN implementation of ScrollView native wrapper and
-    // we expect for RCTScrollView component to contain a subclass of UIScrollview as the only
-    // subview
-    UIScrollView *scrollView = [view.subviews objectAtIndex:0];
-    scrollView.delaysContentTouches = YES;
-  }
-#endif // RCT_NEW_ARCH_ENABLED
+  UIScrollView *scrollView = [self retrieveScrollView:view];
+  scrollView.delaysContentTouches = YES;
 }
 
 - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
--- a/apple/Handlers/RNNativeViewHandler.mm
+++ b/apple/Handlers/RNNativeViewHandler.mm
@@ -39,14 +39,13 @@
 
 - (void)touchesBegan:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
 {
-  [self setStateIfScrollView:UIGestureRecognizerStatePossible];
   [_gestureHandler setCurrentPointerType:event];
   [_gestureHandler.pointerTracker touchesBegan:touches withEvent:event];
 }
 
 - (void)touchesMoved:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
 {
-  [self setStateIfScrollView:UIGestureRecognizerStateChanged];
+  [self updateStateIfScrollView];
   [_gestureHandler.pointerTracker touchesMoved:touches withEvent:event];
 }
 
@@ -57,7 +56,7 @@
 
   // For now, we are handling only the scroll view case.
   // If more views need special treatment, then we can switch to a delegate pattern
-  if (![self isScrollView]) {
+  if ([_gestureHandler retrieveScrollViewFromRecognizer:self] == nil) {
     [self reset];
   }
 }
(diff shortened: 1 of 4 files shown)


apple/RNGestureHandler.m
```diff
@@ -474,6 +474,11 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
     return YES;
   }
 
+  if ([otherGestureRecognizer.view isKindOfClass:[UIScrollView class]] &&
```
j-piasecki, line 477: Checking for `UIScrollView` may not be enough, can we check it by the class?

apple/RNGestureHandler.m
```diff
@@ -474,6 +474,11 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
+      [gestureRecognizer isKindOfClass:[RNDummyGestureRecognizer class]]) {
```
j-piasecki, line 478: This also will not be enough - it makes every dummy recognizer simultaneous with everything that's attached to a scroll view. Can you add a check to make sure they are attached to the same view?

apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -69,6 +76,18 @@ - (void)reset
...
+- (void)setStateIfScrollView:(UIGestureRecognizerState)state
+{
+  if ([self isScrollView]) {
+    self.state = state;
+  }
+}
+
+- (BOOL)isScrollView
+{
+  return [self.view isKindOfClass:[RCTScrollView class]];
```
j-piasecki, line 88: On the new architecture you need to check for `RCTScrollViewComponentView`, I think there's a similar check already in the same file so you can do it similarly.

apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -39,20 +39,27 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
 
 - (void)touchesBegan:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
 {
+  [self setStateIfScrollView:UIGestureRecognizerStatePossible];
```
j-piasecki, line 42: I don't think this is needed. `UIGestureRecognizerStatePossible` should be the starting state by default.

apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -39,20 +39,27 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
...
   [_gestureHandler setCurrentPointerType:event];
   [_gestureHandler.pointerTracker touchesBegan:touches withEvent:event];
 }
 
 - (void)touchesMoved:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
 {
+  [self setStateIfScrollView:UIGestureRecognizerStateChanged];
```
j-piasecki, line 49: Now that I think about it, can we just copy the state of the scroll view's recognizer here?

apple/RNGestureHandler.mm
```diff
@@ -495,6 +506,37 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
+      }
+    }
+#else
+    if ([gestureRecognizerView isKindOfClass:[RCTScrollView class]]) {
+      UIScrollView *scrollView = [gestureRecognizerView.subviews objectAtIndex:0];
+      if (scrollView == otherGestureRecognizerView) {
+        return YES;
+      }
+    }
+#endif
```
j-piasecki, line 534: Assuming copying the state from the scroll's recognizer works, we would need that logic repeated the third time. Can we move this to a helper function that would do these checks and return `UIScrollView` if it can be extracted (`nil` otherwise)?

apple/RNGestureHandler.mm
```diff
@@ -495,6 +506,37 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
   return NO;
 }
 
+- (BOOL)areRecognizersCompatible:(UIGestureRecognizer *)gestureRecognizer
```
j-piasecki, line 509: I think the name should mention scroll view in some way since it's only used in that context.

apple/Handlers/RNNativeViewHandler.mm
coado: Yes, I've just checked and it is the starting state.

apple/RNGestureHandler.mm
coado: Changed to "areScrollViewRecognizersCompatible"

apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -69,6 +75,18 @@ - (void)reset
   [_gestureHandler reset];
 }
 
+- (void)updateStateIfScrollView
+{
+  UIScrollView *scrollView = [_gestureHandler retrieveScrollView:self.view];
+  if (!scrollView)
+    return;
```
j-piasecki, line 82:
```suggestion
  if (!scrollView) {
    return;
  }
```

apple/RNGestureHandler.mm
```diff
@@ -495,6 +506,50 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
+}
+
+// is UIPanGestureRecognizer and has scrollView property
+- (BOOL)isUIScrollViewPanGestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
+{
+  if ([gestureRecognizer isKindOfClass:[UIPanGestureRecognizer class]] &&
+      [gestureRecognizer respondsToSelector:@selector(scrollView)]) {
+    return YES;
+  }
+  return NO;
```
j-piasecki, line 530:
```suggestion
  return [gestureRecognizer isKindOfClass:[UIPanGestureRecognizer class]] &&
      [gestureRecognizer respondsToSelector:@selector(scrollView)];
```

apple/RNGestureHandler.mm
```diff
@@ -495,6 +506,50 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
+}
+
+- (UIScrollView *)retrieveScrollView:(UIView *)view
```
j-piasecki, line 533: I believe this line breaks macOS compilation. You should be able to replace `UIView` with [`RNGHUIView`](https://github.com/software-mansion/react-native-gesture-handler/blob/5a671e9f0b02adfa4c565d0141cb6c01aaa3c0cc/apple/RNGHUIKit.h).

You will also need to do something about `UIScrollView` which will not exist on macOS. You should be able to use the `TARGET_OS_OSX` flag and conditionally replace it with `NSScrollView`.

apple/RNGestureHandler.mm
```diff
@@ -495,6 +506,50 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
...
+    if (scrollView && scrollView == otherGestureRecognizer.view) {
+      return YES;
+    }
+  }
+
+  return NO;
+}
+
+// is UIPanGestureRecognizer and has scrollView property
+- (BOOL)isUIScrollViewPanGestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
```
j-piasecki, line 524: This may also break on macOS.

I guess since there is no touch interaction with scroll views on macOS this entire method can simply accept any `NSGestureRecognizer` and return `NO` on mac (cc. @m-bert)

m-bert: I agree. Right now macOS is still in early stage of development, so we don't have to worry about this logic at the moment 😅

apple/RNGestureHandler.h
```diff
@@ -9,6 +9,12 @@
 #import <Foundation/Foundation.h>
 #import <React/RCTConvert.h>
 
+#if TARGET_OS_OSX
+#define TargetScrollView NSScrollView
+#else
+#define TargetScrollView UIScrollView
```
j-piasecki, line 15:
```suggestion
#define RNGHUIScrollView NSScrollView
#else
#define RNGHUIScrollView UIScrollView
```

I'd stick to the convention that's already in use (maybe dropping UI from the name, not strong opinions here), and move it to the other file with macOS compat stuff.

coado: I've changed it to RNGHUIScrollView. We already have RNGHUIView, so I think we can keep the UI in the name.

j-piasecki (approved):

PR #2962, opened 2024-06-26
Add docs page for Reanimated Swipeable

latekvo:
This PR adds documentation page for the new Reanimated `Swipeable` component.

PR #2992, opened 2024-07-15
Add documentation page for `Pressable` component

latekvo:
## Description

This PR adds a documentation page for the new gesturized version of the `Pressable` component


latekvo:
I'll change the gif to be whiter.

latekvo:
Please let me know if the gif is too big, small, dark or yellow. 
Comparing it to `components/buttons` gif it's just the right color but larger, 
comparing it to `components/swipeable` it's darker than it and similar size width-wise.

latekvo:
To address the mixed capitalization - in other component docs like [Reanimated Swipeable](https://github.com/software-mansion/react-native-gesture-handler/pull/2962), `Swipeable`, `Drawer Layout` and `Buttons`, prop descriptions always (with few exceptions) start with a small letter while all the rest of the article uses letter capitalisation.

I applied this scheme to this page as well - prop descriptions start with a small letter while all the rest of the sentences start with capitalised ones.

example/src/new_api/pressable/index.tsx
m-bert, line 1: A bit out of scope (we can leave it though)

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,162 @@
...
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import GifGallery from '@site/components/GifGallery';
+
+<GifGallery>
+  <img src={useBaseUrl('gifs/pressable.gif')} width="70%" />
+</GifGallery>
+
+:::info
+This component is a drop-in replacement for the `Pressable` component, rewritten using `Gesture Handler`.
+:::
```
m-bert, line 16: Do we need that information? And if we do, I don't think we have to say that it was rewritten using Gesture Handler - it should be pretty obvious.

```suggestion
:::info
This component is a drop-in replacement for the `Pressable` component.
:::
```

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+### `onHoverOut` (Web only)
+
+called when pointer stops hovering over the element.
+
+### `delayHoverIn` (Web only)
+
+:::note
+this prop is not present in the original pressable's React Native documentation.
+:::
```
m-bert, line 75: I don't think we need those admonitions

```suggestion
```

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+duration to wait after hover in before calling `onHoverIn`.
+
+### `delayHoverOut` (Web only)
+
+:::note
+this prop is not present in the original pressable's React Native documentation.
+:::
```
m-bert, line 83:
```suggestion
```

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,162 @@
...
+### `onLongPress`
+
+called immidiately after pointer has been down for at least `delayLongPress` milliseconds (`500` ms by default).
+
+after `onLongPress` has been called, `onPressOut` will be called as soon as the pointer is lifted and `onPress` will no longer be called at all.
+
+### `cancelable`
+
+whether a press gesture can be interrupted by a parent gesture such as a
+scroll event. Defaults to true.
```
m-bert, line 61:
```suggestion
scroll event. Defaults to `true`.
```

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,162 @@
...
+additional distance outside of this view in which a press is detected and `onPressIn` is triggered.
+
+### `pressRetentionOffset`
+
+additional distance outside of this view (or `hitSlop` if present) in which a touch is considered a
+press before `onPressOut` is triggered.
+
+### `android_disableSound` (Android only)
+
+if true, doesn't play system sound on touch.
```
m-bert, line 106:
```suggestion
if `true`, doesn't play system sound on touch.
```

m-bert (commented):
I'm not a fan of having so many admonitions. IMHO in that case they make docs less readable and people will pay less attention to the ones on other pages. I may be wrong though (cc @kacperkapusciak)

docs/docs/components/pressable.mdx
latekvo: With this one I tried to follow `Reanimated Swipeable` documentation which features almost identical info box:
https://github.com/software-mansion/react-native-gesture-handler/blob/673e037b977fa586ac3ada235ee501b91b4e2320/docs/docs/components/reanimated_swipeable.md?plain=1#L15

docs/docs/components/pressable.mdx
latekvo: I agree, removed

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+```js
+import { Pressable } from 'react-native-gesture-handler';
+```
+
+## Properties
+
+### `children`
+
+either children or a render prop that receives a boolean reflecting whether
```
j-piasecki, line 32:
```suggestion
either children or a render function that receives a boolean reflecting whether
```

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+duration (in milliseconds) from `onPressIn` before `onLongPress` is called.
+
+### `disabled`
+
+whether the `Pressable` behavior is disabled.
+
+### `hitSlop`
+
+additional distance outside of this view in which a press is detected and `onPressIn` is triggered.
```
j-piasecki, line 89: Would be nice to mention what is the type of this prop. It reads as if only numbers are supported, but you can also pass a `Rect` here.

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+### `pressRetentionOffset`
+
+additional distance outside of this view (or `hitSlop` if present) in which a touch is considered a
```
j-piasecki, line 93: Same

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,154 @@
...
+press before `onPressOut` is triggered.
+
+### `android_disableSound` (Android only)
+
+if `true`, doesn't play system sound on touch.
+
+### `android_ripple` (Android only)
+
+enables the Android ripple effect and configures its color, radius and other parameters.
```
j-piasecki, line 102: Same

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+### `onPressOut`
+
+called before `onPress` when a touch is released.
+
+### `onLongPress`
+
+called immidiately after pointer has been down for at least `delayLongPress` milliseconds (`500` ms by default).
+
+after `onLongPress` has been called, `onPressOut` will be called as soon as the pointer is lifted and `onPress` will no longer be called at all.
```
j-piasecki, line 56:
```suggestion
after `onLongPress` has been called, `onPressOut` will be called as soon as the pointer is lifted, and `onPress` will not be called.
```

docs/docs/components/pressable.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+<GifGallery>
+  <img src={useBaseUrl('gifs/pressable.gif')} width="70%" />
+</GifGallery>
+
+:::info
+This component is a drop-in replacement for the `Pressable` component.
+:::
+
+`Pressable` is a component that can detect various stages of tap, press and hover interactions on any of its defined children.
```
j-piasecki, line 18:
```suggestion
`Pressable` is a component that can detect various stages of tap, press and hover interactions on any of its children.
```

I know it's the same as in the react native docs, but I don't get the `defined` part. WDYT?

latekvo: I agree, it's not really necessary. @m-bert wdyt?

docs/docs/components/pressable.mdx
latekvo: I'd prefer something more like:
```suggestion
after `onLongPress` has been called, `onPressOut` will be called as soon as the pointer is lifted and `onPress` will no longer be called until the end of the gesture.
```

I think a simple `will not be called.` may be confusing - can be interpreted as `will not be called at this moment`, but what should be implied here is that as soon as `onLongPress` gets called, `onPress` will no longer have a chance to be called until the person lifts all their fingers and placed them down again.

docs/docs/components/pressable.mdx
latekvo: Which one do you prefer?
image
image

I think the first one makes it easier to see that `Rect` is a clickable link.

docs/static/gifs/pressable.gif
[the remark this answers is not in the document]
latekvo: I think removing the entire border would look better. WDYT?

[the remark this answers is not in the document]
m-bert: It is up to you 😅 Border is fine, but it should be the same on all sides. If you want to remove it, feel free to do so.

docs/docs/components/pressable.mdx
m-bert: Let's remove that 😉

example/src/new_api/pressable/index.tsx
latekvo: I wanted the example to be better suited for documentation, and I think it's a change small enough to not be worth it's own PR.

m-bert (commented):
LGTM, next time please make a separate PR to update example 😉

I'll take back _**Request Changes**_ after @j-piasecki approval 😄

docs/docs/components/pressable.mdx
j-piasecki: I see, `will not be called at all` then?

docs/docs/components/pressable.mdx
j-piasecki: +1 for the second one

docs/docs/components/pressable.mdx
latekvo: Sure, that sounds the best out of all of them

j-piasecki (approved):
Looks good

Please don't merge yet, same as with Swipeable - the docs will get published automatically after merge. We want to wait until it's released with that.

m-bert (approved):

Issue #2808, opened 2024-03-15
Android: touch events are executed twice when talkback is enabled


IvanIhnatsiuk:
### Description

Hello, thanks for your library it is really good and very powerful 😎 .

I found an issue that when talkback is enabled, the onPress call occurs multiple times when interacting with the buttons. 

https://github.com/software-mansion/react-native-gesture-handler/assets/86000012/8c37625f-75b7-4a8f-bce5-d20e29171b99

In the video the goBack callback was called twice, so I was navigated to the Accounts tab :)

### Steps to reproduce

1. Enable talkback on android emulator
2. Run your example application on android emulator/real device
3. Press on any button
4. observe that onPress was executed twice

### Snack or a link to a repository

https://github.com/software-mansion/react-native-gesture-handler

### Gesture Handler version

2.15.0

### React Native version

0.72.10

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

Pixel 7 pro

### Acknowledgements

Yes

m-bert:
Hi! I've just tested that on my Samsung S20+ and Pixel 6a emulator - both work fine. To test that, I've added logs into onPress callbacks in `RectButtons`. Could you provide a reproduction that we can look at?

IvanIhnatsiuk:
@m-bert Hello,  I think you can reproduce this in your example app. To reproduce this, I simply updated the App.tsx file to 

```
import React from 'react';
import {
  BorderlessButton,
  GestureHandlerRootView,
} from 'react-native-gesture-handler';
import { Text } from 'react-native';

const press = () => console.log('pressed');

export default function App() {
  return (
    <GestureHandlerRootView style={{ justifyContent: 'center', flex: 1 }}>
      <BorderlessButton onPress={press}>
        <Text>Press me</Text>
      </BorderlessButton>
    </GestureHandlerRootView>
  );
}
```


https://github.com/software-mansion/react-native-gesture-handler/assets/86000012/e529c3f2-9861-4c69-a248-1300606cc831

Please, let me know if you need more details :)

m-bert:
Hello @IvanIhnatsiuk! I've copied your code and everything is fine:

https://github.com/software-mansion/react-native-gesture-handler/assets/63123542/dabf65fe-29b7-41a2-a4c0-4b08580eccb5

[...]

PR #3002, opened 2024-07-23
Fix Android native buttons emitting 2 press events when talkbalk is enabled

latekvo:
## Description

When talkbalk accessability mode is enabled, pressing on any of the buttons provided by gesture handler (`BaseButton`, `RawButton`, `BorderlessButton`, Et Cetera) will cause 2 separate press events to go off.

This PR fixes that invalid behaviour.

Closes: #2808 

## Test plan

- enable talkback mode
- run the attached example
- see how the button labeled `BaseButton` executes twice on `main` and a single time on this branch.

## Attached example

```js
import React from 'react';
import { BaseButton } from 'react-native-gesture-handler';
import { Text, View, StyleSheet } from 'react-native';

const press = () => console.log('JS pressed');
const touch = () => console.log('JS touch');

  return (
    <View style={styles.root}>
      <BaseButton onPress={press}>
        <View
          accessible
          accessibilityRole="button"
          style={styles.button}
          onTouchStart={touch}>
          <Text>BaseButton</Text>
        </View>
      </BaseButton>
    </View>
  );
}

  root: {
    justifyContent: 'center',
    flex: 1,
  },
  button: {
    width: 200,
    height: 100,
    margin: 20,
    backgroundColor: 'cyan',
  },
});
```

### Normal behaviour (no talkback, no fix):
image

### Before fix:
image

### With fix:
image

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
@@ -1,16 +1,16 @@
 package com.swmansion.gesturehandler.core
 
-import android.content.Context
 import android.os.SystemClock
 import android.view.MotionEvent
 import android.view.View
 import android.view.ViewConfiguration
 import android.view.ViewGroup
-import android.view.accessibility.AccessibilityManager
 import android.widget.ScrollView
 import com.facebook.react.views.scroll.ReactScrollView
 import com.facebook.react.views.swiperefresh.ReactSwipeRefreshLayout
 import com.facebook.react.views.textinput.ReactEditText
+import com.swmansion.gesturehandler.react.RNGestureHandlerButtonViewManager
+import com.swmansion.gesturehandler.react.isScreenReaderOn
 
 class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
   private var shouldActivateOnStart = false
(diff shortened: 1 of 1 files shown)


latekvo:
Please ignore the linter issues for now, I'm working on getting openjdk running.

android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -82,6 +84,17 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
...
   override fun onHandle(event: MotionEvent, sourceEvent: MotionEvent) {
     val view = view!!
+
+    val isTouchExplorationEnabled = view.context.isScreenReaderOn()
+
+    if (view is RNGestureHandlerButtonViewManager.ButtonViewGroup && isTouchExplorationEnabled) {
+      // Fix for: https://github.com/software-mansion/react-native-gesture-handler/issues/2808
+      // When TalkBack is enabled, two identical press events are sent, while only one is expected.
+      // The unexpected one is caught by looking at the state of the current handler,
+      // which is UNDETERMINED when receiving the invalid event, and BEGAN when receiving the valid one.
```
m-bert, line 94: Please update this comment as it doesn't check `state` anymore 😅

m-bert (changes requested):
Looks good, but I think we should wait for @j-piasecki opinion 😅

j-piasecki (approved):

m-bert (approved):

PR #3014, opened 2024-07-29
[macOS] Add `LongPressGestureHandler`

m-bert:
## Description

This PR adds `LongPressGestureHandler` on `macOS`.

## Test plan

Tested on example App.


MacOSExample/src/basic/longPress/index.tsx
```diff
@@ -0,0 +1,36 @@
+import { StyleSheet, View } from 'react-native';
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
+
+export default function LongPressExample() {
+  const g = Gesture.LongPress()
+    .onBegin(() => console.log('onBegin'))
+    .onStart(() => console.log('onStart'))
```
j-piasecki, line 7: Why no `onEnd`?

apple/Handlers/RNLongPressHandler.m
```diff
@@ -94,6 +107,83 @@ - (void)touchesCancelled:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)ev
   [self reset];
 }
 
+#else
+- (void)mouseDown:(NSEvent *)event
+{
+  self.state = NSGestureRecognizerStateBegan;
+  startTime = CACurrentMediaTime();
```
j-piasecki, line 114: `previousTime` isn't set anywhere so the duration probably isn't right

apple/Handlers/RNLongPressHandler.m
```diff
@@ -94,6 +107,83 @@ - (void)touchesCancelled:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)ev
...
+
+  [self failOrCancelGesture];
+}
+
+- (void)failOrCancelGesture
+{
+  if (self.state == NSGestureRecognizerStateChanged) {
+    self.state = NSGestureRecognizerStateCancelled;
+  } else {
+    self.state = NSGestureRecognizerStateFailed;
```
j-piasecki, line 170: What happens when you change the state from `ended` to `failed`? I assume that's what's happening when letting go of the mouse button after long press finishes successfully.

apple/Handlers/RNLongPressHandler.m
```diff
@@ -94,6 +107,83 @@ - (void)touchesCancelled:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)ev
...
+  _initPosition = [self locationInView:self.view];
+
+  __weak typeof(self) weakSelf = self;
+
+  block = dispatch_block_create(0, ^{
+    __strong typeof(self) strongSelf = weakSelf;
+
+    if (strongSelf) {
+      strongSelf.state = NSGestureRecognizerStateChanged;
+      strongSelf.state = NSGestureRecognizerStateEnded;
```
j-piasecki, line 127: I'm not sure whether long press should end immediately after activation. I think on other platforms it stays active until the pointer is lifted.

MacOSExample/src/basic/longPress/index.tsx
m-bert: I've thought that `onFinalize` with `success` will be enough. Anyways, I thought of rewriting this example a bit so that it provides more visual experience than looking at console logs. I'll get back here when I finish 😅

apple/Handlers/RNLongPressHandler.m
m-bert: It seemed ok but I'll check that

MacOSExample/src/basic/longPress/index.tsx
m-bert: I've changed this example, let me know what do you think! ([9cfd459](https://github.com/software-mansion/react-native-gesture-handler/pull/3014/commits/9cfd4598c8ca150d649087bc9c8a558e4d98ebbe))

apple/Handlers/RNLongPressHandler.m
m-bert: I'm not exactly sure what happens, but I've noticed that `success` in `onFinalize` is set to `false` even if gesture has ended correctly. I've fixed that and turns out that it was because of this `if`. ([bab5b11](https://github.com/software-mansion/react-native-gesture-handler/pull/3014/commits/bab5b112b8e1161bc0ec81e5a4e1b8305585c8b6))

apple/Handlers/RNLongPressHandler.m
m-bert: Yes, you're right. Changed in [85fa302](https://github.com/software-mansion/react-native-gesture-handler/pull/3014/commits/85fa302b7a8c4f6b177509c2d5abbc689176fb7d)

apple/Handlers/RNLongPressHandler.m
m-bert: Actually it is set inside `handleGesture` method 😅

apple/Handlers/RNLongPressHandler.m
```diff
@@ -94,6 +107,86 @@ - (void)touchesCancelled:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)ev
...
+  }
+}
+
+- (void)failOrCancelGesture
+{
+  if (self.state == NSGestureRecognizerStateChanged) {
+    self.state = NSGestureRecognizerStateCancelled;
+  } else {
+    self.state = NSGestureRecognizerStateFailed;
+  }
```
j-piasecki, line 174: `failOrCancelGesture` is called in one place so I would either inline it or also make a separate method for the other if to stay consistent.

m-bert: I agree 😅

Done in [855c5ca](https://github.com/software-mansion/react-native-gesture-handler/pull/3014/commits/855c5ca93749525293cad51f53cb6c18bf606385)

j-piasecki (approved):

PR #3018, opened 2024-07-30
[macOS] Add `ManualGestureHandler`

m-bert:
## Description

This PR adds `ManualGestureHandler` on `macOS`.

## Test plan

Tested on example App.


apple/Handlers/RNManualHandler.m
```diff
@@ -19,25 +20,27 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
...
     _shouldSendBeginEvent = YES;
   }
+
   return self;
 }
 
-- (void)touchesBegan:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
+- (void)interactionsBegan:(NSSet *)touches withEvent:(UIEvent *)event
 {
-  [_gestureHandler setCurrentPointerType:event];
```
j-piasecki, line 27: Don't we need this?

m-bert: Correct me if I'm wrong, but it is still present in touchesBegan, just git shows this diff in a weird way

j-piasecki: Ah, I see. It makes sense to split that between implementations.

j-piasecki (approved):

PR #3027, opened 2024-08-02
[macOS] Add `HoverGestureHandler`

m-bert:
## Description

This PR adds `HoverGestureHandler` on `macOS`.

### Limitations

It seems that `Hover` doesn't work when wrapped in `ScrollView`. I'm yet to find out what causes that.

## Test plan

Tested on `Hover` example in example app.

### DIFF
--- a/apple/Handlers/RNHoverHandler.m
+++ b/apple/Handlers/RNHoverHandler.m
@@ -17,15 +17,11 @@
   defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_##__VERSION__) && \
       __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_##__VERSION__ && !TARGET_OS_TV
 
-#endif
-
 typedef NS_ENUM(NSInteger, RNGestureHandlerHoverEffect) {
   RNGestureHandlerHoverEffectNone = 0,
   RNGestureHandlerHoverEffectLift,
   RNGestureHandlerHoverEffectHightlight,
 };
-
-#if !TARGET_OS_OSX
 
 #if CHECK_TARGET(13_4)
 
--- a/apple/Handlers/RNHoverHandler.m
+++ b/apple/Handlers/RNHoverHandler.m
@@ -192,6 +192,7 @@ API_AVAILABLE(ios(13.4))
 - (void)unbindFromView
 {
   [_view removeTrackingArea:trackingArea];
+  _view = nil;
 }
 
 - (void)mouseEntered:(NSEvent *)event
--- a/apple/Handlers/RNHoverHandler.m
+++ b/apple/Handlers/RNHoverHandler.m
@@ -164,7 +164,10 @@ API_AVAILABLE(ios(13.4))
 
 #else
 
-@implementation RNHoverGestureHandler
+@implementation RNHoverGestureHandler {
+  NSTrackingArea *trackingArea;
+  RNGHUIView *_view;
+}
 
 - (instancetype)initWithTag:(NSNumber *)tag
 {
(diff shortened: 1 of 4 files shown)


apple/Handlers/RNHoverHandler.m
```diff
@@ -167,13 +168,51 @@ @implementation RNHoverGestureHandler
...
 }
 
+- (void)bindToView:(RNGHUIView *)view
+{
+  for (NSTrackingArea *trackingArea in view.trackingAreas) {
+    [view removeTrackingArea:trackingArea];
+  }
+
+  NSTrackingAreaOptions options =
+      NSTrackingMouseEnteredAndExited | NSTrackingActiveInActiveApp | NSTrackingInVisibleRect;
```
m-bert, line 185: Here we can use `NSTrackingActiveAlways` instead of `NSTrackingActiveInActiveApp` - it will result in `Hover` being active even if window is not focused. 

I'm not sure if we want this behavior though (and making it customizable only on `macOS` doesn't make much sense to me). Let me know what you think @j-piasecki

j-piasecki: I'd like to keep it consistent - if it always works on web on macOS then we can do that. If web requires a window to be focused then we can keep it this way.

apple/Handlers/RNHoverHandler.m
```diff
@@ -163,17 +164,58 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(UIGestureRecognizer *)recogn
...
+
+  trackingArea = [[NSTrackingArea alloc] initWithRect:_view.bounds options:options owner:self userInfo:nil];
+  [_view addTrackingArea:trackingArea];
+}
+
+- (void)unbindFromView
+{
+  [super unbindFromView];
+  [_view removeTrackingArea:trackingArea];
```
j-piasecki, line 195: If we need to keep a separate reference, it should be cleared here so as not to prolong the view's lifetime.

apple/Handlers/RNHoverHandler.m
```diff
@@ -163,17 +164,58 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(UIGestureRecognizer *)recogn
...
+  NSTrackingAreaOptions options =
+      NSTrackingMouseEnteredAndExited | NSTrackingActiveInActiveApp | NSTrackingInVisibleRect;
+
+  trackingArea = [[NSTrackingArea alloc] initWithRect:_view.bounds options:options owner:self userInfo:nil];
+  [_view addTrackingArea:trackingArea];
+}
+
+- (void)unbindFromView
+{
+  [super unbindFromView];
```
j-piasecki, line 194: I assume there's no `super` call in bindToView due to a recognizer not being used. If that's correct, then why do we call `super` here?

I guess the question still stands if the assumption is wrong 😅.

apple/Handlers/RNHoverHandler.m
```diff
@@ -163,17 +164,58 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(UIGestureRecognizer *)recogn
 
 #else
 
-@implementation RNHoverGestureHandler
+@implementation RNHoverGestureHandler {
+  NSTrackingArea *trackingArea;
+  RNGHUIView *_view;
```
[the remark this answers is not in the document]
m-bert: `Hover` does not use any recognizer at the moment 😅

apple/Handlers/RNHoverHandler.m
m-bert: Good point, I've added it automatically, without noticing that it requires recognizer.

Removed in [fd79077](https://github.com/software-mansion/react-native-gesture-handler/pull/3027/commits/fd79077dcd98fe0a78392c4cc7fdae66833bf9fe)

apple/Handlers/RNHoverHandler.m
m-bert: Done in [dd94802](https://github.com/software-mansion/react-native-gesture-handler/pull/3027/commits/dd94802be0e6d89e4ed201eb34f55ad02b351d34)

j-piasecki (approved):
Take a look here: https://github.com/software-mansion/react-native-gesture-handler/pull/3027#discussion_r1701998505

Otherwise, looks good

apple/Handlers/RNHoverHandler.m
m-bert: I've just checked that and it requires window to be active 😅

m-bert:
There still might be some problems with blocking other handlers, but I think that they can be addressed in other PRs if necessary 😅

PR #3028, opened 2024-08-05
[macOS] Add `FlingGestureHandler`

m-bert:
## Description

This PR adds `FlingGestureHandler` on `macOS`. From now, all handlers, except`ForceTouchGestureHandler`, are available on this platform 🥳 

## Test plan

Tested on newly added example.


apple/RNGHVector.h
```diff
@@ -0,0 +1,29 @@
...
+#ifndef RNGHVector_h
+#define RNGHVector_h
+
+@interface Vector : NSObject
+
+@property (atomic, assign) double x;
+@property (atomic, assign) double y;
+@property (atomic, assign) double unitX;
+@property (atomic, assign) double unitY;
+@property (atomic, assign) double magnitude;
```
j-piasecki, line 19: Shouldn't those also be `readonly`?

apple/RNGHVector.h
```diff
@@ -0,0 +1,29 @@
...
+
++ (Vector *)fromDirection:(RNGestureHandlerDirection)direction;
++ (Vector *)fromVelocityX:(double)vx withVelocityY:(double)vy;
+- initWithX:(double)x withY:(double)y;
```
j-piasecki, line 23:
```suggestion
- (nonnull instancetype)initWithX:(double)x withY:(double)y;
```

It's likely that the signature would also need to be updated in the implementation.

apple/Handlers/RNFlingHandler.m
```diff
@@ -158,17 +158,225 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(id)_recognizer
...
+      dispatch_time(DISPATCH_TIME_NOW, (int64_t)(maxDuration * NSEC_PER_SEC)),
+      dispatch_get_main_queue(),
+      failFlingAction);
+}
+
+- (void)mouseDragged:(NSEvent *)event
+{
+  [super mouseDragged:event];
+
+  self.state = NSGestureRecognizerStatePossible;
```
j-piasecki, line 233: I assume `mouseDragged` is called after `mouseDown`, if so then this is going backwards in the state machine.

apple/Handlers/RNFlingHandler.m
```diff
@@ -158,17 +158,225 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(id)_recognizer
...
+  NSPoint currentPosition = [self locationInView:self.view];
+  double currentTime = CACurrentMediaTime();
+
+  NSPoint distance;
+  distance.x = currentPosition.x - lastPosition.x;
+  distance.y = lastPosition.y - currentPosition.y;
+
+  double timeDelta = currentTime - lastTime;
+
+  Vector *velocityVector = [Vector fromVelocityX:(distance.x / timeDelta) withVelocityY:(distance.y / timeDelta)];
```
j-piasecki, line 244: Isn't that very susceptible to sudden changes in velocity/direction?

apple/Handlers/RNFlingHandler.m
```diff
@@ -158,17 +158,225 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(id)_recognizer
...
+{
+  double radians = (degrees * M_PI) / 180;
+  return cos(radians / 2);
+}
+
+@end
+
+#endif
+
 @implementation RNFlingGestureHandler
```
j-piasecki, line 312: It's implemented twice, as the compiler noticed.

apple/RNGestureHandlerDirection.h
```diff
@@ -1,8 +1,33 @@
...
+
 typedef NS_ENUM(NSInteger, RNGestureHandlerDirection) {
   RNGestureHandlerDirectionRight = 1,
   RNGestureHandlerDirectionLeft = 2,
   RNGestureHandlerDirectionUp = 4,
   RNGestureHandlerDirectionDown = 8,
+  RNGestureHandlerDirectionUpLeft = 16,
+  RNGestureHandlerDirectionUpRight = 32,
+  RNGestureHandlerDirectionDownLeft = 64,
+  RNGestureHandlerDirectionDownRight = 128,
```
j-piasecki, line 14: I believe this doesn't work - we use logical `or` to combine directions, so setting `Up | Left` on JS side will result in `6`, while here `16` is expected. Or am I missing something?

apple/RNGestureHandlerDirection.h
```diff
@@ -1,8 +1,33 @@
...
+    RNGestureHandlerDirectionUp,
+    RNGestureHandlerDirectionDown,
 };
+
+static NSInteger axialDirections[] = {
+    RNGestureHandlerDirectionUpLeft,
+    RNGestureHandlerDirectionUpRight,
+    RNGestureHandlerDirectionDownLeft,
+    RNGestureHandlerDirectionDownRight,
+};
```
latekvo, line 29: Shouldn't these be named the other way around?

```suggestion
static NSInteger axialDirections[] = {
    RNGestureHandlerDirectionRight,
    RNGestureHandlerDirectionLeft,
    RNGestureHandlerDirectionUp,
    RNGestureHandlerDirectionDown,
};

static NSInteger diagonalDirections[] = {
    RNGestureHandlerDirectionUpLeft,
    RNGestureHandlerDirectionUpRight,
    RNGestureHandlerDirectionDownLeft,
    RNGestureHandlerDirectionDownRight,
};
```

apple/RNGHVector.m
```diff
@@ -0,0 +1,67 @@
...
+    case RNGestureHandlerDirectionUpRight:
+      return [[Vector alloc] initWithX:1 withY:1];
+    case RNGestureHandlerDirectionDownLeft:
+      return [[Vector alloc] initWithX:-1 withY:-1];
+    case RNGestureHandlerDirectionDownRight:
+      return [[Vector alloc] initWithX:1 withY:-1];
+    default:
+      return [[Vector alloc] initWithX:0 withY:0];
+  }
+}
```
latekvo, line 50: Would it be possible to store static vector objects and cloning them when needed instead of allocating new ones each time?

I believe that's how it's done on Android's native side.

apple/RNGHVector.m
```diff
@@ -0,0 +1,67 @@
...
+- (id)initWithX:(double)x withY:(double)y
+{
+  if (self = [super init]) {
+    self.x = x;
+    self.y = y;
+
+    self.magnitude = hypot(x, y);
+
+    self.unitX = x / self.magnitude;
+    self.unitY = y / self.magnitude;
```
latekvo, line 22: I believe a division by `0` may happen here if a `x: 0, y: 0` `Vector` is constructed

apple/RNGHVector.h
m-bert: If I mark them as `readonly`, I get `Assignment to readonly property` error in `initWithX: withY:` method 😞

apple/RNGHVector.h
m-bert: Done in [3fcac48](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/3fcac48d3af08b460004341f0c62767de48a71ee)

m-bert: Done in [3fcac48](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/3fcac48d3af08b460004341f0c62767de48a71ee). 

IDE also pointed out that I have to update other signatures and specify whether the pointers can be null or not.

apple/Handlers/RNFlingHandler.m
m-bert: The only way to make it work without setting this state to `Possible` is to change it to `Possible` in `mouseDown` and don't change it here. At least this is what I got while experimenting with different state combinations.

Changed in [880e830](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/880e8309ba03a5c35960b086d4d7c63f579aded5)

apple/Handlers/RNFlingHandler.m
m-bert: It may be. This is definitely not the best solution, but the other one that I see is to re-write [VelocityTracker](https://github.com/software-mansion/react-native-gesture-handler/blob/40b4e11efb4bef0417cf2e2f14df5bc17253f8da/src/web/tools/VelocityTracker.ts). Is this something that we want to do?

apple/RNGestureHandlerDirection.h
m-bert: Now you've got me thinking. I agree that it shouldn't, but somehow it does. I'll check why and then get back here.

apple/RNGestureHandlerDirection.h
m-bert: You're absolutely right! Changed in [1570c9e](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/1570c9e5031cf4ad81ca6c28499986c6d98643dd)

apple/RNGHVector.m
m-bert: The thing is, this would require a lot of code and I'm not sure if performance will be much better. Do you think it will be worth to try @j-piasecki?

apple/RNGHVector.m
m-bert: Maybe, but will we ever construct `[0,0]` vector? We have two ways to construct vectors, i.e. from `direction` and from `velocity`. Making vector from direction won't lead to **0**. In case of `velocity`, it is created inside `mouseDragged` and I think it won't be called if mouse stands in place (so that `dx = dy = 0`)

m-bert: On the other hand, there might be some calculations/hardware related issues so that it will be called and translation won't be registered. I'll fix that.

latekvo: > Making vector from direction won't lead to 0.
It may, that's the default case for an invalid value being passed.
Someone would have to explicitly pass an invalid value, but this may occur by accidentally i think, if someone was to combine conflicting directions like `Left | Right`, then i believe that would default to a `0, 0` vector.

In that case the error would be enigmatic and wouldn't help the user fix their issue.

>I think it won't be called if mouse stands in place (so that dx = dy = 0)

I wouldn't entirely count on that.
It's mostly my gut feeling though 😄

m-bert: > It may, that's the default case for an invalid value being passed.

The only possible values that we give to the users are in `RNGestureHandlerDirection` enum, so I don't think invalid value can be passed here.

>if someone was to combine conflicting directions like Left | Right, then i believe that would default to a 0, 0 vector.

Why would `Left | Right` be conflicting? It means that `Fling` should activate with swiping either `Left` or `Right`. What is the conflict here?

apple/Handlers/RNFlingHandler.m
```diff
@@ -87,6 +87,157 @@ - (CGPoint)getLastLocation
...
+  if (!isAligned) {
+    for (int i = 0; i < directionsSize; ++i) {
+      if ([self getAlignment:axialDirections[i]
+              withMinimalAlignmentCosine:axialDeviationCosine
+                      withVelocityVector:velocityVector]) {
+        isAligned = YES;
+        break;
+      }
+    }
+  }
```
latekvo, line 213: Just an opinion, but I think it'd be preferable to check `axialDirections` first.

They are more likely to occur and these iterations are not all that cheap

apple/RNGHVector.m
latekvo: >Why would Left | Right be conflicting? It means that Fling should activate with swiping either Left or Right. What is the conflict here?

`Left | Right` are conflicting only when directly passed to the `fromDirection`, but you're right that this will never occur because `Fling`'s implementations already prevents that from happening.

Here's how such situation would look like:
```
int conflicting = RNGestureHandlerDirectionRight | RNGestureHandlerDirectionLeft;
Vector *directionVector = [Vector fromDirection:conflicting];
```

The code above compiles with no warnings and results in creation of the following vector:
image

Of course as you said, this situation is impossible with how the `Fling` handler works, and so it's not an immediate issue, but it's not preferable to have a global `Vector` class which could be used elsewhere, accepting invalid values without throwing any errors.

apple/RNGestureHandlerDirection.h
m-bert: Okay, in case of diagonals you're right, I've changed that in [a9bb4f6](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/a9bb4f69fd6ddeef5c791b37434cfe2f0145684b)

apple/Handlers/RNFlingHandler.m
m-bert: I agree, done in [6406cd0](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/6406cd06a948b49b43c54ed65926001d6125a5cf)

apple/RNGHVector.m
m-bert: Done in [6900228](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/6900228aa4663c8d2e327dcb5b8a5c39fbbeaaa5)

latekvo (approved):
Looking good, just make sure these are resolved before merging:
https://github.com/software-mansion/react-native-gesture-handler/pull/3028#discussion_r1723126296
https://github.com/software-mansion/react-native-gesture-handler/pull/3028#discussion_r1721276756

m-bert:
>Looking good, just make sure these are resolved before merging:

Those comments require response from @j-piasecki, so I believe he will address them when reviewing this PR 😅

apple/RNGHVector.h
j-piasecki: You would need to write to backing fields (`_x`, `_y`) inside the `Vector` class.

The property will implicitly generate a getter and setter around the backing fields which you can override and setting one to `readonly` will not generate a getter but inside the class, you can still modify it.

apple/Handlers/RNFlingHandler.m
j-piasecki: Weird, I'd expect the recognizer to already be in the `Possible` state at that point.

apple/RNGHVector.m
j-piasecki: Yeah, I'm not sure if it's worth the effort. If Vector was an object with heavy initialization then probably yes, but in this case it would add **a lot** of boilerplate (https://stackoverflow.com/a/40591961).

apple/Handlers/RNFlingHandler.m
j-piasecki: I think we were talking about using the whole distance divided by the entire time of the gesture, would that be hard to implement? It should be a relatively good approximation on the time scale we're operating here.

If not, then using `delta` is better than calculating it ourselves. I see that you already changed it in https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/79ab75c24bdd9a11049249dfddc3ac8aaca00c0f, could you then remove calculations for distance?

m-bert: Oops, I must have accidentally staged these changes as I remember that I wanted to ask you first about it 😅 

>I think we were talking about using the whole distance divided by the entire time of the gesture, would that be hard to implement?

Sounds like something that is possible, I'll check that 😅

m-bert: Okay, I've done it in [6a04fd6](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/6a04fd6ee2e35883a90323b4319310ca29712820), let me know if that is what you had in mind 😅

apple/RNGHVector.h
m-bert: Done in [9a3737c](https://github.com/software-mansion/react-native-gesture-handler/pull/3028/commits/9a3737cebef72e7704033d1327a9443c614f8691)

apple/Handlers/RNFlingHandler.m
m-bert: As discussed in person, we will use calculating total distance instead of difference between `mouseDragged` events.

j-piasecki (approved):

Issue #2990, opened 2024-07-12
`renderScrollComponent` prop of `FlatList` imported from `react-native-gesture-handler` does nothing and needs to be removed


thisisgit:
### Description

Hi,

I found that `FlatList` from RNGH accepts `renderScrollComponent` but it gets overwritten by its internal implementation
https://github.com/software-mansion/react-native-gesture-handler/blob/f0dfef28b2de2b2ec479f86732120b2bf2c2506b/src/components/GestureComponents.tsx#L119-L127

Since it doesn't warn or notify user that this prop isn't supported, I think it's better to omit the prop just like [what reanimated did to `CellRendererComponent` of `Animated.FlatList`](https://github.com/software-mansion/react-native-reanimated/pull/5951).

Found this issue from https://github.com/kirillzyusko/react-native-keyboard-controller/issues/505

### Steps to reproduce

n/a

### Snack or a link to a repository

n/a

### Gesture Handler version

2.16.1

### React Native version

0.74.2

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

PR #3029, opened 2024-08-05
Fix `FlatList` type

j-piasecki:
## Description

Closes https://github.com/software-mansion/react-native-gesture-handler/issues/2990

Removes `renderScrollComponent` from the public API of FlatList exported by Gesture Handler, as its being overwritten with Gesture Handler's scroll view anyway.

## Test plan

Verify that `renderScrollComponent` causes a TypeScript error on FlatList exported from RNGH.

### DIFF
--- a/src/components/GestureComponents.tsx
+++ b/src/components/GestureComponents.tsx
@@ -91,6 +91,12 @@ export const DrawerLayoutAndroid = createNativeWrapper<
 export type DrawerLayoutAndroid = typeof DrawerLayoutAndroid &
   RNDrawerLayoutAndroid;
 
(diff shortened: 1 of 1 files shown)


m-bert (commented):
I think it would be better to use `Omit`:

```tsx
props: PropsWithChildren<
    Omit<RNFlatListProps<ItemT>, 'renderScrollComponent'> &
      RefAttributes<FlatList<ItemT>> &
      NativeViewGestureHandlerProps
```

This way `renderScrollComponent` is truly removed from public API 😅

m-bert (approved):
🚀

PR #3035, opened 2024-08-07
Add keyboard support to gesture handler

latekvo:
## Description

By default, all native android and web components support keyboard navigation and option to press elements with `Enter` or `Spacebar` keys.

Gestures and components provided by Gesture Handler lack the latter option.

This PR adds keyboard support to the entirety of gesture handler, allowing it to replace mouse or finger navigation.

## Test plan

- try navigating around the example app using just the following keys: `Tab`, `Shift-Tab`, `Space`, `Enter`

### DIFF
--- a/src/PointerType.ts
+++ b/src/PointerType.ts
@@ -3,4 +3,5 @@ export enum PointerType {
   STYLUS,
   MOUSE,
   OTHER,
+  KEY,
 }
--- a/src/web/tools/KeyEventManager.ts
+++ b/src/web/tools/KeyEventManager.ts
@@ -1,0 +1,38 @@
+import { AdaptedEvent, EventTypes } from '../interfaces';
+import EventManager from './EventManager';
+import { PointerType } from '../../PointerType';
+
+export default class KeyEventManager extends EventManager<HTMLElement> {
+  private keyDownCallback = (event: unknown): void => {
+    const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.DOWN);
+    this.onPointerDown(adaptedEvent);
+  };
+  private keyUpCallback = (event: unknown): void => {
+    const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.UP);
+    this.onPointerUp(adaptedEvent);
+  };
+
+  public registerListeners(): void {
+    this.view.addEventListener('keydown', this.keyDownCallback);
+    this.view.addEventListener('keyup', this.keyUpCallback);
+  }
+
+  public unregisterListeners(): void {
+    this.view.addEventListener('keydown', this.keyDownCallback);
+    this.view.addEventListener('keyup', this.keyUpCallback);
+  }
+
+  protected mapEvent(event: unknown, eventType: EventTypes): AdaptedEvent {
+    console.log(event);
+    return {
+      x: 0,
+      y: 0,
+      offsetX: 0,
+      offsetY: 0,
+      pointerId: 0,
+      eventType: eventType,
+      pointerType: PointerType.KEY,
+      time: 0,
+    };
+  }
+}
--- a/src/web/tools/KeyEventManager.ts
+++ b/src/web/tools/KeyEventManager.ts
@@ -3,11 +3,22 @@ import EventManager from './EventManager';
 import { PointerType } from '../../PointerType';
 
 export default class KeyEventManager extends EventManager<HTMLElement> {
-  private keyDownCallback = (event: unknown): void => {
+  private activationKeys = ['Enter', 'Space'];
+
+  private keyDownCallback = (event: KeyboardEvent): void => {
+    if (this.activationKeys.indexOf(event.code) === -1) {
+      return;
+    }
+
     const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.DOWN);
     this.onPointerDown(adaptedEvent);
   };
-  private keyUpCallback = (event: unknown): void => {
+
+  private keyUpCallback = (event: KeyboardEvent): void => {
+    if (this.activationKeys.indexOf(event.code) === -1) {
+      return;
+    }
+
     const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.UP);
     this.onPointerUp(adaptedEvent);
   };
@@ -22,17 +33,24 @@ export default class KeyEventManager extends EventManager<HTMLElement> {
     this.view.addEventListener('keyup', this.keyUpCallback);
   }
 
-  protected mapEvent(event: unknown, eventType: EventTypes): AdaptedEvent {
-    console.log(event);
+  protected mapEvent(
+    event: KeyboardEvent,
+    eventType: EventTypes
+  ): AdaptedEvent {
+    const relativeX = this.view.offsetLeft + this.view.offsetWidth / 2;
+    const relativeY = this.view.offsetTop + this.view.offsetHeight / 2;
+    const viewportX = (event.view?.screenX ?? 0) + this.view.offsetWidth / 2;
+    const viewportY = (event.view?.screenY ?? 0) + this.view.offsetHeight / 2;
+
     return {
-      x: 0,
-      y: 0,
-      offsetX: 0,
-      offsetY: 0,
+      x: viewportX,
+      y: viewportY,
+      offsetX: relativeX,
+      offsetY: relativeY,
       pointerId: 0,
       eventType: eventType,
       pointerType: PointerType.KEY,
-      time: 0,
+      time: event.timeStamp,
     };
   }
 }
--- a/src/web/tools/GestureHandlerWebDelegate.ts
+++ b/src/web/tools/GestureHandlerWebDelegate.ts
@@ -11,6 +11,7 @@ import { isPointerInBounds } from '../utils';
 import EventManager from './EventManager';
 import { Config } from '../interfaces';
 import { MouseButton } from '../../handlers/gestureHandlerCommon';
+import KeyEventManager from './KeyEventManager';
 
 export class GestureHandlerWebDelegate
   implements GestureHandlerDelegate<HTMLElement, IGestureHandler>
@@ -46,6 +47,7 @@ export class GestureHandlerWebDelegate
 
     this.eventManagers.push(new PointerEventManager(this.view));
     this.eventManagers.push(new TouchEventManager(this.view));
+    this.eventManagers.push(new KeyEventManager(this.view));
 
     this.eventManagers.forEach((manager) =>
       this.gestureHandler.attachEventManager(manager)


latekvo:
As of [5564563](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/5564563877957383f2dd8bd9de39224131c59e14), it looks like Keyboard is fully functional on `web`

latekvo:
by pressing
- `enter` down
- `tab` down
- `tab` up
- `enter` up

element states gets stuck on `down` position.
I'll have to add `pointerCancelCallback` before pushing.

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,103 @@
...
+import EventManager from './EventManager';
+import { PointerType } from '../../PointerType';
+import { View } from 'react-native';
+
+export default class KeyEventManager extends EventManager<HTMLElement> {
+  private activationKeys = ['Enter', 'Space', 'NumpadEnter'];
+  private cancelationKeys = ['Tab'];
+
+  private keyDownCallback = (event: KeyboardEvent): void => {
+    console.log(event.code);
```
m-bert, line 11:
```suggestion
```

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,103 @@
...
+  private keyUpCallback = (event: KeyboardEvent): void => {
+    if (this.activationKeys.indexOf(event.code) === -1) {
+      return;
+    }
+
+    this.dispatchEvent(event, EventTypes.UP);
+  };
+
+  private dispatchEvent(event: KeyboardEvent, eventType: EventTypes) {
+    (event.target as unknown as View)?.measure?.(
```
m-bert, line 32: We are in `web` environment, can't we use `getBoundingClientRect`?

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,103 @@
...
+      (_x, _y, w, h, pageX, pageY) => {
+        const adaptedEvent = this.adaptEvent(
+          event,
+          eventType,
+          pageX,
+          pageY,
+          w,
+          h
+        );
```
m-bert, line 41: Wouldn't it be possible to somehow use `mapEvent` instead of adding another method?

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,103 @@
...
+import { PointerType } from '../../PointerType';
+import { View } from 'react-native';
+
+export default class KeyEventManager extends EventManager<HTMLElement> {
+  private activationKeys = ['Enter', 'Space', 'NumpadEnter'];
+  private cancelationKeys = ['Tab'];
+
+  private keyDownCallback = (event: KeyboardEvent): void => {
+    console.log(event.code);
+    if (this.cancelationKeys.indexOf(event.code) !== -1) {
```
m-bert, line 12: I'm not sure about using `event.code`. As far as I can see `keyCode` is deprecated, so we have to choose between `code` and `key`. The upside of `key` is that it takes keyboard layout into account (in contrast to `code`, as you can read in the [docs](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent)). On the other hand, `code` provides more information (like `AltLeft` instead of just `Alt`).

I'm not sure which one is better, but maybe that's not something that we should worry about right now (cc @j-piasecki)

m-bert (changes requested):

src/web/tools/KeyEventManager.ts
j-piasecki: I agree that we should use that instead of adding a new method. Measurement on web can be synchronous so that logic can be moved there so different signature shouldn't be a problem.

src/web/tools/KeyEventManager.ts
j-piasecki: Yes, but please use https://github.com/software-mansion/react-native-gesture-handler/blob/acf996d4c022e1c7c911f341521cb707ad7cb24e/src/web/tools/GestureHandlerDelegate.ts#L15

it uses `getBoundingClientRect` under the hood but abstracts away the implementation.

src/web/tools/KeyEventManager.ts
j-piasecki: Given that I know relatively nothing about different keyboard layouts, I'd stick to the one that takes it into account.

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,103 @@
...
+      this.dispatchEvent(event, EventTypes.CANCEL);
```
j-piasecki, line 13: Have you tested a scenario where one view is focused, Enter is pressed down, tab is pressed to focus another view, and then Enter is lifted?

If I understand correctly, the first view would receive DOWN and CANCEL events, but the second one would only receive an UP event. If that's the case, have you tested how different handlers behave in that scenario?

src/web/tools/KeyEventManager.ts
latekvo: Hey @j-piasecki thank you for your review.

I'm not sure if it's possible to use `measureView` inside `KeyEventManager`, as `measureView` is only present on the `GestureHandlerWebDelegate`, which we don't have access to.

I also don't think it would be possible to add access to it either, without creating a circular dependency, as `GestureHandlerWebDelegate` already imports `KeyEventManager`, and to attain this method `KeyEventManager` would have to also import `GestureHandlerWebDelegate`.

Please let me know if it's really necessary to `measureView` instead of `getBoundingClientRect`

src/web/tools/KeyEventManager.ts
latekvo: I agree, changing to `key`

src/web/tools/KeyEventManager.ts
latekvo: This cancelation logic handles the case you've mentioned. 
Without it, it was possible to press `Enter`, switch focus with `Tab` and after that it was impossible to use `Enter` again, and the previously pressed element stayed in a `pressed down` state.

As for an element only receiving the `UP` event, I haven't observed any concerning behaviour.
Afaik both `Gesture Handler` and native buttons handle such cases, but i'll fix this just to be safe.

src/web/tools/KeyEventManager.ts
latekvo: removed

src/web/tools/KeyEventManager.ts
latekvo: As of [1573a90](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/1573a9039ff361ae950150bcd70c22333aee9c33) I moved measurement logic to `mapEvent` and removed `adaptEvent` all together.

src/web/tools/KeyEventManager.ts
m-bert: I'd go with `getBoundingClientRect` - we already use it in [PointerEventManager](https://github.com/software-mansion/react-native-gesture-handler/blob/acf996d4c022e1c7c911f341521cb707ad7cb24e/src/web/tools/PointerEventManager.ts#L241)

src/web/tools/KeyEventManager.ts
latekvo: Changed as of [3cd0886](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/3cd0886fe23294aa50db89949b726db3d1fba487)

src/web/tools/KeyEventManager.ts
latekvo: As of [044cf87](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/044cf87ff6b36207e1794f07d731f6d0f1b4724a) added logic preventing calling UP and CANCEL when element is not pressed down.

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,82 @@
...
+  private keyUpCallback = (event: KeyboardEvent): void => {
+    if (this.activationKeys.indexOf(event.key) === -1 || !this.isPressed) {
+      return;
+    }
+
+    this.dispatchEvent(event, EventTypes.UP);
+  };
+
+  private dispatchEvent(event: KeyboardEvent, eventType: EventTypes) {
+    this.view.getBoundingClientRect();
```
m-bert, line 32:
```suggestion
```

latekvo: Fixed in [324090d](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/324090d69acd0d23f1280245e60a9e8c2cfd1961)

src/PointerType.ts
```diff
@@ -3,4 +3,5 @@ export enum PointerType {
   STYLUS,
   MOUSE,
   OTHER,
+  KEY,
```
m-bert, line 6: I'd swap the order and leave `OTHER` as last element.

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (this.activationKeys.indexOf(event.key) === -1 || !this.isPressed) {
+      return;
+    }
+
+    this.dispatchEvent(event, EventTypes.UP);
+  };
+
+  private dispatchEvent(event: KeyboardEvent, eventType: EventTypes) {
+    const adaptedEvent = this.mapEvent(event, eventType);
+    switch (eventType) {
```
m-bert, line 33:
```suggestion
    const adaptedEvent = this.mapEvent(event, eventType);
    
    switch (eventType) {
```

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,81 @@
...
+    const pointerPosition = {
+      x: viewRect.x + viewRect.width / 2,
+      y: viewRect.y + viewRect.height / 2,
+    };
+
+    return {
+      x: pointerPosition.x,
+      y: pointerPosition.y,
+      offsetX: pointerPosition.x,
+      offsetY: pointerPosition.y,
```
m-bert, line 74: `offsets` are measured with respect to `parent`, not `viewport`. I know it probably doesn't change much in case of keyboard, but I think we should stay consistent with that 😅 

```suggestion
      offsetX: viewRect.width / 2,
      offsetY: viewRect.height / 2,
```

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,81 @@
+import { AdaptedEvent, EventTypes } from '../interfaces';
+import EventManager from './EventManager';
+import { PointerType } from '../../PointerType';
+
+export default class KeyEventManager extends EventManager<HTMLElement> {
```
m-bert, line 5: Since we operate on `KeyboardEvents`, I think it would be better to call it `KeyboardEventManager`

```suggestion
export default class KeyboardEventManager extends EventManager<HTMLElement> {
```

m-bert (commented):
Overall looks good, please take a look at the comments. In the meantime I'm going to test that PR 😅

src/web/tools/KeyEventManager.ts
```diff
@@ -0,0 +1,81 @@
...
+  public unregisterListeners(): void {
+    this.view.addEventListener('keydown', this.keyDownCallback);
+    this.view.addEventListener('keyup', this.keyUpCallback);
+  }
+
+  protected mapEvent(
+    event: KeyboardEvent,
+    eventType: EventTypes
+  ): AdaptedEvent {
+    const viewRect = (event.target as HTMLElement)?.getBoundingClientRect?.();
```
m-bert, line 63: Is `?.` necessary after `getBoundingClientRect`?

src/PointerType.ts
latekvo: This order is intentional. I've seen this enum used on the native side, and I'm not sure web and native side interface `PointerType` at some point. 

I'm also not sure if this enum isn't based on some sort of built-in enum, which may directly interface with this one.
 
I'll have to double-check none of these cases occur before applying this change.
For now, keeping `KEY` on the last position ensures no such interference to occur.

src/web/tools/KeyEventManager.ts
latekvo: Yes, because `event.target` may not always be `HTMLElement` as described by [this documentation](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget).
I've checked if that occurs, and I've actually had that happen to me a couple of time during loading, which without `?.` resulted in a wall of red errors (although these were isolated and no crashes occured).

latekvo: Since we moved from `measure` to `getBoundingClientRect`, i'll add an additional null check to ensure no malformed events are sent.

src/web/tools/KeyEventManager.ts
latekvo: Renamed as of [76bf761](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/76bf76174eb4f1bf186f6ba07b3a368bf6925062)

src/web/tools/KeyEventManager.ts
latekvo: Fixed in [b76a0ef](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/b76a0ef94344932b1669b2dc7b0f5b6d5a813279)

src/web/tools/KeyEventManager.ts
latekvo: Good call, fixed this in [b76a0ef](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/b76a0ef94344932b1669b2dc7b0f5b6d5a813279)

src/web/tools/KeyEventManager.ts
latekvo: Added updated null-check as of [0a5855f](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/0a5855fa4cd9450aa41ff5bdc5999c06e06db4da)

src/PointerType.ts
m-bert: > I'm also not sure if this enum isn't based on some sort of built-in enum, which may directly interface with this one.

It's not 😅 But good catch that the order might matter. I'm not sure if that would be better to keep it, or change the order and then make changes in relevant places (cc @j-piasecki)

latekvo: Couldn't find any causes for concern, and so i applied your request in [aff6119](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/aff611955e6e7b0ae663f4f461becc9fded0dd28)

src/web/tools/KeyboardEventManager.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (this.cancelationKeys.indexOf(event.key) !== -1 && this.isPressed) {
+      this.dispatchEvent(event, EventTypes.CANCEL);
+      return;
+    }
+
+    if (this.activationKeys.indexOf(event.key) === -1) {
+      return;
+    }
+
+    this.dispatchEvent(event, EventTypes.DOWN);
```
m-bert, line 20: As an alternative we could do:

```suggestion
    if (this.cancelationKeys.indexOf(event.key) !== -1 && this.isPressed) {
      this.dispatchEvent(event, EventTypes.CANCEL);
    } else if (this.activationKeys.indexOf(event.key) !== -1) {
      this.dispatchEvent(event, EventTypes.DOWN);
    }
```
which is a bit shorter. It's up to you though 😅

src/web/tools/KeyboardEventManager.ts
```diff
@@ -0,0 +1,93 @@
...
+  };
+
+  private dispatchEvent(event: KeyboardEvent, eventType: EventTypes) {
+    const adaptedEvent = this.mapEvent(event, eventType);
+
+    if (isNaN(adaptedEvent.x)) {
+      // in rare cases when event.target is not a HTMLElement
+      // we want to return before sending out an event
+      return;
+    }
```
m-bert, line 38: In that case, why not simply

```suggestion
    if (!(event.target instanceof HTMLElement)) {
      return;
    }
    const adaptedEvent = this.mapEvent(event, eventType);
```
?

latekvo: Great call, applied in [f72af6c](https://github.com/software-mansion/react-native-gesture-handler/pull/3035/commits/f72af6cd6a0917319c2cc381bc61f3d4d42090ad)

m-bert (approved):
LGTM

Issue #3032, opened 2024-08-06
ReanimatedSwipeable > Jest > getByGestureTestId fails to find component.


markaustinws:
### Description

Trying to retrieve a swipeable component using getByGestureTestId, but it's failing.  Seems that the Test ID does not get rendered

### Steps to reproduce

  test("GetByGestureTestId", async () => {
    // const swipeable = screen.getByTestId("mySwipe");

    render(

      <Swipeable testID="abcd">
        <Text>Hello</Text>
      </Swipeable>

    );

    screen.debug();

    const swipeable = getByGestureTestId("abcd");
  });

Handler with id: 'abcd' cannot be found

Screen debug:

    <View
      collapsable={false}
      jestAnimatedStyle={
        {
          "value": {},
        }
      }
      onLayout={[Function onRowLayout]}
      style={
        {
          "overflow": "hidden",
        }
      }
    >
      <View
        collapsable={false}
        jestAnimatedStyle={
          {
            "value": {
              "pointerEvents": "auto",
              "transform": [
                {
                  "translateX": 0,
                },
              ],
            },
          }
        }
        style={
          {
            "pointerEvents": "auto",
            "transform": [
              {
                "translateX": 0,
              },
            ],
          }
        }
      >
        <Text>
          Hello
        </Text>
      </View>
    </View>

### Snack or a link to a repository

https://snack.expo.dev/@mausti/tenacious-green-tortillas

### Gesture Handler version

2.18.1

### React Native version

0.74.3

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

Debug mode

### Device

None

### Acknowledgements

Yes

Glazzes:
`getGestureById`method will retrieve the gesture object, it's not meant to bring components

```js
// Pan gesture in some component
const pan = Gesture.Pan().withTestID("pan")

test("some tests", () => {
  render(<SomeComponent />) // Some component with the gesture shown above
 const pan = getGestureById("pan");
})
```

With code like this you will get the pan gesture.

markaustinws:
I'm trying to get the gesture from the Swipeable component.  Is this not possible?:

    render(<Swipeable testID="swipeable" />);
    getByGestureTestId("swipeable");
    
    Erroring with   Handler with id: 'swipeable' cannot be found

latekvo:
Hi @markaustinws, 

[...]

PR #3039, opened 2024-08-09
Forward unused props in `ReanimatedSwipeable` and `Pressable` to their inner components

latekvo:
## Description

Currently large portion of the props accepted by `ReanimatedSwipeable` and `Pressable` components remain unused.
This issue mostly refers to all the `accessibility` props.

This PR adds the use of `spread` operator to `ReanimatedSwipeable` and `Pressable` to forward all the remaining props to inner components, which are not explicitly used otherwise.

Fixes: #3032

## Test plan
- turn on radio description
- open the nested `Pressable` example
- add `accessibilityLabel` prop to one of the `Pressables`
- navigate over the modified `Pressable` with your keyboard (use `Tab` and `Shift-Tab` to navigate)
- hear as your label is being read out, which wouldn't occur without this change


latekvo:
I just found that using the following code works just as well as the `ManagedProps` method:

`const { children, style, ...rest } = props;`

The only downside being that it requires to pull out props before being used, as opposed to accessing them directly from the props object. 
And so if we've got 20 unique props in one of our components, that'll require having a 20-line-long header of extracted props as opposed to pulling them on-the-go.  

Would we prefer to stick with the `ManagedProps` method, or remove it in favour of this new one?

m-bert:
`ManagedProps` looks like using a sledgehammer to crack a nut. In my opinion using spread syntax is enough.

> And so if we've got 20 unique props in one of our components, that'll require having a 20-line-long header of extracted props as opposed to pulling them on-the-go.

I think that it is unrealistic edge-case that we shouldn't be concerned about. 

Also, I have a couple of comments on the code, but first let's see if @j-piasecki agrees with me 😅

src/components/utils.ts
```diff
@@ -0,0 +1,33 @@
+class ManagedProps<T> {
+  private usedProps: (keyof T)[] = [];
```
j-piasecki, line 2: This should be a set, accessing the same property multiple times would push it multiple times into the array.

src/components/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  private rawProps: T;
+
+  constructor(props: T) {
+    this.rawProps = props;
+
+    for (const key in props) {
+      // Define getters and setters dynamically
+      Object.defineProperty(this, key, {
+        get: () => {
+          this.usedProps.push(key);
```
j-piasecki, line 12: This wouldn't work as intended when accessing a prop with a delay (i.e. in an event), all renders until that point would have that prop in `remainingProps`.

src/components/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+          return this.rawProps[key];
+        },
+        enumerable: true,
+      });
+    }
+  }
+
+  get remainingProps(): T {
+    const mRemainingProps: Record<string | number | symbol, unknown> = {};
```
j-piasecki, line 21: In this case, it's not a problem, since `remainingProps` is read exactly once but ideally, this would be memoized.

j-piasecki (commented):
I'd go with spreading the props. It may be verbose, but it's explicit in what it does.

latekvo:
>I think that it is unrealistic edge-case that we shouldn't be concerned about.

having around 20 props is not an unrealistic edge-case, it's the exact case this PR is dealing with.
The 2 components utilising `ManagedProps` are `ReanimatedSwipeable` and `Pressable`, both using `17` and `21` unique props respectively.

Although with all that said, I think spread operator approach is indeed simpler and more readable than `ManagedProps`.

src/components/utils.ts
latekvo: Since we decided to go with the spread operator approach, is it ok if i remove this file, and resolve all your suggestion?

j-piasecki: Yes

src/components/Pressable/Pressable.tsx
```diff
@@ -26,9 +26,29 @@ import { GestureTouchEvent } from '../../handlers/gestureHandlerCommon';
...
+    onPressIn,
+    onPressOut,
+    onLongPress,
+    style,
+    children,
+    android_disableSound,
+    android_ripple,
+    disabled,
+    ...remainingProps
+  } = useMemo(() => props, [props]);
```
m-bert, line 49: Do we need to use `useMemo`?

src/components/ReanimatedSwipeable.tsx
```diff
@@ -199,6 +200,26 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+      animationOptions,
+      overshootLeft,
+      overshootRight,
+      onSwipeableWillOpen,
+      onSwipeableWillClose,
+      onSwipeableOpen,
+      onSwipeableClose,
+      testID,
+      ...remainingProps
+    } = useMemo(() => props, [props]);
```
m-bert, line 221: Same as above

m-bert (changes requested):
Please update title and description of this PR.

src/components/ReanimatedSwipeable.tsx
latekvo: I think we do want to keep this one, as the `ForwardedRef` prop may change without changing `SwipeableProps` prop.
While `spread` doesn't take too much resources, we want to avoid triggering multiple `useCallbacks` further down.

src/components/Pressable/Pressable.tsx
latekvo: I think so, we have some `useStates` which may cause rerenders. If we were to remove this `useMemo`, those props would trigger every `useCallback` to rerender each render as well, redering them useless.

j-piasecki: I don't think it's necessary, memoization of props passed to the component should be on the user. For example, we cannot correctly memoize event handlers, without knowing what's in their closure (and we don't have access to it).

Adding to that, I don't think it actually does anything here. React uses [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) to compare the dependencies. You're passing an entire `props` object as a dependency, which is created anew for every render and objects are compared based on reference, not value.

src/components/ReanimatedSwipeable.tsx
j-piasecki: Same here

src/components/Pressable/Pressable.tsx
latekvo: I see, that makes sense. 
I'll remove all the `useMemo` on `props` instances, all `useCallback` instances which depend on any of the `props`.

Do you think there are any other sensible ways to go about optimising the now non-memoized functions, or should they be left as-is?

j-piasecki: Other calls to `useCallback` and `useMemo` are fine. I meant specifically this one where you used the entire `props` object in the dependency array.

j-piasecki: If a user uses `useCallback` to memoize event handlers we want to be able to utilize that and make less updates ourselves.

latekvo: Yes I understand, I kept all the other `useMemo` and `useCallback` instances as all the values they depend on may be primitive values.

Removed prop memoization in [cc3ecdc](https://github.com/software-mansion/react-native-gesture-handler/pull/3039/commits/cc3ecdc486b90e003e9ee1b0a53f58ddc1ded734)

m-bert (approved):

j-piasecki (approved):

PR #3040, opened 2024-08-11
Fix broken image and improve docs

shubh73:
## Description

PR addresses the following issues:

- Typos and improved docs.
- Broken image in testimonials.

## Test plan

CI passes.


m-bert (approved):
Thank you!

docs/docs/fundamentals/manual-gestures/index.md
```diff
@@ -57,6 +57,6 @@ And that's all! As you can see using manual gestures is really easy but as you c
...
 
-While manual gestures open great possibilities we are aware that reimplementing pinch or rotation from scratch just because you need to activate in specific circumstances or require position of the fingers, would be a waste of time as those gestures are already there. Because of that you can use touch events with every gesture so that you can extract more informations about gesture than is sent to you in events. We also added a `manualActivation` modifier on all continous gestures, which prevents the gesture it is applied to from activating by itself thus giving you full control of its behavior.
+While manual gestures open great possibilities we are aware that reimplementing pinch or rotation from scratch just because you need to activate in specific circumstances or require position of the fingers, would be a waste of time as those gestures are already available. Therefore, can use touch events with every gesture to extract more information about gesture than what is provided in the events. We also added a `manualActivation` modifier on all continuous gestures, which prevents the gesture it is applied to from activating automatically, giving you full control over its behavior.
```
patrycjakalinska, line 60:
```suggestion
While manual gestures open great possibilities we are aware that reimplementing pinch or rotation from scratch just because you need to activate in specific circumstances or require position of the fingers, would be a waste of time as those gestures are already available. Therefore, you can use touch events with every gesture to extract more detailed information about the gesture than what the basic events alone provide. We also added a `manualActivation` modifier on all continuous gestures, which prevents the gesture it is applied to from activating automatically, giving you full control over its behavior.
```

patrycjakalinska (changes requested):
Everything looks good, only one small change 👍

patrycjakalinska (approved):
Perfect 🚀

Issue #3010, opened 2024-07-26
GestureDetector blocking ScrollView in Mobile Web (even if explicitly disabled)


mikehobi:
### Description

I'm trying to create a scrolling feed with a draggable/swipable elements.

When I include `GestureDetector` inside a `ScrollView`, scrolling is disabled altogether. I'm trying to implement a solution that allows vertical scrolling and horizontal swiping (on the draggable elements)

The solutions appear to work for iOS/Android, but not for mobile web.

I've provided a simple repro'd snack demo.

### Steps to reproduce

1. Implement GestureRoot, ScrollView, and Gesture components
2. Add `Gesture.Pan`, set to `enabled(false)` or `manualActivation`
3. See that scrolling is blocked on Safari (or when emulating mobile device in Chrome Devtools)

### Snack or a link to a repository

https://snack.expo.dev/@mikehobi/awkward-red-cereal?platform=web

### Gesture Handler version

2.16.1

### React Native version

0.74.3

### Platforms

Web

### JavaScript runtime

None

### Workflow

Expo bare workflow

### Architecture

None

### Build type

None

### Device

Real device

### Device model

iPhone Latest

### Acknowledgements

Yes

m-bert:
Hi @mikehobi! This issue seems closely related to #2617. You can try to add [touchAction](https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/gesture-detector/#touchaction-web-only) property to `GestureDetector` and set it to `pan-y`. This is already done internally in our [Swipeable](https://github.com/software-mansion/react-native-gesture-handler/blob/7050c1b2571a9f6a8213eba0c909b0bcee216a75/src/components/Swipeable.tsx#L544) implementation. Let me know if it helps!

mikehobi:
Thanks, touch-action does indeed solve my issue.

I guess I am still curious why `.enabled(false)` still captures events though, seems as though if it's disabled completely, it should not block other events.

Closing!

m-bert:
> I guess I am still curious why .enabled(false) still captures events though, seems as though if it's disabled completely, it should not block other events.

When handler is created, it sets `touch-action: none;` on underlying view by default. Currently there's no logic that checks if handler is enabled or not, so changing this property won't change `touch-action`.

Do you think it should be added?

PR #3041, opened 2024-08-11
[Web] Reset web-exclusive properties when handler is disabled.

m-bert:
## Description

As mentioned in [this comment](https://github.com/software-mansion/react-native-gesture-handler/issues/3010#issuecomment-2273906102), it may be confusing that Gesture Handler blocks default events even if it is explicitly disabled. This PR adds styles reset when `enabled` property changes.

Currently it changes 3 things:

1. Changes `touch-action` to default value
2. Changes `user-select` to default value
3. Re-enables context menu usage.

I'm not sure if I missed something. If so, please mention it in a comment.

## Test plan

Tested on newly added example.


src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -31,19 +32,16 @@ export class GestureHandlerWebDelegate
       );
     }
 
+    this.ready = true;
+
```
m-bert, line 36: Since `updateGestureConfig` is called before `init` method (I'm referring to Gesture Handler class), `delegate` is not initialized when `onEnabledChange` is called for the first time. This results in crashes since we cannot obtain `config`.

Now, there are two ways of solving it:

1. Check if `handler` was initialized - this way we keep everything inside Gesture Handler class and we don't have to worry about delegate not being ready.

2. Check if `delegate` was initialized - I think it is a bit clearer approach as it keeps `web` specific behavior out of Gesture Handler class. `if` statement inside Gesture Handler looks ok and it doesn't mess with Gesture Handler implementation.

Both of those approaches look fine, but I'm more into the second one so I've decided to start with it. Please let me know what do you think @j-piasecki.

src/web/handlers/GestureHandler.ts
```diff
@@ -41,11 +41,9 @@ export default abstract class GestureHandler implements IGestureHandler {
   protected shouldResetProgress = false;
   protected pointerType: PointerType = PointerType.MOUSE;
 
-  protected delegate: GestureHandlerDelegate<unknown, IGestureHandler>;
+  protected delegate: GestureHandlerWebDelegate;
```
j-piasecki, line 44: This doesn't really seem like keeping web-specific things out of `GestureHandler` 😅.

src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -119,6 +117,47 @@ export class GestureHandlerWebDelegate
...
+    this.view.style['touchAction'] = isHandlerEnabled
+      ? touchAction ?? 'none'
+      : 'auto';
+
+    // @ts-ignore This one disables default events on Safari
+    this.view.style['WebkitTouchCallout'] = isHandlerEnabled
+      ? touchAction ?? 'none'
+      : 'auto';
+  }
```
j-piasecki, line 143: Shouldn't the default value be read from the element first? If one explicitly sets `userSelect` to `none`, this would revert that to `auto`.

src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -31,19 +32,16 @@ export class GestureHandlerWebDelegate
...
 
-    this.addContextMenuListeners(config);
-
-    this.view.style['userSelect'] = config.userSelect ?? 'none';
-    this.view.style['webkitUserSelect'] = config.userSelect ?? 'none';
-
-    this.view.style['touchAction'] = config.touchAction ?? 'none';
-    // @ts-ignore This one disables default events on Safari
-    this.view.style['WebkitTouchCallout'] = 'none';
+    this.setUserSelect(config.enabled!); // `enabled` has default value of true if it was not provided
```
j-piasecki, line 42: In that case, can the type be changed so it's always defined instead of non-null asserts?

src/web/handlers/GestureHandler.ts
```diff
@@ -589,6 +587,10 @@ export default abstract class GestureHandler implements IGestureHandler {
     this.config = { enabled: enabled, ...props };
     this.enabled = enabled;
 
+    if (this.delegate.wasInitialized()) {
+      this.delegate.onEnabledChange(enabled);
+    }
```
j-piasecki, line 592: I think we can move this check inside the delegate, no?

1. It would allow for reactions even if delegate was not initialized yet
2. It keeps the logic outside of `GestureHandler`

src/web/tools/GestureHandlerWebDelegate.ts
j-piasecki: Yep, the second one sounds good.

src/web/handlers/GestureHandler.ts
m-bert: Done in [d80a9eb](https://github.com/software-mansion/react-native-gesture-handler/pull/3041/commits/d80a9eb3381db07b9367ba68cde87689b11ef7d1)

src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -119,6 +131,51 @@ export class GestureHandlerWebDelegate
...
+  private setUserSelect(isHandlerEnabled: boolean) {
+    const { userSelect } = this.gestureHandler.getConfig();
+
+    this.view.style['userSelect'] = isHandlerEnabled
+      ? userSelect ?? 'none'
+      : this.defaultViewStyles.userSelect;
+
+    this.view.style['webkitUserSelect'] = isHandlerEnabled
+      ? userSelect ?? 'none'
+      : this.defaultViewStyles.userSelect;
```
j-piasecki, line 143: I wonder how much we want to push restoring the previous value.

We could save the value here if a handler is enabled and restore it when it's disabled, though it still wouldn't be perfect (I don't think we can do perfect here). WDYT?

m-bert: So that would be situation where someone changes styles during gesture? I'm not exactly sure what are you referring to 😅

m-bert: Okay, we will keep current version as discussed. I'll change that if it starts to cause any problems.

j-piasecki (approved):

latekvo (approved):

PR #3043, opened 2024-08-13
Add `numberOfPointers` to `LongPress`

m-bert:
## Description

`LongPressGestureHandler` doesn't support multiple pointers. Since default value of `numberOfTouchesRequired` on iOS is `1`, handler gets cancelled if you put more than 1 pointer on it. 

This PR adds `numberOfPointers` modifier for `LongPress` so that now users can specify how many pointers are needed for `LongPress` to activate.

### Web

On `web` you could see similar problem, which was probably caused by calling `checkDistanceFail` inside `onPointerMove` - if you added another pointer, start position was still at the first pointer position, not in the midpoint. Therefore it was easy to accidentally cancel handler.

### macOS

I'm not sure if this feature is required on `macOS`.  In many places we just assume that handlers use only `1` pointer (or `2` in case of `pinch`/`rotation`). I've added required property to recognizer, but I have not tested that 😅 

## Test plan

Tested on slightly modified LongPress example from macOS:

<details>
<summary>Test code</summary>

```tsx
import Animated, {
  interpolateColor,
  useAnimatedStyle,
  useSharedValue,
  withTiming,

const Durations = {
  LongPress: 750,
  Reset: 350,
  Scale: 120,
};

const Colors = {
  Initial: '#0a2688',
  Loading: '#6fcef5',
  Success: '#32a852',
  Fail: '#b02525',
};

export default function LongPressExample() {
  const isPressed = useSharedValue(false);
  const color1 = useSharedValue(Colors.Initial);
  const color2 = useSharedValue(Colors.Loading);

  const animatedStyles = useAnimatedStyle(() => {
      colorProgress.value,
      [0, 1],
      [color1.value, color2.value]
    );

    return {
      transform: [
        {
          scale: withTiming(isPressed.value ? 1.2 : 1, {
            duration: Durations.Scale,
          }),
        },
      ],
      backgroundColor,
    };
  });

  const g = Gesture.LongPress()
    .onBegin(() => {
      console.log('onBegin');

      isPressed.value = true;
      colorProgress.value = withTiming(1, {
        duration: Durations.LongPress,
      });
    })
    .onStart(() => console.log('onStart'))
    .onEnd(() => console.log('onEnd'))
    .onFinalize((_, success) => {
      console.log('onFinalize', success);

      isPressed.value = false;

      color1.value = Colors.Initial;
      color2.value = success ? Colors.Success : Colors.Fail;

      colorProgress.value = withTiming(
        0,
        {
          duration: Durations.Reset,
        },
        () => {
          color2.value = Colors.Loading;
        }
      );
    })
    .onTouchesDown(() => console.log('onTouchesDown'))
    .onTouchesMove(() => console.log('onTouchesMove'))
    .onTouchesUp(() => console.log('onTouchesUp'))
    .minDuration(Durations.LongPress)
    .numberOfPointers(2)
    .maxDistance(100);

  return (
      <GestureDetector gesture={g}>
        <Animated.View style={[styles.pressBox, animatedStyles]} />
      </GestureDetector>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
  },

  pressBox: {
    width: 250,
    height: 250,
    borderRadius: 20,
  },
});

```

</details>

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
@@ -18,6 +18,8 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
   private var startTime: Long = 0
   private var previousTime: Long = 0
   private var handler: Handler? = null
+  private var hasScheduledActivation = false
+  private var currentPointers = 0
 
   init {
     setShouldCancelWhenOutside(true)
@@ -44,6 +46,42 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
     return this
   }
 
+  private fun getAverageX(ev: MotionEvent, excludePointer: Boolean = false): Float {
+    if (!excludePointer) {
+      return (0 until ev.pointerCount).map { ev.getX(it) }.average().toFloat()
+    }
+
+    var sumX = 0f
+
+    for (i in 0 until ev.pointerCount) {
+      if (i == ev.actionIndex) {
+        continue
+      }
+
+      sumX += ev.getX(i)
+    }
+
+    return sumX / (ev.pointerCount - 1)
+  }
+
+  private fun getAverageY(ev: MotionEvent, excludePointer: Boolean = false): Float {
+    if (!excludePointer) {
+      return (0 until ev.pointerCount).map { ev.getY(it) }.average().toFloat()
+    }
+
+    var sumY = 0f
+
+    for (i in 0 until ev.pointerCount) {
+      if (i == ev.actionIndex) {
+        continue
+      }
+
+      sumY += ev.getY(i)
+    }
+
+    return sumY / (ev.pointerCount - 1)
+  }
+
   override fun onHandle(event: MotionEvent, sourceEvent: MotionEvent) {
     if (!shouldActivateWithMouse(sourceEvent)) {
       return
@@ -53,43 +91,72 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
       previousTime = SystemClock.uptimeMillis()
       startTime = previousTime
       begin()
-      startX = sourceEvent.rawX
-      startY = sourceEvent.rawY
+      startX = getAverageX(sourceEvent)
+      startY = getAverageY(sourceEvent)
+
+      ++currentPointers
     }
 
-    if (state == STATE_BEGAN && sourceEvent.pointerCount == numberOfPointersRequired) {
+    if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_DOWN) {
+      startX = getAverageX(sourceEvent)
+      startY = getAverageY(sourceEvent)
+
+      if (++currentPointers > numberOfPointersRequired) {
+        fail()
+        currentPointers = 0
+      }
+    }
+    if (state == STATE_BEGAN && currentPointers == numberOfPointersRequired && !hasScheduledActivation) {
       handler = Handler(Looper.getMainLooper())
       if (minDurationMs > 0) {
         handler!!.postDelayed({ activate() }, minDurationMs)
       } else if (minDurationMs == 0L) {
         activate()
       }
+
+      hasScheduledActivation = true
     }
     if (sourceEvent.actionMasked == MotionEvent.ACTION_UP || sourceEvent.actionMasked == MotionEvent.ACTION_BUTTON_RELEASE) {
+      --currentPointers
+
       handler?.let {
         it.removeCallbacksAndMessages(null)
         handler = null
       }
+
       if (state == STATE_ACTIVE) {
         end()
       } else {
         fail()
       }
-    }
-    // Even though action suggests that pointer was lifted, it is still counted in pointerCount. That's why we subtract 1
-    else if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_UP && sourceEvent.pointerCount - 1 < numberOfPointersRequired && state != STATE_ACTIVE) {
-      fail()
+
+      hasScheduledActivation = false
+    } else if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_UP) {
+      if (--currentPointers < numberOfPointersRequired && state != STATE_ACTIVE) {
+        fail()
+        currentPointers = 0
+        hasScheduledActivation = false
+      } else {
+        startX = getAverageX(sourceEvent, true)
+        startY = getAverageY(sourceEvent, true)
+      }
     } else {
       // calculate distance from start
-      val deltaX = sourceEvent.rawX - startX
-      val deltaY = sourceEvent.rawY - startY
+      val x = getAverageX(sourceEvent)
+      val y = getAverageY(sourceEvent)
+
+      val deltaX = x - startX
+      val deltaY = y - startY
       val distSq = deltaX * deltaX + deltaY * deltaY
+
       if (distSq > maxDistSq) {
         if (state == STATE_ACTIVE) {
           cancel()
         } else {
           fail()
         }
+
+        hasScheduledActivation = false
       }
     }
   }
--- a/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
@@ -45,29 +45,15 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
     return this
   }
 
-  private fun getAverageX(ev: MotionEvent, excludePointer: Boolean = false): Float {
+  private fun getAverageCoords(ev: MotionEvent, excludePointer: Boolean = false): Pair<Float, Float> {
     if (!excludePointer) {
-      return (0 until ev.pointerCount).map { ev.getX(it) }.average().toFloat()
+      val x = (0 until ev.pointerCount).map { ev.getX(it) }.average().toFloat()
+      val y = (0 until ev.pointerCount).map { ev.getY(it) }.average().toFloat()
+
+      return Pair(x, y)
     }
 
     var sumX = 0f
-
-    for (i in 0 until ev.pointerCount) {
-      if (i == ev.actionIndex) {
-        continue
-      }
-
-      sumX += ev.getX(i)
-    }
-
-    return sumX / (ev.pointerCount - 1)
-  }
-
-  private fun getAverageY(ev: MotionEvent, excludePointer: Boolean = false): Float {
-    if (!excludePointer) {
-      return (0 until ev.pointerCount).map { ev.getY(it) }.average().toFloat()
-    }
-
     var sumY = 0f
 
     for (i in 0 until ev.pointerCount) {
@@ -75,10 +61,14 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
         continue
       }
 
+      sumX += ev.getX(i)
       sumY += ev.getY(i)
     }
 
-    return sumY / (ev.pointerCount - 1)
+    val x = sumX / (ev.pointerCount - 1)
+    val y = sumY / (ev.pointerCount - 1)
+
+    return Pair(x, y)
   }
 
   override fun onHandle(event: MotionEvent, sourceEvent: MotionEvent) {
@@ -90,8 +80,10 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
       previousTime = SystemClock.uptimeMillis()
       startTime = previousTime
       begin()
-      startX = getAverageX(sourceEvent)
-      startY = getAverageY(sourceEvent)
+
+      val (x, y) = getAverageCoords(sourceEvent)
+      startX = x
+      startY = y
 
       currentPointers++
     }
@@ -99,8 +91,9 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
     if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_DOWN) {
       currentPointers++
 
-      startX = getAverageX(sourceEvent)
-      startY = getAverageY(sourceEvent)
+      val (x, y) = getAverageCoords(sourceEvent)
+      startX = x
+      startY = y
 
       if (currentPointers > numberOfPointersRequired) {
         fail()
@@ -136,13 +129,13 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
         fail()
         currentPointers = 0
       } else {
-        startX = getAverageX(sourceEvent, true)
-        startY = getAverageY(sourceEvent, true)
+        val (x, y) = getAverageCoords(sourceEvent, true)
+        startX = x
+        startY = y
       }
     } else {
       // calculate distance from start
-      val x = getAverageX(sourceEvent)
-      val y = getAverageY(sourceEvent)
+      val (x, y) = getAverageCoords(sourceEvent)
 
       val deltaX = x - startX
       val deltaY = y - startY
--- a/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
@@ -85,7 +85,7 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
       startX = x
       startY = y
 
-      currentPointers++
+      currentPointers = 1
     }
 
     if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_DOWN) {
--- a/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
@@ -12,6 +12,7 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
     get() = (previousTime - startTime).toInt()
   private val defaultMaxDistSq: Float
   private var maxDistSq: Float
+  private var numberOfPointersRequired: Int
   private var startX = 0f
   private var startY = 0f
   private var startTime: Long = 0
@@ -24,6 +25,7 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
     val defaultMaxDist = DEFAULT_MAX_DIST_DP * context.resources.displayMetrics.density
     defaultMaxDistSq = defaultMaxDist * defaultMaxDist
     maxDistSq = defaultMaxDistSq
+    numberOfPointersRequired = 1
   }
 
   override fun resetConfig() {
@@ -34,6 +36,11 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
 
   fun setMaxDist(maxDist: Float): LongPressGestureHandler {
     maxDistSq = maxDist * maxDist
+    return this
+  }
+
+  fun setNumberOfPointers(numberOfPointers: Int): LongPressGestureHandler {
+    numberOfPointersRequired = numberOfPointers
     return this
   }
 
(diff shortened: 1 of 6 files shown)


apple/Handlers/RNLongPressHandler.m
```diff
@@ -89,6 +92,7 @@ - (void)touchesMoved:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
   [_gestureHandler.pointerTracker touchesMoved:touches withEvent:event];
 
   if ([self shouldCancelGesture]) {
+    self.state = UIGestureRecognizerStateFailed;
```
latekvo, line 95: Is `UIGestureRecognizerStateFailed` intentionally being used instead of `UIGestureRecognizerStateCancelled` when gesture cancellation occurs?

latekvo (commented):
One minor comment, looks great overall 👍

j-piasecki (approved):

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -37,6 +41,47 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
+    for (i in 0 until ev.pointerCount) {
+      if (i == ev.actionIndex) {
+        continue
+      }
+
+      sumX += ev.getX(i)
+    }
+
+    return sumX / (ev.pointerCount - 1)
+  }
```
m-bert, line 65: We can merge `getAverageX` and `getAverageY` into one method. For example, we can return object with `x` and `y`, or pass argument which will determine whether we want to calculate `x` or `y`. We can also leave 2 implementations. 

Let me know which approach do you think is better, @j-piasecki, @latekvo

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -46,36 +91,72 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
+
+      hasScheduledActivation = false
+    } else if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_UP) {
+      if (--currentPointers < numberOfPointersRequired && state != STATE_ACTIVE) {
+        fail()
+        currentPointers = 0
+        hasScheduledActivation = false
+      } else {
+        startX = getAverageX(sourceEvent, true)
+        startY = getAverageY(sourceEvent, true)
```
m-bert, line 141: I'm not sure if we want to change `startX` and `startY` in this case, but if we don't, releasing a pointer will basically cancel handler (since it is very hard to avoid little 1px movements) (cc @j-piasecki)

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -53,43 +91,72 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
+    }
+    if (state == STATE_BEGAN && currentPointers == numberOfPointersRequired && !hasScheduledActivation) {
       handler = Handler(Looper.getMainLooper())
       if (minDurationMs > 0) {
         handler!!.postDelayed({ activate() }, minDurationMs)
       } else if (minDurationMs == 0L) {
         activate()
       }
+
+      hasScheduledActivation = true
```
j-piasecki, line 117: Is this property necessary? Wouldn't checking whether action is `DOWN` or `POINTER_DOWN` accomplish the same?

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -53,43 +91,72 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
       previousTime = SystemClock.uptimeMillis()
       startTime = previousTime
       begin()
-      startX = sourceEvent.rawX
-      startY = sourceEvent.rawY
+      startX = getAverageX(sourceEvent)
+      startY = getAverageY(sourceEvent)
+
+      ++currentPointers
```
j-piasecki, line 97: From what I can see, in all but two random places we're using `++` as a suffix operator, can we stick to that?

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -53,43 +91,72 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
     }
 
-    if (state == STATE_BEGAN && sourceEvent.pointerCount == numberOfPointersRequired) {
+    if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_DOWN) {
+      startX = getAverageX(sourceEvent)
+      startY = getAverageY(sourceEvent)
+
+      if (++currentPointers > numberOfPointersRequired) {
```
j-piasecki, line 104: Same

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -53,43 +91,72 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
       if (minDurationMs > 0) {
         handler!!.postDelayed({ activate() }, minDurationMs)
       } else if (minDurationMs == 0L) {
         activate()
       }
+
+      hasScheduledActivation = true
     }
     if (sourceEvent.actionMasked == MotionEvent.ACTION_UP || sourceEvent.actionMasked == MotionEvent.ACTION_BUTTON_RELEASE) {
+      --currentPointers
```
j-piasecki, line 120: Same

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -53,43 +91,72 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
         fail()
       }
-    }
-    // Even though action suggests that pointer was lifted, it is still counted in pointerCount. That's why we subtract 1
-    else if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_UP && sourceEvent.pointerCount - 1 < numberOfPointersRequired && state != STATE_ACTIVE) {
-      fail()
+
+      hasScheduledActivation = false
+    } else if (sourceEvent.actionMasked == MotionEvent.ACTION_POINTER_UP) {
+      if (--currentPointers < numberOfPointersRequired && state != STATE_ACTIVE) {
```
j-piasecki, line 135: Same

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
j-piasecki: One method makes sense imo, given that now their implementation is exactly the same and we always want both values anyway.

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
j-piasecki: I don't think I see what you mean.

> releasing a pointer will basically cancel handler

Isn't that what we want? If there are more pointers than set by the user we fail the handler so at this point there should always be too few pointers, no?

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
latekvo: I think it's better to have them both consolidated into a single `getAveragePosition` function, as I can't really image them being used separately.

But i see that a similar function already exists [here](https://github.com/software-mansion/react-native-gesture-handler/blob/0ece2902d7853b16a352c31d99ca29a58b0de4bb/android/src/main/java/com/swmansion/gesturehandler/core/GestureUtils.kt#L7), and it follows the separate-function approach.
I see that the built-in `MotionEvent` events also use the separate `getX` and `getY` functions approach.

I'm not sure if it'd be better to follow the convention, as it is not strongly set in our codebase.

Let's wait to see what @j-piasecki thinks.

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: It depends what do we understand by `numberOfPointers` are those pointers required for activation, or through whole state changes process? 

Also, the fact that I've introduced this change is based on this:

>it is very hard to avoid little 1px movements

What it really means is that if you lift both fingers at once, you'll most likely receive `ACTION_POINTER_UP` followed by `ACTION_MOVE`. Now, if we don't change starting point it will definitely exceed `maxDistance`, which means that the handler will fail even though we had enough pointers.

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: Yes, it should work the same.

Replaced in [04b4c9f](https://github.com/software-mansion/react-native-gesture-handler/pull/3043/commits/04b4c9fba8269465b499daf8ecfea70f59a2ab4f)

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: Sure, I'm just used to using pre-incrementation when possible since it is faster (however, I don't know if that's the case in Kotlin).

Done in [41495ce](https://github.com/software-mansion/react-native-gesture-handler/pull/3043/commits/41495cebe1a1286b315ea706470441d4b4c2141b)

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: [41495ce](https://github.com/software-mansion/react-native-gesture-handler/pull/3043/commits/41495cebe1a1286b315ea706470441d4b4c2141b)

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: [41495ce](https://github.com/software-mansion/react-native-gesture-handler/pull/3043/commits/41495cebe1a1286b315ea706470441d4b4c2141b)

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: [41495ce](https://github.com/software-mansion/react-native-gesture-handler/pull/3043/commits/41495cebe1a1286b315ea706470441d4b4c2141b)

latekvo (changes requested):
`LongPress` permanently stops working after being cancelled by placing additional finger down while activated.

Tested on `Android` with `numberOfPointers(2)`, `minDuration(2000)` and `maxDistance(99999999)`


https://github.com/user-attachments/assets/e0bbaf40-9679-45e3-ba7f-73c6e29d2ef8

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: Okay, done in [dc90644](https://github.com/software-mansion/react-native-gesture-handler/pull/3043/commits/dc906440416ed11d14320e0807c1b1cf64ed9309)

m-bert:
>LongPress permanently stops working after being cancelled by placing additional finger down while activated.

That's strange, I've just tested it with parameters that you've provided and eI can't see this problem. What I do see though, is that we get `onEnd` and `onFinalize` with `false` triggered, I'll check what's going on. 

Are you sure that it `permanently` stops working?

latekvo:
>Are you sure that it permanently stops working?

Tested it again, it stops working for the very next to-be activation, but works again after that.

### Here's an updated video:

<details>
<summary>
Video
</summary>
https://github.com/user-attachments/assets/62c9becf-83cd-4d58-bce4-3238fe539842
</details>

### And here's the full repro:

<details>
<summary>
Code
</summary>
```tsx
import React from 'react';
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,

  const outer = Gesture.LongPress()
    .numberOfPointers(2)
    .minDuration(1000)
    .maxDistance(99999999)
    // .hitSlop(-100)
    // .onTouchesUp(() => console.log('UP'))
    .onTouchesDown(() => console.log('DOWN'))
    .onStart(() => console.log('START'))
    .onBegin(() => console.log('BEGIN'))
    .onEnd(() => console.log('END'))
    .onFinalize(() => console.log('FINALIZE'));

  return (
    <GestureHandlerRootView>
        <GestureDetector gesture={outer}>
          <View style={styles.nested}>
            {/* <View style={styles.slopper} /> */}
          </View>
        </GestureDetector>
      </View>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
  text: {
    width: 300,
    height: 200,
    backgroundColor: 'tomato',
  },
  nested: {
    width: 300,
    height: 300,
    backgroundColor: 'plum',
  },
  slopper: {
    width: 100,
    height: 100,
    margin: 'auto',
    backgroundColor: 'tomato',
  },
});
```
</details>

latekvo:
Another possibly related issue - given the following code:

<details>
<summary>
Code
</summary>

```tsx
import React from 'react';

  const outer = Gesture.LongPress()
    .numberOfPointers(2)
    .minDuration(1000)
    .maxDistance(99999999)
    .onTouchesUp(() => console.log('UP'))
    .onTouchesDown(() => console.log('DOWN'))
    .onStart(() => console.log('START'))
    .onBegin(() => console.log('BEGIN'))
    .onEnd(() => console.log('END'))
    .onFinalize(() => console.log('FINALIZE'));

  return (
      <GestureDetector gesture={outer}>
        <View style={styles.nested} />
      </GestureDetector>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
  text: {
    width: 300,
    height: 200,
    backgroundColor: 'tomato',
  },
  nested: {
    width: 300,
    height: 300,
    backgroundColor: 'plum',
  },
});

```

</details> 

When gesture is cancelled by dragging the finger out-of-bounds, the next 2 attempts to activate it fail, before the 3rd one is successful.

### To reproduce:

1. press down with one finger, swipe out of bounds
    - 🟢 cancels as anticipated
2. press down with 2 fingers
    - 🔴 gesture immediately failed
    - instead it should have activated after 1000ms
3. press down with 2 fingers
    - 🔴 gesture never activates
    - instead it should have activated after 1000ms
3. press down with 2 fingers
    - 🟢  gesture activates after 1000ms as anticipated
    - successfully activates only after 3rd attempt since the out-of-bounds cancelation

Reproduction video:


https://github.com/user-attachments/assets/cf50a9b7-4326-4897-a64f-c287004a55b3

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
j-piasecki: We may want to consider consolidating other methods that follow this convention

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
j-piasecki: I'm 99% sure the compiler will figure it out 😄

m-bert:
> Tested it again, it stops working for the very next to-be activation, but works again after that.
Everything works fine in my case. @j-piasecki, could you check it if you can reproduce it?

m-bert:
>When gesture is cancelled by dragging the finger out-of-bounds, the next 2 attempts to activate it fail, before the 3rd one is successful.

In my case only the first attempt fails, second one activates as expected. I'll check that.

m-bert:
> In my case only the first attempt fails, second one activates as expected. I'll check that.

Okay, as I thought, `shouldCancelWhenOutside` follows different logic. If handler is to be cancelled by this flag, `onHandle` method is not called (well, it is, but in `GestureHandler`, not in the specific one). Because of that, `currentPointers` were not reseted. 

To avoid much changes in general logic, I've set `currentPointers` to `1` when we first start gesture (i.e. it is in `UNDETERMINED` state).

Let me know if that satisfies you, @j-piasecki @latekvo. 

Changes were made in [this commit](https://github.com/software-mansion/react-native-gesture-handler/pull/3043/commits/a825f2fe780ba61e29b28880e9b70578ddc41043).

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: > Isn't that the same thing? I thought that it would require the exact amount of pointers set to activate and it would fail when adding or removing them before activation (so no activate when too few and no activate when too many).

By "whole state changes process" I meant from `UNDETERMINED` to one of the final states. I'm not talking only about activation 😅 The main reason why I did this is because it is almost impossible to end gesture if those values are not updated - it will fail because `maxDistance` will be exceeded (especially if the pointers are far away from each other).

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: > We may want to consider consolidating other methods that follow this convention

We may. I can do this, but definitely in another PR 😅

latekvo (approved):

latekvo:
I confirm all previously failing test cases now behave correctly on my end  👍

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -37,6 +40,37 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
+    numberOfPointersRequired = numberOfPointers
+    return this
+  }
+
+  private fun getAverageCoords(ev: MotionEvent, excludePointer: Boolean = false): Pair<Float, Float> {
+    if (!excludePointer) {
+      val x = (0 until ev.pointerCount).map { ev.getX(it) }.average().toFloat()
+      val y = (0 until ev.pointerCount).map { ev.getY(it) }.average().toFloat()
+
+      return Pair(x, y)
```
j-piasecki, line 53:
```suggestion
      return x to y
```

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -37,6 +40,37 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
+      }
+
+      sumX += ev.getX(i)
+      sumY += ev.getY(i)
+    }
+
+    val x = sumX / (ev.pointerCount - 1)
+    val y = sumY / (ev.pointerCount - 1)
+
+    return Pair(x, y)
```
j-piasecki, line 71:
```suggestion
    return x to y
```

j-piasecki (commented):
No idea if this looks better. Feel free to ignore it if you want.

m-bert:
> Wouldn't onReset be a better pick here?

You're right. Changed in [d61ed7e](https://github.com/software-mansion/react-native-gesture-handler/pull/3043/commits/d61ed7ea17939908132b7cbee06b45a7b84e1031).

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
m-bert: I'm no expert in Kotlin and using `Pair` seems cleaner to me (I mean it is easier to understand what's going on). I'll leave that as it is now.

j-piasecki (approved):

PR #3054, opened 2024-08-21
docs: Update reanimated_swipeable.md

bryanltobing:
## Description

Fix incorrect import

## Test plan

try to import Swipeable from `import Swipeable from "react-native-gesture-handler/Swipeable"`;

### DIFF
--- a/docs/docs/components/reanimated_swipeable.md
+++ b/docs/docs/components/reanimated_swipeable.md
@@ -22,7 +22,7 @@ Reanimated `Swipeable` allows for implementing swipeable rows or similar interac
 To use it, import it in the following way:
 
 ```js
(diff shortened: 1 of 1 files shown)


m-bert:
Hi! I'm not sure why you try to change import for `ReanimatedSwipeable` component to legacy `Swipeable`. Have we missed something?

latekvo:
Hey,

the file you've edited correctly refers to the `ReanimatedSwipeable` component, thus the import is correct.

Are you sure you have used `>=2.18.0` version of Gesture Handler? 
That's when this component was added, which may be the reason why you weren't able to import it.

bryanltobing:
hi. Thank you for the response. i thought i was in the latest version. turns out the new expo template is still in `v2.16.1`.

my bad for not checking.

PR #3055, opened 2024-08-21
Make osx example utilise the common example app

latekvo:
## Description

This PR makes the MacOS example utilise the common example app.
This change will allow for more in-depth testing of MacOS functionalities.

For now, it's still work in progress


m-bert (commented):
I hope it won't have much conflicts with #2919 😅

MacOSExample/package.json
```diff
@@ -11,18 +11,23 @@
...
+    "postinstall-postinstall": "^2.1.0",
     "react": "18.2.0",
-    "react-native": "0.72.4",
+    "react-native": "^0.74.1",
     "react-native-gesture-handler": "link:../",
-    "react-native-macos": "^0.72.0",
-    "react-native-reanimated": "^3.5.1",
-    "react-native-safe-area-context": "^4.7.2",
-    "react-native-screens": "^3.25.0"
+    "react-native-macos": "^0.73.35",
```
m-bert, line 27: Are you sure those 2 versions are compatible? As far as I can see, `react-native-macos` already has [0.74.1 version](https://github.com/microsoft/react-native-macos/releases/tag/v0.74.1)

m-bert: Of course I meant `react-native` and `react-native-macos` 😅

latekvo:
>I hope it won't have much conflicts with #2919 😅

I hope there won't be any, `App.tsx` should remain minimally changed with just a couple of styles differing, while the rest of the changes should be out of scope of #2919

Regardless, this PR is still WIP, and only works on my end due to some modifications to `App.tsx` structure, as I'm still working out some rendering inconsistencies, which are hopefully caused by default styling differences.

MacOSExample/package.json
latekvo: > Are you sure those 2 versions are compatible?

They certainly aren't, but I think they work together because `react-native` is not used at all.

While I'm not certain about this, I believe `react-native-macos` is a complete replacement for `react-native`, as it's declared as an alias for `react-native` in the `Babel` configuration file

latekvo:
As of now, the common examples are working on MacOS, and I am still working on completely removing the MacOSExample directory, but that may prove challenging as currently we're relying on the `babel.config.js` to make the entire app work on MacOS, which unfortunately would prevent other platforms from using the examples if we were to naively marge both of these folders in the state that they are.

image

MacOSExample/package.json
latekvo: removed `react-native` package in [7399ccc](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/7399ccc0ba8ba1230cf8c614de6b03d9ec86dc08)

latekvo:
With the current state of `MacOS` support, I don't believe it is worth it to keep trying to make it work with `Expo` in this PR, it has proved a bit too challenging for what it does [as it is not currently officially supported](https://github.com/microsoft/react-native-macos/issues/2076#issuecomment-1944849696).

I've previously discussed this with @tsapeta and we couldn't find a working `MacOS` + `Expo` configuration.

I believe the best course of action right now would be to revert progress back to somewhere around [02a96fa](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/02a96fa6f82751b8d53cf8978995ce9d56f7150e), and to cherry-pick some fixes that happened later, for example [the build settings fix](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/6036fe46ad3d92e0b22c2ec1d0b6eaf27bc19379).

### Summary

What has been done:
- have a single app for both `MacOS` and the rest of the platforms

What couldn't be done:
- run both `MacOS` and `iOS` from a single xcode project
- load `MacOS` example off the same `metro` instance as `Android`, `iOS` and `Web` runs off

(cc @m-bert)

m-bert:
> What has been done:
>
> have a single app for both MacOS and the rest of the platforms

I think this is what we wanted to achieve 😅 

> What couldn't be done:
>
> run both MacOS and iOS from a single xcode project
> load MacOS example off the same metro instance as Android, iOS and Web runs off

First one is definitely not a problem. As for second, I have not looked deeply into this configuration, but before you had to start separate metro server for `web`, so it is also not a big deal.

> I think the best course of action right now would be to revert progress back to somewhere around [02a96fa](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/02a96fa6f82751b8d53cf8978995ce9d56f7150e), and to cherry-pick some fixes that happened later, for example [the build settings fix](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/6036fe46ad3d92e0b22c2ec1d0b6eaf27bc19379).

Sounds good 😄

latekvo:
PR functional as of [7717de6](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/7717de698b9e6011920a8cc408b91fe52f4e2897)

currently only working on making the CI tests pass.

MacOSExample/package.json
```diff
@@ -3,26 +3,27 @@
...
+    "postinstall-postinstall": "^2.1.0",
     "react": "18.2.0",
-    "react-native": "0.72.4",
+    "react-native": "^0.73.9",
     "react-native-gesture-handler": "link:../",
-    "react-native-macos": "^0.72.0",
-    "react-native-reanimated": "^3.5.1",
-    "react-native-safe-area-context": "^4.7.2",
-    "react-native-screens": "^3.25.0"
+    "react-native-macos": "^0.73.35",
```
m-bert, line 23: Why not 0.74?

example/package.json
```diff
@@ -32,6 +32,7 @@
     "react-dom": "18.2.0",
     "react-native": "0.74.1",
     "react-native-gesture-handler": "link:..",
+    "react-native-macos": "^0.73.35",
```
m-bert, line 35: Same, why not 0.74?

package.json
```diff
@@ -64,6 +64,7 @@
   "homepage": "https://github.com/software-mansion/react-native-gesture-handler#readme",
   "dependencies": {
     "@egjs/hammerjs": "^2.0.17",
+    "fast-xml-parser": "^4.4.1",
```
m-bert, line 67: Is it necessary?

MacOSExample/macos/.xcode.env.local
```diff
@@ -0,0 +1,2 @@
+export NODE_BINARY=/opt/homebrew/bin/node
```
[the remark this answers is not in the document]
j-piasecki: `.xcode.env` should be commited, `.xcode.env.local` should not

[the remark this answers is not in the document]
latekvo: This file was automatically generated with `pod install` and shouldn't have been committed. Removed it in [dbb75eb](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/dbb75eb92836525ecb823a3dcbbdac07ba258a2e)

MacOSExample/package.json
latekvo: There are multiple breaking changes in `0.74` that break `RNGH`, `Rea`, and bundle loading.

I managed to fix most of them but i'm still working on it.

package.json
latekvo: it's not, removed in [4c12996](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/4c1299690a915c6258c0bf67c0e3a6f14ed9db29)

j-piasecki:
> As far as I know, image assets are not available on macos.
I've tested this by trying to load png images into Image components, while their presence was detected by require, none showed up.

Were the images under the `example` app or `MacOSExample`? This may be caused by the fact that you're effectively loading files from outside the app since it's not a monorepo.

Regarding failing Android build, there are multiple versions of `@react-native/gradle-plugin` in the project
Screenshot 2024-09-19 at 16 31 21
Vs on main:
Screenshot 2024-09-19 at 16 35 10
It may be related (it's possible that iOS failures are caused by a similar problem, though I don't remember any package that could cause it there from the top of my head).

latekvo:
@j-piasecki Image loading fixed in [2d62e3a](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/2d62e3ae81d78d16b9cde3b0e33b910001f13d16) and [896f9be](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/896f9becb15f8efcc97af3b3c599f9779869510b).

image
image

latekvo:
iOS and Android should now both build successfully.
I've ran these two [[Android](https://github.com/software-mansion/react-native-gesture-handler/actions/runs/10960972086), [iOS](https://github.com/software-mansion/react-native-gesture-handler/actions/runs/10960979392)] workflows just to be sure:

[Test Android build](https://github.com/software-mansion/react-native-gesture-handler/actions/workflows/android-build.yml) [Test iOS build](https://github.com/software-mansion/react-native-gesture-handler/actions/workflows/ios-build.yml)

example/src/recipes/simpleRotate/index.tsx
m-bert, line 1: Why not put it inside `simple` directory?

latekvo:
https://github.com/user-attachments/assets/da53c5d3-ec4b-4283-a852-e576dc811f9e

Fixed header being squashed in these commits: [e3c098d](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/e3c098d6eaed68e8f15e74452f3944395fe18b73), [6a6f885](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/6a6f88526be65b1db05fd6cf6c25e21aadae4f2a), [38f861d](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/38f861d31fed23ea0fed630fa63b1824e374b28c) (cherry-picked from [this](https://github.com/software-mansion/react-native-gesture-handler/tree/%40latekvo/test-header-styling-for-macos) branch)

latekvo:
Removed chevrons in [61455cf](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/61455cf5560f9c55ac90bc653cdb19ce06f1b56f), couldn't find a way to properly load their custom font.
Tried:
- the formal method of adding fonts found [here](https://developer.apple.com/documentation/uikit/text_display_and_fonts/adding_a_custom_font_to_your_app?language=objc)
- modifying `react-native-asset` tool to work with `macos`, the tool worked but the fonts still weren't present
- started working on dynamic font loader based on [this 2012 article](https://marco.org/2012/12/21/ios-dynamic-font-loading) and [this StackOverflow answer](https://stackoverflow.com/questions/14048549/how-to-dynamically-load-a-font-under-ios-for-real/16412876#16412876) but decided it's not worth spending too much time on this feature for now.

All other known bugs have been resolved by now as far as i know.

Please let me know if there're any other changes you'd like to see, or if it's ok to merge this PR.

MacOSExample/macos/Podfile
```diff
@@ -5,23 +5,33 @@ prepare_react_native_project!
...
-  # Flags change depending on the env values.
-  flags = get_default_flags()
+  # Override default react-native path for MacOS building
+  config[:reactNativePath] = react_native_path
 
   use_react_native!(
-    :path => '../node_modules/react-native-macos',
-    :hermes_enabled => false,
+    :path => react_native_path,
+    :hermes_enabled => true,
```
m-bert, line 18: Do we have to change it to `true`?

MacOSExample/package.json
```diff
@@ -3,26 +3,31 @@
...
     "@react-navigation/stack": "^6.3.17",
+    "@swmansion/icons": "^0.0.1",
+    "babel-plugin-module-resolver": "^5.0.2",
     "hoist-non-react-statics": "^3.3.2",
     "invariant": "^2.2.4",
+    "patch-package": "^8.0.0",
+    "postinstall-postinstall": "^2.1.0",
     "react": "18.2.0",
-    "react-native": "0.72.4",
+    "react-native": "^0.73.9",
```
m-bert, line 24: Shouldn't `react-native-macos` 0.74 work with `react-native` 0.74?

example/App.tsx
```diff
@@ -347,7 +366,11 @@ const styles = StyleSheet.create({
       },
     }),
     padding: 16,
-    backgroundColor: COLORS.offWhite,
+    // backgroundColor: COLORS.offWhite,
```
m-bert, line 369: I'm not sure if we want to leave it as a comment 😅

example/src/ListWithHeader/Header.tsx
```diff
@@ -56,28 +57,46 @@ function HeaderNative(props: HeaderProps) {
     };
   });
 
+  const collapsedCoefficient = 0.7;
+  const openCoefficient = Platform.OS === 'macos' ? 1 : 0.5;
+  const padding = Platform.OS === 'macos' ? 10 : 0;
+  const horizontalOffset = Platform.OS === 'macos' ? 50 : 0; // todo: relative to width
```
m-bert, line 63: What about this `todo`? Will it be addressed in other PR?

m-bert (changes requested):

example/App.tsx
latekvo: My mistake, removed it in [46a46ac](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/46a46ac95ce6fa6d33334c362cea374a0e972d6a)

example/src/ListWithHeader/Header.tsx
latekvo: Same here

MacOSExample/package.json
latekvo: Fixed in [46a46ac](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/46a46ac95ce6fa6d33334c362cea374a0e972d6a)

MacOSExample/macos/Podfile
latekvo: No, changed to `false` in [46a46ac](https://github.com/software-mansion/react-native-gesture-handler/pull/3055/commits/46a46ac95ce6fa6d33334c362cea374a0e972d6a)

example/src/ListWithHeader/Header.tsx
```diff
@@ -89,20 +108,42 @@ function HeaderNative(props: HeaderProps) {
...
     );
-    const width = (size?.width ?? 0) * (expandFactor.value * 0.2 + 0.4);
+
+    const widthCoefficient = 0.2;
+    const widthBias = Platform.OS === 'macos' ? 0.2 : 0.4;
+
+    const textOpenWidth =
+      (size?.width ?? 0) * (expandFactor.value * widthCoefficient + widthBias);
+    const textCollapsedWidth =
+      (size?.width ?? 0) * (expandFactor.value * widthCoefficient + widthBias);
```
m-bert, line 123: Should those have the same value?

example/src/ListWithHeader/Header.tsx
```diff
@@ -16,7 +16,8 @@ import { COLORS } from '../common';
 const SIGNET = require('./signet.png');
 const TEXT = require('./text.png');
 
-export const HEADER_HEIGHT = Platform.OS === 'web' ? 64 : 192;
+export const HEADER_HEIGHT =
+  Platform.OS === 'web' ? 64 : Platform.OS === 'macos' ? 128 : 192;
```
j-piasecki, line 20: Can we make the header the same as on web? Images seem to always be stretched, it may make sense to set explicit sizes for them instead of relying on the platform to properly contain them.

j-piasecki (commented):
Looks good, few things I noticed:
- `Draggable` example using the old API doesn't work, I don't recall whether it did at some point (cc @m-bert), but it might be worth looking into
- `Pressable` doesn't work, which is not ideal

Bot those can be addressed in follow-ups, as this PR has been going on for quite some time

MacOSExample/macos/Podfile
j-piasecki: Actually, can we keep hermes? It's been the default for quite some time and Reanimated has intentional memory leaks on JSC to prevent crashes IIRC.

(I tried to switch it locally and the `loading bundle` overlay never disappeared for me, I'm not sure whether it was a local issue or not)

latekvo: >I tried to switch it locally and the loading bundle overlay never disappeared for me, I'm not sure whether it was a local issue or not

Occurs on less than 1/10 runs for unknown reason. App re-run fixes this issue.

m-bert:
Okay, I've looked into `Draggable` example and it seems to be the problem with `NativeViewGestureHandler` (something else than I expected). I'll dive into it later.

Also what I've noticed is that sometimes app gets stuck at bundling overlay:

image

example/src/ListWithHeader/Header.tsx
latekvo: Sure, I'll open a separate PR for it to avoid cluttering this one any further.

j-piasecki (approved):
To sum up, I think we can merge this as-is currently (but please, use Hermes) and then create follow-up issues for:
- header looking weird on macOS
- Draggable example not working (for other reasons than expected)
- Pressables not working on macOS
- app being stuck on loading bundle screen

and keep working on these

(use `Maintainer issue` label to stop bot from bothering you).

m-bert (approved):
Remember to address things mentioned by @j-piasecki in follow-ups 😅

PR #3057, opened 2024-08-22
Allow Android Rotation Gesture to continue after second Pointer Lift and Land

DmitriyGalyanov:
## Description

On iOS Rotation Gesture is not getting finished/cancelled when one of two Pointers lifts up, allowing the Gesture to continue when a new second Pointer arrives

On Android this behaviour differs – Rotation Gesture finishes whenever one of two Pointers lifts up, requiring to land 2 completely new Pointers in order to utilise Rotation Gesture again

I understand that this may be closer to default Android Gesture Recogniser Behaviour, but find it unexpected and inconvenient

Thus in this PR I’m adding Ability to tell Rotation Gesture Builder to enable iOS-like Behaviour, which allows Gesture Pause/Remain when second Pointer lifts/lands

Additionally, I added some local toggleable Logs, which I find useful for Debugging, but I’m not sure if it aligns with this Repo Style Guide (or could noticeably reduce Performance) – please notify me, if it’s prohibited

I think this Change may be usable for Gesture Handler Users, thus trying to push it to the Gesture Handler (not only my own Fork) and looking forward to any Feedback regarding both Code – Implementation/Naming and/or local Style Guide – and Docs, which I might’ve misaligned with

Attached Gifs display what I mean (first one shows current Behaviour, second one – with new Modifier)

## Test plan
1. Use `RotationGesture` without new Modifier (`secondPointerLiftFinishesGesture(value: boolean)`) and see if it works as expected (as before – Gesture finishes when one of two key Pointers is lifted)
2. Use `RotationGesture` with new Modifier (`secondPointerLiftFinishesGesture(value: boolean)`) and see if it works as expected (Gesture doesn't finish when second Pointer is lifted and continues when second Pointer is landed again)

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
@@ -1,27 +1,9 @@
 package com.swmansion.gesturehandler.core
 
-import android.util.Log
 import android.view.MotionEvent
 import kotlin.math.atan2
 
-private const val ENABLE_LOGS = false
-private const val BASE_LOG_TAG = "RotationGestureHandler"
-private fun localLog(message: String, isFrequent: Boolean = false) {
-  if (!ENABLE_LOGS) return
-  Log.d(
-    "$BASE_LOG_TAG | ${if (isFrequent) "FREQUENT" else "NON_FREQUENT"}",
-    message
-  )
-}
-
-class RotationGestureDetector(
-  private val gestureListener: OnRotationGestureListener?,
-  private val secondPointerLiftFinishesGesture: Boolean = DEFAULT_SECOND_POINTER_LIFT_FINISHES_GESTURE
-) {
-  companion object {
-    const val DEFAULT_SECOND_POINTER_LIFT_FINISHES_GESTURE = true
-  }
-
+class RotationGestureDetector(private val gestureListener: OnRotationGestureListener?) {
   interface OnRotationGestureListener {
     fun onRotation(detector: RotationGestureDetector): Boolean
     fun onRotationBegin(detector: RotationGestureDetector): Boolean
@@ -71,18 +53,17 @@ class RotationGestureDetector(
   private val pointerIds = IntArray(2)
 
   private fun updateCurrent(event: MotionEvent) {
-    localLog("updateCurrent", true)
-
     previousTime = currentTime
     currentTime = event.eventTime
-
     val firstPointerIndex = event.findPointerIndex(pointerIds[0])
     val secondPointerIndex = event.findPointerIndex(pointerIds[1])
 
     if (
       firstPointerIndex == MotionEvent.INVALID_POINTER_ID ||
       secondPointerIndex == MotionEvent.INVALID_POINTER_ID
-    ) { return }
+    ) {
+      return
+    }
 
     val firstPtX = event.getX(firstPointerIndex)
     val firstPtY = event.getY(firstPointerIndex)
@@ -96,16 +77,11 @@ class RotationGestureDetector(
     // Angle diff should be positive when rotating in clockwise direction
     val angle = -atan2(vectorY.toDouble(), vectorX.toDouble())
 
-    unpause(angle)
+    tryUnpause(angle)
 
     rotation = if (previousAngle.isNaN()) {
       0.0
     } else previousAngle - angle
-
-    localLog(
-      "updateCurrent | rotation: $rotation, angle: $angle, previousAngle: $previousAngle",
-      true
-    )
 
     previousAngle = angle
     if (rotation > Math.PI) {
@@ -121,31 +97,23 @@ class RotationGestureDetector(
   }
 
   /**
-   * Gesture may be paused when second Pointer lifts
-   * with secondPointerLiftFinishesGesture set to `false`
+   * Gesture pauses when second Pointer lifts
    *
    * Then Detector waits for a new second Pointer to arrive to continue Handling
-   * (last Pointer Lifting still finishes the Gesture)
+   * (last Pointer Lifting finishes the Gesture)
    *
-   * @see secondPointerLiftFinishesGesture
-   * @see pause
-   * @see unpause
+   * @see tryPause
+   * @see tryUnpause
    */
   private var isPaused = false
-  private fun pause() {
-    localLog(".pause() requested", true)
-    if (!isPaused) {
-      isPaused = true
-      localLog(".pause() fulfilled")
-    }
+  private fun tryPause() {
+    if (isPaused) return
+    isPaused = true
   }
-  private fun unpause(eventAngle: Double) {
-    localLog(".unpause(eventAngle: $eventAngle) requested", true)
-    if (isPaused) {
-      previousAngle = eventAngle
-      isPaused = false
-      localLog(".unpause(eventAngle: $eventAngle) fulfilled")
-    }
+  private fun tryUnpause(eventAngle: Double) {
+    if (!isPaused) return
+    previousAngle = eventAngle
+    isPaused = false
   }
 
   private fun finish() {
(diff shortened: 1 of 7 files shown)


DmitriyGalyanov:
@wcandillon 

Guys, does somebody see this?

Could you please tell me if it's _not needed / aligned with your opinion_ or will be overseen later

j-piasecki:
Hi! Sorry, I didn't have much time for reviews lately. I should be able to spend more time on reviews soon(-ish, in a week, maybe two).

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
```diff
@@ -53,10 +71,19 @@ class RotationGestureDetector(private val gestureListener: OnRotationGestureList
...
     previousTime = currentTime
     currentTime = event.eventTime
+
     val firstPointerIndex = event.findPointerIndex(pointerIds[0])
     val secondPointerIndex = event.findPointerIndex(pointerIds[1])
+
+    if (
+      firstPointerIndex == MotionEvent.INVALID_POINTER_ID ||
+      secondPointerIndex == MotionEvent.INVALID_POINTER_ID
+    ) { return }
```
j-piasecki, line 85:
```suggestion
    ) {
      return
    }
```

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
```diff
@@ -85,50 +120,109 @@ class RotationGestureDetector(private val gestureListener: OnRotationGestureList
...
+   *
+   * Then Detector waits for a new second Pointer to arrive to continue Handling
+   * (last Pointer Lifting still finishes the Gesture)
+   *
+   * @see secondPointerLiftFinishesGesture
+   * @see pause
+   * @see unpause
+   */
+  private var isPaused = false
+  private fun pause() {
```
j-piasecki, line 135:
```suggestion
  private fun tryPause() {
```

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
```diff
@@ -85,50 +120,109 @@ class RotationGestureDetector(private val gestureListener: OnRotationGestureList
...
+    localLog(".pause() requested", true)
+    if (!isPaused) {
+      isPaused = true
+      localLog(".pause() fulfilled")
+    }
+  }
+  private fun unpause(eventAngle: Double) {
```
j-piasecki, line 142:
```suggestion
  private fun tryUnpause(eventAngle: Double) {
```

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
```diff
@@ -85,50 +120,109 @@ class RotationGestureDetector(private val gestureListener: OnRotationGestureList
...
+        localLog("ActionMove END | ${getLogInfo()}", true)
+      }
+      MotionEvent.ACTION_POINTER_UP -> {
+        localLog("ActionPointerUp START | ${getLogInfo()}")
+        if (isInProgress) {
+          val pointerId = event.getPointerId(event.actionIndex)
+          if (pointerId == pointerIds[0] || pointerId == pointerIds[1]) {
+            // One of the key Pointers has been lifted up, we have to pause or finish the Gesture
+            if (secondPointerLiftFinishesGesture) finish()
+            else pause()
```
j-piasecki, line 216:
```suggestion
            if (secondPointerLiftFinishesGesture) {
              finish()
            } else {
              pause()
            }
```

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
```diff
@@ -85,50 +120,109 @@ class RotationGestureDetector(private val gestureListener: OnRotationGestureList
...
           }
```
j-piasecki, line 217: I think we should keep moving the pointer id to the first index if the gesture is in progress so we don't store a dangling id.

```kotlin
if (pointerId == pointerIds[0]) {
  pointerIds[0] = pointerIds[1]
}
pointerIds[1] = MotionEvent.INVALID_POINTER_ID
```

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
```diff
@@ -85,50 +120,109 @@ class RotationGestureDetector(private val gestureListener: OnRotationGestureList
...
+      MotionEvent.ACTION_POINTER_DOWN -> {
+        localLog("ActionPointerDown START | ${getLogInfo()}")
+        if (!isInProgress) {
+          pointerIds[1] = event.getPointerId(event.actionIndex)
+          isInProgress = true
+          previousTime = event.eventTime
+          previousAngle = Double.NaN
+          updateCurrent(event)
+          gestureListener?.onRotationBegin(this)
+        }
```
j-piasecki, line 192: Similarly here, the pointer id should be updated

```suggestion
        if (!isInProgress || isPaused) {
          pointerIds[1] = event.getPointerId(event.actionIndex)
          updateCurrent(event)
        }

        if (!isInProgress) {
          isInProgress = true
          previousTime = event.eventTime
          previousAngle = Double.NaN
          gestureListener?.onRotationBegin(this)
        }
```

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
```diff
@@ -85,50 +120,109 @@ class RotationGestureDetector(private val gestureListener: OnRotationGestureList
...
-        // All key pointers are up
-        if (!isInProgress && pointerId == pointerIds[0]) {
-          gestureListener?.onRotationEnd(this)
+        if (
+          firstPointerIndex != MotionEvent.INVALID_POINTER_ID &&
+          secondPointerIndex != MotionEvent.INVALID_POINTER_ID
+        ) {
+          updateCurrent(event)
+          gestureListener?.onRotation(this)
         }
```
j-piasecki, line 206: `updateCurrent` already checks for invalid pointers, this way it's clearer what this part does

```suggestion
        updateCurrent(event)

        if (!isPaused) {
          gestureListener?.onRotation(this)
        }
```

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
```diff
@@ -85,50 +120,109 @@ class RotationGestureDetector(private val gestureListener: OnRotationGestureList
...
       }
       MotionEvent.ACTION_MOVE -> if (isInProgress) {
-        updateCurrent(event)
-        gestureListener?.onRotation(this)
-      }
-      MotionEvent.ACTION_POINTER_UP -> {
-        val pointerId = event.getPointerId(event.actionIndex)
+        localLog("ActionMove START | ${getLogInfo()}", true)
+        val firstPointerIndex = event.findPointerIndex(pointerIds[0])
+        val secondPointerIndex = event.findPointerIndex(pointerIds[1])
```
j-piasecki, line 198: Then, this shouldn't be necessary 
```suggestion
```

j-piasecki (changes requested):
> I understand that this may be closer to default Android Gesture Recogniser Behaviour, but find it unexpected and inconvenient

I'm not sure whether it's closer, as a native rotation recognizer doesn't exist on Android AFAIK. Moreover, I'd lean towards saying that it should've been working like this since the beginning and classify that as a bugfix. Instead of making it opt-in, I'd simply make it a default behavior. I'm not sure whether we even need to keep the old one behind a flag, current behavior seems very counterintuitive to me and I don't really see anyone relying on it.

> Additionally, I added some local toggleable Logs, which I find useful for Debugging, but I’m not sure if it aligns with this Repo Style Guide (or could noticeably reduce Performance) – please notify me, if it’s prohibited

Prohibited is a strong word, but I'd remove the logs from the PR. We may consider adding them in the future but to the whole library instead of a single gesture.

Also, please run `yarn format:android` command before pushing to make sure it follows the code style.

Sorry for the delay on this 😅, there's been a lot going on recently.

DmitriyGalyanov:
@j-piasecki 
> Sorry for the delay on this 😅, there's been a lot going on recently.

Thank you for the reply!
Hopefully will update PR accordingly to your review this week

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
DmitriyGalyanov: If we simply toss out second Pointer on any `MotionEvent.ACTION_POINTER_UP` – Gesture will break on a non-key Pointer Lift (eg: third Finger)

I suggest tossing (replacing stored ID with `MotionEvent.INVALID_POINTER_ID`) exactly lifted Pointer

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
DmitriyGalyanov: According to suggested [here](https://github.com/software-mansion/react-native-gesture-handler/pull/3057#discussion_r1791654362) – replacing `MotionEvent.INVALID_POINTER_ID`

DmitriyGalyanov:
> I'm not sure whether it's closer, as a native rotation recognizer doesn't exist on Android AFAIK. Moreover, I'd lean towards saying that it should've been working like this since the beginning and classify that as a bugfix. Instead of making it opt-in, I'd simply make it a default behavior. I'm not sure whether we even need to keep the old one behind a flag, current behavior seems very counterintuitive to me and I don't really see anyone relying on it.

Glad to see that!
And, since both of us think it's closer to a bugfix, I've removed flag as redundant

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
j-piasecki: Ah, you're right. You can also immediately shift the second pointer so that `pointerIds[0]` is always a valid pointer id while the gesture is being recognized.

```kt
if (pointerId == pointerIds[0]) {
  pointerIds[0] = pointerIds[1]
  pointerIds[1] = MotionEvent.INVALID_POINTER_ID
  tryPause()
} else if (pointerId == pointerIds[1]) {
  pointerIds[1] = MotionEvent.INVALID_POINTER_ID
  tryPause()
}
```

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
j-piasecki: And here, if you [shift the pointers](https://github.com/software-mansion/react-native-gesture-handler/pull/3057/files/fec2867e424b465d6d28c025e0c5a436f58a4b9f#r1793052901), you don't need to replace the `MotionEvent.INVALID_POINTER_ID` but simply assume that `pointerIds[1]` should be set (like it's now).

android/src/main/java/com/swmansion/gesturehandler/core/RotationGestureDetector.kt
DmitriyGalyanov: Sounds good, thank you!

[Updated the PR](https://github.com/software-mansion/react-native-gesture-handler/pull/3057/commits/e5528492347e7341406d11e66a2cd8cec4cb9186)

j-piasecki (approved):
Thank you for working on it!

DmitriyGalyanov:
@j-piasecki 

Hello!

Could you, please, tell me when (at least approximately) would it be released?

j-piasecki:
@DmitriyGalyanov Most likely next week.

PR #3065, opened 2024-08-23
Playwright improvements

m-bert:
## Description

I've noticed that our `e2e` tests don't work anymore. Well.. they (almost) did, I've just wrongly assumed that you can open web on standard metro port `8081` like in our example app. The only thing that required fix was default port number.

Nevertheless, this made me look into our `e2e` app and it resulted in few changes:

- Bumped expo version
- Changed scripts:
    - `yarn start` - now this script opens `web` bu default. Also it starts expo without opening browser
    - `yarn test` - it is easier to use this instead of `npx playwright test`
- Changed default port to `19006`

>[!WARNING]
>Previously tests on `webkit` (i.e. `safari`) didn't work. Now after upgrading browsers they do... but tests on `firefox` fail. From what I've found, it is because for some reason [Locator.innerText](https://github.com/software-mansion/react-native-gesture-handler/blob/52c00e486c0637fa03d80c25d09edfab367dda38/e2e/web-tests/utils.ts#L29) returns empty string. Adding timeout doesn't fix it, so it doesn't seem like race condition. Maybe it will be fixed with new `firefox` update.

## Test plan

Go to `e2e/web-tests` and run:

1. `yarn start`
2. `yarn test`


j-piasecki (approved):
It would be nice to check the firefox again in the near future

m-bert:
> It would be nice to check the firefox again in the near future

True, maybe it will be fixed in newer versions

PR #3069, opened 2024-08-26
feat: Clangd support

coado:
## Description

This pull request introduces `clangd` support. You can read more about it in [this](https://github.com/software-mansion/react-native-reanimated/pull/6430) awesome PR that bring it to reanimated. For now it supports iOS only as I am not sure if we need it for android here. To make it work, you have build `FabricExample` or `MacOSExample` first.


coado:
After building the `FabricExample` on android, the `compile_commands.json` file will contain only one translation unit, which is for `cpp-adapter.cpp`, since this is the only used cpp file on android.

.gitignore
```diff
@@ -65,3 +65,8 @@ dist/
 
 # generated by bob
 /lib/
+
+# clangd
+.cache
+CompilationDatabase
+compile_commands.json
```
j-piasecki, line 72: Please keep the new line at the end of the file

scripts/clangd-add-xcode-step.rb
```diff
@@ -0,0 +1,5 @@
+def add_clangd_generation_step()
+  generator_script_path = '../../scripts/clangd-generate-xcode-metadata.sh'
+
+  script_phase :name => 'Generate metadata for clangd', :script => generator_script_path, :shell_path => '/bin/bash', :always_out_of_date => 'true', :execution_position => :after_compile
+end
```
j-piasecki, line 5: Same

android/build.gradle
```diff
@@ -188,6 +193,21 @@ android {
...
+                }
+
+                def rootDir = new File("${project.projectDir}/..")
+                def generated = new File("${compileTask.abi.getCxxBuildFolder()}/compile_commands.json")
+                def output = new File("${rootDir}/compile_commands.json")
+                output.text = generated.text
+
+                println("Generated clangd metadata.")
+            }
+    }
```
j-piasecki, line 210: I think this entire thing should be executed conditionally instead of early return inside.

FabricExample/android/build.gradle
```diff
@@ -1,5 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
+project.ext["isGHExampleApp"] = true
```
j-piasecki, line 3: Can't this be moved to the block below?

coado: Yeah, it should be

android/build.gradle
j-piasecki: Still up to date

j-piasecki (approved):
Looks good, can you merge main?

Issue #2725, opened 2024-01-11
Warning with `Task :react-native-gesture-handler:processDebugManifest`


ciuliene:
### Description

This warning occurs while building for Android

```
> Task :react-native-gesture-handler:processDebugManifest
package="com.swmansion.gesturehandler" found in source AndroidManifest.xml: /Users/gerrico/Documents/EmersonConnected/emerson-connected-app/node_modules/react-native-gesture-handler/android/src/main/AndroidManifest.xml.
Setting the namespace via the package attribute in the source AndroidManifest.xml is no longer supported, and the value is ignored.
Recommendation: remove package="com.swmansion.gesturehandler" from the source AndroidManifest.xml: /Users/gerrico/Documents/EmersonConnected/emerson-connected-app/node_modules/react-native-gesture-handler/android/src/main/AndroidManifest.xml.
```

### Steps to reproduce

Just run

```sh
npm run android
```

or 

```sh
npx react-native run-android
```

### Snack or a link to a repository

Private repository

### Gesture Handler version

2.14.0

### React Native version

0.73.2

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

ciuliene:
It's not so easy, as the repository is private. I'll try to provide you with an example as soon as possible

jacksonlevine:
Seconding this bug. Very annoying.

latekvo:
Hi @ciuliene and @jacksonlevine 

I've opened [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/3077) to hopefully fix the issues you're encountering, please let me know if that PR fixes those issues for you.

ciuliene:
> Hi @ciuliene and @jacksonlevine
> 
> I've opened [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/3077) to hopefully fix the issues you're encountering, please let me know if that PR fixes those issues for you.

Thank you, I will let you know as soon as I have tried it (hopefully tomorrow)

PR #3077, opened 2024-08-29
[Android] Drop support for `AGP version <= 7`

latekvo:
## Description

This PR removes support for Android Gradle Plugin `version <= 7`
`AGP version 7` is unsupported by React Native since `0.73`

Closes: #2725

## Test plan

- build example `Android` app while using `AGP 8.0`
- build example `Android` app while using `AGP 9.0`
- both should run successfully

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,8 +1,5 @@
 import groovy.json.JsonSlurper
 
-import javax.inject.Inject
-import java.nio.file.Files
-
 buildscript {
     def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['RNGH_kotlinVersion']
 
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -98,15 +95,11 @@ repositories {
...
-    }
 
-    if (agpVersion.tokenize('.')[0].toInteger() >= 8) {
-        buildFeatures {
-            buildConfig = true
-        }
+    namespace "com.swmansion.gesturehandler"
+    buildFeatures {
+        buildConfig = true
+        prefab true
```
j-piasecki, line 102:
```suggestion
        prefab = true
```

latekvo: I assumed this notation was intentional, but it looks like the previous notation was just a shorthand for the current one 😄

j-piasecki (approved):

PR #3108, opened 2024-09-17
Add chevron to the example app and center labels

j-piasecki:
## Description

I noticed the button labels were off-center, so I fixed them and added a chevron at the same time as it was looking kind of empty.

Screenshot 2024-09-17 at 11 27 33


m-bert (approved):
🚀

PR #3112, opened 2024-09-18
`stylusData` documentation.

m-bert:
## Description

This PR adds docs for `stylusData` object in events. 

>[!CAUTION]
> This has to be merged after finishing support for all main platforms (`web`/`android`/`iOS`)

## Test plan

Run docs


j-piasecki (approved):
What do you think about linking to MDN since we're targeting the web spec anyway?

m-bert:
@j-piasecki good idea 😄 I'll add it before merging.

PR #3107, opened 2024-09-17
[Web] Stylus support 

m-bert:
## Description

This PR adds more information about `stylus` to event. `Pan` and `Hover` events now have `stylusData` field, which contains the following information:

- `tiltX`
- `tiltY`
- `altitudeAngle`
- `azimuthAngle`
- `pressure`

>[!IMPORTANT]
> Because we receive only one set of data (either `tiltX/tiltY` or `altitudeAngle/azimuthAngle`, we use [this conversion algorithm](https://w3c.github.io/pointerevents/#converting-between-tiltx-tilty-and-altitudeangle-azimuthangle) to calculate second set.

>[!NOTE]
> This PR adds `stylusData` only on web, native platforms will be handled in another PRs
## Test plan

Tested on newly added example.

PR #3113, opened 2024-09-18
[iOS] Stylus support

m-bert:
## Description

This PR adds stylus support on `iOS`.

>[!WARNING]
> This PR lacks support for `Hover` since currently we are unable to test that.

>[!NOTE]
> You can read more about this feature in #3107

## Test plan

Tested on **_StylusData_** example


apple/Handlers/RNPanHandler.h
```diff
@@ -9,4 +9,7 @@
 #import "RNGestureHandler.h"
 
 @interface RNPanGestureHandler : RNGestureHandler
+#if !TARGET_OS_OSX && !TARGET_OS_TV
+@property (atomic, assign) StylusData *stylusData;
```
j-piasecki, line 13: I'm not a fan of keeping reference to the same StylusData both from the recognizer and the handler, especially when they both have access to each other.

apple/Handlers/RNPanHandler.m
```diff
@@ -78,6 +86,24 @@ - (void)setMinimumNumberOfTouches:(NSUInteger)minimumNumberOfTouches
...
   _realMinimumNumberOfTouches = minimumNumberOfTouches;
 }
+
+- (void)tryUpdateStylusData:(UIEvent *)event
+{
+  UITouch *touch = [[event allTouches] anyObject];
+
+  if (touch.type != UITouchTypePencil) {
+    return;
+  }
```
j-piasecki, line 96: Can you use both finger and pencil at the same time?

apple/Handlers/RNPanHandler.m
```diff
@@ -78,6 +86,24 @@ - (void)setMinimumNumberOfTouches:(NSUInteger)minimumNumberOfTouches
...
+
+  _stylusData.altitudeAngle = touch.altitudeAngle;
+  _stylusData.azimuthAngle = [touch azimuthAngleInView:nil];
+  _stylusData.pressure = touch.force / touch.maximumPossibleForce;
+
+  Tilt tilts = spherical2tilt(_stylusData.altitudeAngle, _stylusData.azimuthAngle);
+
+  _stylusData.tiltX = tilts.tiltX;
+  _stylusData.tiltY = tilts.tiltY;
```
j-piasecki, line 105: Why not keep the struct in StylusData?

apple/Handlers/RNPanHandler.m
```diff
@@ -102,6 +128,11 @@ - (void)interactionsBegan:(NSSet *)touches withEvent:(UIEvent *)event
   } else {
     super.minimumNumberOfTouches = _realMinimumNumberOfTouches;
   }
+
+  // We don't want to send stylusData object if stylus is not used. Therefore we set pressure to -1.
+  // If it stays at this value, it means that stylus was not used and we can omit sending stylus data.
+  _stylusData.pressure = -1;
```
j-piasecki, line 134: Shouldn't this be in reset?

apple/RNGHStylusData.h
```diff
@@ -0,0 +1,82 @@
...
+//  RNGHStylusData.h
+//  Pods
+//
+//  Created by Michał Bert on 18/09/2024.
+//
+
+#ifndef RNGHStylusData_h
+#define RNGHStylusData_h
+
+@interface StylusData : NSObject
```
j-piasecki, line 11:
```suggestion
@interface RNGHStylusData : NSObject
```

Everything in ObjC is in the global scope

apple/RNGHStylusData.h
```diff
@@ -0,0 +1,82 @@
...
+- (NSDictionary *)toDictionary;
+
+@end
+
+typedef struct {
+  double tiltX;
+  double tiltY;
+} Tilt;
+
+static Tilt spherical2tilt(double altitudeAngle, double azimuthAngle)
```
j-piasecki, line 28: Same here, either add some prefix or move it to a class

apple/RNGHStylusData.h
```diff
@@ -0,0 +1,82 @@
...
+@property (atomic, assign) double pressure;
+
+- (NSDictionary *)toDictionary;
+
+@end
+
+typedef struct {
+  double tiltX;
+  double tiltY;
+} Tilt;
```
j-piasecki, line 26: Same

apple/RNGHStylusData.m
```diff
@@ -0,0 +1,26 @@
...
+  NSDictionary *stylusDataObject = @{
+    @"tiltX" : @(_tiltX),
+    @"tiltY" : @(_tiltY),
+    @"altitudeAngle" : @(_altitudeAngle),
+    @"azimuthAngle" : @(_azimuthAngle),
+    @"pressure" : @(_pressure),
+  };
+
+  return stylusDataObject;
+}
```
j-piasecki, line 24:
```suggestion
- (NSDictionary *)toDictionary
{
  return @{
    @"tiltX" : @(_tiltX),
    @"tiltY" : @(_tiltY),
    @"altitudeAngle" : @(_altitudeAngle),
    @"azimuthAngle" : @(_azimuthAngle),
    @"pressure" : @(_pressure),
  };
}
```

apple/RNGestureHandlerEvents.h
```diff
@@ -24,12 +25,22 @@
...
                                 withPointerType:(NSInteger)pointerType;
+#if !TARGET_OS_OSX && !TARGET_OS_TV
++ (RNGestureHandlerEventExtraData *)forPan:(CGPoint)position
+                      withAbsolutePosition:(CGPoint)absolutePosition
+                           withTranslation:(CGPoint)translation
+                              withVelocity:(CGPoint)velocity
+                       withNumberOfTouches:(NSUInteger)numberOfTouches
+                           withPointerType:(NSInteger)pointerType
+                            withStylusData:(NSDictionary *)stylusData;
+#else
```
j-piasecki, line 36: Why not keep the one with stylus data and pass `nil` when there's no data? Significantly reduces `#ifs`.

j-piasecki (changes requested):
1. What about hover?
2. Do we want to add support for [rollAngle](https://developer.apple.com/documentation/uikit/uitouch/4391681-rollangle?language=objc)?

m-bert:
Let me answer those questions, @j-piasecki:

>What about hover?

It's hard to say as currently we have no way to test it. The two options that I see is either to implement it and hope that it will work fine, or wait for a device that will allow us to check if it is properly implemented.

Either way, I think we can leave it as TODO and finish it in another PR. Let me know what do you think.

>Do we want to add support for [rollAngle](https://developer.apple.com/documentation/uikit/uitouch/4391681-rollangle?language=objc)?

I'm not sure. If we do, we should also add support for [twist](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/twist) property (I guess they will be similar). Sadly, my phone doesn't support `twist` so I was unable to test that. Also I'll have to check if our device does support `rollAngle`.

The other reason why I have not added it yet is because there's no such thing on Android (at least I haven't seen anything like that in docs).

So in my opinion we can add it, but it is not a big deal if it's missing. What's your stance on that?

apple/Handlers/RNPanHandler.m
m-bert: No, you can't

apple/Handlers/RNPanHandler.h
m-bert: Totally agree with you! Changed in [e120bac](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/e120bacfea13e2af18a0757a977715ce22358a50)[e120bac](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/e120bacfea13e2af18a0757a977715ce22358a50)

apple/Handlers/RNPanHandler.m
m-bert: I wanted to keep it similar across all platforms, though I can change that if you prefer to 😓

apple/RNGestureHandlerEvents.h
m-bert: I'm not sure if I understood correctly, but if you mean simply doing:  `withStylusData: nil;` it will crash the application.

apple/Handlers/RNPanHandler.m
m-bert: Done in [252a7de](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/252a7de9500c43e4eae95af15714af28dcf37033)

apple/RNGHStylusData.m
m-bert: Done in [f482862](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/f48286269904e86688598e135378a5d3850cfdd4)

apple/RNGHStylusData.h
m-bert: Done in [b359bc4](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/b359bc4b89d43d0cfb1bce90c1086792e8d5485e)

apple/RNGHStylusData.h
m-bert: Done in [3b1ac10](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/3b1ac10e458a92ffbb3cbafdbc9024c36f122074)

j-piasecki:
I guess we can add it in the future alongside hover.

apple/RNGestureHandlerEvents.h
j-piasecki: Yes, because it's not handled at all. But you could be checking if `stylusData` exists instead of checking the pressure.

apple/Handlers/RNPanHandler.m
```diff
@@ -31,6 +32,12 @@ @interface RNBetterPanGestureRecognizer : UIPanGestureRecognizer
 @property (nonatomic) CGFloat failOffsetYEnd;
 @property (nonatomic) CGFloat activateAfterLongPress;
 
+#if !TARGET_OS_OSX && !TARGET_OS_TV
+@property (atomic, strong) RNGHStylusData *stylusData;
+
+- (RNGHStylusData *)getStylusData;
```
j-piasecki, line 38:
```suggestion
@property (atomic, readonly, strong) RNGHStylusData *stylusData;
```

Property should synthesize the getter

apple/Handlers/RNPanHandler.m
```diff
@@ -80,6 +88,31 @@ - (void)setMinimumNumberOfTouches:(NSUInteger)minimumNumberOfTouches
...
+  RNGHTilt tilts = ghSpherical2tilt(_stylusData.altitudeAngle, _stylusData.azimuthAngle);
+
+  _stylusData.tiltX = tilts.tiltX;
+  _stylusData.tiltY = tilts.tiltY;
+}
+
+- (RNGHStylusData *)getStylusData
+{
+  return _stylusData;
+}
```
j-piasecki, line 113:
```suggestion
```

apple/RNGHStylusData.h
```diff
@@ -0,0 +1,82 @@
...
+@property (atomic, assign) double pressure;
+
+- (NSDictionary *)toDictionary;
+
+@end
+
+typedef struct {
+  double tiltX;
+  double tiltY;
+} RNGHTilt;
```
j-piasecki, line 26: Wait, why not use `CGPoint`?

j-piasecki (changes requested):

apple/Handlers/RNPanHandler.m
j-piasecki: Nah, it just looked a bit funny to me

apple/RNGHStylusData.h
m-bert: True, we could use that!

Done in [b2a7437](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/b2a7437d18c51829dc29baf7cc62c65b3b419d9c)

apple/Handlers/RNPanHandler.m
m-bert: Done in [b2a7437](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/b2a7437d18c51829dc29baf7cc62c65b3b419d9c)

apple/Handlers/RNPanHandler.m
m-bert: Done in [b2a7437](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/b2a7437d18c51829dc29baf7cc62c65b3b419d9c)

tomekzaw:
Please test this PR on visionOS simulator prior to merging

m-bert:
>Please build this PR for visionOS simulator prior to merging

I've just done it and it builds. I'll check it one more time if the code changes 😅

apple/RNGestureHandlerEvents.h
j-piasecki: cc @m-bert

m-bert: Okay, I think I know what you mean. Let me know if [this change](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/16afa6eb59251f935fb65aa8d2f68762f663d2ca) is ok 😅

apple/Handlers/RNPanHandler.m
```diff
@@ -405,17 +449,21 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(NSPanGestureRecognizer *)rec
                                 withTranslation:[recognizer translationInView:recognizer.view.window.contentView]
                                    withVelocity:[recognizer velocityInView:recognizer.view.window.contentView]
                             withNumberOfTouches:1
-                                withPointerType:RNGestureHandlerMouse];
+                                withPointerType:RNGestureHandlerMouse
```
j-piasecki, line 452: This branch on macOS will fail as the method without `stylusData` no longer exists.

apple/Handlers/RNPanHandler.m
```diff
@@ -405,17 +449,21 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(NSPanGestureRecognizer *)rec
...
+  RNBetterPanGestureRecognizer *panRecognizer = (RNBetterPanGestureRecognizer *)recognizer;
+
   return [RNGestureHandlerEventExtraData forPan:[recognizer locationInView:recognizer.view]
                            withAbsolutePosition:[recognizer locationInView:recognizer.view.window]
                                 withTranslation:[recognizer translationInView:recognizer.view.window]
                                    withVelocity:[recognizer velocityInView:recognizer.view.window]
                             withNumberOfTouches:recognizer.numberOfTouches
-                                withPointerType:_pointerType];
+                                withPointerType:_pointerType
+                                 withStylusData:[panRecognizer.stylusData toDictionary]];
```
j-piasecki, line 466: I know that technically this will work since objC will return nil when trying to invoke a method on a nil object but can it at least get a comment? It's very confusing for my little brain. This is the stupidest feature in a language.

m-bert: Added in [35ce0de](https://github.com/software-mansion/react-native-gesture-handler/pull/3113/commits/35ce0de7b8a8d8bfff860e8b13c16a690ef27324) 😅

j-piasecki (approved):

m-bert:
Hi @RuudBurger, thanks for reporting this. Should be fixed by #3180

PR #3121, opened 2024-09-23
[docs] Improve Reanimated Swipeable documentation.

latekvo:
## Description

Fixed the example provided with the `Reanimated Swipeable` docs, it looks like a fragment of it was missing.

Also, what's the verdict on [this issue](https://github.com/software-mansion/react-native-gesture-handler/pull/2962#discussion_r1666562893)?
Today I had to take a look at these docs again and it really occurred to me the way it's currently written can be really confusing.

Here's my proposition, it removes the `[0, +]` mapping in favour of a verbose explanation:

> ### `renderRightActions`
> 
> accepts a function.  
> The function receives the following arguments:
> 
> - `progress` value representing swiping progress relative to the width of the returned element.  
>   `progress` is `0` when `swipeable` is closed, `1` when `right` action is fully visible.
> 
> - `translation` horizontal translation of `swipeable` in pixels.
> - `swipeable` provides an object exposing the methods listed [here](#methods)
> 
> This function must return a `ReactNode`.

side-by-side comparison:
old|new
---|---
image|image

Of course this change applies to both `renderLeftActions` and `renderRightActions`

### DIFF
--- a/docs/docs/components/reanimated_swipeable.md
+++ b/docs/docs/components/reanimated_swipeable.md
@@ -81,7 +81,8 @@ an argument.
 
 ### `renderLeftActions`
 
-accepts a function which receives the following arguments:
+a function that returns a component which will be rendered under the swipeable after swiping it to the right.
+The function receives the following arguments:
 
 - `progress` value representing swiping progress relative to the width of the returned element.  
   Equals `0` when `swipeable` is closed, `1` when left action is open.  
@@ -95,7 +96,8 @@ To support `rtl` flexbox layouts use `flexDirection` styling.
 
 ### `renderRightActions`
 
-accepts a function which receives the following arguments:
+a function that returns a component which will be rendered under the swipeable after swiping it to the right.
+The function receives the following arguments:
 
 - `progress` value representing swiping progress relative to the width of the returned element.  
   Equals `0` when `swipeable` is closed, `1` when right action is open.  
--- a/docs/docs/components/reanimated_swipeable.md
+++ b/docs/docs/components/reanimated_swipeable.md
@@ -30,7 +30,7 @@ import Swipeable from 'react-native-gesture-handler/ReanimatedSwipeable';
 ### `friction`
 
 a number that specifies how much the visual interaction will be delayed compared to the gesture distance.
-e.g. value of 1 will indicate that the swipeable panel should exactly follow the gesture, 2 means it is going to be two times "slower".
+e.g. value of `1` will indicate that the swipeable panel should exactly follow the gesture, `2` means it is going to be two times "slower".
 
 ### `leftThreshold`
 
@@ -42,11 +42,11 @@ distance from the right edge at which released panel will animate to the open st
 
 ### `dragOffsetFromLeftEdge`
 
-distance that the panel must be dragged from the left edge to be considered a swipe. The default value is 10.
+distance that the panel must be dragged from the left edge to be considered a swipe. The default value is `10`.
 
 ### `dragOffsetFromRightEdge`
 
-distance that the panel must be dragged from the right edge to be considered a swipe. The default value is 10.
+distance that the panel must be dragged from the right edge to be considered a swipe. The default value is `10`.
 
 ### `overshootLeft`
 
@@ -58,7 +58,7 @@ a boolean value indicating if the swipeable panel can be pulled further than the
 
 ### `overshootFriction`
 
-a number that specifies how much the visual interaction will be delayed compared to the gesture distance at overshoot. Default value is 1, it mean no friction, for a native feel, try 8 or above.
+a number that specifies how much the visual interaction will be delayed compared to the gesture distance at overshoot. Default value is `1`, it mean no friction, for a native feel, try `8` or above.
 
 ### `onSwipeableOpen`
 
@@ -112,16 +112,16 @@ To support `rtl` flexbox layouts use `flexDirection` styling.
 
 ### `containerStyle`
 
-style object for the container (Animated.View), for example to override `overflow: 'hidden'`.
+style object for the container (`Animated.View`), for example to override `overflow: 'hidden'`.
 
 ### `childrenContainerStyle`
 
-style object for the children container (Animated.View), for example to apply `flex: 1`.
+style object for the children container (`Animated.View`), for example to apply `flex: 1`.
 
 ### `enableTrackpadTwoFingerGesture` (iOS only)
 
 Enables two-finger gestures on supported devices, for example iPads with trackpads.
-If not enabled the gesture will require click + drag, with enableTrackpadTwoFingerGesture swiping with two fingers will also trigger the gesture.
+If not enabled the gesture will require click + drag, with `enableTrackpadTwoFingerGesture` swiping with two fingers will also trigger the gesture.
 
 ### `mouseButton(value: MouseButton)` (Web & Android only)
 
--- a/docs/docs/components/reanimated_swipeable.md
+++ b/docs/docs/components/reanimated_swipeable.md
@@ -85,10 +85,10 @@ a function that returns a component which will be rendered under the swipeable a
 The function receives the following arguments:
 
 - `progress` - a `SharedValue` representing swiping progress relative to the width of the returned element.
-  - Equals `0` when `swipeable` is closed, `1` when left action is open.
-  - Linearly increases to `Infinity` as left action overshoots it's open position.
-- `translation` - a horizontal offset of the `swipeable` relative to its closed position
-- `swipeableMethods` - provides an object exposing the methods listed [here](#methods)
+  - Equals `0` when `swipeable` is closed, `1` when `swipeable` is open.
+  - Linearly increases to `Infinity` as `swipeable` overshoots it's open position.
+- `translation` - a horizontal offset of the `swipeable` relative to its closed position.
+- `swipeableMethods` - provides an object exposing the methods listed [here](#methods).
 
 This function must return a `ReactNode`.
 
@@ -100,10 +100,10 @@ a function that returns a component which will be rendered under the swipeable a
 The function receives the following arguments:
 
 - `progress` - a `SharedValue` representing swiping progress relative to the width of the returned element.
-  - Equals `0` when `swipeable` is closed, `1` when right action is open.
-  - Linearly increases to `Infinity` as right action overshoots it's open position.
-- `translation` - a horizontal offset of the `swipeable` relative to its closed position
-- `swipeableMethods` - provides an object exposing the methods listed [here](#methods)
+  - Equals `0` when `swipeable` is closed, `1` when `swipeable` is open.
+  - Linearly increases to `Infinity` as `swipeable` overshoots it's open position.
+- `translation` - a horizontal offset of the `swipeable` relative to its closed position.
+- `swipeableMethods` - provides an object exposing the methods listed [here](#methods).
 
 This function must return a `ReactNode`.
 
--- a/docs/docs/components/reanimated_swipeable.md
+++ b/docs/docs/components/reanimated_swipeable.md
@@ -86,7 +86,7 @@ accepts a function which receives the following arguments:
 - `progress` value representing swiping progress relative to the width of the returned element.  
   Equals `0` when `swipeable` is closed, `1` when left action is open.  
   Linearly increases to `Infinity` as left action overshoots it's open position.
-- `translation` horizontal translation of `swipeable` in pixels.
+- `translation` horizontal translation of the `swipeable` in pixels.
 - `swipeable` provides an object exposing the methods listed [here](#methods)
 
 This function must return a `ReactNode`.
@@ -100,7 +100,7 @@ accepts a function which receives the following arguments:
 - `progress` value representing swiping progress relative to the width of the returned element.  
   Equals `0` when `swipeable` is closed, `1` when right action is open.  
   Linearly increases to `Infinity` as right action overshoots it's open position.
-- `translation` horizontal translation of `swipeable` in pixels.
+- `translation` horizontal translation of the `swipeable` in pixels.
 - `swipeable` provides an object exposing the methods listed [here](#methods)
 
 This function must return a `ReactNode`.
--- a/docs/docs/components/reanimated_swipeable.md
+++ b/docs/docs/components/reanimated_swipeable.md
@@ -29,7 +29,8 @@ import Swipeable from 'react-native-gesture-handler/ReanimatedSwipeable';
 
 ### `friction`
 
-a number that specifies how much the visual interaction will be delayed compared to the gesture distance. e.g. value of 1 will indicate that the swipeable panel should exactly follow the gesture, 2 means it is going to be two times "slower".
+a number that specifies how much the visual interaction will be delayed compared to the gesture distance.
+e.g. value of 1 will indicate that the swipeable panel should exactly follow the gesture, 2 means it is going to be two times "slower".
 
 ### `leftThreshold`
 
@@ -61,23 +62,19 @@ a number that specifies how much the visual interaction will be delayed compared
 
 ### `onSwipeableOpen`
 
-method that is called when action panel gets open (either right or left). Takes swipe direction as
-an argument.
+method that is called when action panel gets open (either right or left). Takes swipe direction as an argument.
 
 ### `onSwipeableClose`
 
-method that is called when action panel is closed. Takes swipe direction as
-an argument.
+method that is called when action panel is closed. Takes swipe direction as an argument.
 
 ### `onSwipeableWillOpen`
 
-method that is called when action panel starts animating on open (either right or left). Takes swipe direction as
-an argument.
+method that is called when action panel starts animating on open (either right or left). Takes swipe direction as an argument.
 
 ### `onSwipeableWillClose`
 
-method that is called when action panel starts animating on close. Takes swipe direction as
-an argument.
+method that is called when action panel starts animating on close. Takes swipe direction as an argument.
 
 ### `renderLeftActions`
 
@@ -119,7 +116,8 @@ style object for the children container (Animated.View), for example to apply `f
 
 ### `enableTrackpadTwoFingerGesture` (iOS only)
 
-Enables two-finger gestures on supported devices, for example iPads with trackpads. If not enabled the gesture will require click + drag, with enableTrackpadTwoFingerGesture swiping with two fingers will also trigger the gesture.
+Enables two-finger gestures on supported devices, for example iPads with trackpads.
+If not enabled the gesture will require click + drag, with enableTrackpadTwoFingerGesture swiping with two fingers will also trigger the gesture.
 
 ### `mouseButton(value: MouseButton)` (Web & Android only)
 
@@ -162,7 +160,8 @@ Unlike method `close`, this method does not trigger any animation.
 
 ### Example:
 
-For a more in-depth presentation of differences between the new and the legacy implementations, see [swipeable example](https://github.com/software-mansion/react-native-gesture-handler/blob/main/example/src/release_tests/swipeableReanimation/index.tsx) from GestureHandler Example App.
+For a more in-depth presentation of differences between the new and the legacy implementations,
+see [swipeable example](https://github.com/software-mansion/react-native-gesture-handler/blob/main/example/src/release_tests/swipeableReanimation/index.tsx) from GestureHandler Example App.
 
 ```jsx
 import React from 'react';
--- a/docs/docs/components/reanimated_swipeable.md
+++ b/docs/docs/components/reanimated_swipeable.md
@@ -86,7 +86,7 @@ a function that returns a component which will be rendered under the swipeable a
 The function receives the following arguments:
 
 - `progress` - a `SharedValue` representing swiping progress relative to the width of the returned element.
-  - Equals `0` when `swipeable` is closed, `1` when `swipeable` is open.
+  - Equals `0` when `swipeable` is closed, `1` when `swipeable` is opened.
   - Linearly increases to `Infinity` as `swipeable` overshoots it's open position.
 - `translation` - a horizontal offset of the `swipeable` relative to its closed position.
 - `swipeableMethods` - provides an object exposing the methods listed [here](#methods).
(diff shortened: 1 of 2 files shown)


m-bert:
> Also, what's the verdict on https://github.com/software-mansion/react-native-gesture-handler/pull/2962#discussion_r1666562893?

We could use `Infinity` to make it cleaner. Your suggested version does not say anything about acceptable values and I think it would be better to keep them. However, I'm still not fond of using `[0, -]`, even with `Infinity`.

latekvo:
>I'm still not fond of using `[0, -]`, even with `Infinity`.

Please let me know if the proposal written in description is ok then, it replaces `[0, -]` with a regular description of the input-output ranges.

m-bert:
>Your suggested version does not say anything about acceptable values and I think it would be better to keep them.

latekvo:
>>Your suggested version does not say anything about acceptable values and I think it would be better to keep them.

Do either of these work? I think the input-output range is implicitly stated in both of them.

- `progress` value representing swiping progress relative to the width of the returned element.  
  Equals `0` when `swipeable` is closed, `1` when `right action` is open,  
  and linearly increases to `Infinity` as `right action` overshoots it's open position.

- `progress` value representing swiping progress relative to the width of the returned element.  
  Starts at `0`, increases to `1` as `right action` reaches it's open position,  
  keeps increasing as `right action` overshoots it's open position.

m-bert:
What about:

- `progress` value representing swiping progress relative to the width of the returned element.
Equals 0 when swipeable is closed, 1 when right action is open.
Linearly increases to `Infinity` as right action overshoots it's open position.

?

latekvo:
>progress value representing swiping progress relative to the width of the returned element.
Equals 0 when swipeable is closed, 1 when right action is open.
Linearly increases to Infinity as right action overshoots it's open position.

Sounds good, added it in [2967cea](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/2967ceafe395a6362fed5a2277046a9718a02263)

docs/docs/components/reanimated_swipeable.md
```diff
@@ -81,27 +81,29 @@ an argument.
...
 
-method that is expected to return an action panel that is going to be revealed from the left side when user swipes right.
-This map describes the values to use as inputRange for extra interpolation:
+accepts a function which receives the following arguments:
 
-SharedValue: [startValue, endValue]
+- `progress` value representing swiping progress relative to the width of the returned element.  
+  Equals `0` when `swipeable` is closed, `1` when left action is open.  
+  Linearly increases to `Infinity` as left action overshoots it's open position.
+- `translation` horizontal translation of `swipeable` in pixels.
```
j-piasecki, line 89:
```suggestion
- `translation` horizontal translation of the `swipeable` in pixels.
```

docs/docs/components/reanimated_swipeable.md
```diff
@@ -81,27 +81,29 @@ an argument.
...
-This map describes the values to use as inputRange for extra interpolation:
-
-SharedValue: [startValue, endValue]
+accepts a function which receives the following arguments:
 
-progress: [0, 1]
+- `progress` value representing swiping progress relative to the width of the returned element.  
+  Equals `0` when `swipeable` is closed, `1` when right action is open.  
+  Linearly increases to `Infinity` as right action overshoots it's open position.
+- `translation` horizontal translation of `swipeable` in pixels.
```
j-piasecki, line 103:
```suggestion
- `translation` horizontal translation of the `swipeable` in pixels.
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -144,31 +144,26 @@ export interface SwipeableProps
   onSwipeableCloseStartDrag?: (direction: 'left' | 'right') => void;
 
   /**
-   *
-   * This map describes the values to use as inputRange for extra interpolation:
-   * AnimatedValue: [startValue, endValue]
-   *
-   * progressAnimatedValue: [0, 1] dragAnimatedValue: [0, +]
+   * `progress`: Equals `0` when action is closed, `1` when action is open, linearly increases to `Infinity` as action overshoots it's open position.\
+   * `translation`: translation of swipeable in pixels.
```
j-piasecki, line 148:
```suggestion
   * `translation`: translation of the swipeable in pixels.
```

src/components/ReanimatedSwipeable.tsx
```diff
@@ -144,31 +144,26 @@ export interface SwipeableProps
...
   ) => React.ReactNode;
+
   /**
-   *
-   * This map describes the values to use as inputRange for extra interpolation:
-   * AnimatedValue: [startValue, endValue]
-   *
-   * progressAnimatedValue: [0, 1] dragAnimatedValue: [0, -]
+   * `progress`: Equals `0` when action is closed, `1` when action is open, linearly increases to `Infinity` as action overshoots it's open position.\
+   * `translation`: translation of swipeable in pixels.
```
j-piasecki, line 160:
```suggestion
   * `translation`: translation of the swipeable in pixels.
```

docs/docs/components/reanimated_swipeable.md
latekvo: fixed in [6f68b59](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/6f68b5904f65ca9d81bbca4a8eb8e8d19f070f24)

docs/docs/components/reanimated_swipeable.md
latekvo: fixed in [6f68b59](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/6f68b5904f65ca9d81bbca4a8eb8e8d19f070f24)

src/components/ReanimatedSwipeable.tsx
latekvo: fixed in [6f68b59](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/6f68b5904f65ca9d81bbca4a8eb8e8d19f070f24)

latekvo:
As of [8c3fd7b](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/8c3fd7b4d2f21b112f05de8492844f4c1ef632a9) added a missing argument description to the docstrings.

docs/docs/components/reanimated_swipeable.md
```diff
@@ -81,27 +81,29 @@ an argument.
 
 ### `renderLeftActions`
 
-method that is expected to return an action panel that is going to be revealed from the left side when user swipes right.
-This map describes the values to use as inputRange for extra interpolation:
+accepts a function which receives the following arguments:
```
m-bert, line 84: a function that returns a component that is going to be rendered under `swipeable` after swiping to the right. The function receives the following arguments:

docs/docs/components/reanimated_swipeable.md
```diff
@@ -81,27 +81,29 @@ an argument.
...
 
-SharedValue: [startValue, endValue]
+- `progress` value representing swiping progress relative to the width of the returned element.
```
m-bert, line 86:
```suggestion
- `progress` - a shared value representing the swiping progress relative to the width of the returned element.
```

a tiny grammar change

docs/docs/components/reanimated_swipeable.md
```diff
@@ -81,27 +81,29 @@ an argument.
 
 ### `renderLeftActions`
 
-method that is expected to return an action panel that is going to be revealed from the left side when user swipes right.
-This map describes the values to use as inputRange for extra interpolation:
+accepts a function which receives the following arguments:
 
-SharedValue: [startValue, endValue]
+- `progress` value representing swiping progress relative to the width of the returned element.  
+  Equals `0` when `swipeable` is closed, `1` when left action is open.
```
m-bert, line 87:
```suggestion
  Equals `0` when `swipeable` is closed, `1` when `swipeable` is open.  
```

docs/docs/components/reanimated_swipeable.md
```diff
@@ -81,27 +81,29 @@ an argument.
...
 ### `renderLeftActions`
 
-method that is expected to return an action panel that is going to be revealed from the left side when user swipes right.
-This map describes the values to use as inputRange for extra interpolation:
+accepts a function which receives the following arguments:
 
-SharedValue: [startValue, endValue]
+- `progress` value representing swiping progress relative to the width of the returned element.  
+  Equals `0` when `swipeable` is closed, `1` when left action is open.  
+  Linearly increases to `Infinity` as left action overshoots it's open position.
```
m-bert, line 88:
```suggestion
When an element overshoots the opened position the value tends towards `Infinity`. Goes back to `1` when `swipeable` is released. 
```

docs/docs/components/reanimated_swipeable.md
```diff
@@ -81,27 +81,29 @@ an argument.
...
+- `translation` horizontal translation of the `swipeable` in pixels.
```
m-bert, line 89:
```suggestion
- `translation` - a horizontal offset from the starting position.
```

docs/docs/components/reanimated_swipeable.md
```diff
@@ -81,27 +81,29 @@ an argument.
...
+- `swipeable` provides an object exposing the methods listed [here](#methods)
```
m-bert, line 90: I believe this shouldn't be here in the bullet points of `renderLeftAction` at all

m-bert (changes requested):
Changes to `renderLeftActions`  also apply to `renderRightActions` 😅 Thanks @kacperkapusciak for help ❤️

docs/docs/components/reanimated_swipeable.md
latekvo: >shared value

Wouldn't `SharedValue` work better?

> the

Not sure if it's neccessary, the element it's refferring to is already unique ("the returned element"), and i don't think `progress` is a unique thing itself, as one swipeable has left action `progress`, and right action `progress`. 

> `  `

The double spaces are intentional, they ensure a line break.
While i think `\` would be clearer, it unfortunately doesn't work with our current setup.

docs/docs/components/reanimated_swipeable.md
latekvo: `swipeable` cannot be open, either the `left action` is open, or the `right action` can be open.

docs/docs/components/reanimated_swipeable.md
latekvo: i would keep the `in pixels`, is that ok?

docs/docs/components/reanimated_swipeable.md
latekvo: are you referring to the following?
>- `swipeable` provides an object exposing the methods listed [here](#methods)

If that's the case, it **should** be, it's a reference to the `swipeable` object.

Do you think something like `swipeableReference` or `swipeableMethods` or `swipeableObject` works better?
I think standalone `swipeable` makes the most sense, but i'm not sure if it's the as verbose as the other options.

(cc @kacperkapusciak )

docs/docs/components/reanimated_swipeable.md
latekvo: I think the second sentence is a good addition: ```Goes back to `1` when `swipeable` is released.```.
But for the first one, i think the original works a bit better, it's shorter and i think the wording it uses is a bit clearer

Perhaps the following would work better?
>  As left action `progress` overshoots it's open position, the value linearly increases to `Infinity`.
>  Reverts to `1` when `swipeable` is released.

docs/docs/components/reanimated_swipeable.md
latekvo: Also, while `translation` implicitly suggests a global change, i think `offset` might sound like the change is relative to the action, so would maybe the following work better?
>- `translation` - a horizontal offset of the `swipeable` relative to it's closed position.

docs/docs/components/reanimated_swipeable.md
m-bert: We think that _**under**_ fits better in that case

docs/docs/components/reanimated_swipeable.md
m-bert: > Wouldn't `SharedValue` work better?

I think that both are fine, `SharedValue` is more straightforward.

> Not sure if it's necessary, the element its referring to is already unique (...)

It is not referring to the element itself. Also I believe we can keep it.

> The double spaces are intentional, they ensure a line break.

What about making it a sub-bullet instead?

docs/docs/components/reanimated_swipeable.md
m-bert: > `swipeable` cannot be open

Why is that? For example, it has `onSwipeableWillOpen` and `onSwipeableOpen` props. What I would change here is I'd use passive voice, i.e.:

```suggestion
Equals `0` when `swipeable` is closed, `1` when `swipeable` is opened. 
```

docs/docs/components/reanimated_swipeable.md
m-bert: We wanted to make it as close to natural sentence as possible. Also we'd like to avoid `left action`, that's why we used `element`

docs/docs/components/reanimated_swipeable.md
m-bert: Sounds good

docs/docs/components/reanimated_swipeable.md
m-bert: Okay, now I see. I don't like `swipeable` in that case (even though it could be okay with old implementation). I'd go with either `swipeableRef` or `swipeableMethods` (preferably `swipeableMethods`)

docs/docs/components/reanimated_swipeable.md
latekvo: `onSwipeableWillOpen` distinguishes between `swipeable` opening it's left and right `element`.
Here, the distinction of which side is open is equally necessary.

What about something like this?

> Equals `0` when `swipeable` is closed, `1` when `swipeable` has its left `element` open.

or

> Equals `0` when `swipeable` is closed, `1` when the left `element` is open.

---

`element` could be replaced with `action`, but I think you mentioned you'd prefer to avoid `action` [here](https://github.com/software-mansion/react-native-gesture-handler/pull/3121#discussion_r1782573527), was that a context-specific preference or a general rule?

docs/docs/components/reanimated_swipeable.md
latekvo: Makes sense, done: [e37a656](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/e37a6568d2f6d939cd873bec95a7d103980d8f12)

docs/docs/components/reanimated_swipeable.md
latekvo: > I think that both are fine, SharedValue is more straightforward.

Alright
>It is not referring to the element itself. Also I believe we can keep it.


Alright



>What about making it a sub-bullet instead?

Is this what you meant?

image

docs/docs/components/reanimated_swipeable.md
j-piasecki: Is it in pixels? I think it will be in the scaled units (DIP on Android and Points on iOS) that take screen density into consideration

docs/docs/components/reanimated_swipeable.md
j-piasecki: > Here, the distinction of which side is open is equally necessary.

Isn't it implied by the fact that we're in `renderLeftActions` section?

m-bert: > `onSwipeableWillOpen` distinguishes between `swipeable` opening it's left and right element.

Yes it does, but the name indicates that `Swipeable` itself will be opened.  Why can't you say that it is _**opened**_ if you can say that it is _**closed**_? 

>`element` could be replaced with `action`

I think the best we could replace it with would be `action panel`, as in `tsdocs`. But still, I believe that `Swipeable` fits better here. (cc @j-piasecki, @kacperkapusciak)

docs/docs/components/reanimated_swipeable.md
m-bert: I thought about 2 sub-bullets (2 last sentences)

latekvo: image

docs/docs/components/reanimated_swipeable.md
latekvo: changed to `swipeableMethods` in [6cdf23b](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/6cdf23b6963f35202c0b65225e38e3839fe83c96)

docs/docs/components/reanimated_swipeable.md
j-piasecki: I think we can use that if you want to keep the unit

docs/docs/components/reanimated_swipeable.md
m-bert: That's better. We just have to make adjustments from other comments 😅

docs/docs/components/reanimated_swipeable.md
latekvo: Alright, this is how it currently looks like, and i think it can remain that way, `offset` already implies `DIPs`:

>- `translation` - a horizontal offset of the `swipeable` relative to its closed position

docs/docs/components/reanimated_swipeable.md
latekvo: >That's better. We just have to make adjustments from other comments 😅

I think the current version ([9845471](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/9845471eab4539a1a6d9f4164ab3483a159b296e)) has all the adjustments that were requested.

docs/docs/components/reanimated_swipeable.md
latekvo: I agree, if we want to avoid using terms like left/right action, since these are already implied by the function name, then standalone `swipeable` definitely fits best.

Fixed in [df0ff8c](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/df0ff8cb7be3b0c651b048dced53ffa8b4731b42) throughout [9845471](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/9845471eab4539a1a6d9f4164ab3483a159b296e)

docs/docs/components/reanimated_swipeable.md
latekvo: >We wanted to make it as close to natural sentence as possible. Also we'd like to avoid left action, that's why we used element

I believe this is resolved [here](https://github.com/software-mansion/react-native-gesture-handler/pull/3121#discussion_r1782999151)

docs/docs/components/reanimated_swipeable.md
```diff
@@ -49,59 +50,63 @@ distance that the panel must be dragged from the right edge to be considered a s
...
 ### `renderLeftActions`
 
-method that is expected to return an action panel that is going to be revealed from the left side when user swipes right.
-This map describes the values to use as inputRange for extra interpolation:
+a function that returns a component which will be rendered under the swipeable after swiping it to the right.
+The function receives the following arguments:
 
-SharedValue: [startValue, endValue]
+- `progress` - a `SharedValue` representing swiping progress relative to the width of the returned element.
+  - Equals `0` when `swipeable` is closed, `1` when `swipeable` is open.
```
m-bert, line 89:
```suggestion
  - Equals `0` when `swipeable` is closed, `1` when `swipeable` is opened.
```

WDYT about passive voice?

latekvo: Sounds good, will apply it shortly.

docs/docs/components/reanimated_swipeable.md
latekvo: >Am I missing something, or the only thing that has changed is left action?

I'm not sure which PR you're referencing, in [df0ff8c](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/df0ff8cb7be3b0c651b048dced53ffa8b4731b42) both left and right got replaced.

Are you talking about some PR between [df0ff8c](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/df0ff8cb7be3b0c651b048dced53ffa8b4731b42) and [9845471](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/9845471eab4539a1a6d9f4164ab3483a159b296e)?

image

docs/docs/components/reanimated_swipeable.md
latekvo: Fixed in [b93ec82](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/b93ec82ccd6883768b258354fdf36bb33d42bdb2)

docs/docs/components/reanimated_swipeable.md
latekvo: fixed in [620e247](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/620e247aa4a312d1b649bafe665060594341214f)

docs/docs/components/reanimated_swipeable.md
```diff
@@ -29,7 +29,8 @@ import Swipeable from 'react-native-gesture-handler/ReanimatedSwipeable';
 
 ### `friction`
 
-a number that specifies how much the visual interaction will be delayed compared to the gesture distance. e.g. value of 1 will indicate that the swipeable panel should exactly follow the gesture, 2 means it is going to be two times "slower".
+a number that specifies how much the visual interaction will be delayed compared to the gesture distance.
+e.g. value of 1 will indicate that the swipeable panel should exactly follow the gesture, 2 means it is going to be two times "slower".
```
m-bert, line 33:
```suggestion
e.g. value of `1` will indicate that the swipeable panel should exactly follow the gesture, `2` means it is going to be two times "slower".
```

m-bert (approved):
The time has come ⌛

docs/docs/components/reanimated_swipeable.md
```diff
@@ -115,7 +120,8 @@ style object for the children container (Animated.View), for example to apply `f
 
 ### `enableTrackpadTwoFingerGesture` (iOS only)
 
-Enables two-finger gestures on supported devices, for example iPads with trackpads. If not enabled the gesture will require click + drag, with enableTrackpadTwoFingerGesture swiping with two fingers will also trigger the gesture.
+Enables two-finger gestures on supported devices, for example iPads with trackpads.
+If not enabled the gesture will require click + drag, with enableTrackpadTwoFingerGesture swiping with two fingers will also trigger the gesture.
```
m-bert, line 124:
```suggestion
If not enabled the gesture will require click + drag, with `enableTrackpadTwoFingerGesture` swiping with two fingers will also trigger the gesture.
```

docs/docs/components/reanimated_swipeable.md
latekvo: fixed in [65d1a2b](https://github.com/software-mansion/react-native-gesture-handler/pull/3121/commits/65d1a2bca527024bbd00c90b198b993f61619f6b)

docs/docs/components/reanimated_swipeable.md
latekvo: same

PR #3122, opened 2024-09-23
[macOS] Fix rotation

m-bert:
## Description

Turns out that rotation on macOS works in opposite direction comparing to other platforms. This PR fixes that behavior.

## Test plan

<details>
<summary>Tested on the following code (check `rotation` value in event)</summary>

```tsx
import React from 'react';

  const g = Gesture.Rotation().onChange(console.log);
  return (
      <GestureDetector gesture={g}>
        <View style={{ width: 300, height: 300, backgroundColor: 'red' }} />
      </GestureDetector>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
});
```

</details>

### DIFF
--- a/apple/Handlers/RNRotationHandler.m
+++ b/apple/Handlers/RNRotationHandler.m
@@ -149,7 +149,7 @@
 #if TARGET_OS_OSX
 - (RNGestureHandlerEventExtraData *)eventExtraData:(NSRotationGestureRecognizer *)recognizer
 {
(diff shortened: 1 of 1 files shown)


latekvo (approved):
Tested on `iOS` and `macOS`, they now both function identically 👍

PR #3127, opened 2024-09-25
[Web] Remove `findNodeHandle` function.

m-bert:
## Description

`findNodeHandle` is a utility function that allows us to get native view from React component. The problem is, that under the hood it uses `findDOMNode`, which is deprecated and [will be removed in React 19](https://react.dev/reference/react-dom/findDOMNode). To get rid of this function, I've written our implementation that meets our requirements. 

Also, for new API we had to introduce a little trick (thanks @j-piasecki) - we add wrapper `div` with `display: contents;` style and we pass a ref to this element. With that change, it is much easier to get required `HTMLElement` - we simply have to get `div` from `ref` and find first child (preferably the one without `display: contents;`)

>[!WARNING]
> You may find tat in `StrictMode` most of new API examples doesn't work. That's not really the case. If you check example from [this comment](https://github.com/software-mansion/react-native-gesture-handler/pull/3127#issuecomment-2376061542) it works fine. What really don't work in these examples are the animations.

## Test plan

Tested that example app works and no `findNodeHandle` error is thrown in `StrictMode`


latekvo (approved):
couldn't find any issues 👍

j-piasecki:
Could you check if this still works after the changes:

```jsx
import React from 'react';
import { GestureDetector, Gesture } from 'react-native-gesture-handler';

  const [key, setKey] = React.useState(1);

  const tap = Gesture.Tap()
    .onStart(() => {
      console.log('Tapped');
      setKey((prev) => prev + 1);
    })
    .runOnJS(true);

  return (
        <View
          key={key}
        />
      </GestureDetector>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
});
```

m-bert:
> Could you check if this still works after the changes

Yes it does 😅

src/handlers/gestures/GestureDetector/Wrap.tsx
```diff
@@ -15,12 +18,22 @@ export class Wrap extends React.Component<{
...
       );
+
+      if (Platform.OS === 'web') {
+        return (
+          <div ref={this.props.forwardedRef} style={{ display: 'contents' }}>
+            {clone}
+          </div>
+        );
+      }
+      return clone;
```
j-piasecki, line 36: What if we made an alternative Wrap component for web that was using `forwardRef` and pass the ref to the `div`. This way the second ref in the state for web isn't needed and I think the logic in other files remains unchanged.

src/findNodeHandle.web.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (viewRef?.viewTag !== undefined) {
+    return findNodeHandle(viewRef.viewTag);
+  }
+
+  if (viewRef instanceof HTMLElement) {
+    return viewRef;
+  }
+
+  // In new API, we receive ref object which `current` field points to  wrapper `div` with `display: contents;`.
+  // We want to return first child that doesn't have this property.
```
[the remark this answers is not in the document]
m-bert: Changed in [df1eee7](https://github.com/software-mansion/react-native-gesture-handler/pull/3127/commits/df1eee7f4ecd88c70d63653f01b7dcec29ef7b37)

src/handlers/gestures/GestureDetector/Wrap.tsx
m-bert: Let me know what do you think of current approach 😅

src/findNodeHandle.web.ts
```diff
@@ -0,0 +1,33 @@
+type GestureHandlerRef = {
+  viewTag: GestureHandlerRef;
+  current: HTMLElement;
+};
+
+export default function findNodeHandle(
+  viewRef: GestureHandlerRef | HTMLElement
+): HTMLElement | number | undefined {
```
latekvo, line 8: The `findNodeHandle` returns `null | number | React.Component<any, any> | React.ComponentClass<any>`.

Is there a reason why `undefined` is used here instead of `null`?

src/handlers/gestures/GestureDetector/Wrap.web.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { forwardRef } from 'react';
+import type { LegacyRef } from 'react';
+import { tagMessage } from '../../../utils';
+
+let wrapID = 0;
+
+export const Wrap = forwardRef(
+  ({ children }: { children?: React.ReactNode }, ref) => {
```
j-piasecki, line 8: Can the type be set as generic type of `forwardRef`? I think it would look something like `forwardRef<HTMLElement, PropsWithChildren<{}>`

m-bert: Sure, done in [4fab7ef](https://github.com/software-mansion/react-native-gesture-handler/pull/3127/commits/4fab7eff03080207fea1f8cfcc904958de50fade)

src/handlers/gestures/GestureDetector/Wrap.web.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { forwardRef } from 'react';
+import type { LegacyRef, PropsWithChildren } from 'react';
+import { tagMessage } from '../../../utils';
+
+let wrapID = 0;
```
j-piasecki, line 5: Do we need the ids?

m-bert: Seems like we don't. I've used that to check if view was added by Gesture Handler, but it looks like checking for `display: contents;` is enough

Removed in [7a11ad5](https://github.com/software-mansion/react-native-gesture-handler/pull/3127/commits/7a11ad5b0bb6c24761e34447b0e91b2e1106b3c3)

j-piasecki (approved):

PR #3129, opened 2024-09-27
Rewrite `bottomSheet` example as functional component

latekvo:
## Description

Rewrite `bottomSheet` example to be a function style component, to fix crashes on macOS and make working with this example easier.

## Test plan

- open the `showcase/bottomSheet` example in the example app.
- see how it works like it did previously


latekvo:
This change is a duplicate of `example/src/new_api/bottom_sheet`.
Fixes regarding `Dimensions.get('window').height` crashing on `MacOS` will be done in a separate PR, changing the entire file structure is not necessary to fix these crashes.

PR #3135, opened 2024-10-03
Disable examples on unsupported platforms

m-bert:
## Description

Some of our examples are platform-specific, which means that running them on other platforms doesn't make much sense. In this PR I've disabled navigation buttons so that opening such examples is not possible on unsupported platforms.

E.g.: _**Web styles reset**_ works only on `web`. After this PR you'll no longer be able to open it on other platforms, like `ios` or `android`.

image

## Test plan

Go through example app

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -86,7 +86,7 @@ import { Icon } from '@swmansion/icons';
 interface Example {
   name: string;
   component: React.ComponentType;
-  disabledOn?: Set<String>;
+  unsupportedPlatforms?: Set<String>;
 }
 interface ExamplesSection {
   sectionTitle: string;
@@ -135,12 +135,12 @@ const EXAMPLES: ExamplesSection[] = [
       {
         name: 'Pager & drawer',
         component: PagerAndDrawer,
-        disabledOn: new Set(['web', 'ios', 'macos']),
+        unsupportedPlatforms: new Set(['web', 'ios', 'macos']),
       },
       {
         name: 'Force touch',
         component: ForceTouch,
-        disabledOn: new Set(['web', 'android']),
+        unsupportedPlatforms: new Set(['web', 'android']),
       },
       { name: 'Fling', component: Fling },
     ],
@@ -182,7 +182,7 @@ const EXAMPLES: ExamplesSection[] = [
       {
         name: 'Nested buttons (sound & ripple)',
         component: NestedButtons,
-        disabledOn: new Set(['web', 'ios', 'macos']),
+        unsupportedPlatforms: new Set(['web', 'ios', 'macos']),
       },
       { name: 'Double pinch & rotate', component: DoublePinchRotate },
       { name: 'Double draggable', component: DoubleDraggable },
@@ -194,7 +194,7 @@ const EXAMPLES: ExamplesSection[] = [
       {
         name: 'ContextMenu',
         component: ContextMenu,
-        disabledOn: new Set(['android', 'ios', 'macos']),
+        unsupportedPlatforms: new Set(['android', 'ios', 'macos']),
       },
       { name: 'PointerType', component: PointerType },
       { name: 'Swipeable Reanimation', component: SwipeableReanimation },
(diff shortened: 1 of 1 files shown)


example/App.tsx
```diff
@@ -86,6 +86,7 @@ import { Icon } from '@swmansion/icons';
 interface Example {
   name: string;
   component: React.ComponentType;
+  unsupportedPlatforms?: Set<String>;
```
latekvo, line 89:
```suggestion
  unsupportedPlatforms?: Set<typeof Platform.OS>;
```

example/App.tsx
```diff
@@ -131,8 +132,16 @@ const EXAMPLES: ExamplesSection[] = [
       { name: 'Bouncing box', component: BouncingBox },
       { name: 'Pan responder', component: PanResponder },
       { name: 'Horizontal drawer', component: HorizontalDrawer },
-      { name: 'Pager & drawer', component: PagerAndDrawer },
-      { name: 'Force touch', component: ForceTouch },
+      {
+        name: 'Pager & drawer',
+        component: PagerAndDrawer,
+        unsupportedPlatforms: new Set(['web', 'ios', 'macos']),
+      },
```
latekvo, line 139: Do we want to grey out examples that are supposed to work on all platforms but are broken?

(cc. @j-piasecki)

example/App.tsx
m-bert: Done in [f475fb0](https://github.com/software-mansion/react-native-gesture-handler/pull/3135/commits/f475fb0ad88a2515494b97bb23713a2ee6f7c341)

example/App.tsx
```diff
@@ -131,8 +132,16 @@ const EXAMPLES: ExamplesSection[] = [
...
+      {
+        name: 'Force touch',
+        component: ForceTouch,
+        unsupportedPlatforms: new Set(['web', 'android']),
```
j-piasecki, line 143: Does force touch work on macOS?

example/App.tsx
```diff
@@ -335,13 +354,20 @@ function OpenLastExampleSetting() {
...
 
-function MainScreenItem({ name, onPressItem }: MainScreenItemProps) {
+function MainScreenItem({ name, onPressItem, enabled }: MainScreenItemProps) {
   return (
-    <RectButton style={[styles.button]} onPress={() => onPressItem(name)}>
+    <RectButton
+      enabled={enabled}
+      style={[
+        styles.button,
+        !enabled && { backgroundColor: 'rgba(220, 220, 220, 0.8)' },
```
j-piasecki, line 366: This looks like highlighting not disabling, can you make the text less opaque?

example/App.tsx
m-bert: Theoretically it could. In practice, we get warning that force touch is not available. I think we can disable this example, right?

example/App.tsx
m-bert: Done in [91072ac](https://github.com/software-mansion/react-native-gesture-handler/pull/3135/commits/91072acb376ed50de69547d61f91695ca627bec6)

example/App.tsx
m-bert: I've done it in [9fcc315](https://github.com/software-mansion/react-native-gesture-handler/pull/3135/commits/9fcc31595e5e1092f9d1fb0482bc1e7b3dc32cee)

example/App.tsx
```diff
@@ -335,13 +354,20 @@ function OpenLastExampleSetting() {
...
 
-function MainScreenItem({ name, onPressItem }: MainScreenItemProps) {
+function MainScreenItem({ name, onPressItem, enabled }: MainScreenItemProps) {
   return (
-    <RectButton style={[styles.button]} onPress={() => onPressItem(name)}>
+    <RectButton
+      enabled={enabled}
+      style={[
+        styles.button,
+        !enabled && { backgroundColor: 'rgb(220, 220, 220)', opacity: 0.3 },
```
j-piasecki, line 366: Could this also be moved to `styles` instead of being inline?

j-piasecki (approved):

example/App.tsx
m-bert: Done n [b4a01e3](https://github.com/software-mansion/react-native-gesture-handler/pull/3135/commits/b4a01e3f0dc4d00411205338d0945b8b8cc95957)

PR #3138, opened 2024-10-04
Remove redundant example entry in `App.tsx`

latekvo:
## Description

There are 2 `Pressable` entries, likely as a result of merge conflicts which occurred between these two PRs:

[Update examples](https://github.com/software-mansion/react-native-gesture-handler/pull/2919/files) and [Make osx example utilise the common example app](https://github.com/software-mansion/react-native-gesture-handler/pull/3055)


m-bert (commented):
Why don't we keep it as `Pressable`? I agree that it's new, but we haven't had that before 😅

m-bert (approved):

PR #3139, opened 2024-10-04
[Web] Simplify `PointerTracker` methods

m-bert:
## Description

This PR simplifies some methods found in `PointerTracker`. I've also found a little bug which was not caught because of how 
`tracker` worked before. 

## Test plan

Verified on example app (mostly `Draggable`/`Double Draggable` examples)

### DIFF
--- a/src/web/tools/GestureHandlerOrchestrator.ts
+++ b/src/web/tools/GestureHandlerOrchestrator.ts
@@ -337,26 +337,13 @@ export default class GestureHandlerOrchestrator {
     const isPointerWithinBothBounds = (pointer: number) => {
       const point = handler.getTracker().getLastAbsoluteCoords(pointer);
 
-      // In some rare cases it may happen that this method will be called before handler
-      // has started tracking pointers. When it happens, `point` will be undefined, leading to
-      // silent fail
-      if (!point) {
-        return false;
-      }
-
       return (
         handler.getDelegate().isPointerInBounds(point) &&
         otherHandler.getDelegate().isPointerInBounds(point)
       );
     };
 
-    const handlerPointers: number[] = handler.getTrackedPointersID();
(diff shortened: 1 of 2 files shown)


src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -337,6 +337,13 @@ export default class GestureHandlerOrchestrator {
     const isPointerWithinBothBounds = (pointer: number) => {
       const point = handler.getTracker().getLastAbsoluteCoords(pointer);
 
+      // In some rare cases it may happen that this method will be called before handler
+      // has started tracking pointers. When it happens, `point` will be undefined, leading to
+      // silent fail
+      if (!point) {
+        return false;
+      }
+
```
j-piasecki, line 346: The logic in `checkOverlap` seems a bit weird to me.

`isPointerWithinBothBounds` gets the position of the pointer from `handler` but the method is called for every pointer from `handler` **and** `otherHandler`. The only case when this works is when all the pointers are shared, otherwise, calling the method on a pointer tracked by `otherHandler` but not by `handler` will result in `point` being not found.

I think that removing `otherPointers.some(isPointerWithinBothBounds)` entirely from the return expression will get rid of this problem without a change in behavior. (it will find a pointer only for pointers present in `handlerPointers ∩ otherPointers` which is a subset of `handlerPointers` which was already checked)

m-bert: I see that you've already looked at something that I wanted to take care of after merging this one 😅 I agree that there's something wrong with this logic. I've changed it in [4fdac6f](https://github.com/software-mansion/react-native-gesture-handler/pull/3139/commits/4fdac6f2c4b9d98453a516de2c84d62b679a7887), please have a look and tell me what you think 😅

j-piasecki (approved):

PR #3142, opened 2024-10-07
[MacOS] Fix `Pressable` being unresponsive

latekvo:
## Description

This PR fixes incorrect `Pressable` behaviour on `MacOS`, which was caused by the following sub-issues:

- Pointer tracker on `MacOS` kept returning incorrect relative coordinates.
This was caused by converting the `absolute` coordinates from the `local` coordinate space to the `absolute` coordinate space instead of the other way around.

- There was no workflow set on the `Gesture.Native()` to allow for activation of `Pressable`

## Test plan

- Open any `Pressable` example on `MacOS`
- See how they all should work now.

### DIFF
--- a/src/components/Pressable/Pressable.tsx
+++ b/src/components/Pressable/Pressable.tsx
@@ -302,7 +302,7 @@ export default function Pressable(props: PressableProps) {
       Gesture.Native()
         .onBegin(() => {
           // Android sets BEGAN state on press down
-          if (Platform.OS === 'android') {
+          if (Platform.OS === 'android' || Platform.OS === 'macos') {
             isTouchPropagationAllowed.current = true;
           }
         })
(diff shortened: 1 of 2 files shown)


latekvo:
Broken when testing on nested native buttons.

m-bert (approved):

j-piasecki (approved):

src/components/Pressable/Pressable.tsx
```diff
@@ -358,11 +358,7 @@ export default function Pressable(props: PressableProps) {
     gesture.enabled(isPressableEnabled);
     gesture.runOnJS(true);
     gesture.hitSlop(appliedHitSlop);
-    gesture.shouldCancelWhenOutside(false);
-
-    if (Platform.OS !== 'web') {
-      gesture.shouldCancelWhenOutside(true);
-    }
+    gesture.shouldCancelWhenOutside(Platform.OS === 'web' ? false : true);
```
Romick2005, line 361: Can it be like `gesture.shouldCancelWhenOutside(Platform.OS !== 'web');`

m-bert: Yes, it could 😅

PR #3145, opened 2024-10-09
Fix swipeable swiping to left when no `renderRightActions` is set

latekvo:
## Description

Fixed `Swipeable` being able to be opened to the left when no right element is present.
This was caused by incorrectly set `rightOffset`, equal to `0` when no elements were present, even though it should've been set to `rowWidth`

## Test plan

- Open any swipeable example in the example app
- Remove the `renderRightActions` prop
- See how before this fix, `Swipeable` could've still been opened to the left, and now it can't be.

### DIFF
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -230,7 +230,7 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
     const rowWidth = useSharedValue<number>(0);
     const leftWidth = useSharedValue<number>(0);
     const rightWidth = useSharedValue<number>(0);
-    const rightOffset = useSharedValue<number>(0);
+    const rightOffset = useSharedValue<number | null>(null);
 
     const leftActionTranslate = useSharedValue<number>(0);
     const rightActionTranslate = useSharedValue<number>(0);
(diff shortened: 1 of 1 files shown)


src/components/ReanimatedSwipeable.tsx
```diff
@@ -446,7 +456,6 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       },
       openRight() {
         'worklet';
-        rightWidth.value = rowWidth.value - rightOffset.value;
```
m-bert, line 449: Shouldn't we change it to `updateRightElementWidth`?

latekvo: There's no need, this line is now present in `calculateCurrentOffset()` which is executed in the next line before anything else.

src/components/ReanimatedSwipeable.tsx
```diff
@@ -266,19 +266,29 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+    };
+
     const calculateCurrentOffset = useCallback(() => {
       'worklet';
+      updateRightElementWidth();
       if (rowState.value === 1) {
         return leftWidth.value;
       } else if (rowState.value === -1) {
-        return -rowWidth.value - rightOffset.value;
+        return -rowWidth.value - rightOffset.value!;
```
m-bert, line 283: I'd think about changing it to use ternary operator:

```tsx
      return rowState.value === 1
        ? leftWidth.value
        : rowState.value === -1
        ? -rowWidth.value - rightOffset.value!
        : 0;
```

But it has 3 options, so it may look a bit messy. It's up to you to decide 😅

m-bert (approved):

src/components/ReanimatedSwipeable.tsx
latekvo: It looks ok, done in [5e8ded8](https://github.com/software-mansion/react-native-gesture-handler/pull/3145/commits/5e8ded891497cf000ebb4a8f36ba3cfb05f7d630)

PR #3146, opened 2024-10-09
Create `ReanimatedDrawerLayout` component

latekvo:
## Description

This PR adds `ReanimatedDrawerLayout` component.

## Test plan

- use the newly added Reanimated Drawer Layout example to see how the drawer layout functions
- use the provided sample code to test how the legacy one used to work

<details>
<summary>
Collapsed code - legacy component preview
</summary>

```js

import {
  Gesture,
  GestureDetector,
  DrawerLayout,
import { SharedValue } from 'react-native-reanimated';

const DrawerPage = ({ progress }: { progress?: SharedValue }) => {
  progress && console.log('Drawer opening progress:', progress);
  return <View style={styles.drawerContainer} />;
};

export default function ReanimatedDrawerExample() {
  const drawerRef = useRef<any>(null);

  const tapGesture = Gesture.Tap()
    .runOnJS(true)
    .onStart(() => drawerRef.current?.openDrawer());

  return (
    <DrawerLayout ref={drawerRef} renderNavigationView={() => <DrawerPage />}>
        <View style={styles.innerContainer}>
          <Text>Open drawer</Text>
        </View>
      </GestureDetector>
    </DrawerLayout>
  );
}

  drawerContainer: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: 'pink',
  },
  innerContainer: {
    margin: 'auto',
    padding: 35,
    paddingHorizontal: 25,
    backgroundColor: 'pink',
  },
});
```
</details>

### DIFF
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -713,10 +713,9 @@ const DrawerLayout = forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
               styles.drawerContainer,
               drawerAnimatedStyle,
               drawerContainerStyle,
+              dynamicDrawerStyles,
             ]}>
-            <Animated.View style={dynamicDrawerStyles}>
-              {props.renderNavigationView(openValue)}
-            </Animated.View>
+            {props.renderNavigationView(openValue)}
           </Animated.View>
         </Animated.View>
       </GestureDetector>
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -598,11 +598,22 @@ const DrawerLayout = forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
       const isBack = drawerType === DrawerType.BACK;
       const isIdle = drawerState === DrawerState.IDLE;
 
+      if (isBack) {
+        return {
+          transform: [
+            {
+              translateX: 0,
+            },
+          ],
+          flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
+        };
+      }
+
       let translateX = 0;
 
-      if (!isBack && isIdle) {
+      if (isIdle) {
         translateX = drawerOpened ? 0 : closedDrawerOffset;
-      } else if (!isBack && !isIdle) {
+      } else {
         translateX = interpolate(
           openValue.value,
           [0, 1],
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -215,7 +215,6 @@ const defaultProps = {
 const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
   function DrawerLayout(props: DrawerLayoutProps, ref) {
     const dragX = useSharedValue<number>(0);
-    const touchX = useSharedValue<number>(0);
     const drawerTranslation = useSharedValue<number>(0);
 
     const [containerWidth, setContainerWidth] = React.useState(0);
@@ -236,15 +235,6 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
     } = props;
 
     const fromLeft = drawerPosition === positions.Left;
-
-    // %% this likely shouldn't be reset every render
-    if (drawerPosition !== positions.Left) {
-      // To handle right-side drawers, reverse events coming from gesture handler
-      // in a way they emulate left-side drawer gestures
-      touchX.value = containerWidth;
-    } else {
-      touchX.value = 0;
-    }
 
     const sideCorrectedDragX = useDerivedValue(() =>
       drawerPosition !== positions.Left ? -1 * dragX.value : dragX.value
@@ -352,8 +342,6 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
       (toValue: number, velocity: number, speed?: number) => {
         'worklet';
         dragX.value = 0;
-        touchX.value =
-          props.drawerPosition === positions.Left ? 0 : containerWidth;
 
         const willShow = toValue !== 0;
         isDrawerOpen.value = willShow;
@@ -404,7 +392,6 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
         );
       },
       [
-        containerWidth,
         dragX,
         drawerState,
         drawerTranslation,
@@ -412,12 +399,10 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
         emitStateChanged,
         fromLeft,
         isDrawerOpen,
-        props.drawerPosition,
         props.hideStatusBar,
         props.onDrawerClose,
         props.onDrawerOpen,
         props.statusBarAnimation,
-        touchX,
         updateShowing,
       ]
     );
@@ -498,11 +483,13 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
 
     const overlayDismissGesture = React.useMemo(
       () =>
-        Gesture.Tap().onEnd(() => {
-          if (isDrawerOpen.value && props.drawerLockMode !== 'locked-open') {
-            closeDrawer();
-          }
-        }),
+        Gesture.Tap()
+          .maxDistance(25)
+          .onEnd(() => {
+            if (isDrawerOpen.value && props.drawerLockMode !== 'locked-open') {
+              closeDrawer();
+            }
+          }),
       [closeDrawer, isDrawerOpen.value, props.drawerLockMode]
     );
 
@@ -570,7 +557,6 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
         .onUpdate((event) => {
           // %% removed touchX from the equation, add it back if we want no panning to occur until the finger touches the drawer edge
           dragX.value = event.translationX;
-          touchX.value = event.x;
           drawerTranslation.value =
             drawerType === 'front'
               ? sideCorrectedDragX.value +
@@ -599,7 +585,6 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
       props.mouseButton,
       props.statusBarAnimation,
       sideCorrectedDragX.value,
-      touchX,
     ]);
 
     // When using RTL, row and row-reverse flex directions are flipped.
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -542,9 +542,18 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
           }
         })
         .onUpdate((event) => {
-          dragX.value =
+          dragX.value = Math.max(
+            drawerOpened
+              ? interpolate(
+                  event.x,
+                  [0, drawerWidth, drawerWidth + 1],
+                  [0, drawerWidth, drawerWidth]
+                )
+              : 0,
             event.translationX +
-            (drawerOpened ? drawerWidth * -gestureOrientation : 0);
+              (drawerOpened ? drawerWidth * -gestureOrientation : 0)
+          );
+
           drawerTranslation.value =
             drawerType === 'front'
               ? sideCorrectedDragX.value +
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -209,7 +209,7 @@ export interface DrawerLayoutProps {
 
   /**
    * @default 'auto'
-   * Sets the displayed cursor pictogram when hovering over the drawer.
+   * Sets the displayed cursor pictogram when the drawer is being dragged.
    * Values: see CSS cursor values
    */
   activeCursor?: ActiveCursor;
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -214,6 +214,8 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
     const [drawerState, setDrawerState] = React.useState<DrawerState>(IDLE);
     const [drawerOpened, setDrawerOpened] = React.useState(false);
 
+    console.log(drawerState);
+
     const {
       drawerPosition = defaultProps.drawerPosition,
       drawerWidth = defaultProps.drawerWidth,
@@ -362,11 +364,7 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
             if (finished) {
               emitStateChanged(IDLE, willShow);
               runOnJS(setDrawerOpened)(willShow);
-              if (drawerState !== DRAGGING) {
-                // It's possible that user started drag while the drawer
-                // was settling, don't override state in this case
-                runOnJS(setDrawerState)(IDLE);
-              }
+              runOnJS(setDrawerState)(IDLE);
               if (willShow) {
                 dragX.value = drawerWidth * sideCorrection;
                 props.onDrawerOpen && runOnJS(props.onDrawerOpen)?.();
@@ -380,7 +378,6 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
       },
       [
         dragX,
-        drawerState,
         drawerTranslation,
         edgeWidth,
         emitStateChanged,
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -34,7 +34,9 @@ import {
   ActiveCursor,
   MouseButton,
   HitSlop,
+  GestureStateChangeEvent,
 } from '../handlers/gestureHandlerCommon';
+import { PanGestureHandlerEventPayload } from '../handlers/GestureHandlerEventPayload';
 
 const DRAG_TOSS = 0.05;
 
@@ -408,17 +410,9 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
     );
 
     const handleRelease = React.useCallback(
-      ({
-        nativeEvent,
-      }: {
-        nativeEvent: {
-          translationX: number;
-          velocityX: number;
-          x: number;
-        };
-      }) => {
+      (event: GestureStateChangeEvent<PanGestureHandlerEventPayload>) => {
         'worklet';
-        let { translationX: dragX, velocityX, x: touchX } = nativeEvent;
+        let { translationX: dragX, velocityX, x: touchX } = event;
 
         if (drawerPosition !== DrawerPosition.LEFT) {
           // See description in _updateAnimatedEvent about why events are flipped
@@ -581,9 +575,7 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
                 )
               : sideCorrectedDragX.value;
         })
-        .onEnd((event) => {
-          handleRelease({ nativeEvent: event });
-        });
+        .onEnd(handleRelease);
     }, [
       dragX,
       drawerLockMode,
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -109,7 +109,6 @@ export interface DrawerLayoutProps {
     newState: DrawerState,
     drawerWillShow: boolean
   ) => void;
-  useNativeAnimations?: boolean;
 
   drawerType?: DrawerType;
 
@@ -205,7 +204,6 @@ interface DrawerLayoutMethods {
 const defaultProps = {
   drawerWidth: 200,
   drawerPosition: DrawerPosition.LEFT,
-  useNativeAnimations: true,
   drawerType: DrawerType.FRONT,
   edgeWidth: 20,
   minSwipeDistance: 3,
@@ -238,7 +236,6 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
       minSwipeDistance = defaultProps.minSwipeDistance,
       edgeWidth = defaultProps.edgeWidth,
       drawerLockMode = defaultProps.drawerLockMode,
-      useNativeAnimations = defaultProps.useNativeAnimations,
       overlayColor = defaultProps.overlayColor,
       enableTrackpadTwoFingerGesture = defaultProps.enableTrackpadTwoFingerGesture,
       activeCursor = defaultProps.activeCursor,
@@ -275,11 +272,9 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
     });
 
     const gestureOptions: {
-      useNativeDriver: boolean;
       listener?: (event: { nativeEvent: { translationX: number } }) => void;
     } = React.useMemo(
       () => ({
-        useNativeDriver: useNativeAnimations,
         listener: props.onDrawerSlide
           ? (ev) => {
               const translationX = Math.floor(
@@ -291,7 +286,7 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
             }
           : undefined,
       }),
-      [containerWidth, props, useNativeAnimations]
+      [containerWidth, props]
     );
 
     if (props.onDrawerSlide) {
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -654,7 +654,6 @@ const DrawerLayout = forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
         gesture={panGesture}
         userSelect={props.userSelect}
         enableContextMenu={props.enableContextMenu}>
-        {/* ^^^ "false if MouseButton.RIGHT is specified" either remove that description or add that logic */}
         <Animated.View
           ref={componentRef}
           style={styles.main}
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -123,12 +123,10 @@ export interface DrawerLayoutProps {
   statusBarAnimation?: StatusBarAnimation;
 
   /**
-   * @default black
+   * @default 'rgba(0, 0, 0, 0.7)'
    *
-   * Color of a semi-transparent overlay to be displayed on top of the content
-   * view when drawer gets open. A solid color should be used as the opacity is
-   * added by the Drawer itself and the opacity of the overlay is animated (from
-   * 0% to 70%).
+   * Color of the background overlay.
+   * Animated from `0%` to `100%` as the drawer opens.
    */
   overlayColor?: string;
 
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -517,7 +517,7 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
         .onStart(() => {
           emitStateChanged(DrawerState.DRAGGING, false);
           runOnJS(setDrawerState)(DrawerState.DRAGGING);
-          if (props.keyboardDismissMode === 'on-drag') {
+          if (props.keyboardDismissMode === DrawerKeyboardDismissMode.ON_DRAG) {
             runOnJS(Keyboard.dismiss)();
           }
           if (props.hideStatusBar) {
@@ -691,7 +691,7 @@ const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
           <Animated.View
             pointerEvents="box-none"
             animatedProps={drawerAnimatedProps}
-            accessibilityViewIsModal={isDrawerOpen.value}
+            accessibilityViewIsModal={isDrawerOpen.value} // %% use animated props
             style={[
               styles.drawerContainer,
               drawerAnimatedStyle,
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -717,9 +717,10 @@ const DrawerLayout = forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
               styles.drawerContainer,
               drawerAnimatedStyle,
               drawerContainerStyle,
-              dynamicDrawerStyles,
             ]}>
-            {renderNavigationView(openValue)}
+            <Animated.View style={dynamicDrawerStyles}>
+              {renderNavigationView(openValue)}
+            </Animated.View>
           </Animated.View>
         </Animated.View>
       </GestureDetector>
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -91,14 +91,29 @@ export interface DrawerLayoutProps {
     progressAnimatedValue: SharedValue<number>
   ) => ReactNode;
 
+  /**
+   * Determines the side from which the drawer will open.
+   */
   drawerPosition?: DrawerPosition;
 
+  /**
+   * Width of the drawer.
+   */
   drawerWidth?: number;
 
+  /**
+   * Background color of the drawer.
+   */
   drawerBackgroundColor?: string;
 
+  /**
+   * Set if the drawer should get locked after being opened or closed.
+   */
   drawerLockMode?: DrawerLockMode;
 
+  /**
+   * Determines if system keyboard should be closed upon dragging the drawer.
+   */
   keyboardDismissMode?: DrawerKeyboardDismissMode;
 
   /**
@@ -119,6 +134,9 @@ export interface DrawerLayoutProps {
     drawerWillShow: boolean
   ) => void;
 
+  /**
+   * Type of animation that will play when opening the drawer.
+   */
   drawerType?: DrawerType;
 
   /**
@@ -127,6 +145,9 @@ export interface DrawerLayoutProps {
    */
   edgeWidth?: number;
 
+  /**
+   * Minimal distance to swipe before the drawable starts moving.
+   */
   minSwipeDistance?: number;
 
   /**
@@ -154,8 +175,14 @@ export interface DrawerLayoutProps {
    */
   overlayColor?: string;
 
+  /**
+   * Style wrapping the content.
+   */
   contentContainerStyle?: StyleProp<ViewStyle>;
 
+  /**
+   * Style wrapping the drawer.
+   */
   drawerContainerStyle?: StyleProp<ViewStyle>;
 
   /**
@@ -168,7 +195,10 @@ export interface DrawerLayoutProps {
 
   onDrawerSlide?: (position: number) => void;
 
-  // Implicit `children` prop has been removed in @types/react^18.0.0
+  // Implicit `children` prop has been removed in @types/react^18.0.
+  /**
+   * Children to be rendered normally - not inside the drawer.
+   */
   children?: ReactNode | ((openValue?: SharedValue<number>) => ReactNode);
 
   /**
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -389,8 +389,8 @@ const DrawerLayout = forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
             overshootClamping: true,
             velocity: normalizedInitialVelocity,
             mass: animationSpeed ? 1 / animationSpeed : 1,
-            damping: 50,
-            stiffness: 300,
+            damping: 40,
+            stiffness: 500,
           },
           (finished) => {
             if (finished) {
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -520,9 +520,10 @@ const DrawerLayout = forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
         .simultaneousWithExternalGesture(overlayDismissGesture)
         .enableTrackpadTwoFingerGesture(enableTrackpadTwoFingerGesture)
         .enabled(
-          drawerState !== DrawerState.SETTLING && drawerOpened
-            ? drawerLockMode !== DrawerLockMode.LOCKED_OPEN
-            : drawerLockMode !== DrawerLockMode.LOCKED_CLOSED
+          drawerState !== DrawerState.SETTLING &&
+            (drawerOpened
+              ? drawerLockMode !== DrawerLockMode.LOCKED_OPEN
+              : drawerLockMode !== DrawerLockMode.LOCKED_CLOSED)
         )
         .onStart(() => {
           emitStateChanged(DrawerState.DRAGGING, false);
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -625,11 +625,7 @@ const DrawerLayout = forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
 
       if (isBack) {
         return {
-          transform: [
-            {
-              translateX: 0,
-            },
-          ],
+          transform: [{ translateX: 0 }],
           flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
         };
       }
@@ -648,11 +644,7 @@ const DrawerLayout = forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
       }
 
       return {
-        transform: [
-          {
-            translateX,
-          },
-        ],
+        transform: [{ translateX }],
         flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
       };
     });
--- a/src/components/ReanimatedDrawerLayout.tsx
+++ b/src/components/ReanimatedDrawerLayout.tsx
@@ -199,8 +199,8 @@ export interface DrawerLayoutProps {
 }
 
 export type DrawerMovementOption = {
-  velocity?: number;
-  speed?: number;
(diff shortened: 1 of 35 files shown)


latekvo:
As of [1ebcfa4](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/1ebcfa4537b58bfb2bc81309b62b2159c6e98d60), I'm not aware of any bugs that haven't been fixed already. In the following days I will test some minor parts of the component, like type compatibility with the legacy version, if it works when exported into a real app and not just an example package, etc.
Regardless, as of now it is ready to review.

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+
+    const handleContainerLayout = ({ nativeEvent }: LayoutChangeEvent) => {
+      setContainerWidth(nativeEvent.layout.width);
+    };
+
+    const emitStateChanged = React.useCallback(
+      (newState: DrawerState, drawerWillShow: boolean) => {
+        'worklet';
+        props.onDrawerStateChanged &&
+          runOnJS(props.onDrawerStateChanged)?.(newState, drawerWillShow);
```
j-piasecki, line 297: Can we check whether `onDrawerStateChanged` is a worklet and call it on the UI thread if so?

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+      pointerEvents: isDrawerShowing.value
+        ? ('auto' as const)
+        : ('none' as const),
+    }));
+
+    // While the drawer is hidden, it's hitSlop overflows onto the main view by edgeWidth
+    // This way it can be swiped open even when it's hidden
+    const [edgeHitSlop, setEdgeHitSlop] = React.useState<HitSlop>(
+      isFromLeft ? { right: edgeWidth } : { left: edgeWidth }
+    );
```
j-piasecki, line 318: Can't this be handled by an animated style dependant on `isDrawerOpen`?

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+
+    // gestureOrientation is 1 if the gesture is expected to move from left to right and -1 otherwise
+    const gestureOrientation = React.useMemo(
+      () => (isFromLeft ? 1 : -1) * (drawerOpened ? -1 : 1),
+      [isFromLeft, drawerOpened]
+    );
+
+    const animateDrawer = React.useCallback(
+      (toValue: number, velocity: number, speed?: number) => {
```
j-piasecki, line 327: What's the difference between `speed` and `velocity`?

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+
+    const handleRelease = React.useCallback(
+      ({
+        nativeEvent,
+      }: {
+        nativeEvent: {
+          translationX: number;
+          velocityX: number;
+          x: number;
+        };
```
j-piasecki, line 401: Why pack data into `nativeEvent` when it can be passed directly?

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+      opacity: openValue.value,
+      backgroundColor: props.overlayColor ?? defaultProps.overlayColor,
+    }));
+
+    const renderOverlay = React.useCallback(() => {
+      return (
+        <GestureDetector gesture={overlayDismissGesture}>
+          <Animated.View
+            pointerEvents={isDrawerOpen.value ? 'auto' : 'none'}
+            animatedProps={overlayAnimatedProps}
```
j-piasecki, line 490: This will always overwrite `pointerEvents` above so one of those isn't needed (since `isDrawerOpen` is a shared value it should go through `animatedProps`)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+
+  /**
+   * @default black
+   *
+   * Color of a semi-transparent overlay to be displayed on top of the content
+   * view when drawer gets open. A solid color should be used as the opacity is
+   * added by the Drawer itself and the opacity of the overlay is animated (from
+   * 0% to 70%).
+   */
+  overlayColor?: string;
```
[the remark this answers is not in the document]
latekvo: The description is wrong, it works the way you described it.
Fixed description in [3486ea8](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/3486ea8bdb0b7727140560db1393da493684173c).

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+      drawerPosition = defaultProps.drawerPosition,
+      drawerWidth = defaultProps.drawerWidth,
+      drawerType = defaultProps.drawerType,
+      drawerBackgroundColor,
+      drawerContainerStyle,
+      contentContainerStyle,
+      minSwipeDistance = defaultProps.minSwipeDistance,
+      edgeWidth = defaultProps.edgeWidth,
+      drawerLockMode = defaultProps.drawerLockMode,
+      useNativeAnimations = defaultProps.useNativeAnimations,
```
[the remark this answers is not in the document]
latekvo: Fixed this, and added more default values to the `defaultProps` object in [2ce4ca2](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/2ce4ca2676ae3b2eb74d6540f0e149ab2e91602a)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+
+export type DrawerPosition = 'left' | 'right';
+
+export type DrawerState = 'Idle' | 'Dragging' | 'Settling';
+
+export type DrawerType = 'front' | 'back' | 'slide';
+
+export type DrawerLockMode = 'unlocked' | 'locked-closed' | 'locked-open';
+
+export type DrawerKeyboardDismissMode = 'none' | 'on-drag';
```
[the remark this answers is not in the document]
latekvo: You're right, fixed in [c289958](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/c28995838b7affcfc8dc1598903351bce710a9c8)

src/components/ReanimatedDrawerLayout.tsx
latekvo: Fixed in [5830645](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/58306453e30ac6e5d03bdfdc0b3be9d6526ef131)

src/components/ReanimatedDrawerLayout.tsx
latekvo: `velocity` is the initial animation velocity, `speed` is the animation speed.
This is an artifact from the legacy component, i can rename it internally to `animationSpeed`, 
but since it's already exposed in the `DrawerLayoutProps` interface, i'm not sure if we should change it there.

src/components/ReanimatedDrawerLayout.tsx
latekvo: fixed in [eaba51f](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/eaba51ff0e7590e653a0026e50397675d81af4e1)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+
+export interface DrawerLayoutProps {
+  /**
+   * This attribute is present in the standard implementation already and is one
+   * of the required params. Gesture handler version of DrawerLayout make it
```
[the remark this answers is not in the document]
latekvo: Fixed in [a00785c](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/a00785cff68d27879ec60891cedecef70d4ecd51)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+  onDrawerOpen?: () => void;
+
+  /**
+   * Called when the status of the drawer changes.
+   */
+  onDrawerStateChanged?: (
+    newState: DrawerState,
+    drawerWillShow: boolean
+  ) => void;
+  useNativeAnimations?: boolean;
```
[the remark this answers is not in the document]
latekvo: removed in [c0a79ae](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/c0a79ae8f40cdddcf26b803025e545add9e4e896)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+                : interpolate(
+                    openValue.value,
+                    [0, 1],
+                    [closedDrawerOffset, 0],
+                    Extrapolation.CLAMP
+                  ),
+          },
+        ],
+        flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
+      };
```
m-bert, line 646: Same here, is there a way to rewrite it? For example:


```suggestion
      let translation;

      if (drawerType == DrawerType.BACK) {
        translation = 0;
      } else if (drawerState === DrawerState.IDLE) {
        translation = drawerOpened ? 0 : closedDrawerOffset;
      } else {
        translation = interpolate(
          openValue.value,
          [0, 1],
          [closedDrawerOffset, 0],
          Extrapolation.CLAMP
        );
      }

      return {
        transform: [{ translateX: translation }],
        flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
      };
```

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      [componentRef.current, openDrawer, closeDrawer]
+    );
+
+    return (
+      <GestureDetector
+        gesture={panGesture}
+        userSelect={props.userSelect}
+        enableContextMenu={props.enableContextMenu}>
+        {/* ^^^ "false if MouseButton.RIGHT is specified" either remove that description or add that logic */}
```
m-bert, line 682: I think there's something to be done here :)

m-bert (changes requested):

m-bert (changes requested):
Overall looks fine :)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+
+    // While the drawer is hidden, it's hitSlop overflows onto the main view by edgeWidth
+    // This way it can be swiped open even when it's hidden
+    const [edgeHitSlop, setEdgeHitSlop] = React.useState<HitSlop>(
+      isFromLeft ? { right: edgeWidth } : { left: edgeWidth }
+    );
+
+    // gestureOrientation is 1 if the gesture is expected to move from left to right and -1 otherwise
+    const gestureOrientation = React.useMemo(
+      () => (isFromLeft ? 1 : -1) * (drawerOpened ? -1 : 1),
```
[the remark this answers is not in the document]
latekvo: Fixed in [4a806dd](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/4a806dd21a7132484acbafe1d14b7d0b0c868b61)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+  drawerBackgroundColor?: string;
+
+  drawerLockMode?: DrawerLockMode;
+
+  keyboardDismissMode?: DrawerKeyboardDismissMode;
+
+  /**
+   * Called when the drawer is closed.
+   */
+  onDrawerClose?: () => void;
```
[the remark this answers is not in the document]
latekvo: The ones with missing docs must have been undocumented in the original `DrawerLayout`, 
I'll add all the missing docstrings.

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+    );
+
+    if (props.onDrawerSlide) {
+      gestureOptions.listener = (ev) => {
+        const translationX = Math.floor(Math.abs(ev.nativeEvent.translationX));
+        const position = translationX / containerWidth;
+
+        props.onDrawerSlide?.(position);
+      };
+    }
```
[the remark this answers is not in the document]
latekvo: It is, not sure how that happened, legacy implementation of this callback seems to be completely broken as well :/
This used to be a part of legacy `Animated.event()` code, removed and reimplemented in [daa47f3](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/daa47f372916e89a76544e73227116c675b8c4ae). 
The `onDrawerSlide` callback is being now called upon every `openValue` update.

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,745 @@
...
+              const translationX = Math.floor(
+                Math.abs(ev.nativeEvent.translationX)
+              );
+              const position = translationX / containerWidth;
+
+              props.onDrawerSlide?.(position);
+            }
+          : undefined,
+      }),
+      [containerWidth, props, useNativeAnimations]
```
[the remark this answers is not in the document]
latekvo: removed all together in [daa47f3](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/daa47f372916e89a76544e73227116c675b8c4ae)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+                Math.abs(ev.nativeEvent.translationX)
+              );
+              const position = translationX / containerWidth;
+
+              props.onDrawerSlide?.(position);
+            }
+          : undefined,
+      }),
+      [containerWidth, props]
+    );
```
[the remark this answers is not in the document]
latekvo: reimplemented the `onDrawerSlide` callback logic in [daa47f3](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/daa47f372916e89a76544e73227116c675b8c4ae), more details [here](https://github.com/software-mansion/react-native-gesture-handler/pull/3146#discussion_r1814570851)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+const DrawerLayout = React.forwardRef<DrawerLayoutMethods, DrawerLayoutProps>(
+  function DrawerLayout(props: DrawerLayoutProps, ref) {
+    const dragX = useSharedValue<number>(0);
+    const drawerTranslation = useSharedValue<number>(0);
+
+    const [containerWidth, setContainerWidth] = React.useState(0);
+    const [drawerState, setDrawerState] = React.useState<DrawerState>(
+      DrawerState.IDLE
+    );
+    const [drawerOpened, setDrawerOpened] = React.useState(false);
```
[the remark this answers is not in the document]
latekvo: Fixed in [32f9503](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/32f9503634ee46bb12f05b2627bfac71a93e0013)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+        .enableTrackpadTwoFingerGesture(enableTrackpadTwoFingerGesture)
+        .enabled(
+          drawerLockMode !== DrawerLockMode.LOCKED_CLOSED &&
+            drawerLockMode !== DrawerLockMode.LOCKED_OPEN &&
+            drawerState !== DrawerState.SETTLING
+        )
+        .onStart(() => {
+          emitStateChanged(DrawerState.DRAGGING, false);
+          runOnJS(setDrawerState)(DrawerState.DRAGGING);
+          if (props.keyboardDismissMode === 'on-drag') {
```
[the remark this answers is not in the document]
m-bert: Apart from the fact that this commit introduced new 'todo' it's ok

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+        'worklet';
+        const willShow = toValue !== 0;
+        isDrawerOpen.value = willShow;
+        runOnJS(setEdgeHitSlop)(
+          isFromLeft
+            ? { left: 0, width: isDrawerOpen.value ? undefined : edgeWidth }
+            : { right: 0, width: isDrawerOpen.value ? undefined : edgeWidth }
+        );
+
+        isDrawerShowing.value = willShow;
```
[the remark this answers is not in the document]
latekvo: Both `SharedValues` were completely identical, removed `isDrawerShowing` in [edb2e41](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/edb2e41b06bc90f91ae212db9a20a39aba632e77)

src/components/ReanimatedDrawerLayout.tsx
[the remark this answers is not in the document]
latekvo: It did because I noticed a new issue, while already working on something else, otherwise i'd lose track if the bugs i have to fix.

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+    );
+
+    const panGesture = React.useMemo(() => {
+      return Gesture.Pan()
+        .activeCursor(activeCursor)
+        .mouseButton(mouseButton)
+        .hitSlop(drawerOpened ? fillHitSlop : edgeHitSlop)
+        .minDistance(drawerOpened ? 100 : 0)
+        .activeOffsetX(gestureOrientation * minSwipeDistance)
+        .failOffsetY([-15, 15])
```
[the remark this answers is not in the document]
latekvo: I think the only item that isn't already configurable here is `failOffsetY`, I could add a prop for configuring that, but i don't think there's any use case for doing so. 
`failOffsetY` is completely ignored when opening the drawer.
In all other cases, `failOffsetY` only really triggers when the swipe is almost completely vertical, and I'm actually not sure why that would be useful, since drawer closing is already cancelable by lifting the finger. 
Perhaps it could prevent some bugs with scroll list, but I'd have to test it further to check for that.
Maybe we should just remove it?

Also, aren't `minDistance` and `activeOffsetX` redundant? They both set the minimal distance a touch has to travel before activating the `Pan` gesture.

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+export enum DrawerLockMode {
+  UNLOCKED = 'unlocked',
+  LOCKED_CLOSED = 'locked-closed',
+  LOCKED_OPEN = 'locked-open',
+}
+
+export enum DrawerKeyboardDismissMode {
+  NONE = 'none',
+  ON_DRAG = 'on-drag',
+}
```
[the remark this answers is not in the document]
latekvo: You're right, the props cannot be set using string literals anymore, so these are unnecessary.
Removed strings from the enums in [89ed099](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/89ed099235d518bea24c4f15eef74612c157dd7c)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,742 @@
...
+                  : interpolate(
+                      event.x - containerWidth,
+                      [-drawerWidth - 1, -drawerWidth, 0],
+                      [drawerWidth, drawerWidth, 0]
+                    )
+                : 0,
+              sideCorrection *
+                (event.translationX +
+                  (drawerOpened ? drawerWidth * -gestureOrientation : 0))
+            );
```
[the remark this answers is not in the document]
latekvo: I absolutely agree, fixed in [3060229](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/30602295f3caab026324cc1b4ce51a8dc1d887c2)

src/components/ReanimatedDrawerLayout.tsx
latekvo: Rewritten in [be4979c](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/be4979cf2b01874ddd925e2c47f812ed4482be29)

src/components/ReanimatedDrawerLayout.tsx
latekvo: Not sure what you mean by that, it's only used for configuring the `Pan` gesture, so it's preferable for it to be `State`, and it's only set in one place besides the initialiser, so I'm not sure if it's worth creating an animated style specifically for it if we can just use it's setter.

Could you please elaborate?

src/components/ReanimatedDrawerLayout.tsx
latekvo: This prop automatically switches to `false` when mouseButton is set to `MouseButton.RIGHT`, more details in [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/2676).

Tested on the `ReanimatedDrawerLayout` to confirm it works that way, couldn't test on the legacy one because it crashes on `web`.

Removed this comment in [7918726](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/7918726fa1a90f971704f6daf907b3cc28c19e11)

src/components/ReanimatedDrawerLayout.tsx
j-piasecki: I'll be honest, I don't remember what I meant 😅. But looking at it now, I don't see where the animated style would be used so I think we can safely ignore that one.

src/components/ReanimatedDrawerLayout.tsx
j-piasecki: I'd go further and also change `velocity` to `initialVelocity`. The way it's been in the previous implementation is really confusing.

src/components/ReanimatedDrawerLayout.tsx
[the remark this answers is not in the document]
j-piasecki: Looks good, one nit though - shouldn't `onDrawerSlide` be called as part of the gesture callback?

src/components/ReanimatedDrawerLayout.tsx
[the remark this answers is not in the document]
j-piasecki: I don't think it's necessary and we also shouldn't remove it. It's there so that vertical swipes don't block other gestures in any way (otherwise the pan could activate when swiping up-down).

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,716 @@
...
+      }
+
+      return {
+        transform: [
+          {
+            translateX,
+          },
+        ],
+        flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
+      };
```
m-bert, line 621: Actually in that case we could simply check if `isBack` is `true` and return right away.

```suggestion
      if (isBack) {
        return {
          transform: [{ translateX: 0 }],
          flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
        };
      }

      let translateX = 0;

      if (isIdle) {
        translateX = drawerOpened ? 0 : closedDrawerOffset;
      } else {
        translateX = interpolate(
          openValue.value,
          [0, 1],
          [closedDrawerOffset, 0],
          Extrapolation.CLAMP
        );
      }

      return {
        transform: [{ translateX }],
        flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
      };
```

WDYT?

m-bert:
Also, on web tapping does not close drawer

src/components/ReanimatedDrawerLayout.tsx
latekvo: Changed in [82e3c00](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/82e3c00bc2b66eb576ed23757ccb91f650b102cf)

The exposed `type` still exposes `speed` and `velocity`, let me know if we want to deprecate the exported `velocity` and `speed` props as well, and alias them with `initialVelocity` and `animationSpeed`.

src/components/ReanimatedDrawerLayout.tsx
latekvo: I've tried implementing this with a simple `runOnJS` drop-in replacement function:

```tsx
interface WorkletOrRegularFunction {
  // `unknown` and `never` won't work in this case
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  (...args: any[]): unknown;
  __workletHash?: boolean;
}

const adaptCallback = (callback: WorkletOrRegularFunction) => {
  'worklet';
  return callback.__workletHash ? callback : runOnJS(callback);
};
```
```js
props.onDrawerSlide && adaptCallback(props.onDrawerSlide)(newOpenValue);
```

Where `adaptCallback` would replace `runOnJS` and then be run like `runOnJS`.

Unfortunately, i could not get such mechanism to not crash, executing a returned `runOnJS`-wrapped function seems to crash every single time with the following error:

`ERROR  ReanimatedError: Exception in HostFunction: [Reanimated] Attempted to extract from a HostObject that wasn't converted to a Shareable., js engine: reanimated`

I believe the crashes are related to one of these issues, but none of them exactly replicate their repro:
- https://github.com/software-mansion/react-native-reanimated/issues/5442
- https://github.com/software-mansion/react-native-reanimated/issues/5443

The shortest working alternative i could find looks like this:

```tsx
interface WorkletOrRegularFunction {
  // `unknown` and `never` won't work in this case
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  (...args: any[]): unknown;
  __workletHash?: boolean;
}

const isWorklet = (cb?: WorkletOrRegularFunction) => {
  return !!cb?.__workletHash;
};
```
```js
// eslint-disable-next-line babel/no-unused-expressions
isWorklet(props.onDrawerOpen)
  ? props.onDrawerOpen?.()
  : props.onDrawerOpen && runOnJS(props.onDrawerOpen)();
```

Not sure if we want to use it, it's bulky but at least it's somewhat readable.

src/components/ReanimatedDrawerLayout.tsx
latekvo: Sure, looks great, changed in [ea31eb3](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/ea31eb34c2ba31113b0ab71ec848537b1204d3d6).

src/components/ReanimatedDrawerLayout.tsx
j-piasecki: Why not just remove them and use the new ones? Since this is a new component it wouldn't be a breaking change.

latekvo:
>Also, on web tapping does not close drawer

Fixed in [d2a52dd](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/d2a52dd1b32252e2e175eb3adace301b946b8c12) and [920a869](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/920a869c63f986f491c8696411330d027f09abec)

src/components/ReanimatedDrawerLayout.tsx
latekvo: Alright, just to clarify the component is not entirely new - it's a replacement for existing `DrawerLayout`.
This particular type has been pulled directly from the legacy version. 
Regardless, I'm all for it, the change is minimal, and if anything breaks for anyone, the fix is trivial.

Changed in [5b0237e](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/5b0237e4f843df8d88a2b7d7fb8eaf01f91acd1f)

src/components/ReanimatedDrawerLayout.tsx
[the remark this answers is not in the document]
latekvo: Added all missing docstrings in: [19cabf1](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/19cabf1f93c3cc67e7814c397578a4e379c4b8cf)

I'll open a separate PR for our docs page.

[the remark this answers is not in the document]
latekvo: Opened the separate PR here: [link](https://github.com/software-mansion/react-native-gesture-handler/pull/3188)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -168,7 +195,10 @@ export interface DrawerLayoutProps {
 
   onDrawerSlide?: (position: number) => void;
 
-  // Implicit `children` prop has been removed in @types/react^18.0.0
+  // Implicit `children` prop has been removed in @types/react^18.0.
+  /**
+   * Children to be rendered normally - not inside the drawer.
```
m-bert, line 200: There's something wrong with this sentence. What about:

_**Children that will be rendered directly in the component hierarchy, not within the drawer.**_

src/components/ReanimatedDrawerLayout.tsx
m-bert: >This prop automatically switches to false when mouseButton is set to MouseButton.RIGHT, more details in https://github.com/software-mansion/react-native-gesture-handler/pull/2676.

That's not exactly true. It behaves as it is `false`, but it's value is not set to `false`. You can see it [here](https://github.com/software-mansion/react-native-gesture-handler/blob/da9eed867ff09633c506fc9ad08634eaf707cbdb/src/web/tools/GestureHandlerWebDelegate.ts#L104).

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,756 @@
...
+
+      return {
+        transform: [
+          {
+            translateX,
+          },
+        ],
+        flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
+      };
+    });
```
m-bert, line 658: I know it's just formatting, but can we stick to this:

```tsx
{
  transform: [{ translateX: 0 }],
  flexDirection: reverseContentDirection ? 'row-reverse' : 'row',
};
```

so it is 4 lines long instead of 9? 😅

src/components/ReanimatedDrawerLayout.tsx
j-piasecki: Wouldn't something like `runCallback` instead of `adaptCallback` work? I.e. instead of doing something like this:
```ts
adaptCallback(props.onDrawerSlide)(newOpenValue);
```
do it like this:
```ts
runCallback(props.onDrawerSlide, newOpenValue);
```

Instead of returning the new function, simply call it inside.

src/components/ReanimatedDrawerLayout.tsx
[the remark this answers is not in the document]
j-piasecki: Ah, good point. Yes, I think it should still be called when the drawer is being moved by the animation.

Would adding a flag saying whether a gesture is active to the callback be much of an issue?

src/components/ReanimatedDrawerLayout.tsx
latekvo: Changed in [3e36cbe](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/3e36cbe15fb8f37ec262ef7268026e43cad70eaf).
I had no idea allowed making any style changes without adding a `// prettier-ignore` comment 😅

src/components/ReanimatedDrawerLayout.tsx
latekvo: How about:
>Elements that will be rendered inside the content view.

The term `content view` is most often used for the inner container throughout the documentation, 
and I removed `children` since that would make the definition recursive.

src/components/ReanimatedDrawerLayout.tsx
latekvo: Unfortunately the second approach doesn't work either, `runCallback` would have to be a worklet running on UI, and `__workletHash` doesn't exist on UI. The `isWorkletFunction` function provided by `reanimated` doesn't work on UI either.

As a result, we'd have to check if a function is a worklet on JS, then read that data on UI, I tried doing that with the following method:

```ts
// this runs on JS, returns either a worklet or a remote function
const adaptCallback = (callback: WorkletOrRegularFunction) => {
  return callback.__workletHash ? callback : runOnJS(callback);
};

const onDrawerOpen = useMemo(
  () => adaptCallback(props.onDrawerOpen ?? noop),
  [props.onDrawerOpen]
);

onDrawerOpen();
```

Unfortunately, it looks like `runOnJS` stops working after being returned by another function, and behaves like a regular function which crashes on UI with the following error: 

```py
ReanimatedError: [Reanimated] Tried to synchronously call a non-worklet anonymous function on the UI thread.
```

Now I'm not really ok with creating a large boilerplate wrapper for each callback we want to use, since ideally this is something we want to apply to all the callbacks.
Here's how such boilerplate looks like:
<details>
<summary>
collapsed code
</summary>

```tsx
    const isDrawerStateWorklet = useMemo(
      () => isWorkletFunction(props.onDrawerStateChanged),
      [props.onDrawerStateChanged]
    );
    const onDrawerStateChanged = useCallback(
      (newState: DrawerState, drawerWillShow: boolean) => {
        'worklet';
        if (isDrawerStateWorklet) {
          props.onDrawerStateChanged?.(newState, drawerWillShow);
        } else {
          props.onDrawerStateChanged && runOnJS(props.onDrawerStateChanged);
        }
        // %% lint demands props, says props.onDrawerStateChanged is useless, likely bogus
      },
      [isDrawerStateWorklet, props]
    );

    const isDrawerSlideWorklet = useMemo(
      () => isWorkletFunction(props.onDrawerSlide),
      [props.onDrawerSlide]
    );
    const onDrawerSlide = useCallback(
      (position: number) => {
        'worklet';
        if (isDrawerSlideWorklet) {
          props.onDrawerSlide?.(position);
        } else {
          props.onDrawerSlide && runOnJS(props.onDrawerSlide);
        }
      },
      [isDrawerSlideWorklet, props]
    );

    const isDrawerOpenWorklet = useMemo(
      () => isWorkletFunction(props.onDrawerOpen),
      [props.onDrawerOpen]
    );
    const onDrawerOpen = useCallback(() => {
      'worklet';
      if (isDrawerOpenWorklet) {
        props.onDrawerOpen?.();
      } else {
        props.onDrawerOpen && runOnJS(props.onDrawerOpen);
      }
    }, [isDrawerOpenWorklet, props]);

    const isDrawerCloseWorklet = useMemo(
      () => isWorkletFunction(props.onDrawerClose),
      [props.onDrawerClose]
    );
    const onDrawerClose = useCallback(() => {
      'worklet';
      if (isDrawerCloseWorklet) {
        props.onDrawerClose?.();
      } else {
        props.onDrawerClose && runOnJS(props.onDrawerClose);
      }
    }, [isDrawerCloseWorklet, props]);
```

</details>

Another method (this one crashes):

<details>
<summary>
collapsed code
</summary>

```ts
const onDrawerOpen = useMemo(
  () =>
    isWorkletFunction(props.onDrawerOpen)
      ? props.onDrawerOpen
      : runOnJS(props.onDrawerOpen ?? noop),
  [props.onDrawerOpen]
);

const onDrawerClose = useMemo(
  () =>
    isWorkletFunction(props.onDrawerClose)
      ? props.onDrawerClose
      : runOnJS(props.onDrawerClose ?? noop),
  [props.onDrawerClose]
);
```
</details>
 
I'd prefer to introduce this feature in a separate PR, where it'll apply to both the `ReanimatedDrawerLayout` and `ReanimatedSwipeable`.

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,750 @@
...
+          isFromLeft
+            ? { left: 0, width: willShow ? undefined : edgeWidth }
+            : { right: 0, width: willShow ? undefined : edgeWidth }
+        );
+
+        emitStateChanged(DrawerState.SETTLING, willShow);
+        runOnJS(setDrawerState)(DrawerState.SETTLING);
+
+        if (props.hideStatusBar) {
+          StatusBar.setHidden(willShow, statusBarAnimation);
```
j-piasecki, line 367: Does this work on the UI thread? I don't think `setHidden` is a worklet.

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,750 @@
...
+            stiffness: 300,
+          },
+          (finished) => {
+            if (finished) {
+              emitStateChanged(DrawerState.IDLE, willShow);
+              runOnJS(setDrawerOpened)(willShow);
+              runOnJS(setDrawerState)(DrawerState.IDLE);
+              if (willShow) {
+                dragX.value = drawerWidth * sideCorrection;
+                props.onDrawerOpen && runOnJS(props.onDrawerOpen)?.();
```
j-piasecki, line 391: I'm unsure if we should access the entire `props` object on the UI thread instead of just the things that are needed.

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,750 @@
...
+      typeof props.children === 'function'
+        ? props.children(openValue) // renderer function
+        : props.children;
+
+    const componentRef = useRef(null);
+
+    useImperativeHandle(
+      ref,
+      () => ({
+        ...componentRef,
```
j-piasecki, line 674: The goal here it to add a `current` field to the ref?

src/components/ReanimatedDrawerLayout.tsx
latekvo: Fixed in [900fa4f](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/900fa4f147164805097eb0da6ad83ce2b8c30ca7)

src/components/ReanimatedDrawerLayout.tsx
latekvo: Fixed in [f14eff1](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/f14eff181dc67ff884a25922f6e4b10cd2d7da05)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,751 @@
...
+  initialVelocity?: number;
+  animationSpeed?: number;
+};
+
+export interface DrawerLayoutMethods {
+  openDrawer: (options?: DrawerMovementOption) => void;
+  closeDrawer: (options?: DrawerMovementOption) => void;
+}
+
+const defaultProps = {
```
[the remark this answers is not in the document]
latekvo: Typing `defaultProps` as `Partial<DrawerLayoutProps>` stops it from being automatically typed as an object of  literal props, meaning it can no longer be used as a setter of default values, since nearly all it's props are `| undefined` now.
Not sure if there's a way around it.

image
image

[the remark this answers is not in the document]
latekvo: After some further investigation it looks like both our prettier and babel versions are too dated to use this feature. Changing them would require a separate PR。Is it ok if we ignore this issue for now, and I'll fix it in a separate PR?

src/components/ReanimatedDrawerLayout.tsx
latekvo: This was a mistake, it didn't reveal any additional props to the user.
Removed in [0972256](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/09722561121239b19f28186f4c92fd200f3b7975)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,751 @@
...
+  drawerType?: DrawerType;
+
+  /**
+   * Defines how far from the edge of the content view the gesture should
+   * activate.
+   */
+  edgeWidth?: number;
+
+  /**
+   * Minimal distance to swipe before the drawable starts moving.
```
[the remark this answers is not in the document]
latekvo: Fixed in [3ccfbfe](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/3ccfbfe9e3a446c1f6c4092eca57fd241a01c57b)

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,751 @@
...
+
+  // Implicit `children` prop has been removed in @types/react^18.0.
+  /**
+   * Elements that will be rendered inside the content view.
+   */
+  children?: ReactNode | ((openValue?: SharedValue<number>) => ReactNode);
+
+  /**
+   * @default 'none'
+   * Defines which userSelect property should be used.
```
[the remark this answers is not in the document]
latekvo: Fixed in [3ccfbfe](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/3ccfbfe9e3a446c1f6c4092eca57fd241a01c57b)

src/components/ReanimatedDrawerLayout.tsx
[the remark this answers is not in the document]
j-piasecki: Yep, makes sense

src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -0,0 +1,751 @@
...
+   */
+  drawerWidth?: number;
+
+  /**
+   * Background color of the drawer.
+   */
+  drawerBackgroundColor?: string;
+
+  /**
+   * Set if the drawer should get locked after being opened or closed.
```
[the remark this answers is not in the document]
latekvo: Does this work better?
```
Specifies the lock mode of the drawer. 
Programatic opening/closing isn't affected by the lock mode. Defaults to `UNLOCKED`.
- `UNLOCKED` - the drawer will respond to gestures.
- `LOCKED_CLOSED` - the drawer will stay closed and the gestures will be disabled.
- `LOCKED_OPEN` - the drawer will stay opened and the gestures will be disabled.
```
image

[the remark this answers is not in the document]
latekvo: Fixed in [c0d13b1](https://github.com/software-mansion/react-native-gesture-handler/pull/3146/commits/c0d13b19c1b66ecfcc678d4280df3af9c2fcc6ba) to the following:
```
Specifies the lock mode of the drawer.
Programatic opening/closing isn't affected by the lock mode. Defaults to `UNLOCKED`.
- `UNLOCKED` - the drawer will respond to gestures.
- `LOCKED_CLOSED` - the drawer will move freely until it settles in a closed position, then the gestures will be disabled.
- `LOCKED_OPEN` - the drawer will move freely until it settles in an opened position, then the gestures will be disabled.
``` 
image

j-piasecki (approved):
LGTM! Please wait for GH actions before merging.

m-bert (approved):
Two things that I've noticed:

1. Drawer sometimes flickers (i.e. closes and immediately opens). Unfortunately I'm not exactly sure how to reproduce it. 
2. In the provided example, when `drawer` is opened on `web` it is possible to click buttons in background. It doesn't work like this on `android` and `iOS`

but I think we don't have to worry about it in this PR as it is already quite large 😅

PR #3163, opened 2024-10-17
[web] Add support for two finger pan

### LINKED ISSUE #800: Support panning with touchpads
When providing apps that work across touch devices and laptops with touchpads - which I guess is the common case when using Hammer - the natural way to interact is two-finger swiping to pan and swipe. I'm getting bug reports on this because people feel "swipe isn't working" as they expect, it's simply the UX one would expect (I know it applies to Macs, but I guess it applies to other platforms as well).

I personally feel this might fit best in Hammer - but there seems to be a hesitation to add this to the core library. I had a look at implementing this myself but I'm just not familiar enough with Hammer. Maybe someone that's more familiar with it could have a look? Would be much appreciated :)

m-bert:
## Description

This PR adds support for two finger panning on touchpad on `web`.

>[!WARNING]
> Two finger gestures may be used as system/browser gestures (for example `swipe` to go back). This PR doesn't handle these cases.

## Implementation

Implementation is based on [WheelEvents](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event). This leads to some limitations - whole state flow of `Pan` has to be managed inside one callback (`onWheel`).

### Ending gesture

Starting gesture is easy, in contrast to ending it. To finish gesture we use `timeout`  - if no `wheel event` was received since `setTimeout` was called, we can end gesture. 

### `Mouse` vs `Touchpad`

It is hard to determine whether `mouse` or `touchpad` was used to start the gesture. `WheelEvent` doesn't have any information about the device, therefore we have to use heuristics to check that. 

>[!NOTE]
> You cannot start gesture with mouse scroll.

To see if events were generated with mouse, we check whether `wheelDeltaY` property (which is now [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event#event_properties)) of event is multiple of `120`. In short, this is standard `wheel delta` for mouse. Here you can find useful links that will tell more about why it works:

- https://stackoverflow.com/questions/10744645/detect-touchpad-vs-mouse-in-javascript
- https://devblogs.microsoft.com/oldnewthing/20130123-00/?p=5473  

>[!CAUTION]
> While this will work most of the times, it is possible that user will somehow generate this specific `wheel delta` with touchpad and gesture will not be recognized correctly.

Closes #800

## Test plan

Tested on new _**Two finger Pan**_ example

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -213,7 +213,7 @@ const EXAMPLES: ExamplesSection[] = [
       {
         name: 'Two finger Pan',
         component: TwoFingerPan,
-        unsupportedPlatforms: new Set(['android', 'macos', 'ios']),
+        unsupportedPlatforms: new Set(['android', 'macos']),
       },
     ],
   },
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -39,6 +39,7 @@ import NestedButtons from './src/release_tests/nestedButtons';
 import PointerType from './src/release_tests/pointerType';
 import SwipeableReanimation from './src/release_tests/swipeableReanimation';
 import NestedGestureHandlerRootViewWithModal from './src/release_tests/nestedGHRootViewWithModal';
+import TwoFingerPan from 'src/release_tests/twoFingerPan';
 import { PinchableBox } from './src/recipes/scaleAndRotate';
 import PanAndScroll from './src/recipes/panAndScroll';
 import { BottomSheet } from './src/showcase/bottomSheet';
@@ -209,6 +210,11 @@ const EXAMPLES: ExamplesSection[] = [
         unsupportedPlatforms: new Set(['android', 'ios', 'macos']),
       },
       { name: 'Stylus data', component: StylusData },
+      {
+        name: 'Two finger Pan',
+        component: TwoFingerPan,
+        unsupportedPlatforms: new Set(['android', 'macos', 'ios']),
+      },
     ],
   },
   {
--- a/example/src/release_tests/twoFingerPan/index.tsx
+++ b/example/src/release_tests/twoFingerPan/index.tsx
@@ -1,0 +1,61 @@
+import { StyleSheet, View } from 'react-native';
+import Animated, {
+  SharedValue,
+  useAnimatedStyle,
+  useSharedValue,
+} from 'react-native-reanimated';
+
+import React from 'react';
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
+
+const BOX_SIZE = 270;
+
+export default function TwoFingerPan() {
+  const r = useSharedValue(128);
+  const b = useSharedValue(128);
+
+  const clampColor = (sv: SharedValue<number>) =>
+    sv.value < 0 ? 0 : sv.value > 255 ? 255 : sv.value;
+
+  const pan = Gesture.Pan()
+    .onChange((event) => {
+      r.value -= event.changeY;
+      b.value += event.changeX;
+
+      r.value = clampColor(r);
+      b.value = clampColor(b);
+
+      console.table({ red: r.value, blue: b.value });
+    })
+    .enableTrackpadTwoFingerGesture(true);
+
+  const animatedStyles = useAnimatedStyle(() => {
+    const backgroundColor = `rgb(${r.value}, 128, ${b.value})`;
+
+    return {
+      backgroundColor,
+    };
+  });
+
+  return (
+    <View style={styles.container} collapsable={false}>
+      <GestureDetector gesture={pan}>
+        <Animated.View style={[styles.box, animatedStyles]} />
+      </GestureDetector>
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+    height: '100%',
+  },
+  box: {
+    width: BOX_SIZE,
+    height: BOX_SIZE,
+    borderRadius: BOX_SIZE / 2,
+  },
+});
--- a/example/src/release_tests/twoFingerPan/index.tsx
+++ b/example/src/release_tests/twoFingerPan/index.tsx
@@ -10,20 +10,16 @@ import { Gesture, GestureDetector } from 'react-native-gesture-handler';
 
 const BOX_SIZE = 270;
 
+const clampColor = (v: number) => Math.min(255, Math.max(0, v));
+
 export default function TwoFingerPan() {
   const r = useSharedValue(128);
   const b = useSharedValue(128);
 
-  const clampColor = (sv: SharedValue<number>) =>
-    sv.value < 0 ? 0 : sv.value > 255 ? 255 : sv.value;
-
   const pan = Gesture.Pan()
     .onChange((event) => {
-      r.value -= event.changeY;
-      b.value += event.changeX;
-
-      r.value = clampColor(r);
-      b.value = clampColor(b);
+      r.value = clampColor(r.value - event.changeY);
+      b.value = clampColor(b.value + event.changeX);
 
       console.table({ red: r.value, blue: b.value });
     })
--- a/example/src/release_tests/twoFingerPan/index.tsx
+++ b/example/src/release_tests/twoFingerPan/index.tsx
@@ -19,9 +19,8 @@ export default function TwoFingerPan() {
     .onChange((event) => {
       r.value = clampColor(r.value - event.changeY);
       b.value = clampColor(b.value + event.changeX);
-
-      console.table({ red: r.value, blue: b.value });
     })
+    .runOnJS(true)
     .enableTrackpadTwoFingerGesture(true);
 
   const animatedStyles = useAnimatedStyle(() => {
--- a/example/src/release_tests/twoFingerPan/index.tsx
+++ b/example/src/release_tests/twoFingerPan/index.tsx
@@ -1,6 +1,5 @@
 import { StyleSheet, View } from 'react-native';
 import Animated, {
-  SharedValue,
   useAnimatedStyle,
   useSharedValue,
 } from 'react-native-reanimated';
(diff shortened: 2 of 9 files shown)


example/App.tsx
```diff
@@ -209,6 +210,11 @@ const EXAMPLES: ExamplesSection[] = [
         unsupportedPlatforms: new Set(['android', 'ios', 'macos']),
       },
       { name: 'Stylus data', component: StylusData },
+      {
+        name: 'Two finger Pan',
+        component: TwoFingerPan,
+        unsupportedPlatforms: new Set(['android', 'macos', 'ios']),
```
j-piasecki, line 216:
```suggestion
        unsupportedPlatforms: new Set(['android', 'macos']),
```

Shouldn't this prop also work on iPads?

example/src/release_tests/twoFingerPan/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const clampColor = (sv: SharedValue<number>) =>
+    sv.value < 0 ? 0 : sv.value > 255 ? 255 : sv.value;
+
+  const pan = Gesture.Pan()
+    .onChange((event) => {
+      r.value -= event.changeY;
+      b.value += event.changeX;
+
+      r.value = clampColor(r);
+      b.value = clampColor(b);
```
j-piasecki, line 26:
```suggestion
const clampColor = (v: number) => Math.min(255, Math.max(0, v));

export default function TwoFingerPan() {
  const r = useSharedValue(128);
  const b = useSharedValue(128);

  const pan = Gesture.Pan()
    .onChange((event) => {
      r.value = clampColor(r.value - event.changeY);
      b.value = clampColor(b.value + event.changeX);
```

example/src/release_tests/twoFingerPan/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+      console.table({ red: r.value, blue: b.value });
```
j-piasecki, line 28: Is this log needed?

src/web/handlers/PanGestureHandler.ts
```diff
@@ -351,6 +361,70 @@ export default class PanGestureHandler extends GestureHandler {
     }
   }
 
+  private scheduleWheelEnd(event: AdaptedEvent) {
+    this.endWheelTimeout = setTimeout(() => {
+      if (this.receivedWheelEvent) {
+        return;
+      }
```
j-piasecki, line 368: Why is this needed when the timeout is cleared in `onWheel`? And why not clean it here, before scheduling a new one?

src/web/handlers/PanGestureHandler.ts
```diff
@@ -351,6 +361,70 @@ export default class PanGestureHandler extends GestureHandler {
...
+
+      this.end();
+      this.tracker.removeFromTracker(event.pointerId);
+      this.currentState = State.UNDETERMINED;
+
+      this.wheelDevice = WheelDevice.UNDETERMINED;
+    }, 30);
```
j-piasecki, line 375: Shouldn't this be configurable?

src/web/handlers/PanGestureHandler.ts
```diff
@@ -351,6 +361,70 @@ export default class PanGestureHandler extends GestureHandler {
...
+    clearTimeout(this.endWheelTimeout);
+
+    if (this.currentState === State.UNDETERMINED) {
+      this.wheelDevice =
+        event.wheelDeltaY! % 120 !== 0
+          ? WheelDevice.TOUCHPAD
+          : WheelDevice.MOUSE;
+
+      if (this.wheelDevice === WheelDevice.MOUSE) {
+        this.scheduleWheelEnd(event);
```
j-piasecki, line 398: 1. Do we need to schedule in this case instead of ending outright?
2. Do we need to end here at all? `begin` is called after this condition, so we shouldn't even get there with mouse, no?

src/web/tools/PointerEventManager.ts
```diff
@@ -200,7 +217,16 @@ export default class PointerEventManager extends EventManager<HTMLElement> {
...
   }
 
-  protected mapEvent(event: PointerEvent, eventType: EventTypes): AdaptedEvent {
+  protected mapEvent(
+    event: PointerEvent | WheelEvent,
+    eventType: EventTypes
+  ): AdaptedEvent {
+    return event instanceof PointerEvent
+      ? this.mapPointerEvent(event, eventType)
+      : this.mapWheelEvent(event);
```
j-piasecki, line 226: We may want to redesign the event manager system in the future. We now have `PointerEventManager` that needs to check whether the event it receives is a `PointerEvent` which seems wrong to me.

j-piasecki (commented):
Have you checked if the behavior is the same as on iPad?

latekvo:
As far as I see, `WheelEvent.wheelDeltaY` [isn't supported on firefox](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousewheel_event), would it be possible to resort to other heuristics like having a threshold on `WheelEvent.deltaY` in combination with `WheelEvent.deltaMode` for that browser?

src/web/tools/PointerEventManager.ts
m-bert: There's nothing wrong with adding another event manager, but I'm not exactly sure how would it work. Adding `MouseEventManager` doesn't make much sense at this point for me. Also, this manager would have only one method - `onWheel`.

On the other hand, `KeyboradManager` also has one callback, so maybe it is fine 😅

src/web/handlers/PanGestureHandler.ts
m-bert: I'm not sure, should it? I was to move it to a constant but it seems that I've missed it. Either way, do we want to give users permission to change this value? I think we can stick to one that works ok, but leaving it open for customization also makes sense.

src/web/handlers/PanGestureHandler.ts
m-bert: If we don't schedule cleanup, wheelDevice won't be set to `UNDETERMINED`. I'm not sure if we want to set it here though. However, we can call `end` only if handler was `active`.

example/App.tsx
m-bert: It should 😅

m-bert: Brought back in [dc59483](https://github.com/software-mansion/react-native-gesture-handler/pull/3163/commits/dc59483274fb14a3478a0e95a3958080a44c5bbd)

src/web/handlers/PanGestureHandler.ts
m-bert: At first I wondered if the situation below is possible:

```tsx
this.receivedWheelEvent = true;
// execute `scheduleWheelEnd` here
clearTimeout(this.endWheelTimeout);
```

but we believe that it is not possible because of how event loop works.

So yes, it is not necessary and I've removed it in [a640c0c](https://github.com/software-mansion/react-native-gesture-handler/pull/3163/commits/a640c0c7700d96d66b95ecda06b1bf8c7bd730c3)

src/web/handlers/PanGestureHandler.ts
m-bert: Also changed in [a640c0c](https://github.com/software-mansion/react-native-gesture-handler/pull/3163/commits/a640c0c7700d96d66b95ecda06b1bf8c7bd730c3)

m-bert:
>Overall, would it be possible to detect that mouse is not being used instead of checking if it is?

As discussed, I think there's no better way to check that (but I'm open to discussion if you do find a better approach).

>would it be possible to resort to other heuristics like having a threshold on WheelEvent.deltaY in combination with WheelEvent.deltaMode for that browser?

Values of `deltaY` differ between browsers when it comes to mouse. Also `deltaMode` is the same for touchpad and for mouse.

>The current method seems to be very prone to false positives when it comes to detecting touchpad being used.

Activating on mouse wheel is a side effect of this feature that is not affecting main functionality.

m-bert:
> Have you checked if the behavior is the same as on iPad?

I did and indeed it does (at least the example does) 😅

src/web/tools/PointerEventManager.ts
m-bert: Okay, I've discussed it with @latekvo and we decided that it will be ok to add `WheelEventManager`. Done in [bf008b1](https://github.com/software-mansion/react-native-gesture-handler/pull/3163/commits/bf008b19202cc5d23b4bfa6c68fd4eec32d8c5ce)

example/src/release_tests/twoFingerPan/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+export default function TwoFingerPan() {
+  const r = useSharedValue(128);
+  const b = useSharedValue(128);
+
+  const pan = Gesture.Pan()
+    .onChange((event) => {
+      r.value = clampColor(r.value - event.changeY);
+      b.value = clampColor(b.value + event.changeX);
+    })
+    .runOnJS(true)
```
j-piasecki, line 23: Why does it need to run on the JS thread?

j-piasecki (approved):

src/web/handlers/PanGestureHandler.ts
j-piasecki: I guess we can leave it for now and see if there's a demand for it.

example/src/release_tests/twoFingerPan/index.tsx
m-bert: Because this example is also available on iOS 😅

j-piasecki: What does it change?

Isn't making `clampColor` a worklet enough?

m-bert: You're right, it is. I don't think it needs a PR though 😅

PR #3165, opened 2024-10-17
Optimise `ReanimatedSwipeable` component

latekvo:
## Description

- Memoize all Gestures, and non-memoized functions.
- Reduce number of used shared values
- Fix unoptimal dependency lists on existing `useMemo`s

## Test plan

- use `android`, `iOS` has no performance issues
- use attached code to measure performance
  - tested on physical `Samsung S24 Ultra`
  - pre-optimization average: `19 fps`
  - post-optimization average: `68 fps`

## Test code

Tested on `@shopify/flash-list` using their `useBenchmark` utility hook.

<details>
<summary>
Collapsed code
</summary>

```js
import { Text, StyleSheet, View, Alert } from 'react-native';

import { FlashList, ListRenderItem, useBenchmark } from '@shopify/flash-list';
import { memo, useRef } from 'react';

function RightAction() {
  return <Text style={styles.rightAction}>Text</Text>;
}

type Item = {
  id: string;
  title: string;
};

const generateItems = (count: number): Item[] => {
  return Array.from({ length: count }, () => {
    const randomNumber = Math.random().toString(36).substring(2, 8);
    return {
      id: randomNumber,
      title: `Title ${randomNumber}`,
    };
  });
};

const data = generateItems(200);

const _RenderItemView = (item: Item) => (
  <View style={styles.swipeable}>
    <Text>{item.title}</Text>
  </View>
);

const RenderItemViewMemoed = memo(_RenderItemView);

  const renderItem: ListRenderItem<Item> = ({ item }) => {
    return (
      <ReanimatedSwipeable renderRightActions={RightAction}>
        <RenderItemViewMemoed {...item} />
      </ReanimatedSwipeable>
    );
  };

  const flashListRef = useRef<FlashList<Item> | null>(null);
  useBenchmark(flashListRef, (callback) => {
    Alert.alert('result', callback.formattedString);
  });

  return (
    <FlashList
      ref={flashListRef}
      data={data}
      renderItem={renderItem}
      estimatedItemSize={50}
      keyExtractor={(item) => item.id}
    />
  );
}

  leftAction: { width: 50, height: 50, backgroundColor: 'crimson' },
  rightAction: { width: 50, height: 50, backgroundColor: 'purple' },
  separator: {
    width: '100%',
    borderTopWidth: 1,
  },
  swipeable: {
    height: 50,
    alignItems: 'center',
  },
});
```
</details>

### DIFF
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -251,17 +251,13 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
 
     const rowState = useSharedValue<number>(0);
 
-    // %% cache value, see if they are already provided in the event
     const userDrag = useSharedValue<number>(0);
 
-    // %% same-ish
     const appliedTranslation = useSharedValue<number>(0);
 
-    // %% these likely could be removed, see if rightOffset is neccessary
     const rowWidth = useSharedValue<number>(0);
     const leftWidth = useSharedValue<number>(0);
     const rightWidth = useSharedValue<number>(0);
-    const rightOffset = useSharedValue<number | null>(null);
 
     // %% progression cache, houses withSpring output
     const showLeftProgress = useSharedValue<number>(0);
@@ -285,18 +281,8 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
     const overshootLeftProp = overshootLeft;
     const overshootRightProp = overshootRight;
 
-    const updateRightElementWidth = useCallback(() => {
-      'worklet';
-      if (rightOffset.value === null) {
-        rightOffset.value = rowWidth.value;
-      }
-      rightWidth.value = Math.max(0, rowWidth.value - rightOffset.value);
-    }, [rightOffset, rightWidth, rowWidth.value]);
-
     const updateAnimatedEvent = useCallback(() => {
       'worklet';
-
-      updateRightElementWidth();
 
       const overshootLeft = overshootLeftProp ?? leftWidth.value > 0;
       const overshootRight = overshootRightProp ?? rightWidth.value > 0;
@@ -354,7 +340,6 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       rowState.value,
       showLeftProgress,
       showRightProgress,
-      updateRightElementWidth,
       userDrag.value,
     ]);
 
@@ -539,13 +524,19 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
               swipeableMethods.current
             )}
             <View
-              onLayout={({ nativeEvent }) =>
-                (rightOffset.value = nativeEvent.layout.x)
-              }
+              onLayout={({ nativeEvent }) => {
+                rightWidth.value = rowWidth.value - nativeEvent.layout.x;
+              }}
             />
           </Animated.View>
         ),
-      [appliedTranslation, renderRightActions, rightOffset, showRightProgress]
+      [
+        appliedTranslation,
+        renderRightActions,
+        rightWidth,
+        rowWidth.value,
+        showRightProgress,
+      ]
     );
 
     const handleRelease = useCallback(
@@ -553,8 +544,6 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
         'worklet';
         const { velocityX } = event;
         userDrag.value = event.translationX;
-
-        updateRightElementWidth();
 
         const leftThresholdProp = leftThreshold ?? leftWidth.value / 2;
         const rightThresholdProp = rightThreshold ?? rightWidth.value / 2;
@@ -592,7 +581,6 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
         rightThreshold,
         rightWidth.value,
         rowState.value,
-        updateRightElementWidth,
         userDrag,
       ]
     );
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -285,6 +285,7 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
     const overshootLeftProp = overshootLeft;
     const overshootRightProp = overshootRight;
 
+    // %% should only run once, not every pan update
     const updateRightElementWidth = useCallback(() => {
       'worklet';
       if (rightOffset.value === null) {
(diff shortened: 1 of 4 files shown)


m-bert:
I don't understand this part:

>2.6 | remove all unused variables
4.6 | remove all unused variables < large leap
4.8 | remove all unused variables


How did you manage to remove _**all**_ unused variables, and then do this again and again?

src/components/ReanimatedSwipeable.tsx
```diff
@@ -262,28 +280,20 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
-      friction: 1,
-      overshootFriction: 1,
-    };
-
-    const {
-      friction = defaultProps.friction,
-      overshootFriction = defaultProps.overshootFriction,
-    } = props;
-
+    // %% remove or rename these, may cause 1. reloads, 2. name shadowing
```
m-bert, line 283: What about this? Also, such comments should be marked as `TODO`

src/components/ReanimatedSwipeable.tsx
```diff
@@ -262,28 +280,20 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
     const overshootLeftProp = overshootLeft;
     const overshootRightProp = overshootRight;
 
-    const updateRightElementWidth = () => {
+    // %% should only run once, not every pan update
```
m-bert, line 287: What about those `%%` is this some convention that I'm not familiar with?

src/components/ReanimatedSwipeable.tsx
```diff
@@ -333,37 +343,64 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+      userDrag.value,
+      updateRightElementWidth,
+    ]);
 
     const dispatchImmediateEvents = useCallback(
       (fromValue: number, toValue: number) => {
+        'worklet';
         if (toValue > 0) {
-          onSwipeableWillOpen?.(SwipeDirection.RIGHT);
+          onSwipeableWillOpen &&
```
m-bert, line 365: Why don't we move this check to the `if` above?

```tsx
if (toValue > 0 && onSwipeableWillOpen){
  runOnJS(onSwipeableWillOpen)(SwipeDirection.RIGHT);
}
```

src/components/ReanimatedSwipeable.tsx
latekvo: I use `%%` to mark issues and questions i want to resolve before merging. 
I believe I removed all such comments before requesting your review, 
but as time passed I found new issues i have to address.

src/components/ReanimatedSwipeable.tsx
latekvo: Same here

latekvo:
>How did you manage to remove all unused variables, and then do this again and again?

Upon removing unused values, new, previously used values become unused.
By removing unused values via `VSC`'s `remove all unused declarations` functionality, 
same heaps of unused declarations can be consistently removed.
This helps with sifting out which variables/functions are associated with the largest `fps` changes.

As can be seen, the second `remove all unused variables` step causes the largest `fps` improvement among the early steps, this is due to removal of `updateAnimatedEvent` and `handleRelease` callbacks (`43%` and `25%` improvement respectively), unfortunately neither of these functions had any glaring issues, and i don't think much can be done to optimise them.

src/components/ReanimatedSwipeable.tsx
latekvo: I agree, your approach looks cleaner.

Changed in [155f20f](https://github.com/software-mansion/react-native-gesture-handler/pull/3165/commits/155f20f8a78f0dfd503e4f47bf16f9c2553f2457)

src/components/ReanimatedSwipeable.tsx
latekvo: Removed this comment in [c2ac345](https://github.com/software-mansion/react-native-gesture-handler/pull/3165/commits/c2ac3452729d20a36041a78b68860299472486da)

src/components/ReanimatedSwipeable.tsx
latekvo: Removed this comment in [09f145b](https://github.com/software-mansion/react-native-gesture-handler/pull/3165/commits/09f145b8d0a930d21a219a53d5ab74958d5e7a9c)

m-bert:
>Upon removing unused values, new, previously used values become unused.

Yes, I get that. The thing is, in such report I'd expect only one entry about removing unused variables, in which whole chain is removed, not just first layer 😅

src/components/ReanimatedSwipeable.tsx
```diff
@@ -433,179 +461,196 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
+        onSwipeableOpenStartDrag,
+        rowState,
+        updateAnimatedEvent,
+        updateRightElementWidth,
+        userDrag,
+      ]
+    );
 
-    panGesture.enabled(enabled !== false);
+    useImperativeHandle(ref, () => swipeableMethods.value, [swipeableMethods]);
```
m-bert, line 653: Why do we need to use `useImperativeHandle`?

latekvo: I'm not sure if there's another method for attaching methods to the ref of `ReanimatedSwipeable`.
Why do we want to avoid using it?

m-bert: I forgot that previous implementation was a class component. Either way, chances are that this might not work as expected since `swipeableMethods` is `SharedValue` and its reference does not change between renders. I think @j-piasecki will provide better explanation 😅

j-piasecki: The potential issue here is that shared values (and I think derived values as well) are preferentially stable between renders. In this case, the `useImperativeHandle` hook will execute exactly once as the swipeableMethods itself doesn't change - its value does.

Another problem is that the only time `useImperativeHandle` may be reevaluated is during render, but changes to shared values (and derived values, animated styles, etc.) don't cause a re-render. Due to this, changing the dependency array to `swipeableMethods.value` may or may not work, since it would need to depend on something else to trigger renders.

latekvo: I encountered a dependancy circle when working on memoizing the `swipeableMethods` object.
[`swipeableMethods` -> `animateRow` -> `dispatchEndEvents` -> `swipeableMethods`]
My solution for now is to remove `swipeableMethods` from the `dispatchEndEvents`, this unfortunately has to change the `onSwipeableOpen` and `onSwipeableClose` functions' signature from: 
```js
(
  direction: SwipeDirection.LEFT | SwipeDirection.RIGHT,
  swipeable: SwipeableMethods
) => void
``` 
to 
```js
(
  direction: SwipeDirection.LEFT | SwipeDirection.RIGHT,
) => void
```
I couldn't find an alternative solution for this problem, and I think there might not be a way around it.
Thankfully the now removed `swipeable: SwipeableMethods` argument is completely redundant, and can be replaced with the `swipeableRef` exposed via `useImperativeHandle`, 
this is also the way it currently works in the legacy `Swipeable` component.

m-bert (approved):
I love the fact that I can see avg 68 fps on my phone 😄 

Great job!

PR #3166, opened 2024-10-17
fix: Android event dispatcher change for 0.76.0-rc.6

mysport12:
https://github.com/facebook/react-native/pull/47090/files

## Description
Fixes a build failure on RN 0.76.0-rc.6 as a result of a change in the way the event dispatcher is accessed.

See change here: https://github.com/facebook/react-native/pull/47090/files

Note: This change breaks RN 0.76.0-rc.5 and below

## Test plan

- Android build fails with RN 0.76.0-rc.6 before the change
- Successful build with RN 0.76.0-rc.6 after the change

### DIFF
--- a/FabricExample/package.json
+++ b/FabricExample/package.json
@@ -16,7 +16,7 @@
     "patch-package": "^6.5.0",
     "postinstall-postinstall": "^2.1.0",
     "react": "18.3.1",
-    "react-native": "0.76.0-rc.2",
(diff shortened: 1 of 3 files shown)


m-bert (approved):
Thanks for fixing this ❤️

latekvo (approved):

renchap:
You only updated the fabric files with the change, but forgot to update the non-fabric code. I opened a PR with this change: #3176

PR #3173, opened 2024-10-21
Disable worklet warnings on web

nahn20:
## Description

This PR disables callback worklet warnings on web. These warnings are unnecessary since the web platform doesn't leverage worklets.


m-bert:
HI @nahn20! Unfortunately we don't want to remove this error, even on web 😞 If it is bothering you please use `runOnJS` as error suggests.

nahn20:
Makes sense! We were trying to avoid that since these callbacks are designed to be worklets, but on web, they're not detected and marked as worklets. I'm sure this works for Expo Router projects, but our build system isn't configured for web worklet detection.

PR #3176, opened 2024-10-22
Fixes Android event dispatcher for old arch

renchap:
## Description

This is similar to #3166, but for the old arch

## Test plan

Before this change, our app failed to build with RN 0.76.0-rc.6 without the new arch enabled. With this change, it builds and runs correctly.


m-bert (approved):
Thank you very much for fixing this problem ❤️

haswalt:
I know this is only a recent change but can there be a release cut with this in? 0.76 is currently broken.

m-bert:
Hi @haswalt, we released it in 2.20.2 yesterday 😅

haswalt:
lol missed it by 2 hours. Good work

PR #3183, opened 2024-10-24
Remove Reanimated 2 version from the docs

j-piasecki:
## Description

Removes explicit Reanimated 2 references - the library also works with Rea 3


m-bert (approved):
👍 

Also, it is a bit inconsistent that in some places we have Reanimated and in others `Reanimated`, maybe we should unify that?

PR #3191, opened 2024-10-30
Make the `FabricExample` use the common examples.

latekvo:
## Description

This PR makes the `FabricExample` app use the same set of examples as the paper example app.

## Test plan

- open the `FabricExample` app and see how it looks like the common example app.


latekvo:
closing, making it easier to switch between `fabric` and `paper` directly on the common example is simpler, and will allow us to remove `FabricExample` all together.

PR #3197, opened 2024-11-04
[Web] Adjust `findNodeHandle` to properly detect `SVG`

m-bert:
## Description

#3127 introduced our own implementation of `findNodeHandle` on web. Unfortunately, it doesn't work if someone uses `GestureDetector` on SVG elements, e.g.:

```jsx
<Svg width={200} height={200}>
  <GestureDetector gesture={tapGestureCircle}>
    <Circle r={200} cx={210} cy={210} fill={circleFill} />
  </GestureDetector>
</Svg>
```

Code above would render additional `div` element with `display: 'contents';`, which would break `SVG`.  

This PR does the following things:

1.  Bumps `react-native-svg` version
2. Modifies `Wrap` component such that now if element is part of `SVG` it doesn't wrap it into additional `div`
3. Adjusts `findNodeHandle` function to properly handle `SVG` refs

## Test plan

<details>

```jsx
import {
  GestureHandlerRootView,
  GestureDetector,
  Gesture,
import { Svg, Circle } from 'react-native-svg';
import { useState, useCallback } from 'react';

  const [circleFill, setCircleFill] = useState('blue');
  const switchCircleColor = useCallback(
    () => setCircleFill((old) => (old === 'blue' ? 'brown' : 'blue')),
    [setCircleFill]
  );

  const tapGestureCircle = Gesture.Tap().runOnJS(true).onEnd(switchCircleColor);

  return (
    <GestureHandlerRootView style={{ flex: 1, paddingTop: 200 }}>
      <View style={{ padding: 10, borderWidth: 1, alignSelf: 'flex-start' }}>
        <Svg width={200} height={200}>
          <GestureDetector gesture={tapGestureCircle}>
            <Circle r={200} cx={210} cy={210} fill={circleFill} />
          </GestureDetector>
        </Svg>
      </View>
    </GestureHandlerRootView>
  );
}
```

</details>

### DIFF
--- a/src/findNodeHandle.web.ts
+++ b/src/findNodeHandle.web.ts
@@ -1,11 +1,9 @@
-type GestureHandlerRef = {
-  viewTag: GestureHandlerRef;
-  current: HTMLElement;
-};
+import type { GestureHandlerRef, SVGRef } from './web/interfaces';
+import { isSVGElement } from './web/utils';
 
 export default function findNodeHandle(
-  viewRef: GestureHandlerRef | HTMLElement
-): HTMLElement | number {
+  viewRef: GestureHandlerRef | SVGRef | HTMLElement | SVGElement
+): HTMLElement | SVGElement | number {
   // Old API assumes that child handler is HTMLElement.
   // However, if we nest handlers, we will get ref to another handler.
   // In that case, we want to recursively call findNodeHandle with new handler viewTag (which can also be ref to another handler).
@@ -13,7 +11,7 @@ export default function findNodeHandle(
     return findNodeHandle((viewRef as GestureHandlerRef).viewTag);
   }
 
-  if (viewRef instanceof HTMLElement) {
+  if (viewRef instanceof HTMLElement || viewRef instanceof SVGElement) {
     if (viewRef.style.display === 'contents') {
       return findNodeHandle(viewRef.firstChild as HTMLElement);
     }
@@ -21,9 +19,13 @@ export default function findNodeHandle(
     return viewRef;
   }
 
+  if (isSVGElement(viewRef)) {
+    return (viewRef as SVGRef).elementRef.current;
+  }
+
   // In new API, we receive ref object which `current` field points to  wrapper `div` with `display: contents;`.
   // We want to return the first descendant (in DFS order) that doesn't have this property.
-  let element = viewRef?.current;
+  let element = (viewRef as GestureHandlerRef)?.current;
 
   while (element && element.style.display === 'contents') {
     element = element.firstChild as HTMLElement;
--- a/src/findNodeHandle.web.ts
+++ b/src/findNodeHandle.web.ts
@@ -1,5 +1,5 @@
 import type { GestureHandlerRef, SVGRef } from './web/interfaces';
-import { isSVGElement } from './web/utils';
+import { isRNSVGElement } from './web/utils';
 
 export default function findNodeHandle(
   viewRef: GestureHandlerRef | SVGRef | HTMLElement | SVGElement
@@ -19,7 +19,7 @@ export default function findNodeHandle(
     return viewRef;
   }
 
-  if (isSVGElement(viewRef)) {
+  if (isRNSVGElement(viewRef)) {
     return (viewRef as SVGRef).elementRef.current;
   }
 
--- a/src/findNodeHandle.web.ts
+++ b/src/findNodeHandle.web.ts
@@ -11,7 +11,7 @@ export default function findNodeHandle(
     return findNodeHandle((viewRef as GestureHandlerRef).viewTag);
   }
 
-  if (viewRef instanceof HTMLElement || viewRef instanceof SVGElement) {
+  if (viewRef instanceof Element) {
     if (viewRef.style.display === 'contents') {
       return findNodeHandle(viewRef.firstChild as HTMLElement);
     }
--- a/src/handlers/gestures/GestureDetector/Wrap.web.tsx
+++ b/src/handlers/gestures/GestureDetector/Wrap.web.tsx
@@ -12,10 +12,10 @@ export const Wrap = forwardRef<HTMLDivElement, PropsWithChildren<{}>>(
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const child: any = React.Children.only(children);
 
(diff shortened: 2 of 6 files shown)


src/handlers/gestures/GestureDetector/Wrap.web.tsx
```diff
@@ -12,14 +12,23 @@ export const Wrap = forwardRef<HTMLDivElement, PropsWithChildren<{}>>(
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const child: any = React.Children.only(children);
 
+      const isSVGElement =
+        Object.getPrototypeOf(child?.type)?.name === 'WebShape';
```
jakex7, line 16: Why this check is different from https://github.com/software-mansion/react-native-gesture-handler/pull/3197/files#diff-ec5692c69123b44d3a59cab1b401b52570a631f15fa37631163eb7502424fdbdR259-R265 ?

src/web/utils.ts
```diff
@@ -227,3 +232,34 @@ function spherical2tilt(altitudeAngle: number, azimuthAngle: number) {
...
+  'TextPath',
+  'Use',
+];
+
+export function isSVGElement(viewRef: SVGRef | GestureHandlerRef) {
+  const className = Object.getPrototypeOf(viewRef).constructor.name;
+
+  return (
+    SVGElements.indexOf(className) >= 0 && Object.hasOwn(viewRef, 'elementRef')
+  );
```
j-piasecki, line 264: This is confusing. Where does `SVGRef` come from? Why do we check for `elementRef`?

src/findNodeHandle.web.ts
```diff
@@ -1,29 +1,31 @@
...
+): HTMLElement | SVGElement | number {
   // Old API assumes that child handler is HTMLElement.
   // However, if we nest handlers, we will get ref to another handler.
   // In that case, we want to recursively call findNodeHandle with new handler viewTag (which can also be ref to another handler).
   if ((viewRef as GestureHandlerRef)?.viewTag !== undefined) {
     return findNodeHandle((viewRef as GestureHandlerRef).viewTag);
   }
 
-  if (viewRef instanceof HTMLElement) {
+  if (viewRef instanceof HTMLElement || viewRef instanceof SVGElement) {
```
j-piasecki, line 14: Can't we check simply for `Element`?

j-piasecki (commented):
(whispers into the void) Who's gonna tell him about `MathMLElement`?

src/handlers/gestures/GestureDetector/Wrap.web.tsx
m-bert: Because those are 2 different things. Here we get React node:

image

And in the line that you've linked we get ref to the actual `SVG` component (like `Circle`)

image

src/findNodeHandle.web.ts
m-bert: Seems like we can, both of them extend `Element`. Moreover, `MathMLElement` also  does this 😅 

Done in [dfe53b6](https://github.com/software-mansion/react-native-gesture-handler/pull/3197/commits/dfe53b6ec3a2409b6655ab329a2d51d52170f4df)

src/web/utils.ts
m-bert: Commentary added in [81e85ac](https://github.com/software-mansion/react-native-gesture-handler/pull/3197/commits/81e85ac5ba8769a364c2ac50613587c76829e856)

src/web/utils.ts
```diff
@@ -227,3 +232,43 @@ function spherical2tilt(altitudeAngle: number, azimuthAngle: number) {
...
+// findNodeHandle, we can encounter such element in 2 forms - SVG tag or ref to SVG Element. Since Gesture Handler
+// does not depend on SVG, we use our simplified SVGRef type that has `elementRef` field. This is something that is present
+// in actual SVG ref object.
+//
+// In order to make sure that node passed into this function is in fact SVG element, first we check if its constructor name
+// corresponds to one of the possible SVG elements. Then we also check if `elementRef` field exists.
+// By doing both steps we decrease probability of detecting situations where, for example, user makes custom `Circle` and
+// we treat it as SVG.
+export function isSVGElement(viewRef: SVGRef | GestureHandlerRef) {
+  const className = Object.getPrototypeOf(viewRef).constructor.name;
```
jakex7, line 269: Maybe we should rename it to avoid confusion with the React `className` prop. https://react.dev/learn#adding-styles

src/web/utils.ts
jakex7: It might be helpful to prefix these checks with `RN` for `RNSVG` to avoid potential confusion, as it’s not part of the standard web API.

j-piasecki: > SVGRef is a type added in interfaces.ts file

That I can see, but where does `elementRef` come from?

src/web/utils.ts
m-bert: Changed to `componentClassName` in [0722672](https://github.com/software-mansion/react-native-gesture-handler/pull/3197/commits/0722672fafccfc11363e210126d38e95d2ad2fcd), let me know what do you think 😅

jakex7 (approved):
LGTM :shipit:

src/web/utils.ts
j-piasecki: For future context - it's this https://github.com/software-mansion/react-native-svg/blob/38186e82833d7a7d6bb3efdf272b8919028e7df0/src/web/WebShape.ts#L23-L24

src/handlers/gestures/GestureDetector/Wrap.web.tsx
j-piasecki: Can we rename one of those in this case?

m-bert: What about renaming the one in `Wrap.web.ts` into `isRNSVGComponent`? Or `isRNSVGNode`?

m-bert: Changed to `isRNSVGNode` in [73d7924](https://github.com/software-mansion/react-native-gesture-handler/pull/3197/commits/73d7924ee88c8aff768cdf4dbc36055801fddfc6)

PR #3204, opened 2024-11-08
Add fabric support to the example app

latekvo:
## Description

Adds `fabric` support to the example app. 
To use `fabric`, set the optional `FABRIC_ENABLED` environment variable to `1` during `yarn` installation. 

## Test plan

- test `fabric`
  - `cd example`
  - `FABRIC_ENABLED=1 yarn`
  - `yarn start`
  - run `Android` via `Android Studio`
  - `cd ios`
  - bundle exec pod install`
  - run `iOS` via `XCode`
  - see how when apps load, the metro displays `bridgeless mode enabled` message
- test `paper`
  - `cd example`
  - `yarn`
  - `yarn start`
  - run `Android` via `Android Studio`
  - `cd ios`
  - `bundle exec pod install`
  - run `iOS` via `XCode`


example/package.json
```diff
@@ -14,31 +14,32 @@
...
+    "expo-camera": "~15.0.16",
     "expo-font": "~12.0.10",
     "hoist-non-react-statics": "^3.3.2",
     "invariant": "^2.2.4",
     "patch-package": "^6.5.1",
     "postinstall-postinstall": "^2.1.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "react-native": "0.74.1",
+    "react-native": "0.75.4",
```
m-bert, line 37: Why not 0.76?

example/package.json
m-bert, line 1: Why do we change/remove `^` in some versions?

example/src/release_tests/touchables/index.tsx
```diff
@@ -33,7 +33,9 @@ const renderSampleBox = (color?: string) => (
   />
 );
 
-const toReactNativeTouchable = (touchable: React.ComponentType<unknown>) => {
+const toReactNativeTouchable = (
+  touchable: React.ComponentType<unknown>
+): React.ComponentType<any> => {
```
m-bert, line 38: Is there a way to avoid `any`?

m-bert: Also, why in some functions we remove return types, while in other we add them?

example/package.json
latekvo: For now expo migrating tools for `rn 0.75` -> `rn 0.76` and `sdk 51` -> `sdk 52` are not yet supported, and I could not manually get it working by bumping everything's version to latest.
- `expo install expo@^52.0.0 --fix` < crashes upon usage
- `expo-doctor` < doesn't support `sdk 52` yet
- `expo upgrade helper` < doesn't support `sdk 52` yet

I believe it'd be best to wait for one of these tools to be supported. It could take anywhere from a couple of days to a couple of weeks, so i think it makes the most sense to keep it at `0.75` for now, and open a separate PR bumping the version to `0.76` once `expo` migrating tools will start working.

example/package.json
latekvo: `^` means the minor version is flexible
`~` means the patch version is flexible
most of these changes were applied automatically by `yarn`

example/src/release_tests/touchables/index.tsx
latekvo: >why in some functions we remove return types

I removed all instances of `ViewStyle`, as it's incompatible between `0.75` and `0.74`, in all instances where it's been removed, the type has been automatically inferred.

>while in other we add them?

Which ones are you referring to?
I can only see types being either removed, or replaced by generic ones where they couldn't've been inferred.

>Is there a way to avoid any?

Unfortunately I couldn't find a way to.
`unknown` and `never` won't work, and i couldn't find another type which would prevent the errors being thrown. The type also cannot be automatically inferred in this instance.
This issue is a result of `RN`'s `TouchableHighlight` and `TouchableOpacity` being rewritten to function-style components in `0.75` while `TouchableNativeFeedback` and `TouchableWithoutFeedback` remain class components.

example/package.json
latekvo: I'm not sure why these changes were made.
- could've been automatically set by `yarn`, in which case I don't really have a way of checking why `~` was chosen over `^` or the other way around.
- could've been recommended by `expo-doctor`, but likewise i don't really have a way of knowing for certain why `~` was chosen over `^` or the other way around.

If you'd prefer that, i can revert all changes related to `^` and `~`, but i'm not sure why we'd want to do that.

m-bert: We don't have, if it was done by package manager

example/src/release_tests/touchables/index.tsx
latekvo: >To this one 😉 Previously there was no type annotation for return value.

Right, sorry, `React.ComponentType<any>` in this case means that a react component with `any` props will be returned. I'll have to give it a second thought. I'm not sure if the new function-style components it is now returning can be used in the same way the old class components could've been used.

latekvo: fixed in [273661f](https://github.com/software-mansion/react-native-gesture-handler/pull/3204/commits/273661f97aae6b0e53fb8d700767e79ccc9a683e)

m-bert: What I meant was to ask if TypeScript has a problem with type inference in that case?

example/src/release_tests/touchables/index.tsx
```diff
@@ -297,7 +301,10 @@ const TOUCHABLES: Touchables[] = [
 ];
 
 const screens: Record<string, Touchables> = TOUCHABLES.reduce(
-  (map: Record<string, Touchables>, obj) => ((map[obj.text] = obj), map),
+  (map: Record<string, Touchables>, obj: Touchables) => {
+    map[obj.text] = obj;
+    return map;
+  },
```
m-bert, line 307: Doesn't it do the same thing? If so, why do we change that?

j-piasecki: It explicitly types the `obj` to be `Touchables`, and (IMO) `return` is more readable than using a comma here.

example/src/release_tests/touchables/index.tsx
latekvo: >What I meant was to ask if TypeScript has a problem with type inference in that case?

It does. Here's the error that's thrown when using the element returned by this function:
image

Also, I tested all the problematic components - `RNTouchableOpacity` and `RNTouchableHighlight` - which are the reason for those type errors, and they function as intended. All props are properly applied, the components are rendered.

example/app.config.js
```diff
@@ -0,0 +1,59 @@
...
+    ],
+    [
+      'expo-build-properties',
+      {
+        ios: {
+          newArchEnabled: !!+process.env.FABRIC_ENABLED,
+        },
+        android: {
+          newArchEnabled: !!+process.env.FABRIC_ENABLED,
+        },
```
m-bert, line 55: WDYT @j-piasecki?

m-bert (commented):
LGTM, please wait for @j-piasecki approve and then we can :shipit:

example/app.config.js
j-piasecki: I think that's essentailly `Number(process.env.FABRIC_ENABLED)` but less readable 😅

latekvo: It's a shorthand for `Boolean(Number(process.env.FABRIC_ENABLED))`, without `Boolean()` an error is thrown. 
Previously I wasn't sure if `!!+` was much of a problem, but now I agree the longer form is much more readable.
Fixed in [ae790b3](https://github.com/software-mansion/react-native-gesture-handler/pull/3204/commits/ae790b3673ed7f9016d6b963bbd9857dd606394b)

m-bert (approved):

PR #3207, opened 2024-11-12
Update how Gesture Handler exposes `setGestureState` to the Reanimated UI runtime

j-piasecki:
## Description

Changes how `setGestureState` is exposed to the UI runtime. Instead of the weird conditionally adding Reanimated as a dependency on Android and the weird cast on iOS it uses `_WORKLET_RUNTIME` const injected by Reanimated into the JS runtime. This allows Gesture Handler to decorate the UI runtime without direct dependencies between the libraries.

The new approach relies on two methods being added to the global object:
- `_setGestureStateAsync` on the JS runtime
- `_setGestureStateSync` on the UI runtime

which allows for state manipulation also from the JS thread.

The basic example has been modified to easily test the new functionality.

> [!CAUTION]
> This works only on the New Architecture (and breaks the old one)

## Test plan

Test the expo example app and the modified basic example app


FabricExample/package.json
```diff
@@ -17,7 +17,8 @@
     "postinstall-postinstall": "^2.1.0",
     "react": "19.0.0",
     "react-native": "0.79.0",
-    "react-native-gesture-handler": "link:../"
+    "react-native-gesture-handler": "link:../",
+    "react-native-reanimated": "^3.17.4"
```
tomekzaw, line 21: How about `4.0.0-beta.3` instead? Note that this also requires `react-native-worklets` and `react-native-worklets/plugin` instead of `react-native-reanimated/plugin`.

src/handlers/gestures/gestureStateManager.ts
```diff
@@ -18,51 +18,47 @@ const warningMessage = tagMessage(
...
+declare const global: {
+  _setGestureStateNew: (handlerTag: number, state: State) => void;
+};
+
+const wrappedSetGestureState = (handlerTag: number, state: State) => {
+  'worklet';
+
+  if (REANIMATED_AVAILABLE) {
+    // When Reanimated is available, setGestureState should be defined
+    if (global._setGestureStateNew) {
```
tomekzaw, line 31: Let's use `globalThis` instead of `global`, I believe @tjzel can provide some more insights on that.
```suggestion
    if (globalThis._setGestureState) {
```
Also, I'm not sure if we need to include "new" in the name, how about using `_setGestureState` instead and renaming `setGestureState` to setGestureStateOld` as suggested in another comment?

src/handlers/gestures/gestureStateManager.ts
```diff
@@ -18,51 +18,47 @@ const warningMessage = tagMessage(
 const REANIMATED_AVAILABLE = Reanimated?.useSharedValue !== undefined;
 const setGestureState = Reanimated?.setGestureState;
```
tomekzaw, line 19:
```suggestion
const setGestureStateOld = Reanimated?.setGestureState;
```

src/RNGestureHandlerModule.ts
```diff
@@ -3,3 +3,27 @@
...
 import Module from './specs/NativeRNGestureHandlerModule';
 export default Module;
+
+declare const global: {
+  _WORKLET_RUNTIME: ArrayBuffer;
+};
+
+function tryInstallUIBindings() {
+  if (global._WORKLET_RUNTIME) {
+    console.log('Installing Gesture Handler UI bindings');
```
tomekzaw, line 13: Do we want to keep those console.logs in `.ts` file? Maybe let's convert them to comments.

src/RNGestureHandlerModule.ts
```diff
@@ -3,3 +3,27 @@
...
+    Module.installUIRuntimeBindings();
+  } else {
+    setTimeout(() => {
+      console.log('Waiting for worklet runtime to be available');
+      tryInstallUIBindings();
+    }, 1);
```
tomekzaw, line 19: not good, did this actually happen to you? I believe that `_WORKLET_RUNTIME` is injected synchronously once we import `react-native-reanimated`

apple/RNGestureHandlerModule.mm
```diff
@@ -140,13 +144,43 @@ - (void)setBridge:(RCTBridge *)bridge
...
+              [strongSelf setGestureState:state forHandler:handlerTag];
+            }
+          }
+          return jsi::Value::undefined();
+        });
+
+    uiRuntime->global().setProperty(*uiRuntime, "_setGestureStateNew", std::move(setGestureStateNew));
   };
+
   if (_dispatchToJSThread) {
```
tomekzaw, line 184: Do we still need this? I think we're safe to assume that `RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(installUIRuntimeBindings)` will always be called on the JS queue.

apple/RNGestureHandlerModule.mm
```diff
@@ -140,13 +144,43 @@ - (void)setBridge:(RCTBridge *)bridge
...
-    RCTCxxBridge *cxxBridge = (RCTCxxBridge *)self.bridge;
-    auto runtime = (jsi::Runtime *)cxxBridge.runtime;
-    decorateRuntime(*runtime);
+    auto arrayBufferValue = runtime->global()
+                                .getProperty(*runtime, "_WORKLET_RUNTIME")
+                                .getObject(*runtime)
+                                .getArrayBuffer(*runtime)
+                                .data(*runtime);
+    uintptr_t *uiAddr = reinterpret_cast<uintptr_t *>(&arrayBufferValue[0]);
+    jsi::Runtime *uiRuntime = reinterpret_cast<jsi::Runtime *>(*uiAddr);
```
tomekzaw, line 158:
```suggestion
    jsi::Runtime &uiRuntime = *reinterpret_cast<jsi::Runtime *>(*uiRuntimeAddress);
```

apple/RNGestureHandlerModule.mm
```diff
@@ -35,11 +35,11 @@
 #endif // RCT_NEW_ARCH_ENABLED
 
 #ifdef RCT_NEW_ARCH_ENABLED
-@interface RNGestureHandlerModule () <RNGestureHandlerStateManager, RCTTurboModule>
+@interface RNGestureHandlerModule () <RCTTurboModule, RCTTurboModuleWithJSIBindings>
```
tomekzaw, line 38: `RCTTurboModuleWithJSIBindings`, nice! 👏

android/src/main/jni/CMakeLists.txt
```diff
@@ -11,24 +11,30 @@ endif()
...
   PRIVATE
   "${REACT_NATIVE_DIR}/ReactCommon"
 )
 
 find_package(ReactAndroid REQUIRED CONFIG)
+find_package(fbjni REQUIRED CONFIG)
 if (ReactAndroid_VERSION_MINOR GREATER_EQUAL 76)
   target_link_libraries(
     ${PACKAGE_NAME}
+    fbjni::fbjni
```
tomekzaw, line 37: Why do we need this?

apple/RNGestureHandlerModule.mm
```diff
@@ -140,13 +144,43 @@ - (void)setBridge:(RCTBridge *)bridge
...
+    uintptr_t *uiAddr = reinterpret_cast<uintptr_t *>(&arrayBufferValue[0]);
+    jsi::Runtime *uiRuntime = reinterpret_cast<jsi::Runtime *>(*uiAddr);
+
+    __weak RNGestureHandlerModule *weakSelf = self;
+
+    auto setGestureStateNew = jsi::Function::createFromHostFunction(
+        *uiRuntime,
+        jsi::PropNameID::forAscii(*uiRuntime, "_setGestureStateNew"),
+        2,
+        [weakSelf](jsi::Runtime &runtime, const jsi::Value &thisValue, const jsi::Value *arguments, size_t count)
```
tomekzaw, line 166:
```suggestion
        [weakSelf](jsi::Runtime &runtime, const jsi::Value &, const jsi::Value *args, size_t count)
```

apple/RNGestureHandlerModule.mm
```diff
@@ -140,13 +144,43 @@ - (void)setBridge:(RCTBridge *)bridge
...
+            -> jsi::Value {
+          if (count == 2) {
+            auto handlerTag = (int)arguments[0].asNumber();
+            auto state = (int)arguments[1].asNumber();
```
tomekzaw, line 170:
```suggestion
            const auto handlerTag = static_cast<int>(arguments[0].asNumber());
            const auto state = static_cast<int>(arguments[1].asNumber());
```

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -446,23 +453,18 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
...
-    }
+  private external fun initHybrid(): HybridData
+  private external fun getBindingsInstallerCxx(): BindingsInstallerHolder
+  private external fun decorateUIRuntime()
 
+  override fun getBindingsInstaller() = getBindingsInstallerCxx()
+
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  override fun installUIRuntimeBindings(): Boolean {
+    decorateUIRuntime()
```
tomekzaw, line 464: Can't we just inline `decorateUIRuntime` here?

apple/RNGestureHandlerModule.mm
```diff
@@ -51,6 +51,8 @@ @implementation RNGestureHandlerModule {
 
   // Oparations called after views have been updated.
   NSMutableArray<GestureHandlerOperation> *_operations;
+
+  jsi::Runtime *_runtime;
```
tomekzaw, line 55: Since we're accessing both RN and UI runtimes, I'd suggest using `rnRuntime` and `uiRuntime` instead.
```suggestion
  jsi::Runtime *_rnRuntime;
```

apple/RNGestureHandlerModule.mm
```diff
@@ -91,8 +93,9 @@ - (dispatch_queue_t)methodQueue
 }
 
 #ifdef RCT_NEW_ARCH_ENABLED
-void decorateRuntime(jsi::Runtime &runtime)
+- (void)installJSIBindingsWithRuntime:(jsi::Runtime &)runtime
```
tomekzaw, line 96:
```suggestion
- (void)installJSIBindingsWithRuntime:(jsi::Runtime &)rnRuntime
```

apple/RNGestureHandlerModule.mm
```diff
@@ -140,13 +144,43 @@ - (void)setBridge:(RCTBridge *)bridge
 #endif // RCT_NEW_ARCH_ENABLED
 
 #ifdef RCT_NEW_ARCH_ENABLED
-RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(install)
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(installUIRuntimeBindings)
 {
+  auto runtime = _runtime;
```
tomekzaw, line 149:
```suggestion
  auto &rt = _rnRuntime;
```

android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    void RNGestureHandlerModule::decorateRuntime(jsi::Runtime& runtime) {
+        auto isViewFlatteningDisabled = jsi::Function::createFromHostFunction(
+            runtime,
+            jsi::PropNameID::forAscii(runtime, "isViewFlatteningDisabled"),
+            1,
+            [](jsi::Runtime &runtime,
+               const jsi::Value &thisValue,
+               const jsi::Value *arguments,
```
tomekzaw, line 48:
```suggestion
               const jsi::Value &,
               const jsi::Value *args,
```

android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -0,0 +1,93 @@
...
+               size_t count) -> jsi::Value {
+                if (!arguments[0].isObject()) {
+                    return jsi::Value::null();
+                }
+
+                auto shadowNode = shadowNodeFromValue(runtime, arguments[0]);
+                bool isViewFlatteningDisabled = shadowNode->getTraits().check(ShadowNodeTraits::FormsStackingContext);
```
tomekzaw, line 55:
```suggestion
                const auto shadowNode = shadowNodeFromValue(runtime, arguments[0]);
                const auto isViewFlatteningDisabled = shadowNode->getTraits().check(ShadowNodeTraits::FormsStackingContext);
```

android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -0,0 +1,93 @@
...
+                
+                // This is done using component names instead of type checking because
+                // of duplicate symbols for RN types, which prevent RTTI from working.
+                const char *componentName = shadowNode->getComponentName();
+                bool isTextComponent = strcmp(componentName, "Paragraph") == 0 ||
+                                       strcmp(componentName, "Text") == 0;
```
tomekzaw, line 61:
```suggestion
                const auto componentName = shadowNode->getComponentName();
                const auto isTextComponent = strcmp(componentName, "Paragraph") == 0 ||
                                       strcmp(componentName, "Text") == 0;
```

android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -0,0 +1,93 @@
...
+                return jsi::Value(isViewFlatteningDisabled || isTextComponent);
+            });
+
+        runtime.global().setProperty(
+            runtime, "isViewFlatteningDisabled", std::move(isViewFlatteningDisabled));
+    }
+
+    void RNGestureHandlerModule::decorateUIRuntime() {
+        // TODO: make sure we are on JS?
+        auto arrayBufferValue = runtime->global().getProperty(*runtime, "_WORKLET_RUNTIME").getObject(*runtime).getArrayBuffer(*runtime).data(*runtime);
```
tomekzaw, line 72:
```suggestion
        jsi::Runtime &rt = *runtime;
        auto arrayBufferValue = rt.global().getProperty(rt, "_WORKLET_RUNTIME").getObject(rt).getArrayBuffer(rt).data(rt);
```

android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -0,0 +1,93 @@
...
+        uintptr_t* uiAddr = reinterpret_cast<uintptr_t*>(&arrayBufferValue[0]);
+        jsi::Runtime* uiRuntime = reinterpret_cast<jsi::Runtime*>(*uiAddr);
```
tomekzaw, line 74:
```suggestion
        const auto uiRuntimeAddress = reinterpret_cast<uintptr_t*>(&arrayBufferValue[0]);
        jsi::Runtime &uiRuntime = *reinterpret_cast<jsi::Runtime *>(*uiRuntimeAddress);
```

tomekzaw:
Additionally, I'd like to rename `NativeMethods.h` to `GestureHandler.h` or `SetGestureState.h` in react-native-reanimated, is that okay?

https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/apple/reanimated/apple/native/NativeMethods.h

android/src/main/jni/RNGestureHandlerModule.h
```diff
@@ -0,0 +1,31 @@
...
+
+        jni::global_ref<RNGestureHandlerModule::javaobject> javaPart_;
+        explicit RNGestureHandlerModule(jni::alias_ref<RNGestureHandlerModule::javaobject> jThis);
+        jni::local_ref<BindingsInstallerHolder::javaobject> getBindingsInstallerCxx();
+
+        void setGestureState(int handlerTag, int state);
+        void decorateRuntime(jsi::Runtime& runtime);
+        void decorateUIRuntime();
+    };
+} // namespace newarchtricks
```
tomekzaw, line 31:
```suggestion
} // namespace gesturehandler
```

src/handlers/gestures/gestureStateManager.ts
j-piasecki: I don't think it supports React Native 0.79 yet?

```
In file included from /Users/jakubpiasecki/Projects/gesture-handler/FabricExample/node_modules/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp:1:
In file included from /Users/jakubpiasecki/Projects/gesture-handler/FabricExample/ios/Pods/Headers/Public/React-jsi/jsi/jsi.h:13:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/functional:552:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/__functional/boyer_moore_searcher.h:25:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/array:507:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/algorithm:1846:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/__algorithm/inplace_merge.h:18:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/__algorithm/rotate.h:14:
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/__algorithm/move_backward.h:44:19: error: overload resolution selected deleted operator '='
```

FabricExample/package.json
j-piasecki: I don't think it supports React Native 0.79 yet?

```
In file included from /Users/jakubpiasecki/Projects/gesture-handler/FabricExample/node_modules/react-native-reanimated/Common/cpp/reanimated/NativeModules/ReanimatedModuleProxy.cpp:1:
In file included from /Users/jakubpiasecki/Projects/gesture-handler/FabricExample/ios/Pods/Headers/Public/React-jsi/jsi/jsi.h:13:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/functional:552:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/__functional/boyer_moore_searcher.h:25:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/array:507:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/algorithm:1846:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/__algorithm/inplace_merge.h:18:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/__algorithm/rotate.h:14:
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.4.sdk/usr/include/c++/v1/__algorithm/move_backward.h:44:19: error: overload resolution selected deleted operator '='
```

src/RNGestureHandlerModule.ts
j-piasecki: Nope, but they are useful for development. One of the todos from the PR description is `get rid of the console logs`

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: I guess we could, but the module is a hybrid class and `decorateUIRuntime` is implemented in C++.

j-piasecki: Ideally, I'd like to share code for runtime decoration between platforms, but I really don't want to configure shared c++ code in this PR as well (it's already done in the `display: contents` branch) and resolve conflicts related to that. I'd like to merge those PRs into one branch soon-ish, though.

android/src/main/jni/CMakeLists.txt
j-piasecki: Aren't hybrid classes part of fbjni?

apple/RNGestureHandlerModule.mm
j-piasecki: That sounds like a safe assumption, especially given that Android didn't have a counterpart of that.

src/handlers/gestures/gestureStateManager.ts
j-piasecki: I'd prefer not to change the previous method - it's consumed by RNGH, but it's defined entirely on the Reanimated part. Changing it will very likely break the integration between Reanimated and older versions of RNGH.

And `_setGestureStateNew` is subject to change, though `_setGestureState` will be hard to beat 😅

FabricExample/package.json
tomekzaw: Might be the case indeed

tjzel: Once we have RN0.80 supported we should draft another beta ASAP

android/src/main/jni/CMakeLists.txt
tjzel: They are afaik

android/src/main/jni/RNGestureHandlerModule.h
```diff
@@ -0,0 +1,31 @@
...
+                "Lcom/swmansion/gesturehandler/react/RNGestureHandlerModule;";
+        static jni::local_ref<jhybriddata> initHybrid(jni::alias_ref<jhybridobject> jThis);
+        static void registerNatives();
+        ~RNGestureHandlerModule();
+
+    private:
+        friend HybridBase;
+        jsi::Runtime* rnRuntime = nullptr;
+
+        jni::global_ref<RNGestureHandlerModule::javaobject> javaPart_;
```
tjzel, line 23: You must release `javaPart_` when native modules are invalidated, otherwise you'll create a retain cycle and leak memory.

apple/RNGestureHandlerModule.mm
tjzel: are both necessary?

apple/RNGestureHandlerModule.mm
```diff
@@ -115,7 +118,7 @@ void decorateRuntime(jsi::Runtime &runtime)
 
         return jsi::Value(isViewFlatteningDisabled);
       });
-  runtime.global().setProperty(runtime, "isViewFlatteningDisabled", std::move(isViewFlatteningDisabled));
+  rnRuntime.global().setProperty(rnRuntime, "isViewFlatteningDisabled", std::move(isViewFlatteningDisabled));
```
tjzel, line 121: You might consider adding at least one underscore to prevent potential conflicts.

```suggestion
  rnRuntime.global().setProperty(rnRuntime, "_isViewFlatteningDisabled", std::move(isViewFlatteningDisabled));
```

src/RNGestureHandlerModule.ts
```diff
@@ -3,3 +3,27 @@
 
 import Module from './specs/NativeRNGestureHandlerModule';
 export default Module;
+
+declare const global: {
+  _WORKLET_RUNTIME: ArrayBuffer;
+};
```
tjzel, line 9: I'm not sure, but this probably will pollute the end-user global type.

src/RNGestureHandlerModule.ts
tjzel: Reanimated's `index.ts` doesn't have any side-effects and just requiring it might not trigger the TurboModule to initialize. I think it would be better to require `executeOnUIRuntimeSync`, then invoke it with a HostFunction that will invoke native `tryInstallUIBindings`, to make sure the pipeline is all set.

src/handlers/gestures/gestureStateManager.ts
tjzel: maybe setGestureStateDEPRECATED?

tjzel (commented):
Very nice work overall, I like the removals 😎

apple/RNGestureHandlerModule.mm
```diff
@@ -140,18 +144,36 @@ - (void)setBridge:(RCTBridge *)bridge
...
+          // TODO: expose to JS and dispatch to UI thread if called on JS?
+          RNGestureHandlerModule *strongSelf = weakSelf;
+          if (strongSelf != nullptr) {
+            [strongSelf setGestureState:state forHandler:handlerTag];
+          }
+        }
+        return jsi::Value::undefined();
+      });
+
+  uiRuntime.global().setProperty(uiRuntime, "_setGestureStateNew", std::move(setGestureStateNew));
```
tomekzaw, line 176: Why don't we move this code to the same place where `isViewFlatteningDisabled` is defined?

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Implemented code sharing in this PR. There's no longer a wrapper around the native method.

apple/RNGestureHandlerModule.mm
j-piasecki: Done

src/RNGestureHandlerModule.ts
j-piasecki: Removed entirely.

apple/RNGestureHandlerModule.mm
j-piasecki: It needs to be delayed since there's no guarantee UI runtime will be initialized and injected into the JS one by this point.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -50,6 +58,10 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
 
   @ReactMethod
   override fun createGestureHandler(handlerName: String, handlerTagDouble: Double, config: ReadableMap) {
+    if (!uiRuntimeDecorated) {
+      uiRuntimeDecorated = decorateUIRuntime()
+    }
```
tomekzaw: If, for some reason, `decorateUIRuntime()` always returns false, then we will call `decorateUIRuntime` in each call `createGestureHandler`. How about this instead?
```kt
if (!uiRuntimeDecorated) {
  uiRuntimeDecorated = decorateUIRuntime()
  if (!uiRuntimeDecorated) {
    throw Exception('Failed to decorate UI runtime');
  }
}
```

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
```diff
@@ -9,18 +9,26 @@ else()
 endif()
 
 set(PACKAGE_NAME "gesturehandler")
+set(RNGH_DIR "../../../../")
```
tomekzaw: This is okay but I like to use `"${CMAKE_SOURCE_DIR}../../../../"` in such cases so the paths are always absolute.

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
```diff
@@ -9,18 +9,26 @@ else()
...
 set(REACT_ANDROID_DIR "${REACT_NATIVE_DIR}/ReactAndroid")
 
+file(GLOB_RECURSE gesture_handler_SRCS CONFIGURE_DEPENDS ./*.cpp)
```
tomekzaw: Do we need to wrap the path in double quotes?
```suggestion
file(GLOB_RECURSE gesture_handler_SRCS CONFIGURE_DEPENDS "./*.cpp")
```

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
```diff
@@ -9,18 +9,26 @@ else()
...
   SHARED
-  cpp-adapter.cpp
+  ${gesture_handler_SRCS}
+  ${gesture_handler_shared_SRCS}
 )
 
 target_include_directories(
   ${PACKAGE_NAME}
+  PUBLIC
+  ./
```
tomekzaw: Do we want to wrap it in double quotes? Do we want to make the path absolute?
```suggestion
  "./"
```

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -0,0 +1,57 @@
...
+RNGestureHandlerModule::initHybrid(jni::alias_ref<jhybridobject> jThis) {
+  return makeCxxInstance(jThis);
+}
+
+jni::local_ref<BindingsInstallerHolder::javaobject>
+RNGestureHandlerModule::getBindingsInstallerCxx() {
+  return jni::make_local(
+      BindingsInstallerHolder::newObjectCxxArgs([&](jsi::Runtime &runtime) {
+        this->rnRuntime = &runtime;
+        RuntimeDecorator::installJSRuntimeBindings(
```
tomekzaw: Let's rename `rnRuntime` to `rnRuntime_` and explicitly capture `this`.
```suggestion
      BindingsInstallerHolder::newObjectCxxArgs([this, &](jsi::Runtime &runtime) {
        rnRuntime_ = &runtime;
        RuntimeDecorator::installJSRuntimeBindings(
```

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -0,0 +1,57 @@
...
+            runtime, [&](int handlerTag, int state) {
+              this->setGestureState(handlerTag, state);
```
tomekzaw:
```suggestion
              setGestureState(handlerTag, state);
```

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.h
```diff
@@ -0,0 +1,33 @@
...
+  static auto constexpr kJavaDescriptor =
+      "Lcom/swmansion/gesturehandler/react/RNGestureHandlerModule;";
+  static jni::local_ref<jhybriddata> initHybrid(
+      jni::alias_ref<jhybridobject> jThis);
+  static void registerNatives();
+  ~RNGestureHandlerModule();
+
+ private:
+  friend HybridBase;
+  jsi::Runtime *rnRuntime = nullptr;
```
tomekzaw:
```suggestion
  jsi::Runtime *rnRuntime_ = nullptr;
```

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.h
```diff
@@ -0,0 +1,33 @@
...
+
+  jni::global_ref<RNGestureHandlerModule::javaobject> javaPart_;
+  explicit RNGestureHandlerModule(
+      jni::alias_ref<RNGestureHandlerModule::javaobject> jThis);
+  jni::local_ref<BindingsInstallerHolder::javaobject> getBindingsInstallerCxx();
+
+  void setGestureState(int handlerTag, int state);
```
tomekzaw:
```suggestion
  void setGestureState(const int handlerTag, const int state);
```

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -13,33 +13,33 @@
...
 #import <ReactCommon/CallInvoker.h>
 #import <ReactCommon/RCTTurboModule.h>
-
-#import <react/renderer/components/text/ParagraphShadowNode.h>
-#import <react/renderer/components/text/TextShadowNode.h>
-#import <react/renderer/uimanager/primitives.h>
+#import <ReactCommon/RCTTurboModuleWithJSIBindings.h>
 #endif // RCT_NEW_ARCH_ENABLED
 
+#import "RuntimeDecorator.h"
```
tomekzaw: This can be a bit problematic if there's more than one library that has a header file with the exact same name as CocoaPods puts them all in a single directory and they get overwritten. Given that there are some library authors that like to copy code from our codebase (and that's okay!) without much thinking (and that's not okay), I would suggest using angle-bracket import (if possible) or adding some prefix.

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -91,31 +93,18 @@ - (dispatch_queue_t)methodQueue
...
-        bool isViewFlatteningDisabled = shadowNode->getTraits().check(ShadowNodeTraits::FormsStackingContext);
+  _rnRuntime = &rnRuntime;
+  __weak RNGestureHandlerModule *weakSelf = self;
 
-        return jsi::Value(isViewFlatteningDisabled);
-      });
-  runtime.global().setProperty(runtime, "isViewFlatteningDisabled", std::move(isViewFlatteningDisabled));
+  RuntimeDecorator::installJSRuntimeBindings(rnRuntime, [weakSelf](int handlerTag, int state) {
+    RNGestureHandlerModule *strongSelf = weakSelf;
+    if (strongSelf != nullptr) {
```
tomekzaw:
```suggestion
    if (strongSelf != nil) {
```

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -51,11 +51,13 @@ @implementation RNGestureHandlerModule {
 
   // Oparations called after views have been updated.
   NSMutableArray<GestureHandlerOperation> *_operations;
+
+  facebook::jsi::Runtime *_rnRuntime;
```
tomekzaw:
```suggestion
  jsi::Runtime *_rnRuntime;
```

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -139,29 +129,27 @@ - (void)setBridge:(RCTBridge *)bridge
...
-    _dispatchToJSThread(block);
-  } else {
-    [self.bridge dispatchBlock:block queue:RCTJSThread];
-  }
+  __weak RNGestureHandlerModule *weakSelf = self;
 
-  return @true;
+  return RuntimeDecorator::installUIRuntimeBindings(*_rnRuntime, [weakSelf](int handlerTag, int state) {
+    RNGestureHandlerModule *strongSelf = weakSelf;
+    if (strongSelf != nullptr) {
```
tomekzaw:
```suggestion
    if (strongSelf != nil) {
```

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -139,29 +129,27 @@ - (void)setBridge:(RCTBridge *)bridge
...
-#endif // RCT_NEW_ARCH_ENABLED
 
 RCT_EXPORT_METHOD(createGestureHandler
                   : (nonnull NSString *)handlerName handlerTag
                   : (double)handlerTag config
                   : (NSDictionary *)config)
 {
+  if (!_uiRuntimeDecorated) {
+    _uiRuntimeDecorated = [self installUIRuntimeBindings];
+  }
```
tomekzaw:
```suggestion
  if (!_uiRuntimeDecorated) {
    _uiRuntimeDecorated = [self installUIRuntimeBindings];
    if (!_uiRuntimeDecorated) {
      // throw error
    }
  }
```

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -125,6 +114,7 @@ - (void)initialize
   _manager = [[RNGestureHandlerManager alloc] initWithModuleRegistry:self.moduleRegistry
                                                         viewRegistry:_viewRegistry_DEPRECATED];
   _operations = [NSMutableArray new];
+  _rnRuntime = nullptr;
```
tomekzaw: Do we really need to initialize it?

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -225,6 +213,18 @@ - (void)setBridge:(RCTBridge *)bridge
 }
 
 - (void)setGestureState:(int)state forHandler:(int)handlerTag
+{
+  if (RCTIsMainQueue()) {
+    [self setGestureStateSynchronously:state forHandler:handlerTag];
+  } else {
+    RCTExecuteOnMainQueue(^{
+      RCTAssertMainQueue();
```
tomekzaw: No need for assert I guess as `RCTExecuteOnMainQueue` already guarantees that.

It makes sense to move `RCTAssertMainQueue();` into `setGestureStateSynchronously:forHandler:`, though.

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
+#ifndef ANDROID
+#include <react/renderer/components/text/ParagraphShadowNode.h>
+#include <react/renderer/components/text/TextShadowNode.h>
+#endif
+
+#include <react/renderer/uimanager/primitives.h>
+
+#include "RuntimeDecorator.h"
+
+namespace gesturehandler {
```
tomekzaw: Let's add a newline
```suggestion
namespace gesturehandler {

```

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
+#ifndef ANDROID
+#include <react/renderer/components/text/ParagraphShadowNode.h>
+#include <react/renderer/components/text/TextShadowNode.h>
+#endif
+
+#include <react/renderer/uimanager/primitives.h>
+
+#include "RuntimeDecorator.h"
```
tomekzaw: This can be problematic as well

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+namespace gesturehandler {
+using namespace facebook;
+using namespace facebook::react;
+
+void RuntimeDecorator::installJSRuntimeBindings(
+    jsi::Runtime &rnRuntime,
+    std::function<void(int, int)> &&setGestureState) {
+  auto isViewFlatteningDisabled = jsi::Function::createFromHostFunction(
```
tomekzaw:
```suggestion
  const auto isViewFlatteningDisabled = jsi::Function::createFromHostFunction(
```

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+#ifndef ANDROID
+        if (dynamic_pointer_cast<const ParagraphShadowNode>(shadowNode)) {
+          return jsi::Value(true);
+        }
+
+        if (dynamic_pointer_cast<const TextShadowNode>(shadowNode)) {
+          return jsi::Value(true);
+        }
+#endif
```
tomekzaw: Why do we need those?

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+        const auto isViewFlatteningDisabled = shadowNode->getTraits().check(
```
tomekzaw: The identifier shadows the name from outer scope (with a completely different meaning)
```suggestion
        const auto isFormsStackingContext = shadowNode->getTraits().check(
```

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+            ShadowNodeTraits::FormsStackingContext);
+
+        // This is done using component names instead of type checking because
+        // of duplicate symbols for RN types, which prevent RTTI from working.
+        const auto componentName = shadowNode->getComponentName();
+        const auto isTextComponent = strcmp(componentName, "Paragraph") == 0 ||
+            strcmp(componentName, "Text") == 0;
+
+        return jsi::Value(isViewFlatteningDisabled || isTextComponent);
```
tomekzaw:
```suggestion
        return jsi::Value(isFormsStackingContext || isTextOrParagraphComponent);
```

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+          return jsi::Value(true);
+        }
+#endif
+
+        const auto isViewFlatteningDisabled = shadowNode->getTraits().check(
+            ShadowNodeTraits::FormsStackingContext);
+
+        // This is done using component names instead of type checking because
+        // of duplicate symbols for RN types, which prevent RTTI from working.
+        const auto componentName = shadowNode->getComponentName();
```
tomekzaw:
```suggestion
        const auto &componentName = shadowNode->getComponentName();
```

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+        const auto isTextComponent = strcmp(componentName, "Paragraph") == 0 ||
```
tomekzaw:
```suggestion
        const auto isTextOrParagraphComponent = strcmp(componentName, "Paragraph") == 0 ||
```

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+  auto setGestureStateAsync = jsi::Function::createFromHostFunction(
+      rnRuntime,
+      jsi::PropNameID::forAscii(rnRuntime, "_setGestureStateAsync"),
+      2,
+      [setGestureState](
+          jsi::Runtime &rt,
+          const jsi::Value &,
+          const jsi::Value *args,
+          size_t count) -> jsi::Value {
+        if (count == 2) {
```
tomekzaw: smart, we should do that more often

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+          const auto handlerTag = static_cast<int>(args[0].asNumber());
+          const auto state = static_cast<int>(args[1].asNumber());
+
+          setGestureState(handlerTag, state);
+        }
+        return jsi::Value::undefined();
+      });
+
+  uiRuntime.global().setProperty(
+      uiRuntime, "_setGestureStateSync", std::move(setGestureStateSync));
```
tomekzaw: Let's add a newline
```suggestion
      uiRuntime, "_setGestureStateSync", std::move(setGestureStateSync));

```

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
```diff
@@ -0,0 +1,120 @@
...
+  return true;
+}
```
tomekzaw:
```suggestion
}

```

packages/react-native-gesture-handler/shared/RuntimeDecorator.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+#include <jsi/jsi.h>
+
+namespace gesturehandler {
```
tomekzaw:
```suggestion
namespace gesturehandler {

```

packages/react-native-gesture-handler/shared/RuntimeDecorator.h
```diff
@@ -0,0 +1,16 @@
...
+
+class RuntimeDecorator {
+ public:
+  static void installJSRuntimeBindings(
+      jsi::Runtime &rnRuntime,
+      std::function<void(int, int)> &&setGestureState);
+  static bool installUIRuntimeBindings(
+      jsi::Runtime &rnRuntime,
+      std::function<void(int, int)> &&setGestureState);
+};
```
tomekzaw:
```suggestion
};

```

packages/react-native-gesture-handler/shared/RuntimeDecorator.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+#include <jsi/jsi.h>
+
+namespace gesturehandler {
+using namespace facebook;
+
+class RuntimeDecorator {
+ public:
+  static void installJSRuntimeBindings(
```
tomekzaw: I would suggest using the term "RN runtime" instead of "JS runtime".
```suggestion
  static void installRNRuntimeBindings(
```

packages/react-native-gesture-handler/src/handlers/gestures/gestureStateManager.ts
```diff
@@ -9,60 +7,43 @@ export interface GestureStateManagerType {
...
 
-// Check if reanimated module is available, but look for useSharedValue as conditional
-// require of reanimated can sometimes return content of `utils.ts` file (?)
-const REANIMATED_AVAILABLE = Reanimated?.useSharedValue !== undefined;
-const setGestureState = Reanimated?.setGestureState;
+  if (globalThis._setGestureStateSync) {
+    globalThis._setGestureStateSync(handlerTag, state);
+  } else if (globalThis._setGestureStateAsync) {
+    globalThis._setGestureStateAsync(handlerTag, state);
+  }
```
tomekzaw, line 26: This will fail silently if things go wrong, perhaps we'd like to throw some error instead?

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
j-piasecki: I'd rather rely on dynamic casts to check shadow node types but it's not working on Android, so we do that based on component names there. Hopefully, we'll be able to consistently rely on dynamic casts in the future.

packages/react-native-gesture-handler/src/handlers/gestures/gestureStateManager.ts
j-piasecki: Added in [5238604](https://github.com/software-mansion/react-native-gesture-handler/pull/3207/commits/52386048d0a7d4179698273fc2cd95eacad3a66e)

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
j-piasecki: Went with `"${CMAKE_SOURCE_DIR}"`

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: That's a good catch - it would be called every time with reanimated not installed. I updated it to only call when Reanimated is available and throw on fail. WDYT?

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
j-piasecki: Same as Android

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
j-piasecki: Renamed to `RNGHRuntimeDecorator.h`

android/src/main/jni/RNGestureHandlerModule.h
j-piasecki: Done - [15345ab](https://github.com/software-mansion/react-native-gesture-handler/pull/3207/commits/15345ab03bab88c4418b4f4e517d43eb5a345fb5)

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
tomekzaw: Let's add a comment in the code then

packages/react-native-gesture-handler/src/handlers/gestures/gestureStateManager.ts
tomekzaw: Looks like `tagMessage` is not marked as worklet so this will fail with a different error message than expected

packages/react-native-gesture-handler/shared/RuntimeDecorator.cpp
j-piasecki: There is one, slightly below where the string comparison is done: https://github.com/software-mansion/react-native-gesture-handler/pull/3207/files/d5d7233cbda9587a1c2539b41bc4707932563c7f#diff-c4fc5329eaf1253ca734307193396657554a35aa0827ee46113a982f28065aceR44-R45

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Ok, I think we need to stick to trying to decorate the runtime every time a gesture is created. `_WORKLET_RUNTIME` isn't injected by Reanimated unless an `Animated.View` is rendered (possibly other circumstances as well), in which case we would throw.

I think it's likely to have a situation where on the first render there are no `Animated.Views` rendered at all, but there's a simple gesture to handle tap or a long press, and animations are rendered conditionally. That's a valid use case that would be failing in this scenario.

Is calling `decorateUIRuntime` in every `createGestureHandler` (if Reanimated is installed) that bad? Realistically, I'd assume we'd call it only a few times and it should be quick given `_WORKLET_RUNTIME` doesn't exist.

scripts/format-cpp.js
m-bert, line 1: Do we need this? We have `format-apple` script, the only difference is path passed to `glob`. Maybe it would be better to rename this script and pass path as argument?

.lintstagedrc.json
```diff
@@ -6,5 +6,6 @@
     "yarn format:android"
   ],
   "packages/react-native-gesture-handler/apple/**/*.{h,m,mm,cpp}": "yarn format:apple",
+  "packages/react-native-gesture-handler/{shared,android/src}/**/*.{h,cpp}": "yarn format:cpp",
```
m-bert, line 9: See the comment to `format-cpp.js`

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -1,15 +1,19 @@
...
 import com.facebook.react.module.annotations.ReactModule
+import com.facebook.react.turbomodule.core.interfaces.BindingsInstallerHolder
+import com.facebook.react.turbomodule.core.interfaces.TurboModuleWithJSIBindings
 import com.facebook.soloader.SoLoader
-import com.swmansion.common.GestureHandlerStateManager
 import com.swmansion.gesturehandler.NativeRNGestureHandlerModuleSpec
+import com.swmansion.gesturehandler.ReanimatedProxy
 import com.swmansion.gesturehandler.core.GestureHandler
 
 // UIManagerModule.resolveRootTagFromReactTag() was deprecated and will be removed in the next RN release
```
m-bert, line 19: Is it still relevant?

packages/react-native-gesture-handler/android/paper/src/main/java/com/swmansion/gesturehandler/NativeRNGestureHandlerModuleSpec.java
```diff
@@ -41,9 +41,9 @@ public NativeRNGestureHandlerModuleSpec(ReactApplicationContext reactContext) {
   @DoNotStrip
   public abstract void handleClearJSResponder();
 
-  @ReactMethod
+  @ReactMethod(isBlockingSynchronousMethod = true)
   @DoNotStrip
-  public abstract void createGestureHandler(String handlerName, double handlerTag, ReadableMap config);
+  public abstract boolean createGestureHandler(String handlerName, double handlerTag, ReadableMap config);
```
m-bert, line 46: Could we add a comment about why it was changed to `bool`? 😅

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -0,0 +1,68 @@
...
+    const int handlerTag,
+    const int state) {
+  static const auto method =
+      javaClassLocal()->getMethod<void(jint, jint)>("setGestureHandlerState");
+
+  method(this->javaPart_, handlerTag, state);
+}
+
+bool RNGestureHandlerModule::decorateUIRuntime() {
+  // TODO: make sure we are on JS?
```
m-bert: Is this still relevant?

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -51,11 +51,16 @@ @implementation RNGestureHandlerModule {
 
   // Oparations called after views have been updated.
   NSMutableArray<GestureHandlerOperation> *_operations;
+
+  jsi::Runtime *_rnRuntime;
+
+  bool _checkedReanimated;
+  bool _reanimatedAvailable;
```
m-bert: I think those 2 names may be quite confusing

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -91,31 +96,18 @@ - (dispatch_queue_t)methodQueue
 }
 
 #ifdef RCT_NEW_ARCH_ENABLED
-void decorateRuntime(jsi::Runtime &runtime)
+- (void)installJSIBindingsWithRuntime:(jsi::Runtime &)rnRuntime
+                          callInvoker:(const std::shared_ptr<facebook::react::CallInvoker> &)callinvoker
```
m-bert, line 100: I'm not sure about obj-c convention, but wouldn't it be `withCallInvoker: callInvoker`?

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -139,35 +131,36 @@ - (void)setBridge:(RCTBridge *)bridge
...
+    }
+  });
 }
-#endif // RCT_NEW_ARCH_ENABLED
 
-RCT_EXPORT_METHOD(createGestureHandler
-                  : (nonnull NSString *)handlerName handlerTag
-                  : (double)handlerTag config
-                  : (NSDictionary *)config)
+- (NSNumber *)createGestureHandler:(NSString *)handlerName handlerTag:(double)handlerTag config:(NSDictionary *)config
```
m-bert, line 146: Why do we have `bool` in `cpp` and `NSNumber` in `obj-c`?

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -226,6 +219,18 @@ - (void)setBridge:(RCTBridge *)bridge
...
+  } else {
+    RCTExecuteOnMainQueue(^{
+      [self setGestureStateSynchronously:state forHandler:handlerTag];
+    });
+  }
+}
+
+- (void)setGestureStateSynchronously:(int)state forHandler:(int)handlerTag
+{
+  RCTAssertMainQueue();
```
m-bert, line 233: Based on the `if` statements above, is it possible that we are not on the main queue?

packages/react-native-gesture-handler/package.json
```diff
@@ -10,6 +10,7 @@
     "format:js": "yarn format-js",
     "format:android": "node ../../scripts/format-android.js",
     "format:apple": "node ../../scripts/format-apple.js",
+    "format:cpp": "node ../../scripts/format-cpp.js",
```
m-bert: Same here, look at comment for `format-cpp.js`

packages/react-native-gesture-handler/src/handlers/gestures/gestureStateManager.ts
```diff
@@ -9,60 +8,45 @@ export interface GestureStateManagerType {
   end: () => void;
 }
 
-const warningMessage = tagMessage(
-  'react-native-reanimated is required in order to use synchronous state management'
-);
+// declare methods to keep the TS happy
```
m-bert: I've just seen a comment in diff above and we may stick to one convention 😅 

```suggestion
// Declare methods to keep the TS happy.
```

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
```diff
@@ -0,0 +1,124 @@
...
+  auto setGestureStateSync = jsi::Function::createFromHostFunction(
+      uiRuntime,
+      jsi::PropNameID::forAscii(uiRuntime, "_setGestureStateSync"),
+      2,
+      [setGestureState](
+          jsi::Runtime &rt,
+          const jsi::Value &,
+          const jsi::Value *args,
+          size_t count) -> jsi::Value {
+        if (count == 2) {
```
m-bert: Why is there magic `2`? Can we name it? 😅

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
j-piasecki: Not my call https://github.com/facebook/react-native/blob/f4a9aa352544c9b9387ba3b74a385341cba81414/packages/react-native/ReactCommon/react/nativemodule/core/platform/ios/ReactCommon/RCTTurboModuleWithJSIBindings.h#L20 🤷

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
j-piasecki: This is what codegen generates

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
j-piasecki: I hope not, this assert is debug-only so we will know if there's something we missed

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
j-piasecki: Number of arguments - handlerTag and state

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Yes, it's used to check if reanimated is installed at all before trying to decorate UI runtime

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -104,7 +119,21 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
...
+    if (UiThreadUtil.isOnUiThread()) {
+      setGestureStateSync(handlerTag, newState)
+    } else {
+      UiThreadUtil.runOnUiThread {
+        setGestureStateSync(handlerTag, newState)
+      }
+    }
+  }
+
+  private fun setGestureStateSync(handlerTag: Int, newState: Int) {
```
latekvo, line 134: We're using `setGestureStateSync` in Kotlin, and `setGestureStateSynchronously` in `obj-c`.
This is a nitpick, but I think it'd make sense to stick to a single convention, unless there's a reason for this difference.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -49,10 +58,16 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
   }
 
   @ReactMethod
-  override fun createGestureHandler(handlerName: String, handlerTagDouble: Double, config: ReadableMap) {
+  override fun createGestureHandler(handlerName: String, handlerTagDouble: Double, config: ReadableMap): Boolean {
```
latekvo, line 61: The `createGestureHandler` seems to only ever return `true/1`. 
Is this the case, or am I missing something?
If it is the case, can we get away with returning `void` instead?

j-piasecki: When a method returns **any value** in the codegen spec, the generated method is synchronous. I don't think there's a way to force codegen to generate a synchronous method that returns void.

scripts/format-cpp.js
j-piasecki: Unified the scripts in [da123f7](https://github.com/software-mansion/react-native-gesture-handler/pull/3207/commits/da123f7e32b80ba942e77fd1e40909fbb4a52911). Let me know what you think.

.lintstagedrc.json
j-piasecki: Replied above

packages/react-native-gesture-handler/package.json
j-piasecki: Replied above

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: Renamed in [77ac470](https://github.com/software-mansion/react-native-gesture-handler/pull/3207/commits/77ac470f4895dca66561837eeab246a943f7523c).

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
j-piasecki: Renamed in [4d51da1](https://github.com/software-mansion/react-native-gesture-handler/pull/3207/commits/4d51da1d97b5a9ed142327a459da25a5f04553c4)

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
j-piasecki: I don't think so, removed

packages/react-native-gesture-handler/android/paper/src/main/java/com/swmansion/gesturehandler/NativeRNGestureHandlerModuleSpec.java
j-piasecki: Added in [d4606a0](https://github.com/software-mansion/react-native-gesture-handler/pull/3207/commits/d4606a07654c0c414fd032891a76f5918538015c), though I added it in the spec which dictates the shape of native modules.

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  auto setGestureStateSync = jsi::Function::createFromHostFunction(
+      uiRuntime,
+      jsi::PropNameID::forAscii(uiRuntime, "_setGestureStateSync"),
+      2,
+      [setGestureState](
+          jsi::Runtime &rt,
+          const jsi::Value &,
+          const jsi::Value *args,
+          size_t count) -> jsi::Value {
```
m-bert: It would be great if we could change it to more descriptive name 😅

m-bert (approved):
LGTM (let's see what @tomekzaw thinks) 🎉

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  rnRuntime.global().setProperty(
+      rnRuntime, "_setGestureStateAsync", std::move(setGestureStateAsync));
+}
+
+bool RNGHRuntimeDecorator::installUIRuntimeBindings(
+    jsi::Runtime &rnRuntime,
+    std::function<void(int, int)> &&setGestureState) {
+  const auto runtimeHolder =
+      rnRuntime.global().getProperty(rnRuntime, "_WORKLET_RUNTIME");
```
WoLewicki, line 86: Do I get it correctly that this is added to rnRuntime by Reanimated? From which version of lib is it there? Also, can't we have some race conditions that it is not yet there when you call this method or something like that? Btw, is it safe to have some kind of pointer from one runtime to another?

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
```diff
@@ -0,0 +1,124 @@
...
+  if (runtimeHolder.isUndefined()) {
+    return false;
+  }
+
+  const auto arrayBufferValue =
+      runtimeHolder.getObject(rnRuntime).getArrayBuffer(rnRuntime).data(
+          rnRuntime);
+  const auto uiRuntimeAddress =
+      reinterpret_cast<uintptr_t *>(&arrayBufferValue[0]);
+  jsi::Runtime &uiRuntime =
```
WoLewicki, line 97: Oh ok, so the above is just a bool specifying that there will be a buffer with a pointer to memory with uiRuntime allocated there?

WoLewicki (commented):
Some comments

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
j-piasecki: > Do I get it correctly that this is added to rnRuntime by Reanimated

Yes

> From which version of lib is it there

Pretty sure it was added in 2.x

> Also, can't we have some race conditions that it is not yet there when you call this method or something like that

Yes - when you don't render an `Animated.View`, the Reanimated native module will not be initialized, and this field will not exist. This is handled on the JS side which will throw an exception about this.

> is it safe to have some kind of pointer from one runtime to another

I believe so

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
j-piasecki: I'm not sure what you mean by "the above".

`runtimeHolder` is a jsi value holding an array buffer. `arrayBufferValue` is the data held inside the buffer (the 8 bytes of the pointer to the ui runtime). `uiRuntimeAddress` is a pointer to the first byte of the buffer value, casted to `uintptr_t` which is an integer the size of a pointer. `uiRuntime` is a reference to the ui runtime which we get by casting the `uiRuntimeAddress` to `jsi::Runtime *` and dereferencing it.

WoLewicki: Ok got it, thanks

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
tomekzaw: Just for the context, `_WORKLET_RUNTIME` was added a few years ago to support integration of Reanimated with Expo GL. The latter could be used from the Reanimated UI runtime so it would inject some JSI bindings into the Reanimated UI runtime.

packages/react-native-gesture-handler/shared/RNGHRuntimeDecorator.cpp
tomekzaw: For the context, this is because we can't represent a 64-bit pointer as a `number` so we need to use `ArrayBuffer` for that.

PR #3218, opened 2024-11-16
fix: missing ReanimatedDrawerLayout

r0b0t3d:
## Description

I'm getting error when importing `react-native-gesture-handler/ReanimatedDrawerLayout` because the folder is missing

## Test plan

Add the folder `ReanimatedDrawerLayout` on my local in `node_modules` and it works


m-bert (approved):
Thank you so much for spotting this! ❤️

PR #3222, opened 2024-11-18
Fix the wall of errors when running `lint:js` locally

latekvo:
## Description

This PR fixes the errors which occur when running `lint:js` locally.

Unfortunately, the version incompatibility warnings cannot be fixed without downgrading `eslint` due to  `eslint-config-satya164` dependency which indirectly hard-requires `eslint@^7.1.0`

## Test plan

- run `yarn lint:js` locally
- see how it throws multiple errors on `main` branch, and none on this PR


latekvo:
After further investigation, this turned out to be a local issue.
The `main` branch is not affected.

Issue #3184, opened 2024-10-24
iOS crash on react-native 0.76 with new arch and react StrictMode


0xDing:
### Description

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Termination Reason: SIGNAL 6 Abort trap: 6
Terminating Process: hopi [1353]

Triggered by Thread:  0

Last Exception Backtrace:
0   CoreFoundation                	       0x1804b70e0 __exceptionPreprocess + 160
1   libobjc.A.dylib               	       0x18008ede8 objc_exception_throw + 72
2   Foundation                    	       0x180e73c40 -[NSMutableDictionary(NSMutableDictionary) classForCoder] + 0
3   hopi.debug.dylib              	       0x10e1eb6f0 -[RNGestureHandlerRegistry attachHandlerWithTag:toView:withActionType:] + 300
4   hopi.debug.dylib              	       0x10e1e3e48 -[RNGestureHandlerManager attachGestureHandler:toViewWithTag:withActionType:] + 1184
5   hopi.debug.dylib              	       0x10e1e3f0c __77-[RNGestureHandlerManager attachGestureHandler:toViewWithTag:withActionType:]_block_invoke + 52
6   libdispatch.dylib             	       0x180178de0 _dispatch_client_callout + 16
7   libdispatch.dylib             	       0x18017c780 _dispatch_continuation_pop + 804
8   libdispatch.dylib             	       0x18018fde0 _dispatch_source_invoke + 1620
9   libdispatch.dylib             	       0x180187ad8 _dispatch_main_queue_drain + 880
10  libdispatch.dylib             	       0x180187758 _dispatch_main_queue_callback_4CF + 40
11  CoreFoundation                	       0x18041ae3c __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 12
12  CoreFoundation                	       0x180415534 __CFRunLoopRun + 1944
13  CoreFoundation                	       0x180414960 CFRunLoopRunSpecific + 536
14  GraphicsServices              	       0x190183b10 GSEventRunModal + 160
15  UIKitCore                     	       0x185aa2b40 -[UIApplication _run] + 796
16  UIKitCore                     	       0x185aa6d38 UIApplicationMain + 124
17  hopi.debug.dylib              	       0x10cdf7bcc __debug_main_executable_dylib_entry_point + 112 (main.swift:11)
[...]

PR #3224, opened 2024-11-19
Fix : RN 0.76 attachHandlerWithTag Crash

ChanghyeonYoon:
## Description

I don't know the exact reason for the absence of the handler, but it should work fine if you change it to that code.
I think it's probably a flow issue when using React Native 0.76 New Architecture with React Navigation v7.
Anyway, you can change the code to the following and it should work fine.

### DIFF
--- a/apple/RNGestureHandlerRegistry.m
+++ b/apple/RNGestureHandlerRegistry.m
@@ -37,7 +37,10 @@
               withActionType:(RNGestureHandlerActionType)actionType
 {
   RNGestureHandler *handler = _handlers[handlerTag];
-  RCTAssert(handler != nil, @"Handler for tag %@ does not exists", handlerTag);
+  if (handler == nil) {
+    NSLog(@"Handler for tag %@ does not exist", handlerTag);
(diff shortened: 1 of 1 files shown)


miridih-chyoon:
Related Issue #3184

ChanghyeonYoon:
Related Issue #3184

m-bert:
Hi @ChanghyeonYoon! First of all, thank you for submitting this PR. Unfortunately, this is not a solution that we want to implement, so I have to close it.

> I don't know the exact reason for the absence of the handler

The fact that handler doesn't exist is connected to **_StrictMode_**. The thing is, if you enable _**StrictMode**_ in your application, `renders` and `effects` [will be called twice](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-effects-in-development). For some reason (we are yet to find out why), order of operations on iOS when new architecture is enabled is different than on the old one. In other words, handler is dropped, and then checked if it exists. This will fail, but for now we don't know what causes this difference.

PR #3228, opened 2024-11-19
Update RNGestureHandlerPackage to supported stable API

jakex7:
## Summary

Replace the deprecated `TurboReactPackage` with the newer `BaseReactPackage` to improve compatibility with future React Native releases. Note that this change requires at least react-native@0.74+, so I've added sourceSet to maintain compatibility with `react-native@0.73.x` (similar to https://github.com/software-mansion/react-native-svg/pull/2543).

Similar change has been introduced in the SVG package https://github.com/software-mansion/react-native-svg/pull/2541

## Test plan

CI should pass


android/build.gradle
```diff
@@ -173,6 +173,14 @@ android {
...
                 srcDirs += 'paper/src/main/java'
             }
+
+            if (REACT_NATIVE_MINOR_VERSION >= 74) { // new Package API https://github.com/software-mansion/react-native-svg/pull/2541
+                // use for react-native@0.74 and above
+                srcDirs += 'package74/src/main/java'
+            } else {
+                // maintain compatibility with react-native@0.73
+                srcDirs += 'package73/src/main/java'
+            }
```
j-piasecki, line 183: I don't think this is necessary. We officially support the last three versions of React Native and this makes change will make the package harder to maintain.

android/package74/src/main/java/com/swmansion/gesturehandler/RNGestureHandlerPackage.kt
```diff
@@ -0,0 +1,86 @@
...
+      return ReactModuleInfoProvider {
+        val reactModule: ReactModule = RNGestureHandlerModule::class.java.getAnnotation(ReactModule::class.java)!!
+
+        mutableMapOf(
+          RNGestureHandlerModule.NAME to ReactModuleInfo(
+            reactModule.name,
+            RNGestureHandlerModule::class.java.name,
+            reactModule.canOverrideExistingModule,
+            reactModule.needsEagerInit,
+            true, // Has constants is hardcoded to return true, so replacing it with `true` changes nothing.
```
j-piasecki, line 74: Shouldn't this be removed to align with the SVG PR?

jakex7: You're right, `ReactModuleInfo` also has a new constructor. Fixed in https://github.com/software-mansion/react-native-gesture-handler/pull/3228/commits/20a4dc04b566c2cc5a7ba6a0970f7f984512f9d5

j-piasecki (approved):

tomekzaw (approved):

latekvo (approved):

j-piasecki (approved):

RonRadtke:
Just since I stumbled over it after releasing it: https://github.com/react-native-community/cli/pull/2563

@jakex7 Right now the fixed versions of the cli are not shipped yet with new RN packages, and old packages without updated cli are broken too....

jakex7:
Thanks for the info @RonRadtke! Fortunately RNGH should be safe as it falls into first check which is less strict and has been there for a long time
https://github.com/react-native-community/cli/blob/feba14cf36de2fdefa41c7052841a5ee34c19084/packages/cli-config-android/src/config/findPackageClassName.ts#L71

FYI @j-piasecki

PR #3229, opened 2024-11-19
Remove all `Hammer.JS` usages & references 

latekvo:
## Description

Removed legacy `Hammer.JS` web implementation, all it's usages and references.
Removed `enableLegacyWebImplementation`.
Removed `isNewWebImplementationEnabled`. It's not abundantly clear at first, but it's an internal function.


src/EnableNewWebImplementation.ts
m-bert, line 1: Do we need this file? If this PR is meant to be part of Gesture Handler 3, then I think we can simply remove those functions as they don't do anything anymore.

Even if someone has them in their codebase, the only thing they have to change is to remove one line from one file (assuming they didn't mess anything in their setup) - not a big deal.

latekvo: I added an error message to it, because it wasn't marked as deprecated in `RNGH 2` and I'd prefer to avoid sudden API changes without warning.

But it's not a strong preference - we can remove it if you think it's better to do that instead of giving a warn that `enableLegacyWebImplementation` isn't functional anymore

m-bert: Yes, I'd prefer to remove it. It is pointless to call these functions at this point. Other than that, one has to read the error and adapt to the changes. If the error will be thrown by missing function it will be spotted earlier.  

> I'd prefer to avoid sudden API changes without warning.

It is not a sudden API change. This will be difference between major versions, not minors.

> it wasn't marked as deprecated in RNGH 2

Gesture Handler 2 is still here, you can mark it as deprecated 😉

m-bert (approved):

PR #3230, opened 2024-11-20
Remove deprecated components

latekvo:
## Description

removes the following components:
-
-
-

These components were also removed from all files referencing them, 
such as examples, index files, and the `App.tsx` of the common example app

## Test plan

- see how these components are no longer accessible
- see how there are no errors, and no new warnings throw as compared to the `main` branch


latekvo:
>there's also still documentation for Swipeable component

My bad, I forgot these changes are merged into the `next` branch, not `main`, so they don't trigger the docs publication workflow.

fixed in  [ad0ff9c](https://github.com/software-mansion/react-native-gesture-handler/pull/3230/commits/ad0ff9cdc2fc6abadfff01f537e8961d8e8ae1a8)

>there are still references to the removed components (like in babel.config.js)

Not sure which references you're talking about, in the `babel.config.js` file there are only references to the `ReanimatedSwipeable` and `ReanimatedDrawerLayout` components.



Although I did find another draft for `Swipeable` which hasn't been removed yet.
If you were talking about that, fixed in [46a4c4b](https://github.com/software-mansion/react-native-gesture-handler/pull/3230/commits/46a4c4b51398af3418e34e4bd203616c1747c019)

example/src/release_tests/combo/index.tsx
```diff
@@ -174,41 +176,14 @@ class Combo extends Component<ComboProps> {
...
-              style={styles.rectButton}
-              onPress={() => Alert.alert('Third row clicked')}>
-              <Text style={styles.buttonText}>
-                This one will cancel when you drag outside
-              </Text>
-              {/* Info icon will cancel when you drag your finger outside of its bounds and
-                  then back unlike all the previous icons that would activate when you re-enter
-                  their activation area. This is a typical bahaviour for android but less frequent
-                  for most of the iOS native apps. */}
-              <InfoButton shouldCancelWhenOutside name="third" />
```
j-piasecki, line 211: Would it be possible to reimplement this using the new components? I'd like to keep something along these lines to make sure this behavior doesn't change accidentally.

example/src/release_tests/swipeableReanimation/index.tsx
```diff
@@ -48,55 +47,8 @@ function RightAction(prog: SharedValue<number>, drag: SharedValue<number>) {
...
-        {
-          transform: [{ translateX: trans }],
-        },
-      ]}>
-      Text
-    </Animated.Text>
-  );
-}
-
 export default function Example() {
```
j-piasecki, line 50: If this is the yellowish looking one then I guess we can remove it entirely, no? It was supposed to compare both implementations visually and with one implementation gone it kinda lost its purpose.

latekvo: Makes sense, although right now it serves more as a test for the programatic swipeable controls, which encountered regressions several times since release.

Screenshot 2024-12-06 at 14 35 22

latekvo: done in [5599019](https://github.com/software-mansion/react-native-gesture-handler/pull/3230/commits/55990199c652625937884e15543377fafcc95df0)

example/src/release_tests/combo/index.tsx
latekvo: done in [0a87e5e](https://github.com/software-mansion/react-native-gesture-handler/pull/3230/commits/0a87e5e1986b3182865548bf903bf00ea52e3f21)

j-piasecki (approved):
Make sure CI passes before merging.

latekvo:
To fix the `touchables` tslint error, I had to use a `ts-ignore` due to lack of other options.

There's been a recession added to `TouchableNativeFeedback` in: #3260
I'll fix that issue separately, but for now i think it's out of the scope of this PR.

If that isn't an issue, i'll be merging this PR.

m-bert:
Closing in favor of #3734. Hopefully you don't mind 😅

PR #3244, opened 2024-11-28
[Android] Change `onTouchEvent` to `dispatchTouchEvent` in `NativeViewGestureHandler`

### LINKED ISSUE #2454: Panning gesture cannot working with Webview component
### Description

Unable to detect panning gestures for the Webview(`react-native-webview`) component.
Only onFinalize event is triggered in vertical panning.
Horizontal panning works normally.
Example:
```
  const gesture = Gesture.Pan()
    .onStart(e => {
      console.log('onStart');
    })
    .onUpdate(e => {
      console.log(`onUpdate`);
    })
    .onEnd(() => {
      console.log('onEnd');
    })
    .onFinalize(e => {
      console.log('onFinalize');
    });
  return (
        <View style={{ flex: 1 }}>
          {/* For webview, only onFinalize event is triggered in vertical panning, horizontal panning works normally.  */}
          <WebView source={{ uri: 'https://github.com/software-mansion/react-native-gesture-handler' }} />
          {/* For view, all normal  */}
          <View style={{ height: 200 }}></View>
        </View>
      </GestureDetector>
    </View>
  );
}
```

### Steps to reproduce

Above code.

https://snack.expo.dev/XvDJr8BQq

### Gesture Handler version

2.8.0

### React Native version

0.70.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

### LINKED ISSUE #3196: GestureHandlerRootView interferes with text selection inside webview during Android back gesture
### Description

When doing a back gesture in Android - the `GestureHandlerRootView` causes bugs interacting with the underlying Webview and they do not seem to be avoidable and are without any workaround. This makes things feel beyond janky.

### Steps to reproduce

1. Use [provided repro Expo Snack](https://snack.expo.dev/UcOqm5_29su_NwaSIrg0r)
2. Do back gestures over the text, but don't release the gesture. You'll notice text is selected or the Android text zoom bubble appears (and may persist)
3. Comment out the `BuggyApp` and enable the alternative version `WorkingApp` to see the intended behavior (text selection may change still, but will not set expanded text selections or cause text zoom bubble to appear)

[Snack repro](https://snack.expo.dev/UcOqm5_29su_NwaSIrg0r)

### Gesture Handler version

2.20.2

### React Native version

0.72.x through 0.76.x

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

Paper (Old Architecture)

### Build type

None

### Device

Real device

### Device model

Galaxy S24 (Android 14) -- Reproducible on all Android devices using gestures

### Acknowledgements

Yes

m-bert:
## Description

Currently wrapping `WebView` into `NativeViewGestureHandler`  doesn't work as expected - events from `native` gesture do not reach `WebView`. This happens because inside `NativeViewGestureHandler` we call `onTouchEvent` method. In native hierarchy `WebView` is nested inside `ViewGroup`, which means that calling `onTouchEvent` instead of `dispatchTouchEvent` won't do anything to the `WebView`.

Should fix #2454 

Fixes #3196 

## Test plan

Tested on example app and the code below:

<details>
<summary>Test code</summary>

```jsx
import { WebView } from 'react-native-webview';
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,

function WebViewScreen() {
  const native = Gesture.Native()
    .shouldActivateOnStart(true)
    .disallowInterruption(true);
  const pan = Gesture.Pan().onChange(console.log);

  const g = Gesture.Simultaneous(native, pan);

  return (
    <View style={styles.webViewContainer}>
      <GestureDetector gesture={g}>
        <WebView
          source={{ uri: 'https://templates.tiptap.dev/nw6Cmz6HfD' }}
          style={styles.webView}
          javaScriptEnabled={true}
          domStorageEnabled={true}
          startInLoadingState={true}
          onError={(syntheticEvent) => {
            const { nativeEvent } = syntheticEvent;
            console.warn('WebView error: ', nativeEvent);
          }}
        />
      </GestureDetector>
    </View>
  );
}

export default function BuggyApp() {
  return (
    <GestureHandlerRootView>
      <WebViewScreen />
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  text: {
    fontSize: 18,
  },
  webView: {
    flex: 1,
  },
  webViewContainer: {
    flex: 1,
    width: '100%',
  },
});
```

</details>

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
@@ -99,7 +99,7 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
       if (state == STATE_UNDETERMINED && !hook.canBegin(event)) {
         cancel()
       } else {
-        view.onTouchEvent(event)
+        view.dispatchTouchEvent(event)
         if ((state == STATE_UNDETERMINED || state == STATE_BEGAN) && view.isPressed) {
           activate()
         }
@@ -116,12 +116,12 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
       when {
         shouldActivateOnStart -> {
           tryIntercept(view, event)
-          view.onTouchEvent(event)
+          view.dispatchTouchEvent(event)
           activate()
         }
 
         tryIntercept(view, event) -> {
-          view.onTouchEvent(event)
+          view.dispatchTouchEvent(event)
           activate()
         }
 
@@ -136,7 +136,7 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
         }
       }
     } else if (state == STATE_ACTIVE) {
-      view.onTouchEvent(event)
+      view.dispatchTouchEvent(event)
     }
   }
 
@@ -145,7 +145,7 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
     val event = MotionEvent.obtain(time, time, MotionEvent.ACTION_CANCEL, 0f, 0f, 0).apply {
       action = MotionEvent.ACTION_CANCEL
     }
-    view!!.onTouchEvent(event)
+    view!!.dispatchTouchEvent(event)
     event.recycle()
   }
 
--- a/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
@@ -101,7 +101,7 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
       if (state == STATE_UNDETERMINED && !hook.canBegin(event)) {
         cancel()
       } else {
-        hook.triggerEvent(view, event)
+        hook.sendTouchEvent(view, event)
         if ((state == STATE_UNDETERMINED || state == STATE_BEGAN) && view.isPressed) {
           activate()
         }
@@ -118,12 +118,12 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
       when {
         shouldActivateOnStart -> {
           tryIntercept(view, event)
-          hook.triggerEvent(view, event)
+          hook.sendTouchEvent(view, event)
           activate()
         }
 
         tryIntercept(view, event) -> {
-          hook.triggerEvent(view, event)
+          hook.sendTouchEvent(view, event)
           activate()
         }
 
@@ -138,7 +138,7 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
         }
       }
     } else if (state == STATE_ACTIVE) {
-      hook.triggerEvent(view, event)
+      hook.sendTouchEvent(view, event)
     }
   }
 
@@ -147,7 +147,7 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
     val event = MotionEvent.obtain(time, time, MotionEvent.ACTION_CANCEL, 0f, 0f, 0).apply {
       action = MotionEvent.ACTION_CANCEL
     }
-    hook.triggerEvent(view, event)
+    hook.sendTouchEvent(view, event)
     event.recycle()
   }
 
@@ -203,9 +203,9 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
     fun shouldCancelRootViewGestureHandlerIfNecessary() = false
 
     /**
-     * Decides whether or not target view should pass event to children.
+     * Passes the event down to the underlying view using the correct method.
      */
-    fun triggerEvent(view: View?, event: MotionEvent) = view?.onTouchEvent(event)
+    fun sendTouchEvent(view: View?, event: MotionEvent) = view?.onTouchEvent(event)
   }
 
   private class EditTextHook(
@@ -293,6 +293,6 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
     //
     // Calling onTouchEvent won't do anything on wrappers, therefore we use dispatchTouchEvent. This way
     // not only do we trigger onTouchEvent, but also pass event to children.
-    override fun triggerEvent(view: View?, event: MotionEvent) = view?.dispatchTouchEvent(event)
+    override fun sendTouchEvent(view: View?, event: MotionEvent) = view?.dispatchTouchEvent(event)
   }
 }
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -99,7 +99,7 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
       if (state == STATE_UNDETERMINED && !hook.canBegin(event)) {
         cancel()
       } else {
-        view.onTouchEvent(event)
+        view.dispatchTouchEvent(event)
```
j-piasecki, line 102: Does this impact any other views? `dispatchTouchEvent` sounds like it does something significantly different than `onTouchEvent`.

m-bert: From my long journey through Android source code I can say that it is not entirely different. In short `ViewGroup.dispatchTouchEvent` calls `View.dispatchTouchEvent`, which calls `View. performOnTouchCallback`, which calls `View.onTouchEvent`. So in the end we still call this function. The thing is, beside calling `onTouchEvent` indirectly, android also checks for interceptions and extracts touch target.

- [ViewGroup.java](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/ViewGroup.java)
- [View.java](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/View.java)

m-bert: I'll go through example app one more time and focus more on the examples that use `NativeViewGestureHandler`.

Also, what exactly do you mean by 

> Does this impact any other views? 

What do you think could be affected?

Nantris:
I just wanted to comment to confirm this does appear to resolve #3196! I detailed a finding of my testing there which may be worth reading through for anyone reviewing this

https://github.com/software-mansion/react-native-gesture-handler/issues/3196#issuecomment-2508681149

A big thanks to @m-bert for the great work! I hope this can be merged without breaking anything!

Nantris:
Also, I won't pretend I have the expertise to know much here, but I noticed there's one place that `onTouchEvent` is still used. I assume that's purposeful?

android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -82,6 +83,16 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
...
 
+  fun triggerEvent(event: MotionEvent) {
+    val view = view!!
+
+    if (view is ReactViewGroup) {
+      view.dispatchTouchEvent(event)
+    } else {
+      view.onTouchEvent(event)
+    }
+  }
```
j-piasecki, line 94: We may want this to be part of the native gesture hook, no?

android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -199,6 +201,11 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
      * by this one.
      */
     fun shouldCancelRootViewGestureHandlerIfNecessary() = false
+
+    /**
+     * Decides whether or not target view should pass event to children.
+     */
+    fun triggerEvent(view: View?, event: MotionEvent) = view?.onTouchEvent(event)
```
j-piasecki, line 208:
```suggestion
    /**
     * Passes the event down to the underlying view using the correct method.
     */
    fun sendTouchEvent(view: View?, event: MotionEvent) = view?.onTouchEvent(event)
```

android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -278,4 +285,14 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
...
     override fun shouldCancelRootViewGestureHandlerIfNecessary() = true
   }
+
+  private class ReactViewGroupHook : NativeViewGestureHandlerHook {
+    // In some cases wrapping component with NativeViewGestureHandler doesn't work, because the component
+    // itself has a wrapper. This happens for example with WebView, where it has 2 layers of wrappers above
+    // an actual WebView.
+    //
+    // Calling onTouchEvent won't do anything on wrappers, therefore we use dispatchTouchEvent. This way
+    // not only do we trigger onTouchEvent, but also pass event to children.
```
j-piasecki, line 295:
```suggestion
    // There are cases where a native component is wrapped with a `ReactViewGroup` (the component is rendered 
    // inside a `<View />` component in JS). In such cases, calling `onTouchEvent` wouldn't work as those are 
    // ignored by the wrapper view. Instead `dispatchTouchEvent` can be used, which causes the view to dispatch 
    // the event to its children.
```

I'd avoid explicitly mentioning implementation details of other libraries as those can change independently of RNGH and cause the comment to be obsolete.

j-piasecki (approved):

android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
m-bert: Done in [4ca8726](https://github.com/software-mansion/react-native-gesture-handler/pull/3244/commits/4ca8726e9be7718b2c2211a17bfacb663d80e4a8)

lovegaoshi:
just want to leave a comment that I started seeing ` ScrollView       E  Invalid pointerId=-1 in onTouchEvent` with this commit only on Android 15 (both emulator and pixel 7; Android 14/S21 is fine), when using a RNGH scrollview as [flashlist's renderScrollComponent](https://github.com/Spicy-Sparks/react-native-flashdrag-list/blob/f574b81e913c8429fd4182e61bf23972c7db3c35/src/FlashDragList.tsx#L275). ScrollView wont scroll anymore. I [reverted exactly this commit](https://github.com/lovegaoshi/azusa-player-mobile/commit/69fe775c470cebfff734ff63274f5e7802e3d2a3#diff-dd1d5b1c4b957a9f1060b75c091a9b680c82a4a9de58c55729dcfff103f08ba2) and can scroll fine then.

m-bert:
Hi @lovegaoshi! Do you have any reproducible example? As far as I can see this package is not actively supported and depends on outdated versions of libraries, especially Gesture Handler.

lovegaoshi:
thx for the attention! Let me clarify that the problem doesnt necessarily has to do with flashdraglist being unmaintained - at its core,[ its just an Animated wrapped flashlist with RNGH's ScrollView used as its renderScrollComponent.](https://github.com/lovegaoshi/azusa-player-mobile/blob/9be00ddf8bd01b9f8ca2dca2b9603f90972aedbc/src/App.tsx#L20) You can reproduce with simply that.

while [this](https://github.com/lovegaoshi/azusa-player-mobile/tree/bug/RNGH-2.23.0) is not exactly a minimum reproducer, I [removed everything except the exact bug in index.js's entry point.](https://github.com/lovegaoshi/azusa-player-mobile/commit/9be00ddf8bd01b9f8ca2dca2b9603f90972aedbc#diff-26ad4b834941d9b19ebf9db8082bd202aaf72ea0ddea85f5a8a0cb3c729cc6f2) You can build via git clone, yarn, yarn build, then build via Android Studio. Scrolling this list specifically on android 15 (emulator or real device) will yield this.
image

Issue #3246, opened 2024-11-29
[Pressable] android_ripple / color doesn't work correctly


freeboub:
### Description

on `Pressable` android_ripple={color: XXX} doesn't work correctly.

Here are issues:
- when user configure the  android_ripple={color: "#FF0000"} => the ripple color is grey. What ever the configured color is, the color ripple color is always grey (should be red with this sample)
- when user configure the  android_ripple={color: "#FFFFFF4F"} (With transparency) => the ripple color is blue. depending of transparency level, the color is sometimes grey ...
- I have also notice that a fast refresh doesn't refresh correctly the ripple color. (maybe it can help to identify the issue, or at least don't fall in that trap)

I checked with default react native `Pressable` these 3 use cases works fine.

### Steps to reproduce

Just define a Pressable with android_ripple={{color}}

```JS
<Pressable onPress={() => console.log('onPress')} android_ripple={{color: '#FF0000'}}>
         <Text>'Hello World'</Text>
</Pressable>
```

### Snack or a link to a repository

https://github.com/freeboub/bug-react-native-gesture-handler-android-ripple-color

### Gesture Handler version

2.21.2

### React Native version

0.76.2

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo bare workflow

### Architecture

Fabric (New Architecture)

### Build type

Release mode

### Device

Android emulator

### Device model

reproduced on all android devices

### Acknowledgements

Yes

freeboub:
I tried to debug the issue and I have an interesting point, This issue is reproducible only with new architecture enabled.
JS side looks good, but setRippleColor on kotlin side is never called

freeboub:
Ok, I got the root cause, it looks like processColor should not be called with new architecture: 
doing this in Pressable.tsx fix the issue:
`rippleColor={android_ripple?.color ?? defaultRippleColor} `
But it doesn't work with old architecture 😭

freeboub:
following patch can also be added to test the feature in the sample, but it breaks ios behavior, Not sure I should do it:
```diff
diff --git a/FabricExample/src/Navigator.tsx b/FabricExample/src/Navigator.tsx
index de9759f8e..d2d8227e0 100644
--- a/FabricExample/src/Navigator.tsx
+++ b/FabricExample/src/Navigator.tsx
@@ -1,7 +1,8 @@
 /* eslint-disable react-native/no-inline-styles */
 import React, { useEffect, useState } from 'react';
[...]

PR #3250, opened 2024-11-30
fix: rippleColor shall be declared as integer

freeboub:
## Description

Fix rippleColor declaration on pressable
fix: https://github.com/software-mansion/react-native-gesture-handler/issues/3246
I think colorValue can be used without processColor but here an Int32 is needed

## Test plan

I tested with the sample included in the package.
I didn't retest with old architecture but no impact expected, And I already tested it sucessfully

### DIFF
--- a/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerButtonManagerDelegate.java
+++ b/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerButtonManagerDelegate.java
@@ -35,7 +35,7 @@ public class RNGestureHandlerButtonManagerDelegate<T extends View, U extends Bas
         mViewManager.setEnabled(view, value == null ? true : (boolean) value);
         break;
       case "rippleColor":
-        mViewManager.setRippleColor(view, ColorPropConverter.getColor(value, view.getContext()));
+        mViewManager.setRippleColor(view, value == null ? 0 : ((Double) value).intValue());
         break;
       case "rippleRadius":
         mViewManager.setRippleRadius(view, value == null ? 0 : ((Double) value).intValue());
--- a/src/specs/RNGestureHandlerButtonNativeComponent.ts
+++ b/src/specs/RNGestureHandlerButtonNativeComponent.ts
@@ -11,7 +11,7 @@ interface NativeProps extends ViewProps {
   foreground?: boolean;
   borderless?: boolean;
   enabled?: WithDefault<boolean, true>;
-  rippleColor?: ColorValue;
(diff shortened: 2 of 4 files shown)


android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerButtonManagerDelegate.java
```diff
@@ -35,7 +35,7 @@ public void setProperty(T view, String propName, @Nullable Object value) {
         mViewManager.setEnabled(view, value == null ? true : (boolean) value);
         break;
       case "rippleColor":
-        mViewManager.setRippleColor(view, ColorPropConverter.getColor(value, view.getContext()));
```
latekvo, line 38: Removing `ColorPropConverter.getColor(value, view.getContext()` breaks `paper` builds when trying to set the color by it's name (`green` or `red`).
This only affects `paper`, because `fabric` has a different way of setting colors.

src/specs/RNGestureHandlerButtonNativeComponent.ts
```diff
@@ -11,7 +11,7 @@ interface NativeProps extends ViewProps {
   foreground?: boolean;
   borderless?: boolean;
   enabled?: WithDefault<boolean, true>;
-  rippleColor?: ColorValue;
```
latekvo, line 14: Is it possible to avoid overwriting this type?
Color may also be described as a string (`red` or `green`), replacing this type with `Int32` removes that possibility.

latekvo (commented):
Hi, the overall premise of this PR is great, it removes the necessity to use `processColor` on `paper` when using hexadecimal colors.
Unfortunately these changes cause a crash when you try setting `ripple_color` by the color's name (`red`). I'm currently looking into the `fabric` implementation, to see if it can be used in the `paper` workflow.

freeboub:
@latekvo Sorry for the delayed answer, but I am a bit lost to find the best fix...
I agree keeping ColorValue in the spec file would be better
- If I just remove processColor from the JS, It works as is with new arch (Then can keep the ColorValue) 
- but it fails with paper in ColorPropConverter.getColor because the input is a string, not an integer...

Do you think it is a good idea to remove the processColor in JS ? if yes, do you know which call should be done on paper version to transform the string color to integer color on native side ?

Notice that processColor return a ProcessedColorValue, not a ColorValue, I also tried to replace the type to ProcessedColorValue, not better with fabric ...

freeboub:
replaced by: https://github.com/software-mansion/react-native-gesture-handler/pull/3369 let's close this PR

PR #3252, opened 2024-12-02
Fix transform on GestureHandlerButton on Android

janicduplessis:
## Description

RN now handles view transform in onAfterUpdateTransaction https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java#L634 so we need to make sure to call the super method.

## Test plan

Tested that setting transform on button component now works on Android.


j-piasecki (approved):
Thanks for catching it!

PR #3254, opened 2024-12-02
[docs] Add `Swipe & Scroll` guide

m-bert:
## Description

At the beginning of new web implementation `Swipeable` component was blocking `scroll` on `web`. We did fix it by adding `touch-action: pan-y;` as a property to our handlers, and making it default in `Pan` used  under the hood of `Swipeable`.

However, some of our users prefer to create their own versions of `Swipeable` - that's where they encounter this issue.

This PR adds guide section about fixing this problem.

It also changes font size of `code` inside headers such that it is no longer ridiculously small.
## Test plan

Run docs

### DIFF
--- a/docs/docs/guides/swipe-and-scroll.md
+++ b/docs/docs/guides/swipe-and-scroll.md
@@ -1,9 +1,7 @@
 ---
 id: swipe-and-scroll
-title: Swipeable components inside ScrollView (web)
+title: Custom swipeable components inside ScrollView (web)
 ---
-
-## Adding custom swipeable components into `ScrollView`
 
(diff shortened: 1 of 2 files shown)


patrycjakalinska (commented):
Good fix with small font!  ✨ The guide sounds good, one thing I'd do is to change title to more describing like `Custom swipeable components inside ScrollView (web)` and remove subtitle. 

Also, I understand its placement (under `Guides` section) but it's kinda short for separate page. I'd be perfect if there'll be a better place for it, but I'm not pushing as for now it's pretty acceptable.

patrycjakalinska (approved):
🚀

j-piasecki (approved):

PR #3255, opened 2024-12-02
[Web] Migrate to `get` and `set` API [1/2]

m-bert:
## Description

In this PR I've migrated our handlers into `get`/`set` functions instead of traditional getters and setters convention (like in `Java`). 

>[!NOTE]
> This PR changes only handlers and detectors, `tools` directory is not a part of this PR as it would get too big.

## Test plan

Run web example

### DIFF
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -288,13 +288,6 @@ export default abstract class GestureHandler implements IGestureHandler {
     this.resetProgress();
   }
 
-  public getShouldResetProgress(): boolean {
-    return this.shouldResetProgress;
-  }
-  public setShouldResetProgress(value: boolean): void {
-    this.shouldResetProgress = value;
-  }
-
   public shouldWaitForHandlerFailure(handler: IGestureHandler): boolean {
     if (handler === this) {
       return false;
--- a/src/web/handlers/GestureHandler.ts
+++ b/src/web/handlers/GestureHandler.ts
@@ -698,115 +698,117 @@ export default abstract class GestureHandler implements IGestureHandler {
     );
   }
 
-  get lastSentState() {
+  public get lastSentState() {
     return this._lastSentState;
   }
-  set lastSentState(val: State | null) {
+  public set lastSentState(val: State | null) {
     this._lastSentState = val;
   }
 
-  get state() {
+  public get state() {
     return this._state;
   }
-  set state(newState: State) {
+  public set state(newState: State) {
     this._state = newState;
   }
 
-  get shouldCancelWhenOutside() {
+  public get shouldCancelWhenOutside() {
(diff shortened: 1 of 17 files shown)


src/web/handlers/GestureHandler.ts
```diff
@@ -751,60 +686,130 @@ export default abstract class GestureHandler implements IGestureHandler {
...
   }
 
-  public getConfig() {
-    return this.config;
+  public get state() {
+    return this._state;
+  }
+  public set state(state: State) {
+    this._state = state;
   }
```
latekvo, line 713: Previously the `currentState` was `protected`, and was only used internally. 
Now the `state` is exposed as its getters are `public`.
Same for `tracker`, `hasCustomActivationCriteria`, and multiple other internal variables which were `private`/`protected`, and not exposed by any public getters or setters before your changes.

m-bert: Well, yes and no. Even though those modifiers are public, they are not available to end user. At least doing something like:

```jsx
const pan = Gesture.Pan()
pan.state = State.ACTIVE
```

won't work as the user operates on `Gesture` type.

I've also thought about removing setters/getters that are not used, but I'm not sure about that.

Either way, we can change access modifiers to match previous state of this implementation, but I don't think it makes much sense to create `private` getters and setters. CC @j-piasecki

src/web/tools/GestureHandlerOrchestrator.ts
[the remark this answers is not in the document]
m-bert: As discussed yesterday, I left `tools` directory for another PR

src/web/handlers/GestureHandler.ts
```diff
@@ -751,60 +686,130 @@ export default abstract class GestureHandler implements IGestureHandler {
     this.delegate.destroy(this.config);
   }
 
-  //
-  // Getters and setters
-  //
+  public getTrackedPointersID(): number[] {
+    return this.pointerTracker.getTrackedPointersID();
+  }
```
latekvo, line 691: Is this intentional? I _get_ that `getTrackedPointersID` is just forwarding a function with the same name, but I'm not sure why we can't make it a getter called `trackedPointersID`

m-bert: > Since we're already changing the name of the function, is it ok if we rename it to trackedPointerIDs?

Sure, we can

> I'm not sure why we can't make it a getter called trackedPointersID

Here we have to decide whether we want to have getters to values that are calculated, not stored in a private fields. It is already done `ScaleGestureDetector`, but I'm still not sure if we want that. 

On the other hand, i get that it doesn't look good that we have ES6 and Java-style getters 😞 

What's your stance on that, @j-piasecki?

m-bert:
After further discussions I'm closing this PR in favor of #3263

PR #3256, opened 2024-12-03
Update reanimated_swipeable.md

juicypopbop:
## Description

Fixes import

## Test plan

Usage in app


latekvo:
The `ReanimatedSwipeable` is the default export of `react-native-gesture-handler/ReanimatedSwipeable` - it doesn't matter if it's named `ReanimatedSwipeable`, `Swipeable`, or `AbcdAbcd`, it will function the same.

Also this change breaks the example app you're modifying, and doesn't fix any potential issues.

PR #3257, opened 2024-12-04
[Web] Remove unnecessary overrides

m-bert:
## Description

This PR removes unnecessary overriden functions

## Test plan

Run example app


j-piasecki (approved):

m-bert:
@j-piasecki Yes it can, I've targeted `next` because earlier on I was working on some GH3 part and I remembered that we wanted to clean some things up in 3, hence I did it this way 😅

PR #3263, opened 2024-12-06
[Web] Migrate to `get` and `set`

m-bert:
## Description

In this PR I've migrated our handlers into `get`/`set` functions instead of traditional getters and setters convention (like in `Java`). 

## Test plan

Run web example


src/web/detectors/RotationGestureDetector.ts
```diff
@@ -143,24 +143,33 @@ export default class RotationGestureDetector
...
   }
 
-  public getAnchorX(): number {
-    return this.anchorX;
+  public get anchorX() {
+    return this._anchorX;
+  }
+  private set anchorX(value: number) {
+    this._anchorX = value;
   }
```
j-piasecki, line 156: Are setters actually needed here? 'm not sure if it should be possible to set the anchor externally on a gesture. Also applies to other places, do we need to create setters for **everything**?

m-bert: > 'm not sure if it should be possible to set the anchor externally on a gesture.

Well, it won't be set externally as the setter is private 😅 Also, I'm not sure if we want to write directly into backing field, so I'd leave that 😅

latekvo: What's the use case for a setter that doesn't do anything besides setting the backing field? 
Isn't it equivalent to just having an `anchorX` property available, and removing the `_anchorX` backing field?
I believe this PR was meant to migrate `getSomething` functions into `something` getters, but converting every existing regular property into a `getter`+`setter` doesn't seem useful at all.

src/web/detectors/RotationGestureDetector.ts
```diff
@@ -18,10 +18,10 @@ export default class RotationGestureDetector
   private previousTime = 0;
 
   private previousAngle = 0;
-  private rotation = 0;
+  private _rotation = 0;
```
latekvo, line 21: Why were some private fields like `rotation` converted into a `getter`+`setter`, while other like `previousAngle` weren't?

src/web/detectors/RotationGestureDetector.ts
j-piasecki: After some thinking, I think I agree with @latekvo. Having private setters that essentially don't do anything may do more harm than good just by being there (case in point, I missed the `private` when typing the first comment). I'm not sure there's anything wrong with setting the backing field directly - isn't this what we're doing right now but with a `getX` function?

src/web/detectors/RotationGestureDetector.ts
m-bert: Because rotation [is used outside of RotationGestureDetector](https://github.com/software-mansion/react-native-gesture-handler/blob/56c33510127b26386e5c46cfce8019bcd0b7e61e/src/web/handlers/RotationGestureHandler.ts#L21)

src/web/detectors/RotationGestureDetector.ts
m-bert: > (...) converting every existing regular property into a `getter`+`setter` doesn't seem useful at all.

It is not converting _every existing property_, as you've already mentioned below 😉

> Isn't it equivalent to just having an anchorX property available, and removing the _anchorX backing field? 

It would be equivalent to making it `public` which I don't think we want to do.


> I'm not sure there's anything wrong with setting the backing field directly - isn't this what we're doing right now but with a getX function?

Basically yes, if you prefer to do so, I can change it

src/web/detectors/RotationGestureDetector.ts
latekvo: You're right, I missed that.

src/web/detectors/RotationGestureDetector.ts
j-piasecki: Yeah, I think it would make sense. The diff would be the smallest (I think) this way and we can always iterate further on this.

m-bert: Done in [336f744](https://github.com/software-mansion/react-native-gesture-handler/pull/3263/commits/336f74490fd1ef96ea0596741a27060c91015c26)

src/web/handlers/IGestureHandler.ts
```diff
@@ -7,19 +7,25 @@ import type { GestureHandlerDelegate } from '../tools/GestureHandlerDelegate';
...
+  awaiting: boolean;
+  handlerTag: number;
+  config: Config;
+  delegate: GestureHandlerDelegate<unknown, this>;
+  tracker: PointerTracker;
+  state: State;
+  shouldCancelWhenOutside: boolean;
+  shouldResetProgress: boolean;
+  enabled: boolean;
+  pointerType: PointerType;
```
j-piasecki, line 21: Shouldn't some of those be marked as `readonly`?

m-bert: Done in [4275e16](https://github.com/software-mansion/react-native-gesture-handler/pull/3263/commits/4275e16d4199fbaad768bd6886e6977010a8b267)

m-bert: Also, let me know what di you think about `protected setters`

j-piasecki: I don't have a strong opinion of them. On the one hand, they don't do anything at the moment, but on the other the code looks a bit nicer without `_` and if we ever need to add any side-effects to the setters it would require less changed. The question is whether it's likely that we would be adding any side-effects to those setters.

m-bert: While it doesn't seem like we will add side-effects to them, I think that we can leave it for now 😅 Especially that

> the code looks a bit nicer without `_`

j-piasecki (approved):
LGTM, not sure about @latekvo

m-bert:
> LGTM, not sure about @latekvo

I'll wait for his approval 😅

src/web/detectors/ScaleGestureDetector.ts
```diff
@@ -145,27 +145,27 @@ export default class ScaleGestureDetector implements ScaleGestureListener {
...
-  }
+  public calculateScaleFactor(numOfPointers: number): number {
+    if (numOfPointers < 2) {
+      return 1;
+    }
 
-  public getFocusX(): number {
-    return this.focusX;
+    return this.prevSpan > 0 ? this.currentSpan / this.prevSpan : 1;
   }
```
latekvo, line 154: Since this is a trivial calculation, could we maybe change it to `public get scaleFactor`?
This wouldn't be the first calculating getter, `timeDelta` getter also performs a calculation.

latekvo (approved):
LGTM, just one nit-pick

src/web/detectors/ScaleGestureDetector.ts
m-bert: This name was changed to `calculateScaleFactor` intentionally. we do not have access to `numOfPointers` inside `ScaleGestureDetector` so we can't perform first check without having it as a parameter

latekvo: You're right, missed the parameter

PR #3264, opened 2024-12-06
Fix `continuous <-> discrete` gesture interactions

### LINKED ISSUE #3253: ScrollView doesn't interact properly with GestureHandlerButton on Android
### Description

When a button inside a scroll view is activated it won't be interrupted by scrolling on Android. This seems to be a regression since it used to work fine in 2.20.0.

Current behaviour (rngh 2.21.2):

https://github.com/user-attachments/assets/8a65d72b-acc3-4f8f-a242-2356597f5ce8

Expected behaviour (rngh 2.20.0):

https://github.com/user-attachments/assets/2b92e6c5-ad4a-4c80-bf58-51c1458d0cd3

Note it doesn't repro in the snack since it uses the older 2.16.1 version

### Steps to reproduce

1. Activate the button by touching it but not releasing
2. Scroll up or down

Here the button activation should be canceled and the scroll view should scroll

https://snack.expo.dev/8zhTUCK-4AOzV2LzEJEA8

### Gesture Handler version

2.21.2

### React Native version

0.76.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

latekvo:
## Description

This PR fixes an issue with native views, where the native continuous parent couldn't get activated when a discrete child gesture was already active.
After this PR, only kinds of handlers affected by the [native view parents stealing activity fix](https://github.com/software-mansion/react-native-gesture-handler/issues/3253) will be: `continuous-continuous` and `discrete-discrete` interactions.

Fixes #3253
 
As of [bea5d58](https://github.com/software-mansion/react-native-gesture-handler/pull/3264/commits/bea5d58f637d6b543d954ff511745e7a04b4e2a7) this fix works, but I'm still testing it for edge cases.
Currently it also supposes the `gestureHandlers` is sorted from most to least nested handlers, while this characteristic is already used in other parts of the code (inverting `gestureHandlers` array breaks the Gesture Handler), I'm not sure if this is always the case.

## Test examples

<details>
<summary>
combo example:<br>
+ continuous native parent - continuous child interaction<br>
+ discrete native parent - continuous child interaction<br>
+ continuous native parent (vertical & horizontal) - discrete native child interaction
</summary>

```js
import React from 'react';
import { StyleSheet, Text, View, ScrollView } from 'react-native';
import {
  Gesture,
  GestureDetector,
  GestureUpdateEvent,
  PanGestureHandlerEventPayload,
  RectButton,
import Animated, {
  SharedValue,
  useAnimatedStyle,
  useSharedValue,
  withSpring,

  const externalPosition = useSharedValue<{ x: number; y: number }>({
    x: 0,
    y: 0,
  });

  const nestedPosition = useSharedValue<{ x: number; y: number }>({
    x: 0,
    y: 0,
  });

  const setter = (
    position: SharedValue<{
      x: number;
      y: number;
    }>
  ) => {
    return (event: GestureUpdateEvent<PanGestureHandlerEventPayload>) => {
      'worklet';
      position.value = {
        x: event.translationX,
        y: event.translationY,
      };
    };
  };

  const resetter = (
    position: SharedValue<{
      x: number;
      y: number;
    }>
  ) => {
    return () => {
      'worklet';
      position.value = {
        x: withSpring(0),
        y: withSpring(0),
      };
    };
  };

  const internalPan = Gesture.Pan()
    .onUpdate(setter(externalPosition))
    .onFinalize(resetter(externalPosition));

  const internalTap = Gesture.Tap()
    .onStart(() => console.log('Inner tap'))
    .onFinalize(resetter(externalPosition));

  const nestedPan = Gesture.Pan()
    .onUpdate(setter(nestedPosition))
    .onFinalize(resetter(nestedPosition));

  const externalAnimation = useAnimatedStyle(() => {
    return {
      ...styles.box,
      transform: [
        { translateX: externalPosition.value.x },
        { translateY: externalPosition.value.y },
      ],
    };
  });

  const nestedAnimation = useAnimatedStyle(() => {
    return {
      ...styles.box,
      transform: [
        { translateX: nestedPosition.value.x },
        { translateY: nestedPosition.value.y },
      ],
    };
  });

  return (
      <View style={styles.externalContainer}>
        {/* Continuous gesture inside discrete native button */}
        <RectButton
          onPress={() => console.log('left top outer native button')}
          style={[styles.box, { backgroundColor: 'pink', height: 150 }]}>
          <GestureDetector gesture={internalPan}>
            <Animated.View style={externalAnimation}>
              <Text>Continuous gesture inside discrete native button.</Text>
            </Animated.View>
          </GestureDetector>
        </RectButton>
      </View>

      <View style={styles.externalContainer}>
        {/* Discrete gesture inside discrete native button */}
        <RectButton
          onPress={() => console.log('Outer tap')}
          style={[styles.box, { backgroundColor: 'pink', height: 150 }]}>
          <GestureDetector gesture={internalTap}>
            <Animated.View style={styles.box}>
              <Text>Discrete gesture inside discrete native button.</Text>
            </Animated.View>
          </GestureDetector>
        </RectButton>
      </View>

      {/* Continuous gesture inside continuous native list - horizontal */}
        <ScrollView style={styles.horizontalList} horizontal>
          {new Array(20)
            .fill(1)
            .map((value, index) => value * index)
            .map((value) => (
              // Discrete gesture inside continuous native list
              <RectButton
                onPress={() =>
                  console.log('discrete gesture inside horizontal scroll view')
                }
                style={styles.horizontalElement}>
                <Text>
                  Discrete gesture inside continuous native list no. {value}
                </Text>
              </RectButton>
            ))}
        </ScrollView>
      </GestureDetector>

      {/* Continuous gesture inside continuous native list */}
        <ScrollView style={styles.list}>
          <GestureDetector gesture={nestedPan}>
            <Animated.View style={nestedAnimation} collapsable={false}>
              <Text>Continuous gesture inside continuous native list</Text>
            </Animated.View>
          </GestureDetector>

          {new Array(20)
            .fill(1)
            .map((value, index) => value * index)
            .map((value) => (
              // Discrete gesture inside continuous native list
              <RectButton
                onPress={() => console.log('discrete gesture')}
                style={styles.element}>
                <Text>
                  Discrete gesture inside continuous native list no. {value}
                </Text>
              </RectButton>
            ))}
        </ScrollView>
      </GestureDetector>
    </View>
  );
}

  container: {
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
    gap: 20,
  },
  externalContainer: {
    flexDirection: 'row',
    gap: 20,
  },
  box: {
    backgroundColor: 'tomato',
    height: 70,
  },
  list: {
    width: 250,
    backgroundColor: 'plum',
  },
  element: {
    margin: 1,
    height: 40,
    backgroundColor: 'orange',
  },
  horizontalList: {
    height: 180,
  },
  horizontalElement: {
    margin: 1,
    width: 80,
    backgroundColor: 'orange',
  },
  listColumns: {
    flexDirection: 'row',
  },
});

```

</details>


android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -80,10 +81,13 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
       is ReactEditText -> this.hook = EditTextHook(this, view)
       is ReactSwipeRefreshLayout -> this.hook = SwipeRefreshLayoutHook(this, view)
       is ReactScrollView -> this.hook = ScrollViewHook()
+      is ReactHorizontalScrollView -> this.hook = ScrollViewHook()
       is ReactViewGroup -> this.hook = ReactViewGroupHook()
     }
   }
 
+  override fun isContinuous() = view is ReactScrollView || view is ReactHorizontalScrollView || view is ReactSwipeRefreshLayout
```
j-piasecki, line 89: Move this to the hook, please.

It also may be worth it to consider using ternary logic here (`true`, `false`, `null`) and return null for the views that use the default hook to not make any assumptions.

j-piasecki:
> Currently it also supposes the gestureHandlers is sorted from most to least nested handlers, while this characteristic is already used in other parts of the code (inverting gestureHandlers array breaks the Gesture Handler), I'm not sure if this is always the case.

The handlers are added to the list in the order of DFS traversal of the view hierarchy but they are also later sorted: https://github.com/software-mansion/react-native-gesture-handler/blob/f0604b03995ce06ac9e0a10d00ef9802dd5e1e62/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt#L622-L638

I don't remember from the top of my head the exact consequences of that though.

android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
latekvo: Moved isContinuous to the hook in [475c37d](https://github.com/software-mansion/react-native-gesture-handler/pull/3264/commits/475c37deeda4cef01af0a9a679c26526fa24ce1e).

Regarding the ternary logic, I'm not sure if we can apply it to the `NativeViewGestureHandler`. By default, `isContinuous` has to be set to `false`, while `NativeViewGestureHandlerHook` is both the the default hook, and the common ancestor of all other hooks. 

On the other hand, I think it makes sense to use the ternary logic on the base `GestureHandler` class, as it's not directly used anywhere, and it'd be preferable for every old and new gesture handler to explicitly set the `isContinuous` function to either `true` or `false`.

j-piasecki: Yeah, that's what I meant:
- change `isContinuous` to return `Boolean?`
- make every gesture handler explicitly implement it to return true or false
- make basic hook return null
- update other hooks to either return `true` or `false`

latekvo: I think it'd make more sense to convert `GestureHandler` to be an abstract class. I think that just making `isContinuous` a ternary isn't enough to signal that `GestureHandler`'s children should implement it; same for most of the other GestureHandle functions which behave like `abstract` functions - are left empty, and expected to be implemented, while not being explicitly marked as `abstract`.

I don't think `GestureHandler` class is used anywhere directly, and there are multiple such functions which it does not implement, while expecting it's children to implement them.

Of course if that's a change that you think makes sense to implement, i'll open a separate PR for it.

android/src/main/java/com/swmansion/gesturehandler/core/ManualGestureHandler.kt
```diff
@@ -3,6 +3,8 @@ package com.swmansion.gesturehandler.core
 import android.view.MotionEvent
 
 class ManualGestureHandler : GestureHandler<ManualGestureHandler>() {
+  override fun isContinuous() = false
```
m-bert, line 6: I'm not sure if we should treat `Manual` as `discrete`. This gesture is very specific and could act as both

latekvo: I was planning to add a JS-side property to set if a given `Manual` should be treated as `discrete` or `continuous`.
But I agree that by default it makes more sense for it to be `continuous`, will fix this.

j-piasecki:
This is growing in scope really fast. I've tracked down the problem original PR tried to solve - it's that during a gesture another gesture attached to a child view can become active (via a second finger). This doesn't happen by default on Android.

The specific case from the PR is that during a scroll with one finger, another finger is able to start a gesture inside the scroll view which in turn cancels the scroll.

It may make sense to take a step back and try to solve the original issue instead of stacking more stuff on top of it - after some brief tests it looks like adding information about the view hierarchy to the `shouldBeCancelledByActiveHandler` method may be enough:

```kotlin
private fun GestureHandler<*>.isDescendantOf(of: GestureHandler<*>): Boolean {
    var view = this.view?.parent as? View
    while (view != null) {
        if (view == of.view) {
            return true
        }

        view = view.parent as? View
    }

    return false
}

private fun shouldBeCancelledByActiveHandler(handler: GestureHandler<*>) =
    gestureHandlers.any { handler.hasCommonPointers(it) && it.state == GestureHandler.STATE_ACTIVE && !canRunSimultaneously(handler, it) && handler.isDescendantOf(it) }
```

It something like this were to work it would be a much simpler and understandable approach, I think it's worth testing it more deeply. WDYT?

j-piasecki:
Sure, https://github.com/software-mansion/react-native-gesture-handler/pull/3296

latekvo:
Superseded by #3296

Issue #3265, opened 2024-12-07
Conditional rendering breaks blocksExternalGesture


gaearon:
## Working Example

This piece of code **works correctly**:

```js
import {useRef} from 'react'
import {View} from 'react-native'
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,
  ScrollView,
} from 'react-native-gesture-handler'

export default function App() {
  const scrollRef = useRef(null)
  const pan = Gesture.Pan()
    .blocksExternalGesture(scrollRef)
    .onBegin(() => {
      'worklet'
      console.log('begin')
    })
    .onUpdate(() => {
      'worklet'
      console.log('update')
    })
    .onEnd(() => {
      'worklet'
      console.log('end')
    })
    .onFinalize(() => {
      'worklet'
      console.log('finalize')
    })
    .failOffsetX(-1)
  return (
    <GestureHandlerRootView>
      <GestureDetector gesture={pan}>
        <View style={{flex: 1, backgroundColor: 'blue'}}>
          <ScrollViewExample scrollRef={scrollRef} />
        </View>
      </GestureDetector>
    </GestureHandlerRootView>
  )
}

function ScrollViewExample({scrollRef}) {
  return (
    <ScrollView
      ref={scrollRef}
      horizontal
      pagingEnabled
      style={{
        flex: 1,
        backgroundColor: 'yellow',
      }}>
      <View
        style={{
          width: 200,
          height: '100%',
          backgroundColor: 'green',
        }}
      />
      <View
        style={{
          width: 200,
          height: '100%',
          backgroundColor: 'purple',
        }}
      />
    </ScrollView>
  )
}
```

You can verify that **the scroll gesture gets blocked because the Pan has `blocksExternalGesture`.**

## Broken Example

Add conditional rendering to the example:

```diff
function ScrollViewExample({scrollRef}) {
+  const [show, setShow] = useState(true)
+  useEffect(() => {
+    let id = setInterval(() => setShow(s => !s), 3000)
+    return () => clearInterval(id)
+  }, [])
+  if (!show) {
+    return
+  }
  return (
    <ScrollView
```

Observe that the scroll view is still getting blocked during the first three seconds. Then wait three seconds unmounted. From that point on, **the scroll view is no longer getting blocked at all.**

### Steps to reproduce

1. Run https://snack.expo.dev/dO9UvH1o7Gq7NSMzpvt4A
2. Drag right (finger moving left) while you see green background

[...]

PR #3269, opened 2024-12-09
Fix relations not working when one of the child dependencies remounts without a render in the parent

j-piasecki:
## Description

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/3265

As @gaearon pointed out in the issue:
> I think what's happening here is that the library doesn't consider the fact that the contents of a ref can change over time without the parent knowing about it. Rather, it seems to be operating under assumption that a ref change at the bottom will always be accompanied by a state change near the top (wherever blocksExternalGesture is called).

That's unfortunately true, and the current API doesn't allow to use ref callbacks to handle this (nor would it be easy to update it to do that). This is one more point to the pile of reasons why the API needs an overhaul.

For the time being, I implemented a mechanism that allows `GestureDetector` to listen to the mounting and unmounting of gestures. This way it can react to a remount of a gesture it depends on and update the native side if necessary.

## Test plan

Tested on the Example app and the reproducer from the issue

|Android|iOS|
|-|-|
|<video src="https://github.com/user-attachments/assets/4df2e1d7-3932-4c1e-a305-ac1a0207ba20">|<video src="https://github.com/user-attachments/assets/e522bdb4-98a9-4a78-b9e8-2f0f9b28b62c">|

### DIFF
--- a/src/handlers/createHandler.tsx
+++ b/src/handlers/createHandler.tsx
@@ -24,6 +24,7 @@ import { ActionType } from '../ActionType';
 import { PressabilityDebugView } from './PressabilityDebugView';
 import GestureHandlerRootViewContext from '../GestureHandlerRootViewContext';
 import { ghQueueMicrotask } from '../ghQueueMicrotask';
+import { MountRegistry } from '../mountRegistry';
 
 const UIManagerAny = UIManager as any;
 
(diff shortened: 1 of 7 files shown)


m-bert (approved):

gaearon:
Would be great to start playing with it, can we get a canary release?

j-piasecki:
There's been a release candidate release a while ago - https://www.npmjs.com/package/react-native-gesture-handler/v/2.22.0-rc.1

PR #3270, opened 2024-12-09
Make `stylusData` property of the `GestureUpdateEvent` optional

kkafar:
## Description

Idk what's your policy on types & whether the type `stylusData: StylusData | undefined` type is intended in this form or not, 
but currently when creating event of type `GestureUpdateEvent<PanGestureHandlerEventPayload>` user is forced to specify
`stylusData` even if there is none - in such case `stylusData: undefined` must be passed; the prop can not be just simply omitted.

image

I'm humbly suggesting that you consider relaxing these types a bit to `stylusData?: StylusData | undefined`, so that it 
does not have to be always specified. 

Context: this is just quality of life improvement IMO, not crucial. I've stubmled upon this when I've bumped version of `gesture-handler` in `react-native-screens` [(link)](https://github.com/software-mansion/react-native-screens/pull/2561)
 - current types forced me to add this field, and I've thought this is not pleasant.

## Test plan

N/A, every thing should work as before

### DIFF
--- a/src/handlers/GestureHandlerEventPayload.ts
+++ b/src/handlers/GestureHandlerEventPayload.ts
@@ -126,7 +126,7 @@ export type PanGestureHandlerEventPayload = {
   /**
    * Object containing additional stylus data.
    */
-  stylusData: StylusData | undefined;
+  stylusData?: StylusData | undefined;
 };
 
 export type PinchGestureHandlerEventPayload = {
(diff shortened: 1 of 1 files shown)


kkafar:
Note: it is possible that the type should be adjusted in few other places to account for this change. I'll update the code once you confirm that you would agree for such change; otherwise I don't want to spend time on it 😄

src/handlers/GestureHandlerEventPayload.ts
```diff
@@ -126,7 +126,7 @@ export type PanGestureHandlerEventPayload = {
   /**
    * Object containing additional stylus data.
    */
-  stylusData: StylusData | undefined;
+  stylusData?: StylusData | undefined;
```
latekvo, line 129: Sounds good, I think we can remove `| undefined` now, since it's redundant.

latekvo (approved):

PR #3274, opened 2024-12-10
Update introduction and installation pages in the documentation

j-piasecki:
## Description

Update introduction page:
- Move the part about automatic workletization to troubleshooting, that's not really an intro-level topic
- Updated the package description and benefits
- Removed link to not working app
- Added social media links so that the docs don't explicitly tell you to contact kzzzf 😄

Updated installation page:
- Updated the version compatibility table
- Divided the installation into three-part process:
  - Add package
  - Add root view
  - "Platform-specific"
- Removed part about enabling the new arch, RNGH should *just work*
- Removed stuff referencing React <15 and RN <0.60 (what?)


docs/docs/guides/troubleshooting.md
```diff
@@ -73,3 +73,37 @@ or
 ```bash
 npm install
 ```
+
+### Automatic workletization of gesture callbacks
+
+Reanimated's Babel plugin is setup in a way that automatically marks callbacks passed to gestures in the configuration chain as worklets. This means that as long as all your callbacks are defined in a single chain, you don't need to add a `'worklet';` directive at the beginning of the functions. Here is an example that will be automatically workletized:
```
kacperkapusciak, line 79: i think it could be useful to give people links to concepts like:
**workletization**
https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/glossary#to-workletize
**reanimated babel plugin**
https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/glossary#reanimated-babel-plugin

docs/docs/guides/troubleshooting.md
```diff
@@ -73,3 +73,37 @@ or
...
+const gesture = Gesture.Tap().onBegin(callback);
+```
+
+```jsx
+const callback = () => {
+  console.log(_WORKLET);
+};
+const gesture = Gesture.Tap();
+gesture.onBegin(callback);
+```
```
kacperkapusciak, line 109: you should also write here that adding `"worklet";` directive inside the function makes it work

docs/docs/fundamentals/introduction.md
```diff
@@ -6,77 +6,27 @@ sidebar_position: 1
...
-- The ability to define relations between gesture handlers, e.g. when you have a pan handler in `ScrollView` you can make that `ScrollView` wait until it knows pan won't recognize.
-- Mechanisms to use touchables that run in native thread and follow platform default behavior; e.g. in the event they are in a scrollable component, turning into pressed state is slightly delayed to prevent it from highlighting when you fling.
+- A way to use a platform's native touch handling system for recognizing gestures (like pinch, rotation, pan and a few others).
+- The ability to define relations between gestures to ensure gestures, and possibly native components, will not conflict with each other.
+- Mechanisms to use touchable components that run in native thread and follow platform default behavior; e.g. in the event they are in a scrollable component, turning into pressed state is slightly delayed to prevent it from highlighting when you fling.
+- Close integration with [`react-native-reanimated`](https://docs.swmansion.com/react-native-reanimated/) to process toouch events on the UI thread.
```
kacperkapusciak, line 16:
```suggestion
- Close integration with [`react-native-reanimated`](https://docs.swmansion.com/react-native-reanimated/) to process touch events on the UI thread.
```
typo

docs/docs/fundamentals/installation.md
```diff
@@ -4,98 +4,129 @@ title: Installation
...
+cd ios && pod install && cd ..
 ```
 
-## Fabric
-
-Starting with version 2.3.0, Gesture Handler now supports [Fabric](https://reactnative.dev/docs/fabric-renderer)!. To use Gesture Handler in your Fabric application you have to:
+#### Web
 
-#### on iOS:
+There is no additional configuration required for the web, however, since the Gesture Handler 2.10.0 the new web implementation is enabled by default. It is recommended to check if the gestures in your app are working as expected since their behavior should now resemble the native platforms. If you don't want to use the new implementation, you can still revert back to the legacy one by enabling it at the beginning of your `index.js` file:
```
kacperkapusciak, line 119:
```suggestion
There is no additional configuration required for the web, however, since the Gesture Handler 2.10.0 the new web implementation is enabled by default. We recommend you to check if the gestures in your app are working as expected since their behavior should now resemble the native platforms. If you don't want to use the new implementation, you can still revert back to the legacy one by enabling it at the beginning of your `index.js` file:
```
seems a bit more friendly

docs/docs/fundamentals/installation.md
```diff
@@ -4,98 +4,129 @@ title: Installation
...
-npm install --save react-native-gesture-handler
+npx expo prebuild
 ```
 
-After installation, wrap your entry point with `<GestureHandlerRootView>` or
-`gestureHandlerRootHOC`.
+#### Android
 
-For example:
+Android doesn't require any more steps, but keep in mind that if you want to use gestures in Modals you need to wrap Modal's content with `GestureHandlerRootView`:
```
kacperkapusciak, line 79:
```suggestion
Setting up Gesture Handler on Android doesn't require any more steps. Keep in mind that if you want to use gestures in Modals you need to wrap Modal's content with `GestureHandlerRootView`:
```

docs/docs/fundamentals/installation.md
```diff
@@ -4,98 +4,129 @@ title: Installation
...
+      <ActuallApp />
+    </GestureHandlerRootView>
+  );
+}
 ```
 
-or with `npm` if you prefer:
+If you don't provide anything to the `styles` prop, it will default to `flex: 1`. If you want to customize the styling of the root view, don't forget to also include `flex: 1` in the custom style, otherwise your app won't render anything. Keep `GestureHandlerRootView` as close to the actual root of the app as possible. It's the entry point for all gestures and all gesture relations. The gestures won't be recognized outside of the root view, and relations only work between gestures mounted under the same root view.
+
+If you're unsure if one of your dependencies alredy renders `GestureHandlerRootView` on its own, don't worry and add one at the root anyway. In case of nested root views, Gesture Handler will only use the top-most one and ignore the nested ones.
```
kacperkapusciak, line 61:
```suggestion
If you're unsure if one of your dependencies already renders `GestureHandlerRootView` on its own, don't worry and add one at the root anyway. In case of nested root views, Gesture Handler will only use the top-most one and ignore the nested ones.
```
typo

kacperkapusciak (approved):
wow, what an improvement! 🎉 🔥

docs/docs/guides/troubleshooting.md
j-piasecki: done in [aa06fa3](https://github.com/software-mansion/react-native-gesture-handler/pull/3274/commits/aa06fa33641796612b0785b908b13dc6a11e12a1)

docs/docs/guides/troubleshooting.md
j-piasecki: Done in [1aec22a](https://github.com/software-mansion/react-native-gesture-handler/pull/3274/commits/1aec22a51502862792584f212840f2d23e8d67c4)

PR #3278, opened 2024-12-10
Support React Native 0.77

jakex7:
## Description

* 54710a610505f0279b2a4bebd09f706421d0ace5 - Added `sourceSets` to the package due to changes in ViewManager function signatures, which are no longer nullable
* a20adcaa9916a15939d06309ac398e2cda62d7a8 - Updated to reflect changes in https://github.com/facebook/react-native/pull/47551, `findHostInstance_DEPRECATED` is now a property of the `default` object instead of being a named export. Modified implementation reads from default and fall back to named export as needed
* ac7e786dc230acaf392002cf5c2a2787d01aef2e - Similar change to the https://github.com/software-mansion/react-native-svg/pull/2572 that uses new API to create an association map between JS components and their native implementations, avoiding `codegen` crawling through the entire filesystem for `.mm` files.

## Test plan

App with React Native 0.77 (like https://github.com/software-mansion/react-native-gesture-handler/pull/3277) should compile and work as expected.


package.json
```diff
@@ -144,6 +144,11 @@
     "jsSrcsDir": "./src/specs",
     "android": {
       "javaPackageName": "com.swmansion.gesturehandler"
+    },
+    "ios": {
+      "componentProvider": {
+        "RNGestureHandlerButton": "RNGestureHandlerButtonComponentView"
```
j-piasecki, line 150: Is this: https://github.com/software-mansion/react-native-gesture-handler/blob/6a7a128f600948e51e4c3b5eb9c4c71609666361/apple/RNGestureHandlerButtonComponentView.mm#L33-L37

still needed?

jakex7: I'm not certain, but even if it doesn't, `ios.componentProvider` is only available starting from RN 77. Therefore, to maintain backward compatibility, we still need to leave it there.

j-piasecki: Ah, I knew I must've forgotten something 😆

Issue #1661, opened 2021-09-28
Unable to pass testID prop to RectButton


nextlevelbeard:
`testID` prop on `RectButton` component does not take effect.
Furthermore, `accessible` prop also appears to suffer from the same issue.
With RN 0.64.

```jsx
<RectButton testID="test">
```

bsmith-self:
Any movement on adding one or both of these props? We are also trying to test our RN app with Appium, and this seems like a pretty simple passthrough. I might be willing to make a PR myself if the maintainers would be open to that

fcoradini:
I am working with React Native/Appium and required the ID for my selectors.
I was able to achieve this using `accessibilityLabel` on `RectButton`.

`accessibilityLabel` on React Native generates `accessibility-id` on both Android and iOS ([ref](https://blog.stackademic.com/appium-inspector-demystified-a-guide-to-selecting-elements-on-react-native-2cd3e3c60809))

```
"react-native": "0.72.6",
"react-native-gesture-handler": "~2.12.0",
````

**Comment by dgreasi** [comment]
I am having the same using while trying to test the app with Maestro. The `testID` is available in `iOS` but not in `Android`.

I am guessing something is off.

My workaround is quite similar with the above recommendations. 

For now I am using `accessibilityLabel` in the `View` I have inside `RectButton` since the `RectButton` itself doesn't accept an `accessibilityLabel`.

```
"react-native": "0.72.5",
"react-native-gesture-handler": "^2.16.0",
```

m-bert:
Hi @nextlevelbeard! Does this problem occur only with `appium`? I've just created new RN app and run tests right away. They did pass as expected. Could you tell us more about your setup? Or, if it would be possible, create a reproduction repository?

@bsmith-self 

> I might be willing to make a PR myself if the maintainers would be open to that

We are open to that 😅

nextlevelbeard:
@m-bert It's been 3 years since I reported this and I am no longer working with React Native, cannot help unfortunately.

marcin-bojar:
@m-bert what about other components? I just hit an issue where after wrapping components with `GestureDetector` all occurrences of `testID`s for `<Pressable>` components have been removed 🤦

m-bert:
As far as I know, @latekvo is working in `Pressable` in tests 😅

PR #3287, opened 2024-12-13
Expose testID as resource-id on Android

oblador:
## Description

When using `RectButton` and friends the `testID` prop is accessible in E2E testing frameworks on iOS but not on Android. This PR aligns behaviour of RN core components by exposing `testID` as a `resource-id`. 

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/1661

## Test plan

Been testing this via Maestro Studio.


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -241,6 +243,20 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
...
+
+      // Expose the testID prop as the resource-id name of the view. Black-box E2E/UI testing
+      // frameworks, which interact with the UI through the accessibility framework, do not have
+      // access to view tags. This allows developers/testers to avoid polluting the
+      // content-description with test identifiers.
+      val testId = super.getTag(R.id.react_test_id)
+
+      if (testId is String) {
+        info.setViewIdResourceName(testId)
+      }
```
m-bert, line 257: I can see that in `react-native` they also use this method, but with [slightly different check](https://github.com/facebook/react-native/blob/2fee13094b3d384c071978776fd8b7cff0b6530f/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollView.java#L155). Is there a specific reason behind this, or can we do the same?

oblador: Java vs Kotlin

m-bert (approved):

arcvlad:
This is not working anymore? Can not see id in appium?

PR #3294, opened 2024-12-17
Fix `TouchableNativeFeedback` regression

latekvo:
## Description

#3260 introduced a regression to the `Gesture Handlers`'s `TouchableNativeFeedback` component. 
The component used to be a direct reference to the `TouchableNativeFeedback`, which has a [couple of static methods](https://reactnative.dev/docs/touchablenativefeedback#methods).
When it was replaced with a function component, the static methods got removed, leaving only the render function of the `TouchableNativeFeedback`.

## Test plan

See how the type errors in the `example/src/release_tests/touchables/index.tsx` are resolved.

### DIFF
--- a/src/components/touchables/TouchableNativeFeedback.tsx
+++ b/src/components/touchables/TouchableNativeFeedback.tsx
@@ -1,22 +1,19 @@
-import React, { useEffect } from 'react';
 import { TouchableNativeFeedback as RNTouchableNativeFeedback } from 'react-native';
 import { tagMessage } from '../../utils';
 
 /**
  * @deprecated TouchableNativeFeedback will be removed in the future version of Gesture Handler.
  */
-const TouchableNativeFeedback: React.FC<
-  React.ComponentProps<typeof RNTouchableNativeFeedback>
-> = (props) => {
-  useEffect(() => {
+class TouchableNativeFeedback extends RNTouchableNativeFeedback {
+  componentDidMount() {
     console.warn(
       tagMessage(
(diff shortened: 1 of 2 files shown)


src/components/touchables/TouchableNativeFeedback.tsx
```diff
@@ -1,22 +1,19 @@
...
-const TouchableNativeFeedback: React.FC<
-  React.ComponentProps<typeof RNTouchableNativeFeedback>
-> = (props) => {
-  useEffect(() => {
+class TouchableNativeFeedback extends RNTouchableNativeFeedback {
+  componentDidMount() {
     console.warn(
       tagMessage(
-        'TouchableOpacity component will be removed in the future version of Gesture Handler.'
+        'TouchableNativeFeedback component will be removed in the future version of Gesture Handler.'
```
m-bert, line 11:
```suggestion
```

Let's remove this `console.log` as in other `Touchables`

latekvo: done in [13b32eb](https://github.com/software-mansion/react-native-gesture-handler/pull/3294/commits/13b32eb0a9aa992e9fe72577ca8923157f73c9bc)

m-bert:
This is what I meant when I asked if deprecation note works:

image

latekvo:
fixed in [1e8e07e](https://github.com/software-mansion/react-native-gesture-handler/pull/3294/commits/1e8e07e5313bfbd4249493f27eed5c0bc510969f)

m-bert (approved):

PR #3303, opened 2024-12-20
Migrate `next` to `main`.

m-bert:
## Description

This PR migrates changed from [next](https://github.com/software-mansion/react-native-gesture-handler/tree/next) branch into `main`.

## Test plan
Most of those commits are deprecations, to check `get`/`set` run example on web.

### DIFF
--- a/src/web/tools/Vector.ts
+++ b/src/web/tools/Vector.ts
@@ -30,7 +30,7 @@ export default class Vector {
     return new Vector(velocity.x, velocity.y);
   }
 
-  get magnitude() {
+  public get magnitude() {
     return this._magnitude;
   }
 
--- a/src/web/tools/VelocityTracker.ts
+++ b/src/web/tools/VelocityTracker.ts
@@ -84,7 +84,7 @@ export default class VelocityTracker {
     return null;
   }
 
-  public getVelocity(): [number, number] {
+  public get velocity() {
     const estimate = this.getVelocityEstimate();
     if (estimate !== null) {
       return estimate;
--- a/docs/docs/components/touchables.md
+++ b/docs/docs/components/touchables.md
@@ -4,6 +4,10 @@ title: Touchables
 sidebar_label: Touchables
 ---
 
+:::warning
+Touchables will be removed in the future version of Gesture Handler.
+:::
+
 Gesture Handler library provides an implementation of RN's touchable components that are based on [native buttons](buttons.mdx) and does not rely on JS responder system utilized by RN. Our touchable implementation follows the same API and aims to be a drop-in replacement for touchables available in React Native.
 
 React Native's touchables API can be found here:
--- a/docs/docs/gesture-handlers/about-handlers.md
+++ b/docs/docs/gesture-handlers/about-handlers.md
@@ -6,7 +6,7 @@ sidebar_position: 1
 ---
 
 :::warning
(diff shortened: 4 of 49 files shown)


latekvo:
Do you think it'd make sense to rebase these changes onto main instead of squashing them?
While I don't think it's a big deal, it's a bit messy to squash 4 PRs and some other changes together instead of keeping them separate

m-bert:
> Do you think it'd make sense to rebase these changes onto main instead of squashing them?

In our discussion with @j-piasecki we decided that it would be better to `cherry-pick` those changes into separate commit.

latekvo:
Should we close this PR?

latekvo:
>Why is that?

Just asking, since I understood you plan to either cherrypick those changes directly to main without a need for PR, or open a separate PR for all of these changes:

>In our discussion with @j-piasecki we decided that it would be better to cherry-pick those changes into separate commit.

m-bert:
What I meant was that we decided to cherry-pick those commits into one PR (this one) and then merge it. At least this is what we planned to do back then 😅

latekvo:
That makes sense, should we merge it then? 

Also, what is [this commit](https://github.com/software-mansion/react-native-gesture-handler/pull/3303/commits/09d0e85c788fbbbe507e95cffff60b05b6eceb01) doing in this PR? Is it relevant to the rest of the cherry-picked PRs?

Besides that change, I think all of these PRs have been already extensively tested individually.

m-bert:
> Also, what is [this commit](https://github.com/software-mansion/react-native-gesture-handler/pull/3303/commits/09d0e85c788fbbbe507e95cffff60b05b6eceb01) doing in this PR? It doesn't seem relevant to the rest of the cherry-picked PRs.

How does it not seem relevant? It is strictly connected to `get`/`set` migration, as without this commit there's infinite recursion while trying to access instance of `InteractionsManager`

> That makes sense, should we merge it then?

I think so, maybe before next release (cc @j-piasecki)

src/web/tools/VelocityTracker.ts
```diff
@@ -84,7 +84,7 @@ export default class VelocityTracker {
     return null;
   }
 
-  public getVelocity(): [number, number] {
+  public get velocity() {
```
j-piasecki, line 87: Doesn't this change the return type?

j-piasecki:
> I think so, maybe before next release

We are before the next release 😄

latekvo (approved):
Only one nit-pick

src/web/tools/Vector.ts
```diff
@@ -30,7 +30,7 @@ export default class Vector {
     return new Vector(velocity.x, velocity.y);
   }
 
-  get magnitude() {
+  public get magnitude() {
```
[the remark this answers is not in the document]
m-bert: As you can see in other files, it is just a convention to add access modifier to getters and setters. We can remove that, but then we should do this in every place

src/web/tools/VelocityTracker.ts
m-bert: It does to `number[]`. Changed in [b7b4eaa](https://github.com/software-mansion/react-native-gesture-handler/pull/3303/commits/b7b4eaab58d6732b0fd653b24403af43ffd8ca14)

j-piasecki (approved):
🙈

PR #3312, opened 2024-12-30
fix: pressable style on android

adrianolsk:
## Description
The pressed style is not being applied on Android,
investigating the source code and docs, I managed to get it working by adding 
```
shouldActivateOnStart(value: boolean) (Android only)
```
https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/native-gesture#shouldactivateonstartvalue-boolean-android-only

and removing this early return that was causing `isTouchPropagationAllowed.current` to never be set on android
```
 if (Platform.OS !== 'ios') {
       return;
 }
```

Reproduction of the issue (Pressed style not applied on Android)
https://snack.expo.dev/GGXCh7UBdjhl4KOuKCGXe
## Test plan

Tested fix on iPhone 16, Pixel 9 pro and Moto G8

### DIFF
--- a/src/components/Pressable/Pressable.tsx
+++ b/src/components/Pressable/Pressable.tsx
@@ -299,21 +299,15 @@ export default function Pressable(props: PressableProps) {
   // RNButton is placed inside ButtonGesture to enable Android's ripple and to capture non-propagating events
   const buttonGesture = useMemo(
     () =>
-      Gesture.Native()
-        .onBegin(() => {
-          // Android sets BEGAN state on press down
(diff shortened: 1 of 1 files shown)


sidferreira:
Hey @adrianolsk ! 😄 
I just reported the same: https://github.com/software-mansion/react-native-gesture-handler/issues/3367

latekvo:
I checked your repro, and i believe this should be fixed by https://github.com/software-mansion/react-native-gesture-handler/pull/3368.
Please let me know if that PR fixes this issue for you.

PR #3317, opened 2025-01-09
Bump `example` to `Expo 52`

latekvo:
## Description

This PR fixes some issues that occur when trying to use `RNGH` with `expo sdk 52`.

This bump will help with testing the [new expo features](https://docs.expo.dev/guides/tree-shaking/#recursive-optimizations) and [new react native features](https://reactnative.dev/docs/react-native-devtools) which will be useful for optimising Gesture Handler.


latekvo:
Tested on both `fabric` and `paper`, on `android` and `iOS`.
Example app builds and runs with no issues.

---

Regarding [851a72c](https://github.com/software-mansion/react-native-gesture-handler/pull/3317/commits/851a72cfdf80bb65987acd87790461f204bd5fb5), `SharedValue`'s signature was changed in either `3.15.x` or `3.16.x` versions of `react-native-reanimated`, which causes type linter errors as long as the versions between `Example` `package.json` and `root` `package.json` are misaligned.

example/package.json
```diff
@@ -14,31 +14,30 @@
...
     "hoist-non-react-statics": "^3.3.2",
     "invariant": "^2.2.4",
     "patch-package": "^6.5.1",
     "postinstall-postinstall": "^2.1.0",
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
-    "react-native": "0.75.4",
+    "react": "18.3.1",
+    "react-dom": "18.3.1",
+    "react-native": "0.76.5",
```
m-bert, line 36: Why not `0.76.6`?

m-bert (commented):
In my case I can't run `android` on `fabric`. There are some errors related to `react-native-async-storage` and codegen. Have you seen that before?

m-bert (approved):

example/package.json
m-bert: The solution was to remove caches from `~/.gradle/caches`

j-piasecki (approved):

PR #3319, opened 2025-01-09
[CI] Set Xcode version to 16.1

m-bert:
## Description

For some reason some of our checks on CI are executed on Xcode 16.2 - those checks fail. Setting version to 16.1 resolves this problem (this is temporary solution).

## Test plan

Check that CI passes.


latekvo (approved):
Tested on `Xcode 16.2`, on clean `main` branch, worked perfectly.

This is just a GitHub issue, so I think it's ok to ignore the problem for now.

m-bert:
Yes, it is just a GitHub issue, outside of CI it works correctly - that's the main reason behind this PR

PR #3322, opened 2025-01-10
Make `blocksGesture` symmetric on native iOS

j-piasecki:
## Description

Should fix https://github.com/software-mansion/react-native-gesture-handler/issues/3321

This PR makes the native implementation of `blocksGesture` symmetric. Previously, the logic was implemented only in `shouldBeRequiredToFailByGestureRecognizer`, now, it's also in `shouldRequireFailureOfGestureRecognizer`.

## Test plan

Tested on the code from the issue

### DIFF
--- a/apple/RNGestureHandler.mm
+++ b/apple/RNGestureHandler.mm
@@ -461,8 +461,8 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
     shouldRequireFailureOfGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer
 {
+  RNGestureHandler *handler = [RNGestureHandler findGestureHandlerByRecognizer:otherGestureRecognizer];
   if ([_handlersToWaitFor count]) {
-    RNGestureHandler *handler = [RNGestureHandler findGestureHandlerByRecognizer:otherGestureRecognizer];
     if (handler != nil) {
       for (NSNumber *handlerTag in _handlersToWaitFor) {
         if ([handler.tag isEqual:handlerTag]) {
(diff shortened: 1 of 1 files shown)


apple/RNGestureHandler.mm
```diff
@@ -471,6 +471,15 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
       }
     }
   }
+
+  if (handler != nil) {
+    for (NSNumber *handlerTag in handler->_handlersThatShouldWait) {
+      if ([_tag isEqual:handlerTag]) {
+        return YES;
+      }
```
m-bert, line 479: Shouldn't we also check `[_handlersThatShouldWait count]`, as above?

m-bert (commented):
Looks good! 

What I'd suggest is to move this `nil` check before other checks, like this:

<details>
<summary> code </summary>

```obj-c
- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
    shouldRequireFailureOfGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer
{
  RNGestureHandler *handler = [RNGestureHandler findGestureHandlerByRecognizer:otherGestureRecognizer];

  if(handler == nil) {
    return NO;
  }

  if ([_handlersToWaitFor count]) {
    for (NSNumber *handlerTag in _handlersToWaitFor) {
      if ([handler.tag isEqual:handlerTag]) {
        return YES;
      }
    }
  }


  for (NSNumber *handlerTag in handler->_handlersThatShouldWait) {
    if ([_tag isEqual:handlerTag]) {
      return YES;
    }
  }

  return NO;
}
```
 
</details>

or at least combine it with `and`:

```obj-c
if (handler != nil && [_handlersToWaitFor count]) {
```

apple/RNGestureHandler.mm
j-piasecki: I moved the null check up and removed the check for `[_handlersThatShouldWait count]` entirely - I don't really see the difference it makes when the list is empty.

gaearon:
Yes, this does seem to resolve the issue!

gaearon:
Curiously, moving to `blocksExternalGesture` (now that it's fixed on iOS) has also resolved a gesture conflict on Android. I don't have a minimal repro case for that. Just curious that they're not really symmetrical on Android either.

m-bert (approved):

Issue #2862, opened 2024-04-12
Combining Pan Gesture and Swipeable creates unexpected behavior


Athe81:
### Description

I have a GestureDetector with a pan gesture wraped around a Swipeable element. After wrapping the GestureDetector around the Swipeable element, the Swipeable Element does not work as expected. Before it opened or closed after releasing the finger. Now it stucks in the position.

### Steps to reproduce

1. Create a Swipeable element
2. Wrap it with a GestureDetector (gesture pan)
3. Test it

### Snack or a link to a repository

https://snack.expo.dev/@bigbart/bugreport-swipeable-list

### Gesture Handler version

2.14.0

### React Native version

0.73.6

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo managed workflow

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

m-bert:
Hi @Athe81! Why do you want to wrap `Swipeable` with `GestureDetector`? `Swipeables` already have gesture handlers under the hood. Moreover, they use old API of gesture handler along with Animated API, which is not supported by `GestureDetector` (as you can read in the [docs](https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/gesture-detector/)).

Athe81:
Hi @m-bert!

> Why do you want to wrap `Swipeable` with `GestureDetector`? `Swipeables` already have gesture handlers under the hood.

I want to give a haptic feedback to the user, if the `Swipeable` item is swiped more than a threshold. This is to signalize the user, that the action behind, will be executed if he releases his finger.
I don't know how to do this without `GestureDetector`.

m-bert:
I see. In that case you can try something like this:

```jsx
const Item = ({ id, title }) => {
  const windowWidth = Dimensions.get('window').width;
  const threshold = windowWidth / 3;
  const hasReachedThreshold = useSharedValue(false);
  const panRef = useRef();

  const panGesture = Gesture.Pan()
    .onUpdate((e) => {
      hasReachedThreshold.value = Math.abs(e.translationX) >= threshold;
      console.log(hasReachedThreshold.value);
    })
    .withRef(panRef);

  const renderLeftActions = (progress, dragX) => {
    return <View style={styles.left} />;
  };

  const renderRightActions = (progress, dragX) => {
    return <View style={styles.right} />;
  };

  return (
    <GestureDetector gesture={panGesture}>
      <View>
[...]

PR #3324, opened 2025-01-10
feat: Add `simultaneousWithExternalGesture` to ReanimatedSwipable

MrSltun:
## Description
This PR adds a new prop to `RenimatedSwipable` that allows adding a gesture config that can be used simultaneously with `ReanimatedSwipeable`'s gesture config.

The new prop is called `simultaneousWithExternalGesture`, it uses the the cross-component interaction methodology that's mentioned in the [documentation here](https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/gesture-composition/#simultaneouswithexternalgesture).

I also update the docs for it

### Motivation
I've been recently using ReanimatedSwipeable component, and I tried to add an additional gesture config to enable haptic feedback and "release to run function" kind of interaction

After looking through the issues in the repo, I found this issue #2862, I tried to apply it, but it turned out that it was only applicable to the Swipable component

I tried to find a way to add a gesture config that would work simultaneously with the swipeable one but I couldn't find a way to add it. So I looked through the documentation of RNGH for simultaneous gesture handlers and came up with this solution

## Test plan
I already tested it locally on both iOS and Android, and the callbacks work as expected

Not sure if there's something else to do here since this is my first contribution

### DIFF
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -206,7 +206,7 @@ export interface SwipeableProps
 
   /**
    * A base gesture object containing the configuration and callbacks to be
-   * used simultaneously with `Swipeable`'s gesture object.
+   * used with the swipeable's gesture handler.
    */
   simultaneousGesture?: GestureType;
 }
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -205,7 +205,7 @@ export interface SwipeableProps
   childrenContainerStyle?: StyleProp<ViewStyle>;
 
   /**
-   * A gesture object containing the configuration and callbacks to be
+   * An array of gesture objects containing the configuration and callbacks to be
    * used with the swipeable's gesture handler.
    */
   simultaneousWithExternalGestures?: Exclude<GestureRef, number>[];
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -205,10 +205,10 @@ export interface SwipeableProps
   childrenContainerStyle?: StyleProp<ViewStyle>;
 
   /**
-   * A base gesture object containing the configuration and callbacks to be
+   * A gesture object containing the configuration and callbacks to be
    * used with the swipeable's gesture handler.
    */
-  simultaneousGesture?: GestureType;
+  simultaneousWithExternalGesture?: GestureType;
 }
 
 export interface SwipeableMethods {
@@ -254,7 +254,7 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       onSwipeableClose,
       renderLeftActions,
       renderRightActions,
-      simultaneousGesture,
+      simultaneousWithExternalGesture,
       ...remainingProps
     } = props;
 
@@ -655,61 +655,69 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       [close, rowState]
     );
 
-    const panGesture = useMemo(
-      () =>
-        Gesture.Pan()
-          .enabled(enabled !== false)
-          .enableTrackpadTwoFingerGesture(enableTrackpadTwoFingerGesture)
-          .activeOffsetX([-dragOffsetFromRightEdge, dragOffsetFromLeftEdge])
-          .onStart(updateElementWidths)
-          .onUpdate(
-            (event: GestureUpdateEvent<PanGestureHandlerEventPayload>) => {
-              userDrag.value = event.translationX;
+    const panGesture = useMemo(() => {
+      const gesture = Gesture.Pan()
+        .enabled(enabled !== false)
+        .enableTrackpadTwoFingerGesture(enableTrackpadTwoFingerGesture)
+        .activeOffsetX([-dragOffsetFromRightEdge, dragOffsetFromLeftEdge])
+        .onStart(updateElementWidths)
+        .onUpdate(
+          (event: GestureUpdateEvent<PanGestureHandlerEventPayload>) => {
+            userDrag.value = event.translationX;
 
-              const direction =
-                rowState.value === -1
-                  ? SwipeDirection.RIGHT
-                  : rowState.value === 1
-                    ? SwipeDirection.LEFT
-                    : event.translationX > 0
-                      ? SwipeDirection.RIGHT
-                      : SwipeDirection.LEFT;
+            const direction =
+              rowState.value === -1
+                ? SwipeDirection.RIGHT
+                : rowState.value === 1
+                  ? SwipeDirection.LEFT
+                  : event.translationX > 0
+                    ? SwipeDirection.RIGHT
+                    : SwipeDirection.LEFT;
 
-              if (!dragStarted.value) {
-                dragStarted.value = true;
-                if (rowState.value === 0 && onSwipeableOpenStartDrag) {
-                  runOnJS(onSwipeableOpenStartDrag)(direction);
-                } else if (onSwipeableCloseStartDrag) {
-                  runOnJS(onSwipeableCloseStartDrag)(direction);
-                }
+            if (!dragStarted.value) {
+              dragStarted.value = true;
+              if (rowState.value === 0 && onSwipeableOpenStartDrag) {
+                runOnJS(onSwipeableOpenStartDrag)(direction);
+              } else if (onSwipeableCloseStartDrag) {
+                runOnJS(onSwipeableCloseStartDrag)(direction);
               }
+            }
 
-              updateAnimatedEvent();
-            }
-          )
-          .onEnd(
-            (event: GestureStateChangeEvent<PanGestureHandlerEventPayload>) => {
-              handleRelease(event);
-            }
-          )
-          .onFinalize(() => {
-            dragStarted.value = false;
-          }),
-      [
-        dragOffsetFromLeftEdge,
-        dragOffsetFromRightEdge,
-        dragStarted,
-        enableTrackpadTwoFingerGesture,
-        enabled,
-        handleRelease,
-        onSwipeableCloseStartDrag,
-        onSwipeableOpenStartDrag,
-        rowState,
-        updateAnimatedEvent,
-        updateElementWidths,
-        userDrag,
-      ]
-    );
+            updateAnimatedEvent();
+          }
+        )
+        .onEnd(
+          (event: GestureStateChangeEvent<PanGestureHandlerEventPayload>) => {
+            handleRelease(event);
+          }
+        )
+        .onFinalize(() => {
+          dragStarted.value = false;
+        });
+
+      // Add simultaneousWithExternalGesture only if it has a value
+      if (simultaneousWithExternalGesture) {
+        gesture.simultaneousWithExternalGesture(
+          simultaneousWithExternalGesture
+        );
+      }
+
+      return gesture;
+    }, [
+      dragOffsetFromLeftEdge,
+      dragOffsetFromRightEdge,
+      dragStarted,
+      enableTrackpadTwoFingerGesture,
+      enabled,
+      handleRelease,
+      onSwipeableCloseStartDrag,
+      onSwipeableOpenStartDrag,
+      rowState,
+      updateAnimatedEvent,
+      updateElementWidths,
+      userDrag,
+      simultaneousWithExternalGesture,
+    ]);
 
     useImperativeHandle(ref, () => swipeableMethods, [swipeableMethods]);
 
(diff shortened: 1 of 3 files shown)


latekvo:
I think this is very reasonable.

As far as I know, it's currently not possible to write interactions with components that utilise `Gestures` within themselves, without passing some references to the component. I think this is something worth looking into. 

One thing i'd request, is that you rename the prop to `simultaneousWithExternalGesture`, as we already use `simultaneousWithExternalGesture` for this functionally identical purpose.

I'm not sure if we want to add `requireExternalGestureToFail` and `blocksExternalGesture`, since user usually has access to these function on both the parent, and the child, while I think adding them to `Swipeable` would only allow the user to use them on the child (`Swipeable`) while not allowing for them to be used by the parent (`Gesture` that is wrapping `Swipeable`). Let's wait for @m-bert's and @j-piasecki's opinions on this.

<details>
<summary>
Collapsed Pan + Tap test
</summary>

```tsx
import React from 'react';

  const pan = Gesture.Pan().onStart(() => console.log('Pan'));
  const tap = Gesture.Tap().onStart(() => console.log('Tap'));
  return (
          <Text style={{ fontSize: 64, opacity: 0.25 }}>🎯</Text>
        </View>
      </GestureDetector>
    </GestureDetector>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
});

```

</details>


<details>
<summary>
Collapsed Pan + Tap + Swipeable test
</summary>

```tsx
import React from 'react';
import { StyleSheet, Text } from 'react-native';
import Animated, {
  SharedValue,
  useAnimatedStyle,

function LeftAction(_prog: SharedValue<number>, drag: SharedValue<number>) {
  const styleAnimation = useAnimatedStyle(() => {
    return {
      transform: [{ translateX: drag.value - 50 }],
    };
  });

  return (
    <Animated.View style={styleAnimation}>
      <Text style={styles.leftAction}>Text</Text>
    </Animated.View>
  );
}

function RightAction(_prog: SharedValue<number>, drag: SharedValue<number>) {
  const styleAnimation = useAnimatedStyle(() => {
    return {
      transform: [{ translateX: drag.value + 50 }],
    };
  });

  return (
    <Animated.View style={styleAnimation}>
      <Text style={styles.rightAction}>Text</Text>
    </Animated.View>
  );
}

  const pan = Gesture.Pan().onStart(() => console.log('Pan'));
  const tap = Gesture.Tap().onStart(() => console.log('Tap'));
  return (
        <ReanimatedSwipeable
          containerStyle={styles.swipeable}
          friction={2}
          leftThreshold={80}
          enableTrackpadTwoFingerGesture
          rightThreshold={40}
          renderLeftActions={LeftAction}
          renderRightActions={RightAction}>
          <Text>[Reanimated] Swipe me!</Text>
        </ReanimatedSwipeable>
      </GestureDetector>
    </GestureDetector>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
  swipeable: {
    height: 50,
    alignItems: 'center',
  },
  leftAction: {
    width: 50,
    height: 50,
    backgroundColor: 'crimson',
  },
  rightAction: {
    width: 50,
    height: 50,
    backgroundColor: 'purple',
  },
});

```

</details>

m-bert:
Hi @MrSltun! Thanks for this PR!

> One thing i'd request, is that you rename the prop to simultaneousWithExternalGesture, as we already use simultaneousWithExternalGesture for this functionally identical purpose, [more info here](https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/gesture-composition/#simultaneouswithexternalgesture).

Actually I'm not sure if we want to rename it in current state. Also, now when I look at it, I assume that `ReanimatedSwipeable` is wrapped with `GestureDetector`, which means that `simultaneousGesture` is used inside another `GestureDetector`. For that cases we have `simultaneousWithExternalGesture` prop, as combining gestures in this way may not work. 

What I think should be done here is to call this prop `simultaneousWithExternalGesture` and use it inside:

```diff
    const panGesture = useMemo(
      () =>
        Gesture.Pan()
          .enabled(enabled !== false)
          .enableTrackpadTwoFingerGesture(enableTrackpadTwoFingerGesture)
          .activeOffsetX([-dragOffsetFromRightEdge, dragOffsetFromLeftEdge])
          .onStart(() => {
            updateRightElementWidth();
          })
          .onUpdate(
            (event: GestureUpdateEvent<PanGestureHandlerEventPayload>) => {
              userDrag.value = event.translationX;

              const direction =
                rowState.value === -1
                  ? SwipeDirection.RIGHT
                  : rowState.value === 1
                    ? SwipeDirection.LEFT
                    : event.translationX > 0
                      ? SwipeDirection.RIGHT
                      : SwipeDirection.LEFT;

              if (!dragStarted.value) {
                dragStarted.value = true;
                if (rowState.value === 0 && onSwipeableOpenStartDrag) {
                  runOnJS(onSwipeableOpenStartDrag)(direction);
                } else if (onSwipeableCloseStartDrag) {
                  runOnJS(onSwipeableCloseStartDrag)(direction);
                }
              }

              updateAnimatedEvent();
            }
          )
          .onEnd(
            (event: GestureStateChangeEvent<PanGestureHandlerEventPayload>) => {
              handleRelease(event);
            }
          )
          .onFinalize(() => {
            dragStarted.value = false;
          })
+       .simultaneousWithExternalGesture(simultaneousGesture),
      [
        dragOffsetFromLeftEdge,
        dragOffsetFromRightEdge,
        dragStarted,
        enableTrackpadTwoFingerGesture,
        enabled,
        handleRelease,
        onSwipeableCloseStartDrag,
        onSwipeableOpenStartDrag,
        rowState,
        updateAnimatedEvent,
        updateRightElementWidth,
        userDrag,
      ]
    );
```

Could you please check if this works?

MrSltun:
>Actually I'm not sure if we want to rename it in current state. Also, now when I look at it, I assume that ReanimatedSwipeable is wrapped with GestureDetector, which means that simultaneousGesture is used inside another GestureDetector. For that cases we have simultaneousWithExternalGesture prop, as combining gestures in this way may not work.

My Approach was just to combine an external gesture with the `ReanimatedSwipeable` gesture (without a parent `GestureDetector` wrapping the `ReanimatedSwipeable`)

But I like the approach of having the `ReanimatedSwipeable ` gesture working simultaneously with a parent `GestureDetector`, I will update the code, test it, and report back here :)

latekvo:
Works great, test code used: 
<details>
<summary>
Collapsed repro
</summary>

```tsx
import React from 'react';
import Swipeable from 'react-native-gesture-handler/ReanimatedSwipeable';

  const pan = Gesture.Pan()
    .onBegin(() => console.log('began'))
    .onStart(() => console.log('active'))
    .onFinalize((_, success) => console.log(success ? 'ended' : 'cancelled'));

  return (
      <Swipeable
        leftThreshold={50}
        rightThreshold={50}
        renderLeftActions={() => {
          return (
            <View style={{ height: 80, width: 80, backgroundColor: 'yellow' }}>
              <Text>Left</Text>
            </View>
          );
        }}
        renderRightActions={() => {
          return (
            <View style={{ height: 80, width: 80, backgroundColor: 'magenta' }}>
              <Text>Right</Text>
            </View>
          );
        }}>
        <View
          style={{
            height: 80,
            backgroundColor: 'blue',
          }}
        />
      </Swipeable>
    </GestureDetector>
  );
}

```

</details>

Just one thing - the `Tap` in `Swipeable` isn't made simultaneous with the provided `simultaneousWithExternalGesture` prop. Unless I'm missing something, I think it should also be made simultaneous.

Also, I'm getting a type error when i pass my `pan` gesture to the `simultaneousWithExternalGesture` in the test code I provided, but as far as I see, `GestureType` is the correct type to use in this scenario. I'm not sure if this is a case of package version discrepancy or some other issue?
image

MrSltun:
Thank you for testing the changes @latekvo 🫶 

> Just one thing - the `Tap` in `Swipeable` isn't made simultaneous with the provided `simultaneousWithExternalGesture` prop. Unless I'm missing something, I think it should also be made simultaneous.

Hmm I tested the pan gestures, but didn't check the other types, I'll test all the gestures and see which works and which doesn't, and try to figure out why


> Also, I'm getting a type error when i pass my pan gesture to the simultaneousWithExternalGesture in the test code I provided, but as far as I see, GestureType is the correct type to use in this scenario. I'm not sure if this is a case of package version discrepancy or some other issue?

Oh that's weird! I didn't come across it, not sure why it's happening but I'll check

m-bert:
## TypeScript error

We've spent some time with @tjzel trying to find out what's going on. Turns out that this is a problem with our configuration (thank you @j-piasecki ❤️). 

Could you please update `tsconfig.json` in our example app? Otherwise CIs may fail

```diff
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "noEmit": true,
    "baseUrl": ".",
    "paths": {
      "react-native-gesture-handler": ["../src/index.ts"],
+     "react-native-gesture-handler/ReanimatedSwipeable": [
+       "../src/components/ReanimatedSwipeable.tsx"
+     ],
      "react-native-gesture-handler/jest-utils": ["../src/jestUtils/index.ts"]
    }
  },
  "include": ["src/**/*.ts", "src/**/*.tsx", "App.tsx"]
}

```

## Types

Regarding the code, I'd like to use the type that `simultaneousWithExternalGesture` expects, i.e. 

```tsx
simultaneousWithExternalGesture: Exclude<GestureRef, number>[];
```

And that requires a change in passing it into config:

```tsx
.simultaneousWithExternalGesture(
  ...(simultaneousWithExternalGesture ?? [])
),
```

In that case, example usage looks like this:

```tsx
<ReanimatedSwipeable
  simultaneousWithExternalGesture={[pan, tap]}
  ...
>
```

I know that this may be inconvenient with single gestures - we can change the type to:

```tsx
simultaneousWithExternalGesture: Exclude<GestureRef, number> | Exclude<GestureRef, number>[];
```

This will require a bit more logic when passing this prop into config, but it will be easier to use 😅

MrSltun:
@m-bert thank you for taking the time to test the change :)

I updated the the example app's tsconfig

Re: updating the `simultaneousWithExternalGesture` type,
I like the new approach since it allows multiple gestures to be passed to the `ReanimatedSwipeable` gesture handler :)

I updated the name as well to reflect that change to be `simultaneousWithExternalGestures`

I also tested the new change and it works as it should 🎉

m-bert:
Hi @MrSltun! Thanks for updating `tsconfig`. Could you please share code that you've used to test these changes? You can either add it into _**Test plan**_ section, or here in comments.

src/components/ReanimatedSwipeable.tsx
```diff
@@ -202,6 +203,12 @@ export interface SwipeableProps
    * apply `flex: 1`
    */
   childrenContainerStyle?: StyleProp<ViewStyle>;
+
+  /**
+   * An array of gesture objects containing the configuration and callbacks to be
+   * used with the swipeable's gesture handler.
+   */
+  simultaneousWithExternalGestures?: Exclude<GestureRef, number>[];
```
m-bert, line 211:
```suggestion
  simultaneousWithExternalGesture?: Exclude<GestureRef, number>[];
```

m-bert (changes requested):
Hi @MrSltun! Thank you for changing our `tsconfig` 😅 

### Prop name

Please change prop name to `simultaneousWithExternalGesture`. I know that it sounds better with `s` at the end, but this is how it is currently named in Gesture Handler and we would like to stick to it.

### Prop type

After looking at it I'd prefer to remove array if not necessary. It means that type of the prop should be

```tsx
simultaneousWithExternalGesture?:
  | Exclude<GestureRef, number>
  | Exclude<GestureRef, number>[];
```

To handle it correctly, I suggest adding the following logic into `useMemo`:

```js
      ...
        .onFinalize(() => {
          dragStarted.value = false;
        });

      if (!simultaneousWithExternalGesture) {
        return pan;
      }

      if (Array.isArray(simultaneousWithExternalGesture)) {
        pan.simultaneousWithExternalGesture(...simultaneousWithExternalGesture);
      } else {
        pan.simultaneousWithExternalGesture(simultaneousWithExternalGesture);
      }

      return pan;
```

### Tap gesture

Currently this prop **_is not_** used with `Tap` gesture that is present inside `Swipeable` - please add this 😅

MrSltun:
@m-bert Thank you for the detailed feedback :) 

I completely missed the tab gesture 😅 thank you for pointing it out!

I just updated the name of the variable, as well as the definition of the pan gesture, and I added the gesture to the tab one as well :)

m-bert (approved):
Thank you very much for this PR @MrSltun ❤️

latekvo (approved):

PR #3332, opened 2025-01-13
Fix swipeable on window resize

blazejkustra:

### DIFF
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -256,19 +256,8 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
     const leftWidth = useSharedValue<number>(0);
     const rightWidth = useSharedValue<number>(0);
 
-    // used for synchronizing layout measurements between JS and UI
-    const rightOffset = useSharedValue<number | null>(null);
-
     const showLeftProgress = useSharedValue<number>(0);
     const showRightProgress = useSharedValue<number>(0);
-
-    const updateRightElementWidth = useCallback(() => {
-      'worklet';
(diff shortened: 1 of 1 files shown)


src/components/ReanimatedSwipeable.tsx
```diff
@@ -452,7 +441,7 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
         openRight() {
           'worklet';
           // rightOffset and rowWidth are already much sooner than rightWidth
```
latekvo, line 443: I think this comment can be removed

src/components/ReanimatedSwipeable.tsx
```diff
@@ -463,14 +452,13 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
         },
       }),
       [
-        leftWidth,
-        rightOffset,
-        rowWidth,
```
latekvo, line 468: These dependencies don't need the `.value` prop to work, and having it there significantly degrades performance. More details [here](https://github.com/software-mansion/react-native-gesture-handler/pull/3165)

latekvo (commented):
Works great, just a few minor requests

latekvo (changes requested):
It seems these changes brake the right action of `GmailStyleSwipeableRow` in the `New Swipeable` example in the `New api` section of our example app.

I think this may be caused by `rightWidth.value` being updated only in `onLayout` which might be loaded too early to read the panel width.
Previously it was updated multiple times using the `updateRightElementWidth()` function. 
As far as I remember, this was done to fix non-deterministic element loading.

latekvo:
Fixed in #3341

PR #3340, opened 2025-01-16
Fix swipeable overlap bug

latekvo:
## Description

Reported by @blazejkustra 

Before this PR, the left action panel could've been partially or entirely covered the right action panel, when the left action panel was opened.

Similarily, the left action panel could've been seen below the right action panel, when the right action panel was opened.
 
## Test plan

- open `New swipeable` example in the `Gesture Handler Example App`
- swipe right on the `Gmail` swipeable
- see how it was `red` (right action) before this PR, and how it is `green` (left action) thanks to this PR

### DIFF
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -536,6 +536,7 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       [
         appliedTranslation,
         renderRightActions,
+        rightActionAnimation,
         rightOffset,
         showRightProgress,
         swipeableMethods,
@@ -694,18 +695,6 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
         </Animated.View>
       </GestureDetector>
     );
-
-    useAnimatedStyle(() => {
-      console.log('----------');
-      console.log('rowWidth', rowWidth.value);
-      console.log('leftWidth', leftWidth.value);
-      console.log('rightWidth', rightWidth.value);
-      console.log('rightOffset', rightOffset.value);
-      console.log('showLeftProgress', showLeftProgress.value);
-      console.log('showRightProgress', showRightProgress.value);
-      console.log('userDrag', userDrag.value);
-      return {};
-    });
 
     return testID ? (
       <View testID={testID}>{swipeableComponent}</View>
--- a/src/components/ReanimatedSwipeable.tsx
+++ b/src/components/ReanimatedSwipeable.tsx
@@ -414,11 +414,11 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
           const progressTarget = toValue === 0 ? 0 : 1;
 
           showLeftProgress.value =
-            leftWidth.value > 0
+            showLeftProgress.value > 0
               ? withSpring(progressTarget, progressSpringConfig)
               : 0;
           showRightProgress.value =
-            rightWidth.value > 0
+            showRightProgress.value > 0
               ? withSpring(progressTarget, progressSpringConfig)
               : 0;
 
@@ -481,9 +481,15 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       [rowWidth]
     );
 
+    const leftActionAnimation = useAnimatedStyle(() => {
+      return {
+        transform: [{ translateX: showLeftProgress.value === 0 ? -10000 : 0 }],
+      };
+    });
+
     const leftElement = useCallback(
       () => (
-        <Animated.View style={[styles.leftActions]}>
+        <Animated.View style={[styles.leftActions, leftActionAnimation]}>
           {renderLeftActions?.(
             showLeftProgress,
             appliedTranslation,
@@ -498,6 +504,7 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       ),
       [
         appliedTranslation,
+        leftActionAnimation,
         leftWidth,
         renderLeftActions,
         showLeftProgress,
@@ -505,9 +512,15 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       ]
     );
 
+    const rightActionAnimation = useAnimatedStyle(() => {
+      return {
+        transform: [{ translateX: showRightProgress.value === 0 ? 10000 : 0 }],
+      };
+    });
+
     const rightElement = useCallback(
       () => (
-        <Animated.View style={[styles.rightActions]}>
+        <Animated.View style={[styles.rightActions, rightActionAnimation]}>
           {renderRightActions?.(
             showRightProgress,
             appliedTranslation,
@@ -681,6 +694,18 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
         </Animated.View>
       </GestureDetector>
     );
+
+    useAnimatedStyle(() => {
+      console.log('----------');
+      console.log('rowWidth', rowWidth.value);
+      console.log('leftWidth', leftWidth.value);
+      console.log('rightWidth', rightWidth.value);
+      console.log('rightOffset', rightOffset.value);
+      console.log('showLeftProgress', showLeftProgress.value);
+      console.log('showRightProgress', showRightProgress.value);
+      console.log('userDrag', userDrag.value);
+      return {};
+    });
 
     return testID ? (
       <View testID={testID}>{swipeableComponent}</View>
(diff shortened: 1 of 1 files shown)


src/components/ReanimatedSwipeable.tsx
```diff
@@ -481,9 +481,15 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       [rowWidth]
     );
 
+    const leftActionAnimation = useAnimatedStyle(() => {
+      return {
+        transform: [{ translateX: showLeftProgress.value === 0 ? -10000 : 0 }],
```
m-bert, line 486: I assume that there's no special meaning in `10000` number, but can we at least make it a constant? Also, isn't there something more descriptive that we could use (like `Dimensions.width` for example)?

src/components/ReanimatedSwipeable.tsx
```diff
@@ -498,16 +504,23 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
...
         leftWidth,
         renderLeftActions,
         showLeftProgress,
         swipeableMethods,
       ]
     );
 
+    const rightActionAnimation = useAnimatedStyle(() => {
+      return {
+        transform: [{ translateX: showRightProgress.value === 0 ? 10000 : 0 }],
```
m-bert, line 517: Same as above

src/components/ReanimatedSwipeable.tsx
latekvo: The `10000` number is the number which was used in the original `Swipeable`. ([link](https://github.com/software-mansion/react-native-gesture-handler/commit/ce5298025bf6e2d1ea48219bc0fd52365e4ef2c0))

I think using `Dimensions.width + 1px` makes perfect sense, I'll also add `overflow: hidden` style to those elements, so that [removeClippedSubviews](https://reactnative.dev/docs/view#removeclippedsubviews) may be used to optimise them.

Of course, I'll make this number a constant as well.

latekvo: Done in [a64d29b](https://github.com/software-mansion/react-native-gesture-handler/pull/3340/commits/a64d29b46002e0ddcd71ff32339fd9ab57c80ded)

src/components/ReanimatedSwipeable.tsx
latekvo: Done in [a64d29b](https://github.com/software-mansion/react-native-gesture-handler/pull/3340/commits/a64d29b46002e0ddcd71ff32339fd9ab57c80ded)

src/components/ReanimatedSwipeable.tsx
```diff
@@ -481,9 +482,18 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
       [rowWidth]
     );
 
+    // React docs: "call this function on every render, rather than caching the value"
```
m-bert, line 485: Could you paste link to the docs instead of adding _**React docs:**_? 😅

latekvo: This string is only available in the docstring of the `Dimensions.get(...)` function, the docs don't mention this recommendation: https://reactnative.dev/docs/dimensions#get
<details>
<summary>
Collapsed docstring screenshot
</summary>

image

</details>


Should i change it to something like:
`React docstring: "call this function on every render, rather than caching the value"`?

I left this comment in the first place, because initially i wanted to memoize this function, but randomly stumbled onto the docstring recommending agains caching it. I hope this comment will prevent any bugs that may occur in future if someone tries to optimize `Swipeable` by caching this value.

src/components/ReanimatedSwipeable.tsx
m-bert: Why is `overflow` based on progress? Can't it be just `'hidden'`?

src/components/ReanimatedSwipeable.tsx
latekvo: Sounds good, updated the string in [8e12c9e](https://github.com/software-mansion/react-native-gesture-handler/pull/3340/commits/8e12c9ec38d097cf43c30e0da621ac150923b97d)

src/components/ReanimatedSwipeable.tsx
latekvo: It can. Fixed in [07b212b](https://github.com/software-mansion/react-native-gesture-handler/pull/3340/commits/07b212b3a3d0a94113979072a713a3b1bce3e7f6)

src/components/ReanimatedSwipeable.tsx
```diff
@@ -700,9 +729,11 @@ const styles = StyleSheet.create({
   leftActions: {
     ...StyleSheet.absoluteFillObject,
     flexDirection: I18nManager.isRTL ? 'row-reverse' : 'row',
+    overflow: 'hidden',
```
m-bert, line 732: I know I've requested this, but as far as I remember, when we were talking about it you've said that it may break scroll inside side panels. Have you tested that?

latekvo: When there's a scroll list inside one of the actions, the `overflow: hidden` is not applied to the scroll, but to the `View` that is wrapping said scroll, resulting in no issues. 
Tested on:

<details>
<summary>
Collapsed repro
</summary>

```tsx
import React, { memo, ReactElement, useCallback } from 'react';
import { StyleSheet, View, FlatList, Pressable, Text } from 'react-native';
import {
  GestureHandlerRootView,
  ScrollView,
import Swipeable from 'react-native-gesture-handler/ReanimatedSwipeable';

export interface Client {
  id: string;
  firstName: string;
  lastName?: string;
}

interface SearchResultProps {
  client: Client;
  renderRightActions: (client: Client) => ReactElement;
}

export const SearchResult = memo(
  ({ client, renderRightActions }: SearchResultProps) => {
    return (
      // This Swipeable seems to cause the issue
      <Swipeable
        renderRightActions={() => renderRightActions(client)}
        containerStyle={styles.searchResultWrapper}>
        <View style={styles.searchResult}>
          <Pressable onPress={() => console.log('press')}>
            <View style={styles.searchResultRow}>
              <Text style={styles.searchResultText}>
                {client.firstName} {client.lastName}
              </Text>
            </View>
          </Pressable>
        </View>
      </Swipeable>
    );
  },
  (prevProps, nextProps) => {
    return (
      prevProps.client.firstName === nextProps.client.firstName &&
      prevProps.client.lastName === nextProps.client.lastName
    );
  }
);

/**
 * FlatList
 */
export default function EmptyExample(): ReactElement {
  // Create a render function for the left swipe action (delete)
  const renderRightActions = useCallback(
    () => (
      <ScrollView>
        <Pressable
          onPress={() => {
            console.log('tried deleting');
          }}
          style={styles.actionButton}
        />
        <Pressable
          onPress={() => {
            console.log('tried deleting too');
          }}
          style={styles.actionButton}
        />
      </ScrollView>
    ),
    []
  );

  const renderitem = useCallback(
    ({ item }: { item: Client; index: number }) => (
      <SearchResult client={item} renderRightActions={renderRightActions} />
    ),
    [renderRightActions]
  );

  return (
    <GestureHandlerRootView>
      <View style={styles.searchResultsContainer}>
        <FlatList
          data={CLIENTS}
          keyExtractor={(i) => i.id}
          renderItem={renderitem}
        />
      </View>
    </GestureHandlerRootView>
  );
}

  searchResultsContainer: {
    flex: 4,
  },
  actionButton: {
    width: 80,
    height: 80,
    borderRadius: 999,
    backgroundColor: 'red',
  },
  searchResultWrapper: {
    height: 80,
    borderRadius: 999,
    backgroundColor: 'gray',
    marginBottom: 12,
  },
  searchResult: {
    marginBottom: 12, // Make up for 'gap' not working in FlatList
  },
  searchResultRow: {
    height: 80,
    flexDirection: 'row',
    gap: 16,
    alignItems: 'center',
    backgroundColor: 'white',
    borderRadius: 999,
  },
  searchResultText: {
    fontSize: 16,
    lineHeight: 24,
    color: 'black',
  },
});

const CLIENTS = [
  {
    id: 'a',
    firstName: 'a a a a a a a a',
    lastName: 'a a a a a a a a',
  },
  {
    id: 'b',
    firstName: 'b b b b b b b b',
    lastName: 'b b b b b b b b',
  },
  {
    id: 'c',
    firstName: 'c c c c c c c c',
    lastName: 'c c c c c c c c',
  },
  {
    id: 'd',
    firstName: 'd d d d d d d d',
    lastName: 'd d d d d d d d',
  },
  {
    id: 'e',
    firstName: 'e e e e e e e e',
    lastName: 'e e e e e e e e',
  },
  {
    id: 'f',
    firstName: 'f f f f f f f f',
    lastName: 'f f f f f f f f',
  },
];

```

</details>

m-bert (approved):

src/components/ReanimatedSwipeable.tsx
m-bert: One thing I'd add here is that on web it is hard to close this swipeable after opening it, but I guess it is because of `ScrollView`, could you check that?

PR #3343, opened 2025-01-17
Don’t pass gesture handler wrapper props to child

oblador:
## Description

When using useEvent from Reanimated together with PureNativeButton, I noticed that I would receive duplicate events. Upon further investigation I found that this was because createHandler renames the handlers and when reanimated attaches the worklet it does so twice because the same handler is available under different props.

This PR fixes this problem by simply not passing wrapper specific props to the child.

### DIFF
--- a/src/handlers/createNativeWrapper.tsx
+++ b/src/handlers/createNativeWrapper.tsx
@@ -29,17 +29,23 @@ export default function createNativeWrapper<P>(
     P & NativeViewGestureHandlerProps
   >((props, ref) => {
     // Filter out props that should be passed to gesture handler wrapper
-    const gestureHandlerProps = Object.keys(props).reduce(
+    const { gestureHandlerProps, childProps } = Object.keys(props).reduce(
       (res, key) => {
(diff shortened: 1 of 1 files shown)


vovacodes:
Also ran into this issue, so would appreciate merging this in 🙏

m-bert (approved):

latekvo:
I confirm this issue occurs on `RectButton` with `enabled={false}`.
Also, it **doesn't** occur on `Gesture.Tap().enabled(false);`.

oblador:
Thanks! Seems this prop was used by both place (checked, and it seems like this is the only prop behaving this way). Updated PR - please re-check!

m-bert (approved):
Looks good now, thanks for this PR!

vovacodes:
@m-bert is there a chance to make a patch release with this change (obviously unless you planned a release soon anyway)? I'd love to make a change to use this component in prod but currently blocked by this bug in the latest version. I already verified that this commit works perfectly for us 🙏

m-bert:
Yes, we plan to release patch version in this week 😅

PR #3346, opened 2025-01-20
exports PressableStateCallbackType

sidferreira:
## Description

This is an absolute small fix, just exporting the `PressableStateCallbackType` type.

The type was already marked as `export` but not included in the other files.

### DIFF
--- a/src/components/Pressable/PressableProps.tsx
+++ b/src/components/Pressable/PressableProps.tsx
@@ -1,22 +1,15 @@
 import {
-  ColorValue,
   AccessibilityProps,
   ViewProps,
   Insets,
   StyleProp,
   ViewStyle,
+  PressableStateCallbackType as RNPressableStateCallbackType,
+  PressableAndroidRippleConfig as RNPressableAndroidRippleConfig,
 } from 'react-native';
 
-export interface PressableStateCallbackType {
-  readonly pressed: boolean;
-}
-
-export interface PressableAndroidRippleConfig {
-  color?: null | ColorValue | undefined;
-  borderless?: null | boolean | undefined;
-  radius?: null | number | undefined;
-  foreground?: null | boolean | undefined;
-}
+export type PressableStateCallbackType = RNPressableStateCallbackType;
+export type PressableAndroidRippleConfig = RNPressableAndroidRippleConfig;
 
 export type InnerPressableEvent = {
   changedTouches: InnerPressableEvent[];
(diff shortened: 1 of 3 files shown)


latekvo:
Thank you for spotting this! 
Just one request - could you please run `prettier` over these two files?

sidferreira:
> Thank you for spotting this! Just one request - could you please run `prettier` over these two files?

sorry the delay, it is done :)

latekvo:
Sorry for another request, but could you please also change this prop's name to `PressableState`?
I think it's more descriptive, but of course I'm open to other suggestions.

sidferreira:
@latekvo I hate the name too, but it is the same as used in RN (https://github.com/facebook/react-native/blob/82cc465645d7a765707f703a14b4af809a018df8/packages/react-native/Libraries/Components/Pressable/Pressable.d.ts#L24) so I suggest keeping it like this for the sake of "drop-in replacement".

We can also remove RNGH own type and use RN one tho...

latekvo:
You're right, I missed that. In that case, I agree it's better to keep it how it is

latekvo (dismissed):

latekvo:
> We can also remove RNGH own type and use RN one tho

We could do that for `PressableAndroidRippleConfig` as well, I'm not sure if there was some a reason for not doing that in the first place though.

(CC @m-bert)

sidferreira:
@latekvo I like that :)

latekvo:
Hey, talked with @m-bert, you can proceed with the following notation:

```tsx
import {
  ...
  PressableStateCallbackType as RNPressableStateCallbackType,
  PressableAndroidRippleConfig as RNPressableAndroidRippleConfig,
} from 'react-native';

export type PressableStateCallbackType = RNPressableStateCallbackType;
export type PressableAndroidRippleConfig = RNPressableAndroidRippleConfig;

```

After that, this PR will be ready to be merged 👍

sidferreira:
@latekvo may I use 

```

latekvo (approved):

latekvo:
Once again, thanks a lot for this fix :)

sidferreira:
> Once again, thanks a lot for this fix :)

Glad to help. To be sure: you guys handle the merging, right?

latekvo:
>Glad to help. To be sure: you guys handle the merging, right?

Yes, I was just waiting for the CIs to pass

PR #3348, opened 2025-01-21
Emit changes to pointersInside

oblador:
## Description

Currently `onHandlerStateChange` will only send events whenever the handler state changes. However for the use case where one wants to resign highlight from the button whenever the user moves their pointer away from it, the only way is to use the `onGestureEvent`. This handler is however exceptionally noisy especially on Android, to the extent where I get UI thread frame drops when using worklets.

This MR emits additional events for the quite rare case where `pointersInside` prop changes for active gesture handlers to support this use case in a performant way.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -363,6 +363,7 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
     x = adaptedTransformedEvent.x
     y = adaptedTransformedEvent.y
     numberOfPointers = adaptedTransformedEvent.pointerCount
+    val wasWithinBounds = isWithinBounds
     isWithinBounds = isWithinBounds(view, x, y)
     if (shouldCancelWhenOutside && !isWithinBounds) {
       if (state == STATE_ACTIVE) {
@@ -386,6 +387,8 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
       sourceEvent.action == MotionEvent.ACTION_HOVER_EXIT
     ) {
       onHandleHover(adaptedTransformedEvent, adaptedSourceEvent)
+    } else if (sourceEvent.action == MotionEvent.ACTION_MOVE && state == STATE_ACTIVE && wasWithinBounds != isWithinBounds) {
+      orchestrator!!.onHandlerStateChange(this, state, state)
     } else {
       onHandle(adaptedTransformedEvent, adaptedSourceEvent)
     }
--- a/apple/RNGestureHandler.mm
+++ b/apple/RNGestureHandler.mm
@@ -273,7 +273,7 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
            forViewWithTag:(nonnull NSNumber *)reactTag
             withExtraData:(RNGestureHandlerEventExtraData *)extraData
 {
-  if (state != _lastState) {
+  if (state != _lastState || state == RNGestureHandlerStateActive) {
     // don't send change events from END to FAILED or CANCELLED, this may happen when gesture is ended in `onTouchesUp`
     // callback
     if (_lastState == RNGestureHandlerStateEnd &&


latekvo:
Hey, do i understand correctly that you want to detect a touch leaving the handled area, while the handler is still active?
Calling `onHandlerStateChange` despite lack of state change sounds like a significant anti-pattern, and will likely lead to some issues. 

I think you should be able to achieve this effect using `Gesture.Pan`, and it's `onUpdate` callback, which gives you access to touch coordinates on each touch movement. You should be able to use those coordinates to determine if the touch is inside, or outside your desired area.

oblador:
Yes exactly. Since the `pointersInside` prop is emitted in the `onHandlerStateChange` I expect it to get updated whenever it changes. `onUpdate` is incredibly spammy on Android and tanks performance when using a worklet.

latekvo:
I did some tests comparing `onUpdate`, and I see what you mean - using `onUpdate` to perform UI updates is approximately 80% more expensive than doing so without using `onUpdate`. This likely shouldn't be the case, and I'll look into why this issue might be occurring.

<details>
<summary>
Collapsed repro
</summary>

```tsx
import {
  Gesture,
  GestureDetector,
  ScrollView,
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withSpring,

  console.log('rerendering');
  let measurementStart = useRef(performance.now());
  let measurementCount = useRef(0);

  const spamCounter = useSharedValue(0);
  const highlight = useSharedValue(false);

    measurementCount.current++;
    if (measurementCount.current > 10) {
      // used to confirm same amount of updates-per-frame occurs both from `onUpdate` and from `spamCounter`
      console.log(
        '10 updates took:',
        (performance.now() - measurementStart.current).toFixed(2),
        'ms'
      );
      measurementCount.current = 0;
      measurementStart.current = performance.now();
    }

    const randomOffset = Math.random() * 2;

    if (spamCounter.value > 0) {
      spamCounter.value++;
      highlight.value = !highlight.value;
    }

    return {
      backgroundColor: highlight.value ? '#f2f2f2' : '#ffcfcf',
      transform: [
        {
          translateX: highlight.value
            ? withSpring(-randomOffset)
            : randomOffset,
        },
        {
          translateY: highlight.value
            ? -randomOffset
            : withSpring(randomOffset),
        },
      ],
    };
  });

  const pan = Gesture.Pan().onUpdate(() => {
    highlight.value = !highlight.value;
  });

  const tap = Gesture.Tap().onStart(() => {
    if (spamCounter.value > 0) {
      spamCounter.value = 0;
    } else {
      spamCounter.value++;
    }
  });

  return (
        <Animated.View>
          <Text style={{ fontSize: 32, opacity: 0.25 }}>
            Toggle REA-ONLY spam
          </Text>
        </Animated.View>
      </GestureDetector>
        <Animated.View style={[animatedStyle]}>
          <Text style={{ fontSize: 128, opacity: 0.25 }}>🎯</Text>
        </Animated.View>
      </GestureDetector>
      <ScrollView>
        {Array(100)
          .fill(0)
          .map((_, idx) => (
            <Animated.View style={[animatedStyle]} key={idx}>
              <Text style={{ fontSize: 128, opacity: 0.25 }}>🎯</Text>
            </Animated.View>
          ))}
      </ScrollView>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
});

```

</details>

Despite this, i don't think sending out `onHandlerStateChange` despite no state changes is a good idea. 

(CC @m-bert @j-piasecki)

m-bert:
> Despite this, i think sending out onHandlerStateChange despite no state changes is a bad idea.

Yes, as the name suggests, this callback should be associated with state change. Moreover, this would be a change in old API, which will be deprecated in next minor release.

oblador:
> Moreover, this would be a change in old API, which will be deprecated in next minor release.
😅 I hope you meant next major release. What specifically will you deprecate? The native button components?

m-bert:
I meant next _**minor**_ release. Those deprecations are already on `main` branch and include deprecations of:

- Outdated components (`Swipeable`, `DrawerLayout`, `Touchables`)
- Old API (i.e. API known from Gesture Handler v1)

j-piasecki:
I'll close this in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/4102. Update events for `NativeViewGestureHandler` will be dispatched only if the payload differs from the previously sent one.

PR #3350, opened 2025-01-21
Cancel failed buttons in reset

oblador:
## Description

This PR is an attempt at fixing bug reported in https://github.com/software-mansion/react-native-gesture-handler/issues/2160. 

Essentially when having a button with `shouldCancelWhenOutside={true}` in a `ScrollView` and scrolling quickly, especially when at the bottom where the scrollview exhibit rubber band effect, all buttons becomes unresponsive. This is because `touchResponder` is never freed after it goes into `FAILED` state, which typically otherwise happens when `ACTION_UP` is emitted – something that never happens. 

I'm not sure of the ideal solution, but I'd figured I share my half-baked workaround to emit `CANCEL` in `onReset` if the button is stuck in `FAILED`. Curious to hear suggestions on how to improve this.

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -770,6 +770,7 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
     }
   }
   fun reset() {
+    onReset()
     view = null
     orchestrator = null
     Arrays.fill(trackedPointerIDs, -1)
@@ -777,7 +778,6 @@ open class GestureHandler<ConcreteGestureHandlerT : GestureHandler<ConcreteGestu
     trackedPointersCount = 0
     trackedPointers.fill(null)
     touchEventType = RNGestureHandlerTouchEvent.EVENT_UNDETERMINED
-    onReset()
   }
 
   fun withMarkedAsInBounds(closure: () -> Unit) {
--- a/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
@@ -161,6 +161,9 @@ class NativeViewGestureHandler : GestureHandler<NativeViewGestureHandler>() {
   }
 
   override fun onReset() {
+    if (state == STATE_FAILED && view != null) {
+      onCancel()
+    }
     this.hook = defaultHook
   }
 


oblador:
Hi I made a minimal reproduction here: https://snack.expo.dev/@oblador/gesture-handler-list-reproduction

Note that this is Android only, see a video of the reproduction where you can notice that before I scroll the first list item responds with a highlight, but afterwards none do:

https://github.com/user-attachments/assets/345f7d82-269f-4066-add4-09cd1b865b55

oblador:
For what it's worth, I've managed to get into this state even without `shouldCancelWhenOutside={true}`, but it's just much easier/frequent in this mode. Basically it seems to apply to any case where the handler enter failed state. It's also possible to "resurrect" the app if you know which button caused it to die by scrolling slowly on that specific button, which is in line with above explanation because then `ACTION_UP` is emitted.

RohovDmytro:
I also rendomly experience when in a ScrollView there is two buttons and the second one overlap the first one (it's FAB).

Sometimes pressing a FAB buttons also triggers the underlying button (the ripple animation starts). FAB triggers navigation. Then the app becomes unresponsive to onPress events. Navigating back I can see that ripple effect got stuck (never finished) on a first button.

This PR might help with that too.

latekvo (commented):
Hey, I tested your solution on your repro, and I still managed to get the `RectButton`s stuck in the active state. 

It looks like this issue is caused by `NativeViewGestureHandler` either never leaving the `ACTIVE` state, or being removed from `gestureHandlers` array before being reset or finished.

When i managed to get `RectButton` stuck, `onCancel`, `onFail` and `onReset` were never called, suggesting `NativeViewGestureHandler` was never `finished` in the first place (`onFail` doesn't exist on main, added it for debugging). 

I've also tried different approaches, but haven't solved this issue yet. 
I'll dig into this deeper, and keep you updated when I find any solution.

oblador:
@RohovDmytro I think your issue is a different one from this, but I came across it today as well, and created a separate issue for it here: https://github.com/software-mansion/react-native-gesture-handler/issues/3370

oblador:
@latekvo Any progress on your end?

williamsaar:
Hey @j-piasecki! We're on RNGH 2.28 and still experiencing an issue. We are using a modal lib (@gorhom/bottom-sheet) with a pressable button to close the modal. When clicking it on android it sometimes "gets stuck" with the ripple on it and all other pressables stops responding to onClick-events.

m-bert:
Hi! This seems to be fixed (at least on `main` branch). Let me know if this issue still persists and if we should reopen this PR. 

@williamsaar if it still happens, could you prepare reproduction? Once ready, you can open new issue.

PR #3357, opened 2025-01-27
Add `testing-library` compatibility props

latekvo:
## Description

This PR adds 3 props to the `Pressable` used for compatibility with [react-native-testing-library](https://github.com/callstack/react-native-testing-library)

Full discussion: https://github.com/callstack/react-native-testing-library/issues/1738

### DIFF
--- a/src/components/GestureButtonsProps.ts
+++ b/src/components/GestureButtonsProps.ts
@@ -50,6 +50,24 @@ export interface RawButtonProps
    * Style object, use it to set additional styles.
    */
   style?: StyleProp<ViewStyle>;
+
+  /**
+   * Used for testing-library compatibility, not passed to the native component.
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  testOnly_onPress?: Function | null;
+
+  /**
+   * Used for testing-library compatibility, not passed to the native component.
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  testOnly_onPressIn?: Function | null;
+
+  /**
+   * Used for testing-library compatibility, not passed to the native component.
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  testOnly_onPressOut?: Function | null;
 }
 interface ButtonWithRefProps {
   innerRef?: React.ForwardedRef<React.ComponentType<any>>;
--- a/src/components/GestureButtonsProps.ts
+++ b/src/components/GestureButtonsProps.ts
@@ -68,6 +68,12 @@ export interface RawButtonProps
    */
   // eslint-disable-next-line @typescript-eslint/ban-types
   testOnly_onPressOut?: Function | null;
+
+  /**
+   * Used for testing-library compatibility, not passed to the native component.
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  testOnly_onLongPress?: Function | null;
 }
 interface ButtonWithRefProps {
   innerRef?: React.ForwardedRef<React.ComponentType<any>>;
--- a/src/components/Pressable/Pressable.tsx
+++ b/src/components/Pressable/Pressable.tsx
@@ -380,7 +380,7 @@ export default function Pressable(props: PressableProps) {
       ? children({ pressed: pressedState })
       : children;
 
-  const inJestEnv = typeof jest !== 'undefined';
+  const inTestEnv = process.env.NODE_ENV === 'test';
 
   return (
     <GestureDetector gesture={gesture}>
@@ -393,9 +393,9 @@ export default function Pressable(props: PressableProps) {
         rippleColor={processColor(android_ripple?.color ?? defaultRippleColor)}
         rippleRadius={android_ripple?.radius ?? undefined}
         style={[pointerStyle, styleProp]}
-        testOnly_onPress={inJestEnv ? onPress : undefined}
-        testOnly_onPressIn={inJestEnv ? onPressIn : undefined}
-        testOnly_onPressOut={inJestEnv ? onPressOut : undefined}>
+        testOnly_onPress={inTestEnv ? onPress : undefined}
+        testOnly_onPressIn={inTestEnv ? onPressIn : undefined}
+        testOnly_onPressOut={inTestEnv ? onPressOut : undefined}>
         {childrenProp}
         {__DEV__ ? (
           <PressabilityDebugView color="red" hitSlop={normalizedHitSlop} />
--- a/src/components/Pressable/Pressable.tsx
+++ b/src/components/Pressable/Pressable.tsx
@@ -20,7 +20,7 @@ import {
 } from './utils';
 import { PressabilityDebugView } from '../../handlers/PressabilityDebugView';
 import { GestureTouchEvent } from '../../handlers/gestureHandlerCommon';
-import { INT32_MAX } from '../../utils';
+import { INT32_MAX, isTestEnv } from '../../utils';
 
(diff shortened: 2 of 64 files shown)


src/components/GestureButtonsProps.ts
```diff
@@ -50,6 +50,24 @@ export interface RawButtonProps
    * Style object, use it to set additional styles.
    */
   style?: StyleProp<ViewStyle>;
+
+  /**
+   * Used for testing-library compatibility, not passed to the native component.
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  testOnly_onPress?: Function | null;
```
mdjastrzebski, line 58: What kind of argument should be passed to `testOnly_onPress`?

src/components/Pressable/Pressable.tsx
```diff
@@ -380,6 +380,8 @@ export default function Pressable(props: PressableProps) {
       ? children({ pressed: pressedState })
       : children;
 
+  const inJestEnv = typeof jest !== 'undefined';
```
mdjastrzebski, line 383: Perhaps we should use `process.env.NODE_ENV === 'test';` as more typical option for checking test env.

mdjastrzebski (commented):
Generally looks good, I've left some recommended tweaks

src/components/Pressable/Pressable.tsx
latekvo: Fixed in [e51fdcd](https://github.com/software-mansion/react-native-gesture-handler/pull/3357/commits/e51fdcd44bd18e52d79843266f18c07312dea4b5)

latekvo:
Note on [c63a561](https://github.com/software-mansion/react-native-gesture-handler/pull/3357/commits/c63a561547acbf8ec715b036091b5e504b02d33d) - I tried using `ReturnType<typeof setTimeout>`, manually creating `globals.d.ts` type resolutions, and finding ways of giving one of the typing libraries higher priority. Unfortunately none of these methods worked for all instances of conflicting types, with `node` types always overwriting `react-native` types.

src/components/Pressable/Pressable.tsx
m-bert: We already have `isJest` function (defined [here](https://github.com/software-mansion/react-native-gesture-handler/blob/5b535c9a7507eb72e53c0c99b7fc7be6e420b31c/src/utils.ts#L37)). Can't we use that?

latekvo: `NODE_ENV === 'test'` also works with `Mocha` and `Jasmine`, while `isJest` which uses `JEST_WORKER_ID` only works with `Jest`. I think it'd actually be worth changing `isJest` function to the implementation present in this PR.

src/components/Pressable/Pressable.tsx
```diff
@@ -380,6 +380,10 @@ export default function Pressable(props: PressableProps) {
       ? children({ pressed: pressedState })
       : children;
 
+  // @ts-ignore Adding type definitions for @types/node causes multiple conflicts with react-native/types,
+  //            and as far as i know, there is no simple way of automatically resolving them.
+  const inTestEnv = process.env.NODE_ENV === 'test';
```
j-piasecki, line 385: Can we use this https://github.com/software-mansion/react-native-gesture-handler/blob/5b535c9a7507eb72e53c0c99b7fc7be6e420b31c/src/utils.ts#L37-L40?

latekvo: Mentioned here: https://github.com/software-mansion/react-native-gesture-handler/pull/3357#discussion_r1931920705

src/components/Pressable/Pressable.tsx
mdjastrzebski: In the future RNTL would like to support other test runners, like `vitest` or `bun`. If you hardcode to Jest instead of testing, this would break in such case.

src/components/Pressable/Pressable.tsx
mdjastrzebski: THis would tie the implementation only to Jest test runner. In case of user using other test runners Mocha, Vitest, Bun (not all of these are supported by RNTL atm) this props will not be set. The `process.env.NODE_ENV === 'test';` should support all test runners.

src/components/Pressable/Pressable.tsx
m-bert: In that case why don't we change `isJestEnv` implementation instead of adding new check?

latekvo: I agree the `isJestEnv` implementation can be changed, mentioned this in https://github.com/software-mansion/react-native-gesture-handler/pull/3357#discussion_r1931920705 as well. 
Done in [f13e7b0](https://github.com/software-mansion/react-native-gesture-handler/pull/3357/commits/f13e7b041070a03a0a97e241cadb979f1eba75a7)

src/components/Pressable/Pressable.tsx
latekvo: Done in [4d942e8](https://github.com/software-mansion/react-native-gesture-handler/pull/3357/commits/4d942e8d200aff9b35c0d866d5addb04a54480be)

mdjastrzebski:
@latekvo could you also expose `onLongPress` in a similar way?

mdjastrzebski:
I've implemented RNTL counterpart here: https://github.com/callstack/react-native-testing-library/pull/1741

src/components/Pressable/Pressable.tsx
```diff
@@ -390,7 +390,10 @@ export default function Pressable(props: PressableProps) {
         touchSoundDisabled={android_disableSound ?? undefined}
         rippleColor={processColor(android_ripple?.color ?? defaultRippleColor)}
         rippleRadius={android_ripple?.radius ?? undefined}
-        style={[pointerStyle, styleProp]}>
+        style={[pointerStyle, styleProp]}
+        testOnly_onPress={isTestEnv() ? onPress : undefined}
+        testOnly_onPressIn={isTestEnv() ? onPressIn : undefined}
+        testOnly_onPressOut={isTestEnv() ? onPressOut : undefined}>
```
m-bert, line 396: We could do something similar to _**Reanimated**_, i.e. call this only once and store result in constant

```js
const IS_TEST_ENV = isTestEnv()
```

latekvo: I though about that, but `isTestEnv()` is just a property read, and both notations are just as readable, so i don't think there's a point in caching the value.

m-bert: Well, it calls `hasProperty` and then compares values, so it's more than just property read. While it's not crucial in test environment, I think it is better to memoize this value.

latekvo: You're right, done in [7a31f5e](https://github.com/software-mansion/react-native-gesture-handler/pull/3357/commits/7a31f5ec2a42aa51ab2d4fa3f9744f5361b020d1)

latekvo:
>@latekvo could you also expose onLongPress in a similar way?

@mdjastrzebski done in [6184078](https://github.com/software-mansion/react-native-gesture-handler/pull/3357/commits/618407816e8ef604597be3b706e9c72125a27891)

mdjastrzebski (approved):
LGTM 🙌

m-bert (approved):

mdjastrzebski:
@latekvo could you make a release with this PR?

latekvo:
@mdjastrzebski, FYI, this PR has just been released in [2.23.0](https://github.com/software-mansion/react-native-gesture-handler/releases/tag/2.23.0).

Issue #3320, opened 2025-01-09
[Reanimated] Reading from `value` during component render error


danturn:
### Description

I am using Swipeable from "react-native-gesture-handler/ReanimatedSwipeable"
i've been trying to make a swipable flat list and am getting errors:

```
 (NOBRIDGE) WARN  [Reanimated] Reading from `value` during component render. Please ensure that you do not access the `value` property or use `get` method of a shared value while React is rendering a component.
```

I have created a totally minimal repro of this here: https://github.com/danturn/swipe_test

this is what the component looks like:

```
import { useFocusEffect } from "expo-router";
import { useState, useCallback, useEffect } from "react";
import { Text } from "react-native";
import Swipeable from "react-native-gesture-handler/ReanimatedSwipeable";
import { GestureHandlerRootView } from "react-native-gesture-handler";
import Reanimated, {
  SharedValue,
  useAnimatedStyle,
} from "react-native-reanimated";

export default function Index() {
  useFocusEffect(
    useCallback(() => {
      console.log("test");
    }, [])
  );

  return (
    <GestureHandlerRootView>
      <Swipeable renderRightActions={RightAction}>
        <Text style={{ color: "white" }}>ANYTHING IN HERE</Text>
      </Swipeable>
    </GestureHandlerRootView>
  );
}

function RightAction(prog: SharedValue<number>, drag: SharedValue<number>) {
  return (
    <Reanimated.View>
      <Text>Text</Text>
    </Reanimated.View>
  );
}
```

and I'm getting this output:

```
Android Bundled 523ms node_modules/expo-router/entry.js (1188 modules)
 (NOBRIDGE) LOG  Bridgeless mode is enabled
 INFO
 💡 JavaScript logs will be removed from Metro in React Native 0.77! Please use React Native DevTools as your default tool. Tip: Type j in the terminal to open (requires Google Chrome or Microsoft Edge).
 (NOBRIDGE) LOG  test
 (NOBRIDGE) WARN  [Reanimated] Reading from `value` during component render. Please ensure that you do not access the `value` property or use `get` method of a shared value while React is rendering a component.

If you don't want to see this message, you can disable the `strict` mode. Refer to:
https://docs.swmansion.com/react-native-reanimated/docs/debugging/logger-configuration for more details.
 (NOBRIDGE) WARN  [Reanimated] Reading from `value` during component render. Please ensure that you do not access the `value` property or use `get` method of a shared value while React is rendering a component.

[...]

Issue #3170, opened 2024-10-19
ReanimatedSwipeable generating warnings


jbrodriguez:
### Description

trying the example for ReanimatedSwipeable generates warnings on the logs

in my case, i just used RightAction, however those SharedValues seem to be causing the warnings

![Screenshot 2024-10-19 at 17 20 17](https://github.com/user-attachments/assets/a2ecb4a4-629c-4c1a-8bc1-0f5c3584fb9a)

### Steps to reproduce

run the example from the repo

### Snack or a link to a repository

https://github.com/software-mansion/react-native-gesture-handler/blob/main/example/src/release_tests/swipeableReanimation/index.tsx

### Gesture Handler version

2.22.0

### React Native version

0.76.0-rc.6

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Fabric (New Architecture)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

jbrodriguez:
for the time being, im just setting strict to false 🤷‍♂️

latekvo:
Hi @jbrodriguez, thank you for reporting this issue.

Would it be possible for you to share a branch or a repository on which you reproduced this issue?

I could not find this issue occurring on: 
- `rn 0.74.1` + `rngh 2.20.0` + `rea 3.10.0` + `oldArch`.
- `rn 0.74.1` + `rngh 2.20.0` + `rea 3.16.0` + `oldArch`.
- `rn 0.74.1` + `rngh 2.20.0` + `rea 3.16.0` + `newArch`.

Unfortunately I couldn't test on `rn 0.76.0-rc.6`, there are multiple breaking changes when upgrading from `rn 0.74.1` to `rn 0.76.0-rc.6`, both on `android` and `iOS`.

Tested everything in `StrictMode`.

I also couldn't manually find any instance of `.value` anywhere in the render body of either the `ReanimatedSwipeable` component, or the example you linked. 

Are you certain this is an issue with our example or component?

jbrodriguez:
sure, ill provide a repo later today

BrightFoxApps:
[...]

PR #3365, opened 2025-01-29
Fix `ReanimatedSwipeable` reanimated warnings 

juliendevleeschauwer:
## Description

Fix -  WARN  [Reanimated] Reading from `value` during component render.  in react-native-gesture-handler version 2.22.1

This warning appear when you use Swipeable component.

Change : 
in src/components/ReanimatedSwipeable.tsx
replace rowWidth.value to rowWidth in the dependency array of updateElementWidths callback.

#3170 #3320

## Test plan

Tested on Samsung Galaxy A52, android emulator and Iphone 13


latekvo (approved):

latekvo:
Thanks for the fix!

Issue #3367, opened 2025-01-30
Pressable won't show background color


sidferreira:
### Description

I tested this in a new repo (below), and the first build has the issue. We face the same issue in our main app

I works as expected on `expo projects` even updating RNGH

May be related to https://github.com/software-mansion/react-native-gesture-handler/pull/3312

### Steps to reproduce

Add `backgroundColor` to `Pressable`

<img src="https://github.com/user-attachments/assets/3c488872-af55-4fb6-a59b-facd0c41b98a" width="300" />

### Snack or a link to a repository

https://github.com/sidferreira/rngh-pressable-no-background

### Gesture Handler version

2.22.1

### React Native version

0.77.0

### Platforms

Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

Debug mode

### Device

Real device

### Acknowledgements

Yes

tasty0brains:
I have the same thing

latekvo:
Can confirm. Since `RN 0.77`, all `RectButtons`, `BaseButtons`, `RawButtons` and `PureNativeButtons` no longer work with the `backgroundColor` style. Will look into it further, this is a serious issue.

Repro:

<details>
<summary>
Collapsed repro
</summary>

```tsx
import React from 'react';
import { Pressable as RNPressable, StyleSheet, Text } from 'react-native';
import {
  BaseButton,
  GestureHandlerRootView,
  PureNativeButton,
  RawButton,
  RectButton,
} from 'react-native-gesture-handler';

const App = () => {
  return (
    <GestureHandlerRootView>
      <RectButton style={styles.wrapperCustom}>
        <Text style={styles.text}>RectButton</Text>
      </RectButton>
      <BaseButton style={styles.wrapperCustom}>
        <Text style={styles.text}>BaseButton</Text>
      </BaseButton>
      <RawButton style={styles.wrapperCustom}>
        <Text style={styles.text}>RawButton</Text>
      </RawButton>
      <PureNativeButton style={styles.wrapperCustom}>
        <Text style={styles.text}>PureNativeButton</Text>
      </PureNativeButton>
      <RNPressable style={styles.wrapperCustom}>
        {({ pressed }) => (
          <Text style={styles.text}>
            {pressed ? 'Pressed!' : 'Pressable from react-native'}
          </Text>
        )}
      </RNPressable>
    </GestureHandlerRootView>
  );
};

const styles = StyleSheet.create({
  container: {
[...]

PR #3368, opened 2025-01-30
Fix `backgroundColor` not being applied to `RNGestureHandlerButton` 

latekvo:
## Description

`RNGestureHandlerButton` consists of 2 nested `ViewGroups` - `ViewGroupManager` and `ButtonViewGroup`

Before React Native `0.77`, `backgroundColor` was applied to `RNGestureHandlerButton` by `ButtonViewGroup.setBackgroundColor(color)`, which was called by unknown React Native mechanism.

In React Native `0.77` something changed, and  `ButtonViewGroup.setBackgroundColor(color)` stopped being called. 

This PR overrides the setter for `backgroundColor` on `ViewGroupManager`, and forwards the argument to `ButtonViewGroup.setBackgroundColor(color)`.

I have not found the root cause (RN commit) of this issue, and i have no idea what are the consequences of calling both the `BackgroundStyleApplicator.setBackgroundColor()` and `ButtonViewGroup.setBackgroundColor(color)`, but it seems like those two functions were already called sequentially before `0.77`, as `BackgroundStyleApplicator` was introduced in RN `0.76` and most likely called `ButtonViewGroup.setBackgroundColor(color)` itself. 

Fixes https://github.com/software-mansion/react-native-gesture-handler/issues/3367

## Test plan

<details>

<summary>
Repro code
</summary>

```tsx
import React from 'react';
import { Pressable as RNPressable, StyleSheet, Text } from 'react-native';
import {
  BaseButton,
  GestureHandlerRootView,
  PureNativeButton,
  RawButton,
  RectButton,

const App = () => {
  return (
    <GestureHandlerRootView>
      <RectButton style={styles.wrapperCustom}>
        <Text style={styles.text}>RectButton</Text>
      </RectButton>
      <BaseButton style={styles.wrapperCustom}>
        <Text style={styles.text}>BaseButton</Text>
      </BaseButton>
      <RawButton style={styles.wrapperCustom}>
        <Text style={styles.text}>RawButton</Text>
      </RawButton>
      <PureNativeButton style={styles.wrapperCustom}>
        <Text style={styles.text}>PureNativeButton</Text>
      </PureNativeButton>
      <RNPressable style={styles.wrapperCustom}>
        {({ pressed }) => (
          <Text style={styles.text}>
            {pressed ? 'Pressed!' : 'Pressable from react-native'}
          </Text>
        )}
      </RNPressable>
    </GestureHandlerRootView>
  );
};

  container: {
    flex: 1,
    justifyContent: 'center',
    gap: 16,
    padding: 16,
  },
  text: {
    fontSize: 32,
  },
  wrapperCustom: {
    borderRadius: 8,
    padding: 6,
    flex: 1,
    borderWidth: 2,
  },
  logBox: {
    flex: 1,
    padding: 20,
    margin: 10,
    borderWidth: StyleSheet.hairlineWidth,
    borderColor: '#f0f0f0',
    backgroundColor: '#f9f9f9',
  },
});

export default App;

```
</details>

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -55,6 +55,12 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
     view.useDrawableOnForeground = useDrawableOnForeground
   }
 
+  @ReactProp(name = "backgroundColor")
+  override fun setBackgroundColor(view: ButtonViewGroup, backgroundColor: Int) {
+    super.setBackgroundColor(view, backgroundColor)
+    view.setBackgroundColor(backgroundColor)
+  }
+
   @ReactProp(name = "borderless")
   override fun setBorderless(view: ButtonViewGroup, useBorderlessDrawable: Boolean) {
     view.useBorderlessDrawable = useBorderlessDrawable
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -57,7 +57,6 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
 
   @ReactProp(name = "backgroundColor")
   override fun setBackgroundColor(view: ButtonViewGroup, backgroundColor: Int) {
-    super.setBackgroundColor(view, backgroundColor)
     view.setBackgroundColor(backgroundColor)
   }
 


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -55,6 +55,12 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
     view.useDrawableOnForeground = useDrawableOnForeground
   }
 
+  @ReactProp(name = "backgroundColor")
+  override fun setBackgroundColor(view: ButtonViewGroup, backgroundColor: Int) {
+    super.setBackgroundColor(view, backgroundColor)
```
j-piasecki, line 60: Is the super call needed? Isn't the next line effectively overriding whatever is set here?

latekvo: I think you're right. I previously had my concerns because of [this](https://github.com/facebook/react-native/commit/cc23aea2fc95947b19dbd28061f748c13ba6344b) experimental feature introduced in `0.77`, which changes the background updating system, but i re-checked, and `updateBackgroundColor()` works independently of that feature. 

Another concern, which is not ruled out, is that the aforementioned `NewBackgroundAndBorderDrawables` feature may deprecate `LayerDrawable` method used in  `updateBackgroundColor()`, but since there isn't any documentation regarding that feature yet, i think it'd make the most sense to add a comment mentioning this possibility, in case something breaks in the future, and move on.

I'm not sure if we should be finding a way to support `enableNewBackgroundAndBorderDrawables` feature flag, I personally think it's a low-priority change.

Besides those things, `super.setBackgroundColor()` ([implementation](https://github.com/facebook/react-native/blob/566a45e28c76865326aa7f920701a3489d8c166e/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/BackgroundStyleApplicator.kt#L53)) seems to do more or less the same things as `updateBackgroundColor()` ([implementation](https://github.com/software-mansion/react-native-gesture-handler/blob/b3d8fd91dca195267bdc33aa20fd6f5cd37d7fe2/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt#L307)).

m-bert (approved):

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
j-piasecki: I doubt that setBackgroundColor can be deprecated by React Native - https://developer.android.com/reference/android/view/View#setBackgroundColor(int) 😄.

> I'm not sure if we should be finding a way to support enableNewBackgroundAndBorderDrawables right now

I'd leave that be for the time being, it may be easier to do without supporting the old arch.

sidferreira:
Tyvm

PR #3369, opened 2025-01-31
Fix android ripple color bug on fabric

### LINKED ISSUE #3246: [Pressable] android_ripple / color doesn't work correctly

latekvo:
## Description

Android ripple currently does not work on `Fabric`, as all values passed to `RawButton` and it's inheritors are passed through `processColor`, which is crucial on `Paper`, and broken on `Fabric`.

This PR disables usage of `processColor`, when running on `Fabric`.

Note: `isFabric` cannot be moved out of the body of the `Pressable`, as it likely won't be initialised before the `Pressable` starts being rendered. More details [here](https://github.com/software-mansion/react-native-gesture-handler/blob/b3d8fd91dca195267bdc33aa20fd6f5cd37d7fe2/src/utils.ts#L47).

Fixes #3246
Fixes #3312
Supersedes #3250
Likely could add a workaround for https://github.com/software-mansion/react-native-reanimated/issues/6935

## Test plan

<details>
<summary>
Collapsed test code
</summary>

```tsx
import React from 'react';
import { Pressable as RNPressable, StyleSheet, Text } from 'react-native';
import {
  BaseButton,
  GestureHandlerRootView,
  RectButton,
  Pressable,

const App = () => {
  return (
    <GestureHandlerRootView>
      <RectButton style={styles.wrapperCustom} rippleColor={'blue'}>
        <Text style={styles.text}>RectButton</Text>
      </RectButton>
      <BaseButton style={styles.wrapperCustom} rippleColor={'blue'}>
        <Text style={styles.text}>BaseButton</Text>
      </BaseButton>
      <Pressable
        style={styles.wrapperCustom}
        android_ripple={{ color: 'blue' }}>
        {({ pressed }) => (
          <Text style={styles.text}>
            {pressed ? 'Pressed!' : 'Pressable from react-native'}
          </Text>
        )}
      </Pressable>
      <RNPressable
        style={styles.wrapperCustom}
        android_ripple={{ color: 'blue' }}>
        {({ pressed }) => (
          <Text style={styles.text}>
            {pressed ? 'Pressed!' : 'Pressable from react-native'}
          </Text>
        )}
      </RNPressable>
      <Pressable
        style={styles.wrapperCustom}
        android_ripple={{ color: '#00f' }}>
        {({ pressed }) => (
          <Text style={styles.text}>
            {pressed ? 'Pressed!' : 'Pressable from react-native'}
          </Text>
        )}
      </Pressable>
      <RNPressable
        style={styles.wrapperCustom}
        android_ripple={{ color: '#00f' }}>
        {({ pressed }) => (
          <Text style={styles.text}>
            {pressed ? 'Pressed!' : 'Pressable from react-native'}
          </Text>
        )}
      </RNPressable>
    </GestureHandlerRootView>
  );
};

  container: {
    flex: 1,
    justifyContent: 'center',
    gap: 16,
    padding: 16,
  },
  text: {
    fontSize: 32,
  },
  wrapperCustom: {
    borderRadius: 8,
    padding: 6,
    flex: 1,
    borderColor: 'red',
    borderWidth: 2,
  },
  logBox: {
    flex: 1,
    padding: 20,
    margin: 10,
    borderWidth: StyleSheet.hairlineWidth,
    borderColor: '#f0f0f0',
    backgroundColor: '#f9f9f9',
  },
});

export default App;

```

</details>

### DIFF
--- a/src/components/GestureButtons.tsx
+++ b/src/components/GestureButtons.tsx
@@ -121,16 +121,16 @@ class InnerBaseButton extends React.Component<BaseButtonWithRefProps> {
   };
 
   render() {
-    const { rippleColor, style, ...rest } = this.props;
+    const { rippleColor: unprocessedRippleColor, style, ...rest } = this.props;
 
-    const processedRippleColor = isFabric()
-      ? rippleColor
-      : processColor(rippleColor ?? undefined);
+    const rippleColor = isFabric()
+      ? unprocessedRippleColor
+      : processColor(unprocessedRippleColor ?? undefined);
 
     return (
       <RawButton
         ref={this.props.innerRef}
-        rippleColor={processedRippleColor}
+        rippleColor={rippleColor}
         style={[style, Platform.OS === 'ios' && { cursor: undefined }]}
         {...rest}
         onGestureEvent={this.onGestureEvent}
--- a/src/components/GestureButtons.tsx
+++ b/src/components/GestureButtons.tsx
@@ -25,6 +25,8 @@ export const RawButton = createNativeWrapper(GestureHandlerButton, {
   shouldActivateOnStart: false,
 });
 
+let IS_FABRIC: null | boolean = null;
+
 class InnerBaseButton extends React.Component<BaseButtonWithRefProps> {
   static defaultProps = {
     delayLongPress: 600,
@@ -123,7 +125,11 @@ class InnerBaseButton extends React.Component<BaseButtonWithRefProps> {
   render() {
     const { rippleColor: unprocessedRippleColor, style, ...rest } = this.props;
 
-    const rippleColor = isFabric()
+    if (IS_FABRIC === null) {
+      IS_FABRIC = isFabric();
+    }
+
+    const rippleColor = IS_FABRIC
       ? unprocessedRippleColor
       : processColor(unprocessedRippleColor ?? undefined);
 
--- a/src/components/GestureButtons.tsx
+++ b/src/components/GestureButtons.tsx
@@ -18,6 +18,7 @@ import type {
   BorderlessButtonWithRefProps,
   BorderlessButtonProps,
 } from './GestureButtonsProps';
+import { isFabric } from '../utils';
 
 export const RawButton = createNativeWrapper(GestureHandlerButton, {
   shouldCancelWhenOutside: false,
@@ -122,10 +123,14 @@ class InnerBaseButton extends React.Component<BaseButtonWithRefProps> {
   render() {
     const { rippleColor, style, ...rest } = this.props;
 
+    const processedRippleColor = isFabric()
+      ? rippleColor
+      : processColor(rippleColor ?? undefined);
+
     return (
       <RawButton
         ref={this.props.innerRef}
-        rippleColor={processColor(rippleColor)}
+        rippleColor={processedRippleColor}
         style={[style, Platform.OS === 'ios' && { cursor: undefined }]}
         {...rest}
         onGestureEvent={this.onGestureEvent}
--- a/src/components/Pressable/Pressable.tsx
+++ b/src/components/Pressable/Pressable.tsx
@@ -379,15 +379,15 @@ export default function Pressable(props: PressableProps) {
       ? children({ pressed: pressedState })
       : children;
 
-  const memoIsFabric = useCallback(() => isFabric, []);
+  const isFabricValue = useMemo(() => isFabric(), []);
 
   const rippleColor = useMemo(() => {
     const defaultRippleColor = android_ripple ? undefined : 'transparent';
     const unprocessedRippleColor = android_ripple?.color ?? defaultRippleColor;
-    return memoIsFabric()
+    return isFabricValue
       ? unprocessedRippleColor
       : processColor(unprocessedRippleColor);
-  }, [android_ripple, memoIsFabric]);
+  }, [android_ripple, isFabricValue]);
 
   return (
     <GestureDetector gesture={gesture}>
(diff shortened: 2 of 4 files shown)


src/components/Pressable/Pressable.tsx
```diff
@@ -380,6 +378,18 @@ export default function Pressable(props: PressableProps) {
       ? children({ pressed: pressedState })
       : children;
 
+  const defaultRippleColor = useMemo(
```
freeboub, line 381: I think both memo can be merged into only 1 ..

latekvo: You're right, fixed in [a9d2031](https://github.com/software-mansion/react-native-gesture-handler/pull/3369/commits/a9d203156e759bf5fd8af644f0609b5821a5236a)

src/components/Pressable/Pressable.tsx
```diff
@@ -380,6 +378,14 @@ export default function Pressable(props: PressableProps) {
       ? children({ pressed: pressedState })
       : children;
 
+  const rippleColor = useMemo(() => {
+    const defaultRippleColor = android_ripple ? undefined : 'transparent';
+    const unprocessedRippleColor = android_ripple?.color ?? defaultRippleColor;
```
m-bert, line 383: Shouldn't it be 

```jsx
const unprocessedRippleColor = android_ripple?.color ?? 'transparent';
```
?

I don't really see the point of `defaultRippleColor` in that case.

latekvo: `defaultRippleColor` is required to properly replicate the behaviour of RN `Pressable`.
When `android_ripple` is set, but the `color` field is not provided, we want to set `color` to `undefined`, so that it uses the default system value (whiteish opaque).
When `android_ripple` is not set, we want to remove the ripple altogether, which is done via `'transparent'`.

src/components/GestureButtons.tsx
```diff
@@ -122,10 +123,14 @@ class InnerBaseButton extends React.Component<BaseButtonWithRefProps> {
   render() {
     const { rippleColor, style, ...rest } = this.props;
 
+    const processedRippleColor = isFabric()
```
m-bert, line 126: What about `maybeProcessedRippleColor`?

m-bert: Also, why don't we memoize value of `isFabric` instead of calling this everytime?

latekvo: >What about maybeProcessedRippleColor?

Not a big fan of that name, the word `processed` already allows for lack of change.
But I agree that `processed` may be confusing due to `processColor()` function.
Is it ok if instead, I do the following?
`rippleColor` -> `unprocessedRippleColor`
`processedRippleColor` -> `rippleColor`

>Also, why don't we memoize value of isFabric instead of calling this everytime?

Sure, done in [ff48533](https://github.com/software-mansion/react-native-gesture-handler/pull/3369/commits/ff48533eacbf8641c7005744fbf12e6d89dc01a5)

m-bert: > Sure, done in [ff48533](https://github.com/software-mansion/react-native-gesture-handler/pull/3369/commits/ff48533eacbf8641c7005744fbf12e6d89dc01a5)

Not exactly what I meant. I thought about storing it in a variable. Now we still call function in each render.

> Maybe this will do better:
rippleColor -> unprocessedRippleColor
processedRippleColor -> rippleColor

Looks good 👍

latekvo: >Not exactly what I meant. I thought about storing it in a variable. Now we still call function in each render.

Can't do that, `isFabric()` will return incorrect data outside of the function component, as the value it reads will not be initialised before the first render.

latekvo: > Looks good 👍 

Done in [2513b9e](https://github.com/software-mansion/react-native-gesture-handler/pull/3369/commits/2513b9e5bc1a71d974d4b0c5fbd52d2985676013)

m-bert (approved):
Looks ok 👍 

What about using the same `IS_FABRIC` trick in `Pressable`? 😅

latekvo:
> What about using the same IS_FABRIC trick in Pressable?

Done in [f8b2de5](https://github.com/software-mansion/react-native-gesture-handler/pull/3369/commits/f8b2de56030319877a74b96d5260cd235d867aa2)

m-bert (approved):

PR #3371, opened 2025-02-02
fix: overlapping buttons cause buttons to stop being responsive

### LINKED ISSUE #3370: Overlapping buttons cause buttons to stop being responsive
### Description

Similar scenario as described in #3350 but a different cause. Basically when having to buttons on top of each other and tapping in that overlapping area, they both trigger (see video) but only the frontmost button is properly cancelled (the backmost is stuck in highlighted mode). 

This cause all buttons to become unresponsive, the only way to get back into responsive state is if you tap the highlighted button again, which is not always possible if you navigate to a new screen for example. 

https://github.com/user-attachments/assets/e14fffab-792d-44e3-8cb3-3de6bde32224

### Steps to reproduce

1. Make two buttons overlap each other (partially or fully)
2. Press the topmost button
3. Taps don't work anymore

This happens only on Android, but both in debug/release, on JSC/Hermes and on Bare/Expo.

https://snack.expo.dev/@oblador/gesture-handler-overlap-reproduction

### Gesture Handler version

2.20.2

### React Native version

0.76

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

DorianMazur:
## Description

Partially resolves problem in #3370

This PR fixes an issue where a behind button could stay stuck in the "pressed" state if another overlapping button took the responder. By adding a check in onTouchEvent, we ensure that only one button can stay pressed at a time. If another button is already the responder, the behind button cancels itself and won't remain highlighted.

It will not solve the issue if both button are not exclusive. For this edge case better solution is needed.

## With this fix:
https://github.com/user-attachments/assets/b88ff1d1-cf1c-4a62-9967-7dd25e2f3dc0

## Without this fix:
https://github.com/user-attachments/assets/b69771d2-655c-4fa6-b5b2-9edd37d32e9c

## Test plan

https://snack.expo.dev/@oblador/gesture-handler-overlap-reproduction

<details>

<summary>Collapsed test code</summary>

```tsx
import { Text, View, StyleSheet, ScrollView } from 'react-native';
import {
  BaseButton,
  GestureHandlerRootView,

  return (
    <GestureHandlerRootView>
        <BaseButton
          style={styles.button}
          shouldCancelWhenOutside={true}
          onHandlerStateChange={(event) => console.log(event.nativeEvent)}>
          <Text>Behind</Text>
        </BaseButton>
        <BaseButton
          style={[styles.button, styles.onTop]}
          shouldCancelWhenOutside={true}
          onHandlerStateChange={(event) => console.log(event.nativeEvent)}>
          <Text>On top</Text>
        </BaseButton>
      </View>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    backgroundColor: '#ecf0f1',
    justifyContent: 'center',
  },
  button: {
    padding: 30,
    borderWidth: 1,
    backgroundColor: '#ffffff99',
  },
  onTop: {
    position: 'relative',
    top: -40,
    left: 100,
    right: 20,
  },
});
```
</details>

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -260,11 +260,12 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
     }
 
     override fun onInterceptTouchEvent(ev: MotionEvent): Boolean {
+      if (touchResponder != null && touchResponder !== this) {
+        return false
+      }
       if (super.onInterceptTouchEvent(ev)) {
         return true
       }
-      // We call `onTouchEvent` and wait until button changes state to `pressed`, if it's pressed
-      // we return true so that the gesture handler can activate.
       onTouchEvent(ev)
       return isPressed
     }
@@ -288,6 +289,16 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
      * [com.swmansion.gesturehandler.NativeViewGestureHandler.onHandle]  */
     @SuppressLint("ClickableViewAccessibility")
     override fun onTouchEvent(event: MotionEvent): Boolean {
+
+      if (touchResponder != null && touchResponder !== this) {
+        if (isPressed) {
+          setPressed(false)
+        }
+        lastEventTime = event.eventTime
+        lastAction = event.action
+        return false
+      }
+      
       val eventTime = event.eventTime
       val action = event.action
 
@@ -435,7 +446,7 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
       }
     }
 
-    private fun tryGrabbingResponder(): Boolean {
+        private fun tryGrabbingResponder(): Boolean {
       if (isChildTouched()) {
         return false
       }
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -260,9 +260,6 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
     }
 
     override fun onInterceptTouchEvent(ev: MotionEvent): Boolean {
-      if (touchResponder != null && touchResponder !== this) {
-        return false
-      }
       if (super.onInterceptTouchEvent(ev)) {
         return true
       }
--- a/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -288,18 +288,17 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
      * [com.swmansion.gesturehandler.NativeViewGestureHandler.onHandle]  */
     @SuppressLint("ClickableViewAccessibility")
     override fun onTouchEvent(event: MotionEvent): Boolean {
+      val eventTime = event.eventTime
+      val action = event.action
 
       if (touchResponder != null && touchResponder !== this) {
         if (isPressed) {
           setPressed(false)
         }
-        lastEventTime = event.eventTime
-        lastAction = event.action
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -291,6 +291,15 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
       val eventTime = event.eventTime
       val action = event.action
 
+      if (touchResponder != null && touchResponder !== this) {
```
m-bert, line 294: Wouldn't this be enough?

```suggestion
      if (touchResponder !== this) {
```

m-bert (commented):
Thank you for this PR!

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
DorianMazur: Using only if (touchResponder !== this) would also be true when touchResponder is null, so when no other button has grabbed the responder. In that case, we want to proceed normally.

m-bert: Oh, you're right 😅

m-bert (changes requested):
I've looked more into it and it seems that this change breaks buttons with `exclusive={false}`.

Consider this example:

```jsx
<GestureHandlerRootView>
    <BaseButton
      exclusive={false}
      style={styles.button}
      shouldCancelWhenOutside={true}
      onHandlerStateChange={(event) => console.log(event.nativeEvent)}>
      <Text>Behind</Text>
    </BaseButton>
    <BaseButton
      exclusive={false}
      style={styles.button}
      shouldCancelWhenOutside={true}
      onHandlerStateChange={(event) => console.log(event.nativeEvent)}>
      <Text>On top</Text>
    </BaseButton>
  </View>
</GestureHandlerRootView>
```

Now buttons are not stacked on each other and with `exclusive` set to `false` you should be able to click both of them at the same time. However,  with these changes it is not possible.

DorianMazur:
@m-bert I added a check to verify if the button is exclusive.

android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -296,6 +296,17 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
...
+        if (touchResponder!!.exclusive) {
```
m-bert, line 300: I'd combine it with outer `if` to avoid unnecessary nesting

m-bert (changes requested):
Unfortunately, when both buttons have `exclusive={false}`, this problem is still reproducible 😞

DorianMazur:
@m-bert What do you mean? Your example is working for me

https://github.com/user-attachments/assets/1a2ce52e-9020-4e2c-b30a-b515ee3350b1

m-bert:
>@m-bert What do you mean? Your example is working for me

I meant the original one, when the buttons are stacked one on top of the other. The button below gets blocked as in the original issue.

DorianMazur:
@m-bert Yes, that's already happening in the current version of react-native-gesture-handlers 🤷 . If we want to fix this bug for two buttons with exclusive=false, we might need a different solution, but I’m not sure what that would be. Maybe some ray tracing and checking if there is another button on top? That sounds expensive.

My PR already fixes the bug for regular buttons, and the case where two non-exclusive buttons overlap is a rare edge case.

m-bert:
>@m-bert Yes, that's already happening in the latest version of react-native-gesture-handler 🤷 . 

Yes, we are aware. We should find better solution for that. But I agree that it improves current behavior, so I'll give it a green light once You address [this comment](https://github.com/software-mansion/react-native-gesture-handler/pull/3371#discussion_r1940891423)

Also, You can either remove _**fixes**_ from description, or open new issue so that we will know that it is still a problem. 

If you could do it quickly, we may include this in 2.23.0 (cc @latekvo)

DorianMazur:
@m-bert Thanks, I updated description and pushed new commit

m-bert (approved):

latekvo (approved):
Looks good, will include it in the 2.23.0 👍

PR #3374, opened 2025-02-03
Unregister from oldGestureHandler when component unmounts

gavrix:
## Description

We might have discovered a memory leak associated with oldGestureHandler registry. Nothing ever gets removed from it. We noticed a lot of hanging GenericTouchable components at runtime associated with this registry.

## Test plan

After applying this patch on our repo, the same profiling workflow did not show the same amount of hanging GenericTouchables. However, I don't know whats the proper way to verify and test this.

### DIFF
--- a/src/handlers/createHandler.tsx
+++ b/src/handlers/createHandler.tsx
@@ -8,7 +8,11 @@ import {
 import { customDirectEventTypes } from './customDirectEventTypes';
 import RNGestureHandlerModule from '../RNGestureHandlerModule';
 import { State } from '../State';
-import { handlerIDToTag, registerOldGestureHandler, unregisterOldGestureHandler } from './handlersRegistry';
+import {
+  handlerIDToTag,
+  registerOldGestureHandler,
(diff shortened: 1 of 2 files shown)


m-bert (approved):
Good catch, thank you for fixing this! ❤️ 

<details>

```jsx
import { Button, View } from 'react-native';
import {
  GestureHandlerRootView,
  PanGestureHandler,

  const [show, setShow] = useState(false);

  return (
    <GestureHandlerRootView
      style={{ flex: 1, alignItems: 'center', justifyContent: 'space-around' }}>
      <Button onPress={() => setShow((prev) => !prev)} title="Click me!" />

      {show && (
        <PanGestureHandler>
          <View style={{ width: 100, height: 100, backgroundColor: 'red' }} />
        </PanGestureHandler>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>

with 

```jsx
setInterval(() => console.log(oldHandlers), 1000);
```

added in `handlersRegistry.tsx`

PR #3377, opened 2025-02-04
Return false if gesture recognizers are present but all disabled

### LINKED ISSUE #3376: RNGestureHandlerButton Stop Responding When Child View Contains Disabled Accessibility Gesture Recognizers
### Description

I noticed that some BorderlessButtons would stop responding to touches on iOS, and investigating, the cause seems to be that disbled accessibility gesture recognizers are being added to certain child views of RNGestureHandlerButton. This in turn causes the child view to return true from shouldHandleTouch, despite this view not having any tappable gesture recognizers.

It looks like a bandaid fix here might be to check for enabled gesture recognizers, rather than checking that the count is greater than zero, so that these strange, untappable, disabled gesture recognizers don't break the parent button's touch handling

### Steps to reproduce

1. Use BorderlessButton with a child view <View accessible accessibilityRole="button">, which in turn has some <Text> child view
2. Update the state of the app, generally just use it, this happens randomly
3. Observe that some buttons become permanently unable to respond to taps
4. With the debugger, observe that the affected RNGestureHandlerButton has a child view RCTViewComponentView which has two disabled accessibility gesture recognizers

https://snack.expo.io/

### Gesture Handler version

2.21.1

### React Native version

0.76.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

jcolicchio:
## Description

This PR adjusts `shouldHandleTouch` to check for enabled gesture recognizers instead of returning true if the count of gesture recognizers is greater than zero.

The motivation behind this PR is to address a bug where buttons become unresponsive if, I guess, iOS is inserting disabled accessibility gesture recognizers into the child views??

Fixes #3376

## Test plan

I was able to reproduce this reliably on a private repo, and used the debugger to observe `shouldHandleTouch` returning too early from a descendant of the button meant to be tapped. With this fix, I confirmed that the correct button returns to handle to touch event and the buttons all behave as expected

I also was able to reproduce the issue with this sample code:
1. Tap `Courses`
2. Tap `Hello World 1`
3. Tap `All Courses`

The `All Courses` button has an `onPress` which `alert`s, and in this snippet, I observed no alert occurring

<details>
<summary>Click to expand</summary>

```
import { BorderlessButton, FlatList } from 'react-native-gesture-handler';
import { useNavigation } from 'expo-router';
import { useSafeAreaInsets } from 'react-native-safe-area-context';
import {useEffect} from 'react';
import {
    FadeInLeft,
    FadeInUp,
    FadeOutLeft,
    FadeOutUp,

const AnimatedBorderlessButton = Animated.createAnimatedComponent(BorderlessButton);

const OptionsScreen = ({
    options,
    onSelect
}: {
    options: string[];
    onSelect: (option?: string) => void;
}) => {
    const insets = useSafeAreaInsets();
    const navigation = useNavigation();
        container: {
          paddingTop: insets.top,
            backgroundColor: 'rgba(0,0,0,0.7)',
            flex: 1,
            paddingHorizontal: 24,
            position: 'absolute',
            top: 0,
            left: 0,
            right: 0,
            bottom: 0,
        },
        text: {
            paddingTop: 16,
            color: 'white',
        },
    });
    useEffect(() => {
        navigation.setOptions({
          tabBarStyle: { display: 'none' },
        });
        return () => {
          navigation.setOptions({
            tabBarStyle: { display: 'flex' },
          });
        };
    }, [navigation]);
    return (
        <Animated.View
            style={styles.container}
            entering={FadeInUp}
            exiting={FadeOutUp}
        >
            <FlatList
                data={options}
                renderItem={({item}) => (
                    <AnimatedBorderlessButton onPress={() => onSelect(item)}>
                      <View accessible accessibilityRole="button">
                      </View>
                    </AnimatedBorderlessButton>
                )}
            />
        </Animated.View>
    );
};

function HomeScreen() {
    const insets = useSafeAreaInsets();
    const [selectedCategory, setSelectedCategory] = useState('');
    const [options, setOptions] = useState<string[]>([]);

        container: {
            flex: 1,
            paddingTop: insets.top,
            backgroundColor: 'white',
        },
        categoryContainer: {
            gap: 16,
            paddingHorizontal: 24,
        },
    });
    const dummyData = {
        categoryOptions: (Array.from({length: 14}, (_, i) => `Hello World ${i + 1}`)),
    };

    return (
            <AnimatedBorderlessButton
              onPress={() => setOptions(dummyData.categoryOptions)}
              entering={FadeInLeft}
              exiting={FadeOutLeft}
            >
                <View accessible accessibilityRole="button">
                    <Text>{selectedCategory.length > 0 ? selectedCategory : 'Courses'}</Text>
                </View>
            </AnimatedBorderlessButton>
            {selectedCategory.length > 0 && (
                <AnimatedBorderlessButton
                    onPress={() => alert('It Worked')}
                    entering={FadeInLeft}
                    exiting={FadeOutLeft}
                >
                  <View accessible accessibilityRole="button">
                      <Text>All Courses</Text>
                  </View>
                </AnimatedBorderlessButton>
            )}
            {options.length > 0 && (
                <OptionsScreen
                    options={options}
                    onSelect={(selectedOption) => {
                        setSelectedCategory(selectedOption ?? '');
                        setOptions([]);
                    }}
                />
            )}
        </View>
    );
}

export default function TabTwoScreen() {
  return (
    <HomeScreen />
  );
}
```

</details>

### DIFF
--- a/apple/RNGestureHandlerButton.mm
+++ b/apple/RNGestureHandlerButton.mm
@@ -68,9 +68,9 @@
   // accessibility gesture recognizers such as _UIAccessibilityHUDGateGestureRecognizer,
   // ostensibly set by iOS. Such gesture recognizers cause this function to return YES
   // even when the passed view is static text and does not respond to touches. This in
-  // turn prevents the button from receiving touches, breaking functionality. Therefore,
-  // as a bandaid we can count only the enabled gesture recognizers when determining
-  // whether a view should receive touches
+  // turn prevents the button from receiving touches, breaking functionality. To handle
+  // such case, we can count only the enabled gesture recognizers when determining
+  // whether a view should receive touches.
   NSPredicate *isEnabledPredicate = [NSPredicate predicateWithFormat:@"isEnabled == YES"];
   NSArray *enabledGestureRecognizers = [view.gestureRecognizers filteredArrayUsingPredicate:isEnabledPredicate];
 
--- a/apple/RNGestureHandlerButton.mm
+++ b/apple/RNGestureHandlerButton.mm
@@ -64,24 +64,20 @@
     return button.userEnabled;
   }
 
-#if !TARGET_OS_OSX
   // Certain subviews such as RCTViewComponentView have been observed to have disabled
   // accessibility gesture recognizers such as _UIAccessibilityHUDGateGestureRecognizer,
-  // ostensibly set by iOS. Such gesture recognizers cause us to return early, despite
-  // the returned view not actually responding to touches. This in turn breaks button
-  // touch handling. Therefore, as a bandaid we can check to ensure we don't return
-  // true here if the only gesture recognizers we have are all disabled.
-  BOOL hasEnabledGestureRecognizer = NO;
-  for (UIGestureRecognizer *gestureRecognizer in view.gestureRecognizers) {
-    if (gestureRecognizer.isEnabled) {
-      hasEnabledGestureRecognizer = YES;
-      break;
-    }
-  }
+  // ostensibly set by iOS. Such gesture recognizers cause this function to return YES
+  // even when the passed view is static text and does not respond to touches. This in
+  // turn prevents the button from receiving touches, breaking functionality. Therefore,
+  // as a bandaid we can count only the enabled gesture recognizers when determining
+  // whether a view should receive touches
+  NSPredicate *isEnabledPredicate = [NSPredicate predicateWithFormat:@"isEnabled == YES"];
+  NSArray *enabledGestureRecognizers = [view.gestureRecognizers filteredArrayUsingPredicate:isEnabledPredicate];
 
-  return [view isKindOfClass:[UIControl class]] || hasEnabledGestureRecognizer;
+#if !TARGET_OS_OSX
+  return [view isKindOfClass:[UIControl class]] || [enabledGestureRecognizers count] > 0;
 #else
-  return [view isKindOfClass:[NSControl class]] || [view.gestureRecognizers count] > 0;
+  return [view isKindOfClass:[NSControl class]] || [enabledGestureRecognizers count] > 0;
 #endif
 }
 
--- a/apple/RNGestureHandlerButton.mm
+++ b/apple/RNGestureHandlerButton.mm
@@ -64,24 +64,24 @@
     return button.userEnabled;
   }
 
+#if !TARGET_OS_OSX
   // Certain subviews such as RCTViewComponentView have been observed to have disabled
   // accessibility gesture recognizers such as _UIAccessibilityHUDGateGestureRecognizer,
   // ostensibly set by iOS. Such gesture recognizers cause us to return early, despite
   // the returned view not actually responding to touches. This in turn breaks button
   // touch handling. Therefore, as a bandaid we can check to ensure we don't return
   // true here if the only gesture recognizers we have are all disabled.
-  BOOL hasEnabledGestureRecognizer = false;
+  BOOL hasEnabledGestureRecognizer = NO;
   for (UIGestureRecognizer *gestureRecognizer in view.gestureRecognizers) {
     if (gestureRecognizer.isEnabled) {
-      hasEnabledGestureRecognizer = true;
+      hasEnabledGestureRecognizer = YES;
       break;
     }
   }
 
-#if !TARGET_OS_OSX
   return [view isKindOfClass:[UIControl class]] || hasEnabledGestureRecognizer;
 #else
-  return [view isKindOfClass:[NSControl class]] || hasEnabledGestureRecognizer;
+  return [view isKindOfClass:[NSControl class]] || [view.gestureRecognizers count] > 0;
 #endif
 }
 
--- a/apple/RNGestureHandlerButton.mm
+++ b/apple/RNGestureHandlerButton.mm
@@ -64,10 +64,24 @@
     return button.userEnabled;
   }
 
+  // Certain subviews such as RCTViewComponentView have been observed to have disabled
+  // accessibility gesture recognizers such as _UIAccessibilityHUDGateGestureRecognizer,
+  // ostensibly set by iOS. Such gesture recognizers cause us to return early, despite
+  // the returned view not actually responding to touches. This in turn breaks button
+  // touch handling. Therefore, as a bandaid we can check to ensure we don't return
+  // true here if the only gesture recognizers we have are all disabled.
+  BOOL hasEnabledGestureRecognizer = false;
+  for (UIGestureRecognizer *gestureRecognizer in view.gestureRecognizers) {
+    if (gestureRecognizer.isEnabled) {
+      hasEnabledGestureRecognizer = true;
+      break;
+    }
+  }
+
 #if !TARGET_OS_OSX
-  return [view isKindOfClass:[UIControl class]] || [view.gestureRecognizers count] > 0;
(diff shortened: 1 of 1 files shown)


apple/RNGestureHandlerButton.mm
```diff
@@ -64,10 +64,24 @@ - (BOOL)shouldHandleTouch:(RNGHUIView *)view
...
   }
 
+  // Certain subviews such as RCTViewComponentView have been observed to have disabled
+  // accessibility gesture recognizers such as _UIAccessibilityHUDGateGestureRecognizer,
+  // ostensibly set by iOS. Such gesture recognizers cause us to return early, despite
+  // the returned view not actually responding to touches. This in turn breaks button
+  // touch handling. Therefore, as a bandaid we can check to ensure we don't return
+  // true here if the only gesture recognizers we have are all disabled.
+  BOOL hasEnabledGestureRecognizer = false;
+  for (UIGestureRecognizer *gestureRecognizer in view.gestureRecognizers) {
```
m-bert, line 74: I'm pretty sure that this will break on `macos` 😅

apple/RNGestureHandlerButton.mm
```diff
@@ -64,10 +64,24 @@ - (BOOL)shouldHandleTouch:(RNGHUIView *)view
     return button.userEnabled;
   }
 
+  // Certain subviews such as RCTViewComponentView have been observed to have disabled
+  // accessibility gesture recognizers such as _UIAccessibilityHUDGateGestureRecognizer,
+  // ostensibly set by iOS. Such gesture recognizers cause us to return early, despite
```
m-bert, line 69: I'm not sure if "return early" is correct in that case

m-bert (changes requested):
Hi @jcolicchio! Thanks for submitting this PR! I have some remarks though:

1. Let's use `YES` and `NO` instead of `true`/`false`. I'm also not a fan of this, but for now I'd like to stay consistent with our codebase.
2. I think that this PR will break `macos`.
3. Before we merge it, we have to test those changes, so please provide a reproduction that we can use

apple/RNGestureHandlerButton.mm
jcolicchio: In my case, the handler for a touch was the native representation of `<View accessible accessibilityRole="button"><Text>Static Text</Text></View>`. It didn't have any valid tap gesture recognizers to handle touch events, but it was still chosen by this algorithm to be the view which "shouldHandleTouch"

If the algorithm does not return this child view "early", it instead traverses up to the parent `RNGestureHandlerButton` and returns the parent view, which _does_ have gesture handlers due to, in this case, my `BorderlessButton`'s `onPress`, which faithfully gets invoked on tap

So, I think this algorithm should not return a child view which happens to have two disabled accessibility gesture recognizers which were mysteriously added from some unknown source, and should instead ignore these disabled gesture recognizers so that the parent `RNGestureHandlerButton` can be returned and handle touches

apple/RNGestureHandlerButton.mm
jcolicchio: Should I go ahead and move this implementation into `#if !TARGET_OS_OSX` for the sake of this fix, which may be an iOS-specific issue to begin with?

m-bert: Actually I would prefer something like this:

```obj-c
#if !TARGET_OS_OSX
  for (UIGestureRecognizer *gestureRecognizer in view.gestureRecognizers) {
#else
  for (NSGestureRecognizer *gestureRecognizer in view.gestureRecognizers) {
#endif
```

Now when I checked this `macos` did compile without issues, but I'm not sure why it does recognize `UIGestureRecognizer` if it has its own `NSGestureRecognizer`, so I believe it is better to split it into two parts. Also making it `iOS` exclusive may not be the best as it may also occur on `macos` (but for that we would need a reproduction)

apple/RNGestureHandlerButton.mm
m-bert: I see what you mean. What I meant was that the way it is phrased in this comment suggests something like early return in this specific function, not in the whole process. 

> (...) view which happens to have two disabled accessibility gesture recognizers, which were mysteriously added from some unknown source (...)

My guess is that they are added because of `View` with `accessibilityRole="button"`, but I have not checked that yet.

Anyway, I think that we should rephrase this comment when we have more information

m-bert: Also, do you use `Text` from React Native, or from Gesture Handler?

apple/RNGestureHandlerButton.mm
jcolicchio: Got it, I'll take a look at such an implementation tomorrow

apple/RNGestureHandlerButton.mm
m-bert: As far as I remember `Text` component was introduced form 2.22. I asked because that could explain why those views have additional gesture recognizers. 

I've checked example with only `View` with props `accessible` and `accessibilityRole="button"` and it seems that it doesn't have any additional recognizer by default

jcolicchio: Hmm... Using `Text` from RNGH 2.22 actually causes the button to not be tappable at all, as in the following:

```
<AnimatedBorderlessButton onPress={() => onSelect(item)}>
    <View accessible accessibilityRole="button">
    </View>
 </AnimatedBorderlessButton>
```

According to the debugger, the `RCTParagraphTextView` has a `RNDummyGestureRecognizer` initially, even before taking the steps which produce the bug in React Native's `Text`

apple/RNGestureHandlerButton.mm
jcolicchio: I actually went with an implementation using `NSPredicate` to filter on `isEnabled == YES`, which I believe should work equally well on OSX as iOS? I don't have a good test setup to validate on OSX though, so please take a look and let me know if it works alright.

I also updated the comment, and confirmed that my snippet indeed produces the bug on 2.22.0 (Gestures' Text was unfortunately a non-starter), and that this proposed change addresses the bug using Xcode's debugger, so I think this PR should be good to go! Requesting re-review

apple/RNGestureHandlerButton.mm
m-bert: > According to the debugger, the RCTParagraphTextView has a RNDummyGestureRecognizer initially, even before taking the steps which produce the bug in React Native's Text

That's exactly what I meant by:

> I asked because that could explain why those views have additional gesture recognizers.

jcolicchio: Got it — I’d thought you might have been suggesting that NOT using that particular Text might explain why I was seeing additional gesture recognizers

m-bert:
The code looks ok, have you been able to prepare a reproduction? It would be great to check what causes this issue 😅

jcolicchio:
> The code looks ok, have you been able to prepare a reproduction? It would be great to check what causes this issue 😅

I added it to the description of the PR, at the bottom inside of `Click to expand`!

m-bert (approved):
I've just tested that on the repro that you've provided and it indeed works 😅 

What's interesting, it does work inside our example app without this fix. I thought it may be related to React old/new architectures, but it seems that it is not a case. 

Also, I took the liberty of slighgtly changing comment, hope you don't mind 😅

PR #3387, opened 2025-02-06
chore(docs): add actionable advice to deprecation warnings

DomiR:
## Description

Add actionable advice to deprecation warnings.

See: https://github.com/software-mansion/react-native-gesture-handler/pull/3260

## Test plan

- None needed


m-bert (approved):
Good idea, thank you! ❤️

Issue #3375, opened 2025-02-03
useLayoutEffect error on the terminal


deouws:
### Description

λ  ERROR  Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.%s 
  GestureDetector (node_modules/react-native-gesture-handler/lib/commonjs/handlers/gestures/GestureDetector/index.js:76:47)
  DraggableBox (app/index.tsx:14:35)
  hrefAttrs (node_modules/react-native-web/dist/exports/View/index.js:35:24)
  hrefAttrs (node_modules/react-native-web/dist/exports/View/index.js:35:24)
  onScroll (node_modules/react-native-web/dist/exports/ScrollView/ScrollViewBase.js:57:23)
  ScrollView (node_modules/react-native-web/dist/exports/ScrollView/index.js:33:4)
  Route (node_modules/expo-router/build/Route.js:52:17)
  route (node_modules/expo-router/build/useScreens.js:128:4)
  StaticContainer (node_modules/@react-navigation/core/lib/commonjs/StaticContainer.js:14:15)
  EnsureSingleNavigator (node_modules/@react-navigation/core/lib/commonjs/EnsureSingleNavigator.js:19:2)
  SceneView (node_modules/@react-navigation/core/lib/commonjs/SceneView.js:20:2)
  hrefAttrs (node_modules/react-native-web/dist/exports/View/index.js:35:24)
  AnimatedHeaderHeightProvider (node_modules/@react-navigation/native-stack/lib/commonjs/views/NativeStackView.js:117:2)
  hrefAttrs (node_modules/react-native-web/dist/exports/View/index.js:35:24)
  hrefAttrs (node_modules/react-native-web/dist/exports/View/index.js:35:24)
  <unknown> (node_modules/react-native-web/dist/vendor/react-native/Animated/createAnimatedComponent.js:25:44)
  Background (node_modules/@react-navigation/elements/lib/commonjs/Background.js:14:2)
  Screen (node_modules/@react-navigation/elements/lib/commonjs/Screen.js:19:69)
  hrefAttrs (node_modules/react-native-web/dist/exports/View/index.js:35:24)
[...]

PR #3389, opened 2025-02-06
Fix `useLayoutEffect` throwing warnings when using SSR

latekvo:
## Description

Using `useLayoutEffect` in combination with SSR results in a warning.
This PR adds a `useSafeLayoutEffect`, which acts as `useLayoutEffect` in regular apps, and defaults to `useEffect` when using SSR.

fixes https://github.com/software-mansion/react-native-gesture-handler/issues/3375

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,3 +1,5 @@
+import { useLayoutEffect, useEffect } from 'react';
+
 export function toArray<T>(object: T | T[]): T[] {
   if (!Array.isArray(object)) {
     return [object];
@@ -100,3 +102,6 @@ export function deepEqual(obj1: any, obj2: any) {
 }
 
 export const INT32_MAX = 2 ** 31 - 1;
+
+export const useSafeLayoutEffect =
+  typeof window !== 'undefined' ? useLayoutEffect : useEffect;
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -100,3 +102,6 @@ export function deepEqual(obj1: any, obj2: any) {
 }
 
 export const INT32_MAX = 2 ** 31 - 1;
+
+export const useSafeLayoutEffect =
+  typeof window !== 'undefined' ? useLayoutEffect : useEffect;
```
satya164, line 107: `window` isn't a good way to detect a browser environment. `Platform.OS === 'web' && typeof document !== 'undefined'` would better detect browsers.

latekvo: Hey @satya164, thank you for Your feedback.

We've found [this](https://github.com/reduxjs/react-redux/blob/7e2fdd4ee2021e4282e12ba9fc722f09124e30cd/src/utils/useIsomorphicLayoutEffect.ts#L36) implementation, which seems the most comprehensive out of the ones we've seen.

Despite this, we decided to halt any changes in this PR for now, as using `useIsomorphicEffect` would likely only hide the SSR warning, while not preventing the issue that caused us to use the `useLayoutEffect` in the first place.

You can find the full explanation for our decision [here](https://github.com/software-mansion/react-native-gesture-handler/issues/3375#issuecomment-2707808893).

PR #3409, opened 2025-02-14
Fix border styling issues in `RNGestureHandlerButton`

latekvo:
## Description

This PR adds missing prop styles fields to the RNGestureHandlerButton delegate,
and rewrites `ButtonViewGroup` to use `ReactViewGroup` instead of `ViewGroup`.


j-piasecki:
I'm not sure whether this is the right approach. React Native overrides the native `onTouchEvent` in its ViewGroup: https://github.com/facebook/react-native/blob/326467c856de9a8038688828c0b34c084e1a07dd/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.kt#L269-L280, and we would have no way of restoring the default behavior. I'd assume this would break some native interactions (like delaying the press inside a scroll view).

I think an approach with two views, where the outer one applies styling and the inner one is the actual button, would be safer and work on both platforms (iOS suffers from the same problem). We tried that in the past and failed, but now we could build it relying on `display: contents` not to break any layout.

m-bert:
Closing this as #3634 was merged.

PR #3416, opened 2025-02-20
Remove unnecessary version checks

m-bert:
## Description

This PR removes checks for RN versions that are no longer necessary in our codebase.

## Test plan

Run example apps


android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -548,7 +548,7 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
         val event = RNGestureHandlerEvent.obtain(
           handler,
           handlerFactory.createEventBuilder(handler),
-          useTopPrefixedName = BuildConfig.REACT_NATIVE_MINOR_VERSION >= 71
+          true
```
j-piasecki, line 551: Can we just remove `useTopPrefixedName` entirely or is it used somewhere else?

j-piasecki (approved):

PR #3424, opened 2025-02-25
Update versions in compatibility tables

m-bert:
## Description

This PR updates minimal required version of `react-native` in compatibility tables. It also adds note about supporting 3 latest RN versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,11 @@ You will need to have an Android or iOS device or emulator connected.
 
 ## React Native Support
 
+In general, `react-native-gesture-handler` supports the three latest stable releases of `react-native`.
+
 | version | react-native version |
 | ------- | -------------------- |
+| 2.24.0+ | 0.75.0+              |
 | 2.21.0+ | 0.74.0+              |
 | 2.18.0+ | 0.73.0+              |
 | 2.16.0+ | 0.68.0+              |
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ You will need to have an Android or iOS device or emulator connected.
 
 ## React Native Support
 
-In general, `react-native-gesture-handler` supports the three latest stable releases of `react-native`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,8 +35,11 @@ You will need to have an Android or iOS device or emulator connected.
 
 ## React Native Support
 
+In general, `react-native-gesture-handler` supports the three latest stable releases of `react-native`.
```
latekvo, line 38: I think it'd make sense to clarify that we support 3 latest `minor` releases of `react-native`.

>In general, `react-native-gesture-handler` supports the three latest stable minor releases of `react-native`.

Also not sure if `In general` fits here, maybe it could be replaced with a more formal `unless noted otherwise`?

>Unless noted otherwise, `react-native-gesture-handler` supports the three latest stable minor releases of `react-native`.

m-bert: > Also not sure if In general fits here,

I think it does, but on the other hand we don't really need it. I've removed it in [02713aa](https://github.com/software-mansion/react-native-gesture-handler/pull/3424/commits/02713aaf6472c7a8dc612ed9747a6c0cb38a2a85), along with changing _**stable**_ to _**minor**_.

docs/docs/fundamentals/installation.md
```diff
@@ -85,15 +88,12 @@ import { GestureHandlerRootView } from 'react-native-gesture-handler';
 export function CustomModal({ children, ...rest }) {
   return (
     <Modal {...rest}>
-      <GestureHandlerRootView>
-        {children}
-      </GestureHandlerRootView>
+      <GestureHandlerRootView>{children}</GestureHandlerRootView>
```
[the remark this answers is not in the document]
m-bert: Yep, that's formatter. Brought back in [bc9a537](https://github.com/software-mansion/react-native-gesture-handler/pull/3424/commits/bc9a537ab89faffb5e6f2e36a632bf33dd4c233f)

PR #3425, opened 2025-02-26
Fix: race condition on attachedGestures in updateHandlers

uyt95:
## Description

Our production apps are regularly crashing with the following stacktracke:
```
Non-fatal Exception: io.invertase.firebase.crashlytics.JavaScriptError: Cannot read property 'handlers' of undefined
       at .anonymous(address at node_modules/react-native-gesture-handler/src/handlers/gestures/GestureDetector/updateHandlers.ts:51:ghQueueMicrotask$argument_0)
       at .anonymous(address at node_modules/react-native/Libraries/Core/Timers/immediateShim.js:46:global.queueMicrotask$argument_0)
```
This seems to be caused by a race condition on the number of gestures in `preparedGesture.attachedGestures`. This PR fixes that race condition by storing the value of `preparedGesture.attachedGestures` from when the micro task was scheduled.

## Test plan

Run example apps


m-bert (approved):
Thanks for this PR! ❤️

PR #3431, opened 2025-03-03
Wrap `Pressable` with `forwardRef`

### LINKED ISSUE #3427: Ref not forwarded in Pressable
### Description

The `Pressable` component does not forward refs, so we can't use the native element to do stuff. Especially in the browser this would be useful. And the `Pressable` component from React Native itself also forwards refs.

### Steps to reproduce

Try to pass a ref to the `Pressable` component

not needed

### Gesture Handler version

2.22.1

### React Native version

0.76.7

### Platforms

iOS, Web, Android, MacOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

This PR wraps `Pressable` component with `forwardRef` such that `NativeButton` can be accessible via `ref`.

Fixes #3427

## Test plan

<details>

```jsx
import React, { useEffect } from 'react';
import { Pressable } from 'react-native-gesture-handler';

  const pressableRef = React.useRef(null);

  useEffect(() => {
    console.log(pressableRef);
  }, []);

  return (
      <Pressable
        ref={pressableRef}
        onPress={() => console.log('click')}
        style={{ width: 100, height: 100, backgroundColor: 'crimson' }}
      />
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
});
```

</details>


src/components/Pressable/Pressable.tsx
```diff
@@ -27,412 +35,417 @@ const IS_TEST_ENV = isTestEnv();
...
+
+    const pressAndTouchGesture = useMemo(
+      () =>
+        Gesture.LongPress()
+          .minDuration(INT32_MAX) // Stops long press from blocking native gesture
+          .maxDistance(INT32_MAX) // Stops long press from cancelling after set distance
+          .cancelsTouchesInView(false)
+          .onTouchesDown((event) => {
+            handlingOnTouchesDown.current = true;
+            (pressableRef as RefObject<View>).current?.measure(
```
m-bert, line 242: I don't really like this cast, but it does the job.

src/components/Pressable/Pressable.tsx
```diff
@@ -27,412 +35,417 @@ const IS_TEST_ENV = isTestEnv();
...
-            return;
-          }
-          onHoverOut?.(gestureToPressableEvent(event));
-        }),
-    [delayHoverIn, delayHoverOut, onHoverIn, onHoverOut]
-  );
-
-  const pressDelayTimeoutRef = useRef<number | null>(null);
-  const isTouchPropagationAllowed = useRef<boolean>(false);
+const Pressable = forwardRef(
```
m-bert, line 38: We can also do something similar as in [buttons](https://github.com/software-mansion/react-native-gesture-handler/blob/324a236f7dc6d0234c878f35764b4f2938c37192/src/components/GestureButtons.tsx#L149)

src/components/Pressable/Pressable.tsx
```diff
@@ -27,412 +35,417 @@ const IS_TEST_ENV = isTestEnv();
...
-                longPressMinDuration
-              );
+            pressOutHandler(gestureTouchToPressableEvent(event));
+          }),
+      [
+        activateLongPress,
+        longPressMinDuration,
+        normalizedHitSlop,
+        pressInHandler,
+        pressOutHandler,
```
latekvo, line 324: Now that `pressableRef` is a prop, shouldn't it also be included in the dep array?

latekvo (commented):
I'm also not a big fan of that cast, but it doesn't look like there's a much cleaner solution.

src/components/Pressable/Pressable.tsx
m-bert: `ref` object itself won't change, what could change is its `.current` value. Also, changing `current` won't trigger re-render, so it won't trigger `useMemo` recomputation. Correct me if I'm wrong (cc @j-piasecki)

latekvo: But `pressableRef` is not a `ref` anymore like it was before, it's a prop that's passed from outside. Won't changing the `ref` prop passed to `Pressable` change the `pressableRef` object itself?

latekvo (changes requested):
Looks like `Pressable` now crashes on press when `ref` is not provided due to `pressableRef.current` being accessed even when `pressableRef` is null.

image


Tested on **Android** on **example app > Pressable**

src/components/Pressable/Pressable.tsx
j-piasecki: > Also, changing current won't trigger re-render, so it won't trigger useMemo recomputation.

True, therefore adding it doesn't hurt in any way.

> Won't changing the ref prop passed to Pressable change the pressableRef object itself?

It will.

src/components/Pressable/Pressable.tsx
```diff
@@ -27,412 +35,433 @@ const IS_TEST_ENV = isTestEnv();
...
+          }, unstable_pressDelay);
+        } else {
+          pressInHandler(gestureTouchToPressableEvent(event));
+        }
+
+        onEndHandlingTouchesDown.current?.();
+        onEndHandlingTouchesDown.current = null;
+        handlingOnTouchesDown.current = false;
+      },
+      []
```
latekvo, line 279: Dep array is missing variables which were previously included in the `pressAndTouchGesture` memo:

```
activateLongPress,
longPressMinDuration,
normalizedHitSlop,
pressInHandler,
unstable_pressDelay,
```

src/components/Pressable/Pressable.tsx
```diff
@@ -27,412 +35,433 @@ const IS_TEST_ENV = isTestEnv();
...
+          }),
+      [
+        activateLongPress,
+        longPressMinDuration,
+        normalizedHitSlop,
+        pressInHandler,
+        pressOutHandler,
+        unstable_pressDelay,
+        pressableRef,
+      ]
```
latekvo, line 342: This array may be reduced to `[pressableRef, measureCallback, pressOutHandler]`
(note that `measureCallback` is completely missing from it as of now)

latekvo (commented):
Just 2 small requests

src/components/Pressable/Pressable.tsx
m-bert: Done in [6029346](https://github.com/software-mansion/react-native-gesture-handler/pull/3431/commits/6029346e26bd0f130b8ae893c63683f2c4d70fc1)

src/components/Pressable/Pressable.tsx
m-bert: Also done in [6029346](https://github.com/software-mansion/react-native-gesture-handler/pull/3431/commits/6029346e26bd0f130b8ae893c63683f2c4d70fc1)

latekvo (approved):
Looks good to me 👍

PR #3432, opened 2025-03-05
Bump `actions` on CI

m-bert:
## Description

As stated in [github blog](https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down), `actions/cache@v2` is now deprecated. Also, our CIs fail for this reason. This PR bumps given action to `v4`.

## Test plan

Check that Ci passes.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -23,14 +23,14 @@ jobs:
       cancel-in-progress: true
     steps:
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       - name: Use Java 17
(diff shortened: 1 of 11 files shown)


latekvo (commented):
I think we can go ahead and bump other actions as well, like: `actions/checkout@v4` or `actions/setup-node@v4` (I'm sure there are more that I missed).

We're using a lot of deprecated actions, and it looks like the newer ones are backwards-compatible.

latekvo (approved):
Looking good, just please wait for the CIs to pass.

PR #3457, opened 2025-03-21
Migrate to Yarn 4

m-bert:
## Description

This PR changes package manager to _**Yarn 4**_ (**_modern_**)

I think `workspaces` can be set up along with monorepo setup

## Test plan

Build and run example apps


example/.gitignore
```diff
@@ -77,3 +77,12 @@ buck-out/
...
 ios/
+
+# Yarn
+.yarn/*
+!.yarn/cache
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
```
tjzel, line 88:
```suggestion
!.yarn/versions

```

tjzel (approved):
Looks good, I strongly suggest adding a Yarn validation CI to check for potential `yarn.lock` poisoning on CI.

https://github.com/software-mansion/react-native-reanimated/blob/main/.github/workflows/yarn-validation.yml

.github/workflows/static-example-apps-checks.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           node-version: 18
           cache: 'yarn'
       - name: Install root node dependencies
-        run: yarn
+        run: yarn && yarn prepare
```
j-piasecki, line 30: Shouldn't this also use `yarn install --immutable`?

j-piasecki (approved):
I would do as señor @tjzel said.

PR #3468, opened 2025-03-29
Upgrade `react-native-builder-bob`

satya164:
## Description

I noticed that the version used is quite old. The newer versions have a bunch of fixes and useful enhancements such as:

- Error checking: esnures paths in `package.json` don't deviate from build output
- Generation of declarationMap: so consumers can jump to source
- Parallel builds: the build is ~1s faster in gesture-handler repo
- Support for custom babel config
- Updated browserlist data

**Before**

CleanShot 2025-03-29 at 18 07 51@2x

**After**

CleanShot 2025-03-29 at 18 08 28@2x

## Test plan

Run `yarn bob build` and verify that the project builds


m-bert (approved):
Thank you!

PR #3473, opened 2025-04-03
Add relation props to `Pressable` and `ReanimatedSwipeable`

m-bert:
## Description

It has been reported that if `Pressable` is placed under the view with attached `GestureDetector`, it doesn't work since it gets cancelled by other gestures. This PR adds relation properties to `Pressable` and `ReanimatedSwipeable`, so we can make interactions with external handlers possible.

> [!NOTE]
> This change was partially added into `ReanimatedSwipeable` in #3324 

## Test plan

<details>

```jsx
import React from 'react';
import {
  Gesture,
  Pressable,
  GestureDetector,
  GestureHandlerRootView,

  const gesture = Gesture.Pan();

  return (
    <GestureHandlerRootView>
        <View style={styles.buttonContainer} collapsable={false}>
          <Pressable
            style={styles.button}
            onPressIn={() => console.log('Pressable onPressIn')}
            onPressOut={() => console.log('Pressable onPressOut')}
            simultaneousWithExternalGesture={gesture}
          />
        </View>
      </GestureDetector>
    </GestureHandlerRootView>
  );
}

  buttonContainer: {
    flex: 1,
    alignItems: 'center',
    backgroundColor: '#535353',
  },
  button: {
    width: 200,
    height: 200,
    borderRadius: 30,
    backgroundColor: '#1db954',
  },
});
```

</details>

### DIFF
--- a/src/components/Pressable/Pressable.tsx
+++ b/src/components/Pressable/Pressable.tsx
@@ -29,6 +29,11 @@ import {
 import { PressabilityDebugView } from '../../handlers/PressabilityDebugView';
 import { GestureTouchEvent } from '../../handlers/gestureHandlerCommon';
 import { INT32_MAX, isFabric, isTestEnv } from '../../utils';
+import {
+  applyRelationProp,
+  RelationPropName,
+  RelationPropType,
+} from '../utils';
 
 const DEFAULT_LONG_PRESS_DURATION = 500;
 const IS_TEST_ENV = isTestEnv();
@@ -58,8 +63,16 @@ const Pressable = forwardRef(
       disabled,
       accessible,
       simultaneousWithExternalGesture,
+      requireExternalGestureToFail,
+      blocksExternalGesture,
       ...remainingProps
     } = props;
+
+    const relationProps = {
+      simultaneousWithExternalGesture,
+      requireExternalGestureToFail,
+      blocksExternalGesture,
+    };
 
     const [pressedState, setPressedState] = useState(testOnly_pressed ?? false);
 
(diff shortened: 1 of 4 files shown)


j-piasecki:
Why only simultaneous?

m-bert:
> Why only simultaneous?

Since mentioned PR added only `simultaneous` 😅 This one was necessary for given issue. 

I thought about adding others as well (`requireExternalGestureToFail` and `blocksExternalGesture`), but I've decided that it can be done in the follow-up in `Swipeable` and `Pressable`.

However, if you think that it's better, we can kill two birds with one stone and add it in this PR to both components.

j-piasecki:
Yeah, I don't see a reason not to have all of them available.

src/components/Pressable/Pressable.tsx
```diff
@@ -57,9 +62,18 @@ const Pressable = forwardRef(
...
+      requireExternalGestureToFail,
+      blocksExternalGesture,
       ...remainingProps
     } = props;
 
+    const relationProps = {
+      simultaneousWithExternalGesture,
+      requireExternalGestureToFail,
+      blocksExternalGesture,
+    };
```
latekvo, line 75: Lint warnings: `Unsafe assignment of an 'any' value`
image

m-bert: As discussed, those errora were results of one of the vscode extensions and are probably irrelevant

latekvo (approved):
Works great 👍

PR #3484, opened 2025-04-11
Update how handler configuration is handled on Android

j-piasecki:
## Description

All handler factories were stored in the `RNGestureHandlerModule` which always seemed like an odd place to keep them in to me. This PR moves each factory to be a nested class of the relevant gesture handler. This gives us private access to the handler inside the factory, which in turn allows to get rid of the setter methods and assign directly instead. Those were with use since RNGH was written in java, but we have Kotlin now. It's time to let go and use proper syntax.

It also adds default constants for each config property that didn't have it before.

## Test plan

Example app

### DIFF
--- a/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
@@ -186,11 +186,11 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
 
     override fun configure(handler: LongPressGestureHandler, config: ReadableMap) {
       super.configure(handler, config)
-      if (config.hasKey(KEY_LONG_PRESS_MIN_DURATION_MS)) {
-        handler.minDurationMs = config.getInt(KEY_LONG_PRESS_MIN_DURATION_MS).toLong()
+      if (config.hasKey(KEY_MIN_DURATION_MS)) {
+        handler.minDurationMs = config.getInt(KEY_MIN_DURATION_MS).toLong()
       }
-      if (config.hasKey(KEY_LONG_PRESS_MAX_DIST)) {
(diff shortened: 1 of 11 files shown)


android/src/main/java/com/swmansion/gesturehandler/core/TapGestureHandler.kt
```diff
@@ -27,46 +31,19 @@ class TapGestureHandler : GestureHandler<TapGestureHandler>() {
...
-    maxDeltaY = deltaY
-  }
-
-  fun setMaxDist(maxDist: Float) = apply {
-    maxDistSq = maxDist * maxDist
-  }
-
-  fun setMinNumberOfPointers(minNumberOfPointers: Int) = apply {
-    this.minNumberOfPointers = minNumberOfPointers
+    shouldCancelWhenOutside = DEFAULT_SHOULD_CANCEL_WHEN_OUTSIDE
```
j-piasecki, line 46: It looks like `shouldCancelWhenOutside` was reset to `false` on every config update so the default in the initializer wasn't doing anything 🙈

android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
```diff
@@ -21,28 +24,19 @@ class LongPressGestureHandler(context: Context) : GestureHandler<LongPressGestur
...
-  fun setMaxDist(maxDist: Float): LongPressGestureHandler {
-    maxDistSq = maxDist * maxDist
-    return this
-  }
-
-  fun setNumberOfPointers(numberOfPointers: Int): LongPressGestureHandler {
-    numberOfPointersRequired = numberOfPointers
-    return this
+    maxDist = defaultMaxDist
+    shouldCancelWhenOutside = DEFAULT_SHOULD_CANCEL_WHEN_OUTSIDE
```
j-piasecki, line 39: Same

m-bert (approved):

PR #3487, opened 2025-04-14
[Monorepo] Move config files 

m-bert:
## Description

This PR moves config files to root level in monorepo.

> [!NOTE]
> `prepare` was changed to `postinstall`, however this change is already on `main` so it shouldn't be critical.

## Test plan

- [x] Perform test commit to check if commit hooks work
- [x] Build package and install it on external app


packages/react-native-gesture-handler/package.json
```diff
@@ -3,11 +3,10 @@
   "version": "2.25.0",
   "description": "Declarative API exposing native platform touch and gesture system to React Native",
   "scripts": {
-    "prepare": "bob build && husky install",
+    "postinstall": "bob build && cd ../.. && husky install",
```
tjzel, line 6: Ideally you'd call `husky install` in the monorepo root.

tjzel (approved):
LGTM

.github/workflows/static-root-checks.yml
```diff
@@ -12,6 +12,8 @@ jobs:
   check:
     if: github.repository == 'software-mansion/react-native-gesture-handler'
     runs-on: ubuntu-latest
+    env:
+      RNGH_DIR: 'packages/react-native-gesture-handler'
```
j-piasecki, line 16: I assume this should be also applied to other actions, and will come in the future?

README-Fabric.md
j-piasecki, line 1: I think we can get rid of this one.

package.json
```diff
@@ -1,160 +1,26 @@
...
     "lint-staged": "^12.3.2",
-    "madge": "^6.1.0",
     "prettier": "3.3.3",
-    "react": "19.0.0",
-    "react-native": "0.78.0",
-    "react-native-builder-bob": "^0.39.0",
-    "react-native-reanimated": "^3.12.0",
-    "react-test-renderer": "19.0.0",
-    "release-it": "^13.6.5",
     "typescript": "5.0.4"
```
j-piasecki, line 23: Are dependencies needed at all in the root?

.gitignore
```diff
@@ -1 +1,78 @@
-node_modules
\ No newline at end of file
+# OSX
+#
+.DS_Store
```
j-piasecki, line 3: I assume this is the same gitignore as in the package. Can we de-duplicate it?

tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
```
j-piasecki, line 3: Doesn't this have a potential to break stuff? https://www.typescriptlang.org/tsconfig/#baseUrl

.gitignore
tjzel: Actually not, `.gitignore` isn't upstreamed the file tree. IMO it's okay to granular `.gitignore` files.

package.json
tjzel: Prettier, eslint, tsconfig.json, husky etc. are still useful to work on the files in the root. Also, you probably want the root to dictate what version of typescript, prettier etc. is used.

.github/workflows/static-root-checks.yml
m-bert: I decided that it will be better to fix CIs at the end, since they require example apps to work. This one doesn't need  them so I've changed it.

README-Fabric.md
m-bert: I agree, done in [03f1785](https://github.com/software-mansion/react-native-gesture-handler/pull/3487/commits/03f1785299f64ad6dd4e877e3e0ecdc4ec322c5a). Also, I think we can take a closer look at main `README`, but I'd leave it for now

tsconfig.json
m-bert: I don't think so (but I could be wrong). Reanimated also has it in top-level `tsconfig`. Moreover, all nested `tsconfigs` extend this one, so I assume they will treat it as the `baseUrl` points to their directory.

package.json
j-piasecki: Do they also need to be set in the packages?

j-piasecki (approved):
Take a look at https://github.com/software-mansion/react-native-gesture-handler/pull/3487#discussion_r2058553292. If the answer is yes, then go ahead with merge.

package.json
m-bert: If you want to run those from the package (for example `eslint`), then yes. I think we can leave it as it is for now - at this point I've only wanted to duplicate necessary config and move it up to the root as it was easier to work with at the beginning. I'll try to unify them in upcoming PRs.

PR #3490, opened 2025-04-18
[Monorepo] Add `BasicExample` app

m-bert:
## Description

This PR removes currently existing example apps and adds `BasicExample` app - previously `FabricExample`

## Test plan

Build and run `BasicExample` on android and ios


apps/BasicExample/src/App.tsx
```diff
@@ -49,7 +49,7 @@ Stack.setRoutes({
 
 export default function App() {
   return (
-    <GestureHandlerRootView style={{ flex: 1 }}>
+    <GestureHandlerRootView style={{ flex: 1, paddingTop: 50 }}>
```
j-piasecki, line 52: Can this be on `SafeAreaView` (and only on Android) instead?

apps/BasicExample/android/app/build.gradle
```diff
@@ -45,7 +45,7 @@ react {
 
     /* Hermes Commands */
     //   The hermes compiler command to run. By default it is 'hermesc'
-    // hermesCommand = "$rootDir/my-custom-hermesc/bin/hermesc"
+    hermesCommand = "../../node_modules/react-native/sdks/hermesc/osx-bin/hermesc"
```
j-piasecki, line 48: Why only two levels up here, as opposed to four above?

apps/BasicExample/android/gradle.properties
```diff
@@ -10,7 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx512m -XX:MaxMetaspaceSize=256m
-org.gradle.jvmargs=-Xmx4096m -XX:MaxMetaspaceSize=512m
+org.gradle.jvmargs=-Xmx2048m -XX:MaxMetaspaceSize=512m
```
j-piasecki, line 13: TBH, I don't remember why FabricExample had more memory given to it. We can try reducing it.

apps/BasicExample/ios/Podfile
```diff
@@ -2,11 +2,9 @@
 require Pod::Executable.execute_command('node', ['-p',
   'require.resolve(
     "react-native/scripts/react_native_pods.rb",
-    {paths: [process.argv[1]]},
+    {paths: ["../../../node_modules"]},
   )', __dir__]).strip
 
-require_relative '../../scripts/clangd-add-xcode-step.rb'
```
j-piasecki, line 8: I guess, this will ultimately change location, no?

m-bert: Yes, it will 😅

apps/BasicExample/android/gradle.properties
m-bert: Seems that 512 is enough 😅 

Done in [81737e9](https://github.com/software-mansion/react-native-gesture-handler/pull/3490/commits/81737e92962e78d514821af80239f88eabd21e18)

apps/BasicExample/src/App.tsx
m-bert: Yup, changed in [3788cdd](https://github.com/software-mansion/react-native-gesture-handler/pull/3490/commits/3788cdd5096de08d7cd89edda23c342a38b16e59)

apps/BasicExample/android/app/build.gradle
m-bert: I was following [this tutorial] of setting RN app inside monorepo. However, it looks like that this path doesn't make any sense since `react-native` directory is not present in pointed `node_modules` (at least not at this point). I've changed it in [512fc8e](https://github.com/software-mansion/react-native-gesture-handler/pull/3490/commits/512fc8e6dd66fe69e3d8f28647814a0c69628e92)

PR #3489, opened 2025-04-18
Refactor RNGestureHandlerModule on android

j-piasecki:
## Description

- simplifies `RNGestureHandlerRootHelper` a bit
- moves logic related to sending events to a separate class
- moves logic related to resolving handler factories to a separate class

## Test plan

Example & FabricExample apps

PR #3491, opened 2025-04-18
Make base `GestureHandler` on Android non generic

j-piasecki:
## Description

Depends on https://github.com/software-mansion/react-native-gesture-handler/pull/3489

On Android, `GestureHandler` was a generic class specialized by the concrete type of the handler. The only place it was used was `applySelf` method, which allowed for inherited methods to return the concrete handler:

```
tapHandler.setEnabled(true) // returns TapGestureHandler instead of GestureHandler
```

...but we don't use this pattern anywhere. It made much more sense in Java, than it does in Kotlin, where we can simply use `apply`, `let`, etc. when there's a need to do multiple operations on the same object. Now, it simply makes the logic dealing with gesture handlers more complicated as it also needs to keep track of the generic type, which in most cases is `*` so it doesn't give us more type safety.

This PR makes it so that the base `GestureHandler` class is no longer generic, which should make life easier when dealing with non-concrete gesture handlers in the future.

## Test plan

Build Example and FabricExample apps


m-bert (approved):
What an amazing idea ❤️

PR #3494, opened 2025-04-22
[Monorepo] Add `ExpoExample`

m-bert:
## Description

This PR adds `ExpoExample` app into our repository. This will replace old `example`.

## Test plan

Test `ExpoExample` on `iOS`, `android` and `web`


j-piasecki:
Android works, though that's a new one:

https://github.com/user-attachments/assets/d6d795f3-0d39-4b26-b9e2-c4805de683fb

m-bert:
> Ball with velocity is no longer animated

That's a problem with Reanimated. It has already been fixed and released in 3.17.5, but I haven't bumped version eyt 😅

m-bert:
Also for the failed build on `iOS`, I've just tried it on my machine and it seems that I forgot to commit newest `yarn.lock` - in my case it works fine with it (and the outdated one doesn't). Could you please check that?

m-bert:
Okay, it seems that "Chat head" example was crashing because of `safe-area-context` ([see here](https://github.com/AppAndFlow/react-native-safe-area-context/pull/610)), it should be ok now.

j-piasecki:
> Is it deterministic? I've been playing with it for a bit and I couldn't break it 😅

IIRC the message was exactly the same as in the Chat heads, if you were testing after bumping `safe-area-context`, then it was likely fixed.

j-piasecki (approved):
Both platforms build and the issues I've seen previously are resolved 🪨

PR #3496, opened 2025-04-24
[Monorepo] Add `CommonApp`

m-bert:
## Description

This PR adds `CommonApp` that holds shared source code for our examples.

## Test plan

Run `ExpoExample` and `MacOSExample`


j-piasecki:
ExpoExample iOS doesn't build here as well.

ExpoExample Android and MacOSExample build & start.

apps/CommonApp/Gemfile
```diff
@@ -0,0 +1,16 @@
+source 'https://rubygems.org'
```
j-piasecki, line 1: I think most of the config files (ruby, metro, app.json, ...) aren't necessary here as this app won't be built by itself, no?

m-bert: Probably not, main point of this PR was to add shared source for examples and deal with configs in next PRs. However I see that it makes sense to do this right away 😅

m-bert: Done in [ba4ad91](https://github.com/software-mansion/react-native-gesture-handler/pull/3496/commits/ba4ad91a31d918543d7a7b0a6f82e88673ed063e)

j-piasecki (approved):

PR #3290, opened 2024-12-16
Infer accessibility label from children in RNGestureHandlerButton on iOS

oblador:
## Description

When using RectButton and friends without an explicit `accessibilityLabel` prop set, the accessibility label is empty on iOS, whereas on Android it's correctly inferred from its children like other react-native components work. I tracked this down to this component not inheriting from `RCTView` where this algorithm is defined. So to fix this, I simply vendored that function with a new prefix. 

## Test plan

Validated via Maestro Studio.

PR #3499, opened 2025-04-29
[iOS] Avoid unnecessary string allocations

m-bert:
## Description

#3290 introduced recursive algorithm to obtain accessibility label from children. However, unlike in React Native, it allocates `String` at every recursive call. This may lower the performance as those allocations come at high cost. 

## Test plan

Accessibility labels are inferred as before.

### DIFF
--- a/apple/RNGestureHandlerButton.mm
+++ b/apple/RNGestureHandlerButton.mm
@@ -135,13 +135,16 @@
 // Vendored from RCTView.m to infer accessibility label from children
 static NSString *RNGHRecursiveAccessibilityLabel(UIView *view)
 {
-  NSMutableString *str = [NSMutableString stringWithString:@""];
+  NSMutableString *str = nil;
   for (UIView *subview in view.subviews) {
     NSString *label = subview.accessibilityLabel;
     if (!label) {
       label = RNGHRecursiveAccessibilityLabel(subview);
     }
     if (label && label.length > 0) {
+      if (str == nil) {
+        str = [NSMutableString string];
+      }
       if (str.length > 0) {
         [str appendString:@" "];
       }
--- a/apple/RNGestureHandlerButton.mm
+++ b/apple/RNGestureHandlerButton.mm
@@ -151,7 +151,8 @@ static NSString *RNGHRecursiveAccessibilityLabel(UIView *view)
       [str appendString:label];
     }
   }
-  return str.length == 0 ? nil : str;
+
+  return str;
 }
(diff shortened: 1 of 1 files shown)


apple/RNGestureHandlerButton.mm
```diff
@@ -135,13 +135,16 @@ - (NSString *)accessibilityLabel
 // Vendored from RCTView.m to infer accessibility label from children
 static NSString *RNGHRecursiveAccessibilityLabel(UIView *view)
 {
-  NSMutableString *str = [NSMutableString stringWithString:@""];
+  NSMutableString *str = nil;
```
j-piasecki, line 138: The `return` expression should be updated as well - `str` may not be initialized by then.

m-bert: Good catch, done in [026f2e7](https://github.com/software-mansion/react-native-gesture-handler/pull/3499/commits/026f2e7e426faad4d16165962e25faa8d272703b)

j-piasecki (approved):

hirbod:
And what about the infinite recursion? It’s extremely hard to repro, but it can happen. In our case, it was happening randomly when I touched a RectButton inside our LegendList. Sometimes, just backgrounding and foregrounding the app (and likely re-rendering the RectButton in the list) would trigger it. After about 60 seconds of hanging, memory would spike from 300MB to 7GB and the app would crash. It’s hard to tell when and how this happens, but it feels like some kind of bailout mechanism is missing, no?

Screenshot 2025-04-30 at 12 50 44

m-bert:
> And what about the infinite recursion?

We'd like to take a closer look into this issue. Sure, setting max depth is a solution, but maybe we will be able to figure out how to fix it in a better way. Given that, we will do this in another PR. Also it would be great if you could provide a reproduction (if possible).

If you need these changes asap, you can make a patch which reverts [this commit](https://github.com/software-mansion/react-native-gesture-handler/commit/ba942f74fad21a5c9192d35eb915f63004218375). This is where these changes were introduced, so reverting it should make it work again.

hirbod:
I fixed the recursion issue for now by just providing an accessibilityLabel to every RectButton, so I’m good. But I saw a few people on X saying they ran into the same issue, so it might affect more people.

I'll try my best to come up with a repro but this is a tough one for sure.

hirbod:
FYI, the issue is reproducible with `Pressable` by `react-native` as well. So this is a core issue!

PR #3503, opened 2025-04-30
[Monorepo] Setup linters

m-bert:
## Description

This PR sets up `eslint`, `prettier` and `tsc` so that we can finally run lint checks in monorepo.

Each workspace has `ts-check` and `lint-js` scripts which run in given workspace. Running those scripts in `root` results in running them in each workspace. 

> [!NOTE]
> Running those scripts from `root` does not trigger any checks in `ExpoExample` and `MacOSExample`, as they only export app from `CommonApp`

## Test plan

Run `yarn ts-check` and `yarn lint-js` from `root`.


apps/CommonApp/src/ListWithHeader/ListWithHeader.tsx
```diff
@@ -97,46 +97,50 @@ interface ScrollComponentWithOffsetProps extends ScrollViewProps {
...
+    },
+    (offset) => {
+      props.scrollOffset.value = offset;
+    }
+  );
 
-    useEffect(() => {
+  useEffect(() => {
+    if (ref) {
       ref.current = scrollRef.current;
```
j-piasecki, line 121: I don't think this ref is being cleaned up at all, but it's hard to judge how much it matters in the internal test app.

apps/CommonApp/src/new_api/camera/AnimatedCameraView.tsx
```diff
@@ -1,55 +1,56 @@
...
+  const [permission, requestPermission] = useCameraPermissions();
 
-    const animatedProps = useAnimatedProps(() => {
-      return {
-        zoom: props.zoom.value - 1,
-      };
-    });
+  const animatedProps = useAnimatedProps(() => {
+    return {
+      // @ts-ignore zoom value is correct
```
j-piasecki, line 23: What was the error here?

apps/CommonApp/src/release_tests/mouseButtons/index.tsx
```diff
@@ -27,6 +27,7 @@ function Test({ name, gestureHandlers }: TestProps) {
         ]}>
         {gestureHandlers.map((handler, index) => {
           return (
+            // eslint-disable-next-line @eslint-react/no-array-index-key
```
j-piasecki, line 30: I feel like we can disable this for the test apps - I'm pretty sure all of our lists are static, so using index as the key should be fine.

apps/CommonApp/src/release_tests/rectButton/index.tsx
```diff
@@ -71,6 +71,7 @@ type ButtonProps = {
 function Button({ style, text }: ButtonProps) {
   const rectButtonStyles = [styles.button, style];
 
+  // eslint-disable-next-line no-alert
```
j-piasecki, line 74: What's wrong with `alert`?

package.json
```diff
@@ -11,20 +11,23 @@
...
     "eslint-import-resolver-babel-module": "^5.2.0",
-    "eslint-plugin-prettier": "^5.2.1",
+    "eslint-plugin-jest": "27.4.3",
+    "eslint-plugin-prettier": "^5.0.1",
+    "eslint-plugin-react": "^7.37.5",
     "husky": "^8.0.1",
     "lint-staged": "^12.3.2",
     "prettier": "3.3.3",
-    "typescript": "5.0.4"
+    "typescript": "~5.8.3"
```
j-piasecki, line 30: ❤️

packages/react-native-gesture-handler/src/components/GestureComponents.web.tsx
```diff
@@ -31,7 +31,7 @@ export const DrawerLayoutAndroid = () => {
 export const RefreshControl = createNativeWrapper(View);
 
 export const FlatList = React.forwardRef(
-  <ItemT extends any>(props: FlatListProps<ItemT>, ref: any) => (
+  <ItemT,>(props: FlatListProps<ItemT>, ref: any) => (
```
j-piasecki, line 34: Why is a comma there?

m-bert: Because if you don't have comma, it will be treated as `tag` instead of type.

apps/CommonApp/src/release_tests/mouseButtons/index.tsx
m-bert: Actually I've turned this rule into warning, so this comment is not necessary. Removed in [af1663e](https://github.com/software-mansion/react-native-gesture-handler/pull/3503/commits/af1663e2746be9747ed049fa0f891327e7a3fe09)

apps/CommonApp/src/new_api/camera/AnimatedCameraView.tsx
m-bert: Seems like this one was also turned into a warning (I believe it was `@typescript-eslint/no-unsafe-member-access`, but I could be wrong)

apps/CommonApp/src/release_tests/rectButton/index.tsx
j-piasecki: > Furthermore, alert is often used while debugging code, which should be removed before deployment to production.

This is a test app, it's never going to production. Actually, it's not going anywhere. #FreeAlert

apps/BasicExample/package.json
```diff
@@ -7,7 +7,9 @@
     "ios": "react-native run-ios",
     "lint": "eslint .",
     "start": "react-native start",
-    "test": "jest"
+    "test": "jest",
+    "ts-check": "yarn tsc --noEmit",
+    "lint-js": "eslint --ext '.js,.ts,.tsx' src/ && yarn prettier --check './src/**/*.{js,jsx,ts,tsx}'"
```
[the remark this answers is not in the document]
m-bert: Sure, done in [499d751](https://github.com/software-mansion/react-native-gesture-handler/pull/3503/commits/499d7511ccb996c77c6eb9df6a5c98dc0533e37f)

apps/BasicExample/package.json
```diff
@@ -8,7 +8,10 @@
     "lint": "eslint .",
     "start": "react-native start",
     "test": "jest",
-    "postinstall": ""
+    "postinstall": "",
```
j-piasecki, line 11:
```suggestion
```

package.json
```diff
@@ -11,20 +11,24 @@
   ],
   "scripts": {
     "postinstall": "husky",
-    "ts-check": "cd ./packages/react-native-gesture-handler && yarn tsc --noEmit"
+    "ts-check": "yarn workspaces foreach --all --parallel --topological-dev run ts-check",
+    "lint-js": "yarn workspaces foreach --all --parallel --topological-dev run lint-js",
+    "format-js": "yarn workspaces foreach --all --parallel --topological-dev run format-js"
```
j-piasecki, line 16: This is sick

packages/react-native-gesture-handler/package.json
```diff
@@ -8,11 +8,12 @@
     "build": "yarn tsc -p tsconfig.build.json",
     "precommit": "lint-staged",
     "ts-check": "yarn tsc --noEmit",
-    "format:js": "prettier --write --list-different './{src,example,FabricExample,MacOSExample}/**/*.{js,jsx,ts,tsx}'",
+    "format-js": "prettier --write --list-different './src/**/*.{js,jsx,ts,tsx}'",
+    "format:js": "prettier --write --list-different './src/**/*.{js,jsx,ts,tsx}'",
```
j-piasecki, line 12: Can we make this (and `lint:js`) call the other command with `-`?

packages/react-native-gesture-handler/android/spotless.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: "com.diffplug.spotless"
 
 spotless {
     kotlin {
-        target "src/**/*.kt", "reanimated/**/*.kt", "noreanimated/**/*.kt", "svg/**/*.kt", "nosvg/**/*.kt", "common/**/*.kt"
+        target "**/*.kt"
```
j-piasecki, line 5: Since we're changing basically everything, can you check if updating spotless causes any problems?

apps/BasicExample/package.json
m-bert: Well, I was sure that `yarn` disabled calling `pre` and `post` scripts on CI. Turns out that only the output is silenced so I guess we can get rid of this and remove explicitly calling `postinstall`. Done in [ca5aba3](https://github.com/software-mansion/react-native-gesture-handler/pull/3503/commits/ca5aba39acf2a1aeede917fc95a17f78246e3f91).

packages/react-native-gesture-handler/package.json
m-bert: Sure, done in [18ce1a6](https://github.com/software-mansion/react-native-gesture-handler/pull/3503/commits/18ce1a69497a6d96bbd26cb9ec1f179ed3fff189).

packages/react-native-gesture-handler/android/spotless.gradle
m-bert: Well, we can use newer `spotless` 😅 But since this PR is already quite large I've done it in #3507

j-piasecki (approved):

PR #3506, opened 2025-05-07
[Monorepo] Move `scripts` directory to root

m-bert:
## Description

This PR moves `scripts` directory to root of the monorepo. It also

- Brings back `clangd` support,
- Renames some of scripts to be [global](https://yarnpkg.com/features/workspaces#global-scripts)

## Test plan

Check that CI passes and that building iOS app generates `compile_commands.json` file.


packages/react-native-gesture-handler/package.json
```diff
@@ -10,14 +10,14 @@
...
     "format:apple": "find apple/ -iname *.h -o -iname *.m -o -iname *.cpp -o -iname *.mm | xargs clang-format -i",
     "lint-js": "eslint --ext '.js,.ts,.tsx' src/ && yarn prettier --check './src/**/*.{js,jsx,ts,tsx}'",
     "lint:js": "yarn lint-js",
     "lint:android": "./android/gradlew -p android spotlessCheck -q",
-    "circular-dependency-check": "yarn madge --extensions js,ts,tsx --circular src",
-    "architectures-consistency-check": "node ./scripts/codegen-check-consistency.js",
-    "sync-architectures": "node ./scripts/codegen-sync-archs.js"
+    "circular-dependency:check": "yarn madge --extensions js,ts,tsx --circular src",
+    "architectures:consistency-check": "node ../../scripts/codegen-check-consistency.js",
+    "architectures:sync": "node ../../scripts/codegen-sync-archs.js"
```
j-piasecki, line 20: I'm not sure about the names of the scripts. Those colons don't make much sense to me.

m-bert: Well well well... I agree. But also don't 🤷‍♂️

I think that it is because we are used to the fact that colons were meant to be used as "extension" for the script, while now they represent [global scripts](https://yarnpkg.com/features/workspaces#global-scripts). I've added that because I thought that it will be easier to have those scripts as global ones and be able to run them from any place in the repository. On the other hand, I think we rarely call them by hand, so we can leave them with `-` and use them either by navigating to `packages/-react-native-gesture-handler`, or by using `yarn workspace` command. WDYT?

m-bert: Okay, done in [cd22d2f](https://github.com/software-mansion/react-native-gesture-handler/pull/3506/commits/cd22d2f98b5c87fca5e034a88625aa62460dfc58)

j-piasecki (approved):

PR #3507, opened 2025-05-08
[Monorepo] Update `spotless`

m-bert:
## Description

This PR bumps `spotless` (android formatter) to version 7.0.3.

## Test plan

Run `yarn lint:android`


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -82,14 +82,11 @@ class GestureHandlerOrchestrator(
...
-    gestureHandlers.any { !isFinished(it.state) && shouldHandlerWaitForOther(handler, it) }
+  private fun hasOtherHandlerToWaitFor(handler: GestureHandler<*>) = gestureHandlers.any { !isFinished(it.state) && shouldHandlerWaitForOther(handler, it) }
 
-  private fun shouldBeCancelledByFinishedHandler(handler: GestureHandler<*>) =
-    gestureHandlers.any { shouldHandlerWaitForOther(handler, it) && it.state == GestureHandler.STATE_END }
+  private fun shouldBeCancelledByFinishedHandler(handler: GestureHandler<*>) = gestureHandlers.any { shouldHandlerWaitForOther(handler, it) && it.state == GestureHandler.STATE_END }
 
-  private fun shouldBeCancelledByActiveHandler(handler: GestureHandler<*>) =
-    gestureHandlers.any { handler.hasCommonPointers(it) && it.state == GestureHandler.STATE_ACTIVE && !canRunSimultaneously(handler, it) && handler.isDescendantOf(it) }
+  private fun shouldBeCancelledByActiveHandler(handler: GestureHandler<*>) = gestureHandlers.any { handler.hasCommonPointers(it) && it.state == GestureHandler.STATE_ACTIVE && !canRunSimultaneously(handler, it) && handler.isDescendantOf(it) }
```
j-piasecki, line 89: I'm not sure I like those being in a single line. Is there a way to enforce max line width?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -200,7 +200,7 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
 
     var exclusive = true
 
-    private var _backgroundColor = Color.TRANSPARENT
+    private var backgroundColor = Color.TRANSPARENT
```
j-piasecki, line 203: IIRC this was on purpose so as not to conflict with View's `backgroundColor`. It doesn't necessarily mean than it needs the underscore but I don't think we can keep it as `backgroundColor`.

m-bert: So I guess the same goes with `_enabled`? We can't keep `_` as the prefix as the linter throws error, but we may change to to something like `buttonBackgroundColor`, what about that?

j-piasecki: Sounds good, add a comment about avoiding name conflict there.

m-bert: Okay, I've changed `backgroundColor` in [dff404d](https://github.com/software-mansion/react-native-gesture-handler/pull/3507/commits/dff404d902c6c90a74c5dd2aea994d86b980a656) and `enabled` in [585ac44](https://github.com/software-mansion/react-native-gesture-handler/pull/3507/commits/585ac44e2472fbe4cf6b7920dd445327147602fe)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -325,7 +323,15 @@ class RNGestureHandlerButtonViewManager : ViewGroupManager<ButtonViewGroup>(), R
         colorDrawable.setCornerRadii(buildBorderRadii())
       }
 
-      val layerDrawable = LayerDrawable(if (selectable != null) arrayOf(colorDrawable, selectable, borderDrawable) else arrayOf(colorDrawable, borderDrawable))
+      val layerDrawable = LayerDrawable(
+        if (selectable !=
+          null
+        ) {
```
j-piasecki, line 329: 🤦, can we have it on a single line, or is the new spotless that special?

m-bert: Well, it is special in some cases 😅 I've went through changed files and fixed some of those issues, let me know if there's anything else that should be changed. ([9cf0cb9](https://github.com/software-mansion/react-native-gesture-handler/pull/3507/commits/9cf0cb9e1f3c7af79c8fa717dc98530ae1d087cb))

PR #3508, opened 2025-05-09
Update issue template

j-piasecki:
## Description

It's this time of year again - time to update the issue template 🎉 
- Updates the repro section also to mention Gists, points to RN issue template, and moves the snack link
  - I think we should require all issues with a custom reproducer to be based on [RN reproducer template](https://github.com/react-native-community/reproducer-react-native). This will allow us to clearly see all the changes and exact versions of dependencies that need to change compared to an established baseline (including transitive dependencies!) that are necessary for a repro.
- Bumps the placeholder versions to the current ones
- Updates the workflow section as the bare & managed workflows are a thing of the past (and the link was leading to a non-existent page)
- Updates the architecture section to say that the new arch is now enabled by default
  - I also moved the emphasis from codenames to the new/old arch, as it seems more widely used nowadays


m-bert (approved):
I like the idea of mentioning gists 🚀

PR #3519, opened 2025-05-19
[Monorepo] Rename apps

m-bert:
## Description

This PR renames example apps an the ones generated by CLI are not valid (namely `name` field in `package.json` cannot be uppercased).

It also:

- Adds `yarn paper` script into `expo-example` app
- Adds `yarn macos` into `macos-example` app
- Removes `selfReferences` from `common-app` as without it `macos-example` doesn't start (expo seems to handle it better)

## Test plan

Checked that example apps work


apps/common-app/App.tsx
```diff
@@ -307,11 +307,13 @@ function navigate(
...
 function MainScreen({ navigation }: StackScreenProps<ParamListBase>) {
   useEffect(() => {
-    AsyncStorage.multiGet([OPEN_LAST_EXAMPLE_KEY, LAST_EXAMPLE_KEY]).then(
+    void AsyncStorage.multiGet([OPEN_LAST_EXAMPLE_KEY, LAST_EXAMPLE_KEY]).then(
       ([openLastExample, lastExample]) => {
         if (openLastExample[1] === 'true' && lastExample[1]) {
           navigate(navigation, lastExample[1]);
         }
+
+        return;
```
j-piasecki, line 316: Those are funny changes.

j-piasecki (approved):

apps/common-app/App.tsx
m-bert: Yep, this is because `eslint` for some reason started yelling at me and I wanted to avoid adding another ignores/warns 😅

PR #3523, opened 2025-05-21
[Monorepo] Add `pod install` to `postinstall` script

m-bert:
## Description

This PR adds `yarn pods` into `postinstall` script in order to preapre example apps during `yarn` in root.

## Test plan

Check that CI passes.


m-bert:
After further investigation I've found out that CI fails when `ubuntu-latest` is used. Since we don't want to change every CI to use `macos`, I'm closing this PR.

PR #3524, opened 2025-05-22
[Monorepo] Fix `yarn validation` CI

m-bert:
## Description

`yarn validation` CI performs `yarn install` in all changed directories that contain `package.json`. However, we have some directories which should not undergo this procedure:

- DrawerLayout
- ReanimatedDrawerLayout
- Swipeable
- ReanimatedSwipeable
- jest-utils

## Test plan

Hopefully CI on main monorepo PR passes


tjzel (approved):
LGTM

PR #3532, opened 2025-05-26
Update usage of types from React Native to make them compatible with Strict API

j-piasecki:
## Description

Updates usage of some types from `react-native` across RNGH codebase.

Note: typecheck with the strict api enabled will fail on the codegen types in specs - those are available from the package exports since 0.79. Once this is our lowest supported version, we can update those as well.

## Test plan

`yarn ts-check`


packages/react-native-gesture-handler/src/components/ReanimatedSwipeable.tsx
m-bert, line 1: Wouldn't `hitSlop` be in `remainingProps`?

j-piasecki: Yeah, but it's typed as `HitSlop | null | undefined` while new types expect `HitSlop | undefined`

m-bert (approved):

PR #3536, opened 2025-05-28
Add functionality to track subsequent taps

lededje:
## Description

This PR adds support for tracking multiple sequential taps in the Tap Gesture Handler by introducing a new `maxNumberOfTaps` property. This specific use case I was trying to cover are more discrete video interactions, such as skipping forwards or backwards in a video over and over.

Key changes:
- Added new `maxNumberOfTaps` property to the Tap Gesture Handler configuration
- Added support for unlimited taps by allowing `Math.Infinity` as a value

The implementation maintains the existing behavior when `maxNumberOfTaps` is not specified, ensuring backward compatibility with existing code.

## Test plan

Tested on both iOS (iPhone 14 Pro 18.3 - Device) and Android (Pixel 9 13 API 33 - Simulator) platforms the following tap patterns
- Single tap (default behavior)
- Double tap with max 3 taps
- Triple tap with unlimited taps

The changes are exclusively within typescript utilising existing features for the most part, I don't forsee too many issues

### DIFF
--- a/packages/react-native-gesture-handler/src/handlers/TapGestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/handlers/TapGestureHandler.ts
@@ -9,6 +9,7 @@ export const tapGestureHandlerProps = [
   'maxDurationMs',
   'maxDelayMs',
   'numberOfTaps',
+  'maxNumberOfTaps',
   'maxDeltaX',
   'maxDeltaY',
   'maxDist',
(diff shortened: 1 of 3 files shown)


akwasniewski:
Gesture-handler sets the state of tap gesture to active as soon as the minimum number of taps required is reached, which is set using numberOfTaps prop. Adding a prop for max number of taps would require us to change this behaviour, and wait for more taps before moving to active. This is unwanted, as in most use cases we want to register the tap as soon as possible.

Here is a solution for your use case, which uses only existing functionalities:
```
  const isTimeout = useSharedValue(false);
  const tap = Gesture.Tap()
    .onStart(() => { if (!isTimeout.value) { console.log("tapped") } })
    .numberOfTaps(1)
  const failTap = Gesture.Tap()
    .onStart(() => {
      isTimeout.value = true;
      setTimeout(() => {
        isTimeout.value = false;
      }, 500);
    })
    .numberOfTaps(3)
    .runOnJS(true)
  const failOnThree = Gesture.Exclusive(failTap, tap);
  return (
    <GestureDetector
      gesture={failOnThree}>
        <View style={styles.box} />
      </View>
    </GestureDetector>
  );
```

PR #3538, opened 2025-05-30
[Android] Update build tools

m-bert:
## Description

Seems like we fallen behind with our build tools dependencies. This PR:

- Bumps `Kotlin` version to 2.0.21
- Rewrites build scripts into `Kotlin`
- Updates `gradle`
- Removes build warnings

>[!WARNING]
> Looks like `spotless` is a bit picky when it comes to `Kotlin` scripts, so I had to move it to `build.gradle.kts`

## Test plan

- Run `yarn android` in `basic-example` and `expo-example` (build apps with `--warning-mode all`)
- Run `yarn format:android`


packages/react-native-gesture-handler/android/build.gradle.kts
```diff
@@ -0,0 +1,260 @@
...
+
+spotless {
+    kotlin {
+        target("**/*.kt")
+        ktlint().editorConfigOverride(mapOf("indent_size" to "2", "max_line_length" to "120"))
+        trimTrailingWhitespace()
+        leadingTabsToSpaces()
+        endWithNewline()
+    }
+}
```
j-piasecki, line 260: We don't want to use spotless when we're a dependency

packages/react-native-gesture-handler/android/build.gradle.kts
```diff
@@ -0,0 +1,260 @@
+import java.util.Properties
+import java.io.File
+
+plugins {
+    id("com.android.library")
+    id("org.jetbrains.kotlin.android")
+    id("com.diffplug.spotless") version "7.0.4"
```
j-piasecki, line 7: Same

packages/react-native-gesture-handler/android/build.gradle.kts
```diff
@@ -0,0 +1,260 @@
...
+    implementation("org.jetbrains.kotlin:kotlin-stdlib:${project.findProperty("kotlinVersion") ?: project.property("RNGH_kotlinVersion")}")
+}
+
+fun getExternalLibVersion(project: Project): Triple<Int, Int, Int> {
+    val inputFile = File(project.projectDir, "../package.json")
+    val json = groovy.json.JsonSlurper().parseText(inputFile.readText()) as Map<*, *>
+    val libVersion = json["version"] as String
+    val (major, minor, patch) = libVersion.split(".")
+
+    return Triple(major.toInt(), minor.toInt(), patch.toInt())
```
j-piasecki, line 183: In what way does this break when one of the segments is not a number (i.e. `0-rc.1`)?

packages/react-native-gesture-handler/android/build.gradle.kts
```diff
@@ -0,0 +1,260 @@
...
+    id("com.android.library")
+    id("org.jetbrains.kotlin.android")
+    id("com.diffplug.spotless") version "7.0.4"
+}
+
+if (isNewArchitectureEnabled()) {
+    apply(plugin = "com.facebook.react")
+}
+
+fun safeExtGet(prop: String, fallback: Any?): Any? =
```
j-piasecki, line 14: Can this be generic?

packages/react-native-gesture-handler/android/build.gradle.kts
```diff
@@ -0,0 +1,260 @@
...
+fun buildListOfJavaSrcDirs(): List<String> {
+    val dirs = mutableListOf<String>()
+
+    if (shouldUseCommonInterfaceFromReanimated()) {
+        dirs += "reanimated/src/main/java"
+    } else {
+        // Include "common/" only when it's not provided by Reanimated to mitigate
+        // multiple definitions of the same class preventing build
+        dirs += "common/src/main/java"
+        dirs += "noreanimated/src/main/java"
```
j-piasecki, line 218: Actually, can those two be merged into `noreanimated` (likely another PR)?

packages/react-native-gesture-handler/android/build.gradle.kts
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+    
+    if (REACT_NATIVE_MINOR_VERSION >= 77) {
+        // With RN 0.77, ViewManager related functions in the package has different signatures as they 
+        // are no longer nullable
+        dirs += "package77/src/main/java"
+    } else {
+        // TODO: Delete this block:
+        // It's safe to delete this block once we drop support for RN 0.76
```
j-piasecki, line 245: I think this is the moment (in another PR).

packages/react-native-gesture-handler/android/build.gradle.kts
```diff
@@ -0,0 +1,260 @@
...
+    // users to customize the NDK path/version from their root project (e.g. for M1 support)
+    if (rootProject.hasProperty("ndkPath")) {
+        ndkPath = rootProject.extra["ndkPath"] as String
+    }
+    if (rootProject.hasProperty("ndkVersion")) {
+        ndkVersion = rootProject.extra["ndkVersion"] as String
+    }
+
+    defaultConfig {
+        minSdk = safeExtGet("minSdkVersion", 21) as Int
```
j-piasecki, line 76: We also should be able to default to 24 - https://github.com/react-native-community/template/blob/8e6fc9056b491d0a7229286dafdbd3327881f1c9/template/android/build.gradle#L4 (in another PR)

packages/react-native-gesture-handler/android/settings.gradle.kts
```diff
@@ -0,0 +1,23 @@
+pluginManagement {
+    repositories {
+        google()
+        mavenCentral()
+        gradlePluginPortal()
+    }
+
+    plugins {
+        id("com.android.library") version "8.5.0"
+        id("org.jetbrains.kotlin.android") version "2.0.21"
```
j-piasecki, line 10: Shouldn't that be read from the apps config first?

j-piasecki, line 10: Also, assuming this corresponds to the `buildscript` block, don't we need spotless here?

packages/react-native-gesture-handler/android/settings.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+dependencyResolutionManagement {    
+    repositories {
+        google()
+        mavenCentral()
+        gradlePluginPortal()
+    }
+}
+
+include("lib")
+project(":lib").projectDir.mkdirs()
```
j-piasecki, line 23: newline

packages/react-native-gesture-handler/android/build.gradle.kts
m-bert: Yes, I didn't want to make those changes here, but it is definitely right time to do this!

m-bert:
Suppressed by #3544

PR #3551, opened 2025-06-11
fix: replace 'transparent' string with 0x00000000 double to support Fabric

SudoPlz:
## Description

Fabric doesn't run processColor on props, so passing 'transparent' as a string causes a crash in ColorPropConverter.getColor. This change replaces it with the equivalent numeric double (0x00000000), which works both when passed raw and when processed via processColor in non-Fabric environments.

### DIFF
--- a/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
+++ b/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
@@ -455,7 +455,7 @@ const Pressable = forwardRef(
         IS_FABRIC = isFabric();
       }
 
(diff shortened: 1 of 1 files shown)


m-bert:
Hi @SudoPlz, could you provide reproducible example in which `Pressable` crashes?

latekvo:
@SudoPlz I tested this default case on both fabric and paper, and didn't experience a crash.
If you are sure this issue occurs, please provide a minimal repro, and change `Number(0x000...)` to just `0`.

m-bert:
I'm closing this PR as we cannot reproduce the crash. If this issue persists you can open an issue. In that case remember to provide reproduction.

PR #3565, opened 2025-06-17
[Web] Make `PointerTracker` elements nullable

m-bert:
## Description

It was reported that in some cases _Gesture Handler_ crashes on web on the following [line](https://github.com/software-mansion/react-native-gesture-handler/blob/a19c683fa2d2849c5ed72f5d85c1ff9528fe8f6f/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts#L704). Seems like elements in `PointerTracker` may not exist, even though we believed that during our workflow they always do. Casts in `PointerTracker` were introduced because `Map.get` may return `undefined` if the element doesn't exist. Turns out that it actually may be the case, so we decided to remove those casts and handle these cases across our codebase.

While we don't know what exactly causes the fact that elements are not inside `trackedPointers` map, we have two possibilities:

1. Pointer is removed form `PointerTracker` before handler enters `BEGAN` state - removing may happen for example when one lifts pointer from the view (calling `onPointerUp`, etc.).
2. `PointerTracker` is reset before handler enters `BEGAN` state - this one can happen when `pointercancel` event is received.

## Test plan

Tested on `expo-example` on `web`

### DIFF
--- a/packages/react-native-gesture-handler/src/web/tools/PointerTracker.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/PointerTracker.ts
@@ -61,9 +61,9 @@ export default class PointerTracker {
   }
 
   public track(event: AdaptedEvent): void {
-    const element = this.trackedPointers.get(event.pointerId);
+    const pointerData = this.trackedPointers.get(event.pointerId);
 
-    if (!element) {
+    if (!pointerData) {
       return;
     }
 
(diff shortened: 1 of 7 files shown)


packages/react-native-gesture-handler/src/web/tools/PointerTracker.ts
```diff
@@ -114,20 +112,24 @@ export default class PointerTracker {
   }
 
   public getVelocity(pointerId: number) {
-    return {
-      x: this.trackedPointers.get(pointerId)?.velocityX as number,
-      y: this.trackedPointers.get(pointerId)?.velocityY as number,
-    };
+    const element = this.trackedPointers.get(pointerId);
```
j-piasecki:
```suggestion
    const pointer = this.trackedPointers.get(pointerId);
```

m-bert: Well, I'm not sure if I like this idea 😅 When I see `pointer` my instant thought would be that it will be number, but in reality it is an object:

```ts
private readonly _trackedPointers: Map<number, TrackerElement>
```

however, I agree that `element` may not be the best. Since it is similar to events I'd like to call it `PointerEvent` but well... 😅

What about changing it to, for example, `pointerData`?

PR #3571, opened 2025-06-20
Complete `Pressable` refactor.

latekvo:
## Description

This PR introduces a complete `Pressable` refactor.
New `Pressable` version uses a the `StateMachine` class, to define sequences of states that the `Pressable` should transitions between.
The new implementation allows for fewer edge-controls, and features much cleaner code & execution flow.
The new implementation also makes detection and handling of new, unexpected cases much simpler.

### Completion checklist

- [x] `onPress*` callbacks
  - [ ] and modifiers (`hitSlop`, `unstable_pressDelay`)
- [x] `onHover*` callbacks
- [ ] `onLongPress` callback
  - [ ] and modifiers (`delayLongPress`)
- [ ] ref passthrough support
- [x] `Android` support
- [ ] `iOS` support
- [ ] `Web` support
- [ ] `MacOS` support
- [ ] `MacOS` support
- [ ] `style` support

## Test plan

To be written.


packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  constructor(flowDefinitions: FlowDefinition[], label?: string) {
+    this.flows = [];
+    this.label = label;
+    for (const flow of flowDefinitions) {
+      if (flow.isActive === false) {
+        continue;
+      }
+      this.flows.push(new Flow(flow));
+    }
```
m-bert: Correct me if I'm wrong, but only active flows are considered, so we basically disabled `Pressable` on `paper` (except on web) 🤔  

Also aren't those mutually exclusive conditions? I don't see how platform could be both `android` and `iOS` at the same time. In that case,  do we really need array of `Flows`?

latekvo: >Correct me if I'm wrong, but only active flows are considered, so we basically disabled Pressable on paper (except on web) 🤔

This is still a draft, I haven't come around to testing on paper yet.

>Also aren't those mutually exclusive conditions? I don't see how platform could be both android and iOS at the same time. In that case, do we really need array of Flows?

I initially anticipated multiple flows being required per platform.
I'll fix this once i finish mapping flows for all platforms, as the current implementation is simpler to work with and more versatile than any optimised version that i tried to come up with.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,58 @@
...
+        this.reset();
+        this.sendSignal(signal, event);
+      }
+      return;
+    }
+
+    if (step.callback && !this.latestEvent) {
+      // This case indicates an unexpected edge-case that has to be patched
+      console.warn(UNDEFINED_EVENT_ERROR_MESSAGE);
+    }
```
m-bert: I'm pretty sure this will get easily ignored

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,58 @@
...
+  signal: string;
+  callback?: (event: PressableEvent) => void;
+}
+
+const RNGH_ISSUE_URL =
+  'https://github.com/software-mansion/react-native-gesture-handler/issues/new?template=bug-report.yml';
+
+const GH_TAG = '[GESTURE HANDLER] Pressable StateMachine:';
+
+const UNDEFINED_EVENT_ERROR_MESSAGE = `${GH_TAG} Tried calling CB with undefined event argument! Please report this bug: ${RNGH_ISSUE_URL}`;
```
m-bert: I don't think we need URL here. Also why **CB**? Let's just use full form.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+interface StepDefinition {
+  signal: string;
+  callback?: (event: PressableEvent) => void;
+}
+
+const RNGH_ISSUE_URL =
+  'https://github.com/software-mansion/react-native-gesture-handler/issues/new?template=bug-report.yml';
+
+const GH_TAG = '[GESTURE HANDLER] Pressable StateMachine:';
```
m-bert: We have function called [tagMessage](https://github.com/software-mansion/react-native-gesture-handler/blob/36d302e15328d68e31b40e1aec34dd75ce21495f/packages/react-native-gesture-handler/src/utils.ts#L43).

m-bert (commented):
I have mixed feelings about this `undefined` event warning. Something tells me that it will not be reported anyway, unless it breaks something. Have you tested what's happening if it is `undefined`? (cc @j-piasecki)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Fixed in [d0d84ae](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/d0d84ae35e0be951f8d92d63e1771065cd320137)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: > I don't think we need URL here.

Url makes it easier to report the issue.
I don't see a good reason to not include it.

> Also why CB? Let's just use full form.

Fixed in [f75ec15](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/f75ec15efd5588531f3b93bf3103fab03b56f9cb)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,59 @@
...
+const RNGH_ISSUE_URL =
+  'https://github.com/software-mansion/react-native-gesture-handler/issues/new?template=bug-report.yml';
+
+const GH_TAG = tagMessage('Pressable StateMachine:');
+
+const UNDEFINED_EVENT_ERROR_MESSAGE = `${GH_TAG} Tried calling callback with an undefined event argument! Please report this bug: ${RNGH_ISSUE_URL}`;
+
+class StateMachine {
+  private steps: StepDefinition[];
+  private stepIndex: number;
```
m-bert: What about adding `current` into this name?

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,59 @@
...
+  private latestEvent: PressableEvent | undefined;
```
m-bert: For me `lastEvent` would sound better, but feel free to keep it.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  constructor(steps: StepDefinition[]) {
+    this.steps = steps;
+    this.stepIndex = 0;
+    this.latestEvent = undefined;
```
m-bert:
```suggestion
```

latekvo:
> I have mixed feelings about this undefined event warning. Something tells me that it will not be reported anyway, unless it breaks something. Have you tested what's happening if it is undefined?

I did test it. When the `event` is `undefined`, the affected `onPress*` callback simply doesn't get called.
Although, this case used to be more likely to occur, but now that the events are required to progress the state machine, I'm not sure if it's even possible for this warning to be activated.

I think this warning could be removed if you don't have anything against it.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public reset() {
+    this.stepIndex = 0;
+    this.latestEvent = undefined;
```
m-bert: I'm not a huge fan of assigning `undefined`, probably `null` would look better. But as above, it's up to you.

m-bert:
>I think this warning could be removed if you don't have anything against it.

Yes, I agree. If it breaks then I'm pretty sure that we will receive issue anyway 😅  

Also, by "Something tells me that it will not be reported anyway, unless it breaks something." I meant issues, just to clarify 😅

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Done - [d53ddaa](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/d53ddaa34e3ed1dd1dc314098775b181b075867f)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Let me know if the name changes in [43890a7](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/43890a738b36e77f03287da870f8e76cc5d753de) make sense.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Done in [43890a7](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/43890a738b36e77f03287da870f8e76cc5d753de).

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Using `undefined` allows us to use `?:` notation and removes the need for initialisation in the constructor.
Not sure which one is cleaner, I have no strong opinion on this.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Removed all together. Marking as resolved.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Removed all together. Marking as resolved.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,46 @@
+import { PressableEvent } from './PressableProps';
+
+interface StepDefinition {
+  signal: string;
+  callback?: (event: PressableEvent) => void;
+}
+
+class StateMachine {
```
m-bert:
```suggestion
class PressableStateMachine {
```

I'd make it less general

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,46 @@
+import { PressableEvent } from './PressableProps';
+
+interface StepDefinition {
+  signal: string;
```
m-bert: I think we should also call it `event` in this interface 😅

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,46 @@
...
+  callback?: (event: PressableEvent) => void;
+}
+
+class StateMachine {
+  private steps: StepDefinition[];
```
m-bert: What about calling it `states`?

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,46 @@
...
+      }
+      return;
+    }
+
+    this.eventPayload && step.callback?.(this.eventPayload);
+    this.currentStepIndex++;
+
+    if (this.currentStepIndex === this.steps.length) {
+      this.reset();
+    }
```
m-bert: I don't like mixing these 2 conventions in one place. Why don't we change it to 

```suggestion
    if (this.eventPayload && step.callback) {
      step.callback(this.eventPayload);
    }

    this.currentStepIndex++;

    if (this.currentStepIndex === this.steps.length) {
      this.reset();
    }
```

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
m-bert: As discussed, we will use `null`

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
m-bert: I still have some suggestions, mentioned below

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Done in [690fcf3](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/690fcf373195512689f7f3f423f43085aa4eeb09)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Done in [690fcf3](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/690fcf373195512689f7f3f423f43085aa4eeb09)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Done in [690fcf3](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/690fcf373195512689f7f3f423f43085aa4eeb09)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Makes sense, changed in [9371db2](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/9371db28f0c47598c5dd0e657d58d984bb2e93ba)

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -20,473 +22,373 @@ import {
...
+      requestAnimationFrame(() => {
+        (ref ?? fallbackRef).current?.measure((_x, _y, width, height) => {
+          dimensions.current = {
+            width,
+            height,
+          };
+        });
       });
     }
+  }, [dimensionsAfterResize, ref]);
```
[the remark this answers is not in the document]
latekvo: The measurement only occurs on the first load, and on dependency updates (rare).
The `onLayout` prop does not exist on the native button we're using.
Closing as resolved.

packages/react-native-gesture-handler/src/components/Pressable/PressableProps.tsx
```diff
@@ -158,4 +166,11 @@ export interface PressableProps
    * used with the Pressable's gesture handlers.
    */
   blocksExternalGesture?: RelationPropType;
+
+  /**
+   * Defines the dimensions of the Pressable after it's been resized.
+   * This property does not affect Pressable's physical appearance
```
m-bert:
```suggestion
   * This property does not affect Pressable's physical appearance.
```

packages/react-native-gesture-handler/src/components/Pressable/stateDefinitions.ts
```diff
@@ -0,0 +1,90 @@
...
+import { PressableStateMachine } from './StateMachine';
+
+export enum StateMachineEvent {
+  NATIVE_BEGIN = 'nativeBegin',
+  NATIVE_START = 'nativeStart',
+  NATIVE_FINALIZE = 'nativeFinalize',
+  LONG_PRESS_TOUCHES_DOWN = 'longPressTouchesDown',
+}
+
+export function getConfiguredStateMachine(
```
m-bert, line 109: What I'd do is to probably split it into smaller functions, like `getAndroidStateMachine` and so on, and use this general function as a wrapper. But I'll leave that to you to decide.

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  public handleEvent(eventName: string, eventPayload?: PressableEvent) {
+    const step = this.states[this.currentStepIndex];
+    this.eventPayload = eventPayload || this.eventPayload;
+
+    if (step.eventName !== eventName) {
+      if (this.currentStepIndex > 0) {
+        // retry with position at index 0
+        this.reset();
+        this.handleEvent(eventName, eventPayload);
```
m-bert, line 32: Does it work? When you call `reset`, `eventPayload` will be `null`, so if I understand correctly no callback will be called. I can see that those `StateDefinitions` mostly don't have callbacks at the beginning, but I'd like to point out that it may be confusing 😅

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -20,473 +22,372 @@ import {
...
+  } = props;
+
+  const relationProps = {
+    simultaneousWithExternalGesture,
+    requireExternalGestureToFail,
+    blocksExternalGesture,
+  };
+
+  // used only if `ref` is undefined
+  const fallbackRef = useRef<View>(null);
```
m-bert, line 84: What about:

```suggestion
  const pressableRef = ref ?? useRef<View>(null);
```
?

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -20,473 +22,372 @@ import {
...
+
+  const startLongPress = useCallback(
+    (event: PressableEvent) => {
+      if (onLongPress) {
+        cancelLongPress();
+        longPressTimeoutRef.current = setTimeout(() => {
+          isOnPressAllowed.current = false;
+          onLongPress(event);
+        }, delayLongPress ?? DEFAULT_LONG_PRESS_DURATION);
+      }
```
m-bert, line 151: Both are ok

```suggestion
      if (!onLongPress) {
        return;
      }
      
      cancelLongPress();
      longPressTimeoutRef.current = setTimeout(() => {
        isOnPressAllowed.current = false;
        onLongPress(event);
      }, delayLongPress ?? DEFAULT_LONG_PRESS_DURATION);
```

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -20,473 +22,372 @@ import {
...
+        innerHandlePressIn(event);
+      }
+    },
+    [innerHandlePressIn, normalizedHitSlop, unstable_pressDelay]
+  );
+
+  const handlePressOut = useCallback(
+    (event: PressableEvent, success: boolean = true) => {
+      if (!isCurrentlyPressed.current) {
+        // Some prop configurations may lead to handlePressOut being called mutliple times.
```
m-bert, line 204: Do we know these configurations?

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -20,473 +22,372 @@ import {
...
+  );
+
+  const pressAndTouchGesture = useMemo(
+    () =>
+      Gesture.LongPress()
+        .minDuration(INT32_MAX) // Stops long press from blocking Gesture.Native()
+        .maxDistance(INT32_MAX) // Stops long press from cancelling on touch move
+        .cancelsTouchesInView(false)
+        .onTouchesDown((event) => {
+          const pEvent = gestureTouchToPressableEvent(event);
```
m-bert: Why this strange `p` prefix? can we use full form?

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -20,473 +22,372 @@ import {
...
+            stateMachine.reset();
+            handlePressOut(pEvent, false);
+          }
+        })
+        .onBegin(() => {
+          stateMachine.handleEvent(StateMachineEvent.NATIVE_BEGIN);
+        })
+        .onStart(() => {
+          if (Platform.OS !== 'android') {
+            // Gesture.Native().onStart() is broken with Android + hitSlop
```
m-bert, line 319: Is it still the case?

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -20,473 +22,372 @@ import {
...
+
+  const isPressableEnabled = disabled !== true;
+
+  const gestures = [buttonGesture, pressAndTouchGesture, hoverGesture];
+
+  for (const gesture of gestures) {
+    gesture.enabled(isPressableEnabled);
+    gesture.runOnJS(true);
+    gesture.hitSlop(appliedHitSlop);
+    gesture.shouldCancelWhenOutside(Platform.OS === 'web' ? false : true);
```
m-bert:
```suggestion
    gesture.shouldCancelWhenOutside(Platform.OS !== 'web');
```

m-bert:
Also, I can see that web behaves a bit differently:


### Old
https://github.com/user-attachments/assets/c6c30aed-6778-49c9-8a25-ae3cc36384d0

### New
https://github.com/user-attachments/assets/04f0ea66-452c-4494-8cd8-ac116d8d2191

packages/react-native-gesture-handler/src/components/Pressable/PressableProps.tsx
latekvo: Fixed in [0840636](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/084063621813886438dfdfd8209b778c814cd163)

packages/react-native-gesture-handler/src/components/Pressable/stateDefinitions.ts
latekvo: Sure, done in [801b0ec](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/801b0ec318a6e83f338dfd4c0cebb1284359493e)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
latekvo: Yes, it works. If the event turns out to be invalid for the current sequence, the sequence is reset to it's starting state. The event is then applied again to the restarted sequence. This ensures that the state machine never gets stuck, if it wasn't terminated properly.

>`eventPayload` will be `null`

No, it won't.
`this.eventPayload` will be `undefined`, but `eventPayload` could have an event, if the `handleEvent` was originally called with one.


Example of how it works:

Sequence: `A -> B -> C -> D`, where `A` is start, `D` is the terminator.

Let's say the sequence wasn't terminated properly, so we're currently at step `C`.
Event `A` (with payload) is received.
We call `handleEvent('A', event)`, the `handleEvent` recognizes `"A" === "C"` is `false`, so it calls `this.reset()`, then calls itself recursively once again: `this.handleEvent('A', event)`.

This time, everything is behaving as if the sequence was terminated properly. 
Event `A` is received and compared agains expected event `A`. Expression `"A" === "A"` is `true`, so the rest of the workflow is executed as if the `A` event was always expected.

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
latekvo: This will violate the rules of hooks if the `ref` is initially available but then gets removed.

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
latekvo: I think I'll keep the current one. Despite the extra indentation, it seems a little clearer to me.

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
latekvo: Sometimes when one gesture gets cancelled and another passes, `onPressOut` may be called multiple times.
Theoretically all these cases could be picked-out individually, but this is a catch-all solution that doesn't really break anything.

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
latekvo: 'ixed in [600ae44](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/600ae4442e321cd3fa9d62703be7c52cd5375d54)

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
latekvo: Yes, which is why we are disabling it on `android`.
Using `Gesture.Native().onStart()` on `android` turned out to be redundant.

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
m-bert: You're right, I've missed that

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
latekvo: Done in [76dda0a](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/76dda0a3459403129accd0b8acbf91c6a489234e)

packages/react-native-gesture-handler/src/components/Pressable/StateMachine.tsx
m-bert: > then calls itself recursively once again: this.handleEvent('A', event).

Okay, I've missed the fact that it uses `eventPayload` parameter, not `this.eventPayload`

latekvo:
>Also, I can see that web behaves a bit differently:

Fixed in [3b3ed61](https://github.com/software-mansion/react-native-gesture-handler/pull/3571/commits/3b3ed61c0bf7011999e620cbad3fb3f8ae8d5cc8)

m-bert (approved):
I trust you that you've tested those changes 😅 

:shipit::shipit::shipit:

latekvo:
Did last rounds of tests. Merging.

techied:
Thanks for the improvement here. We haven't found any issues yet! Is there a release planned that will include this?

pierroo:
great news!
eagerly waiting for a new release / npm release with this fix, thank you guys!

m-bert:
> Thanks for the improvement here. We haven't found any issues yet! 

Yes, it should be released today / tomorrow 😅

latekvo:
@techied @pierroo

The `Release 2.27.0` is now live with these changes.

PR #3579, opened 2025-06-27
Cleanup `ReanimatedSwipeable`

latekvo:
## Changes

- Replace `forwardRef` with prop `ref`
- Fix possible infinite re-render bug
- Fix invalid dependency lists

## Test plan

Use available Swipeable examples to test if it works.

### DIFF
--- a/packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
+++ b/packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
@@ -1,238 +1,34 @@
 // Similarily to the DrawerLayout component this deserves to be put in a
 // separate repo. Although, keeping it here for the time being will allow us to
 // move faster and fix possible issues quicker
-
-import React, {
-  ForwardedRef,
-  useCallback,
-  useImperativeHandle,
-  useMemo,
-} from 'react';
-import { GestureObjects as Gesture } from '../handlers/gestures/gestureObjects';
-import { GestureDetector } from '../handlers/gestures/GestureDetector';
+import { useMemo, useCallback, useImperativeHandle, ForwardedRef } from 'react';
+import { LayoutChangeEvent, View, I18nManager, StyleSheet } from 'react-native';
+import Animated, {
+  useSharedValue,
+  interpolate,
+  runOnJS,
+  ReduceMotion,
+  withSpring,
+  useAnimatedRef,
+  measure,
+  runOnUI,
+  useAnimatedStyle,
+} from 'react-native-reanimated';
+import { SwipeableProps, SwipeableMethods, SwipeDirection } from '.';
+import { Gesture } from '../..';
 import {
   GestureStateChangeEvent,
   GestureUpdateEvent,
-} from '../handlers/gestureHandlerCommon';
-import type { PanGestureHandlerProps } from '../handlers/PanGestureHandler';
-import type { PanGestureHandlerEventPayload } from '../handlers/GestureHandlerEventPayload';
-import Animated, {
-  ReduceMotion,
-  SharedValue,
-  interpolate,
-  measure,
-  runOnJS,
-  runOnUI,
-  useAnimatedRef,
-  useAnimatedStyle,
-  useSharedValue,
-  withSpring,
-} from 'react-native-reanimated';
+} from '../../handlers/gestureHandlerCommon';
+import { PanGestureHandlerEventPayload } from '../../handlers/GestureHandlerEventPayload';
+import { GestureDetector } from '../../handlers/gestures/GestureDetector';
 import {
-  I18nManager,
-  LayoutChangeEvent,
-  StyleProp,
-  StyleSheet,
-  View,
-  ViewStyle,
-} from 'react-native';
-import { applyRelationProp, RelationPropName, RelationPropType } from './utils';
+  applyRelationProp,
+  RelationPropName,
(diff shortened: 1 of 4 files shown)


packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
```diff
@@ -0,0 +1,606 @@
...
+          close();
+        }
+      });
+
+    Object.entries(relationProps).forEach(([relationName, relation]) => {
+      applyRelationProp(
+        tap,
+        relationName as RelationPropName,
+        relation as RelationPropType
+      );
```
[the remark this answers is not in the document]
latekvo: Good catch.

packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
```diff
@@ -0,0 +1,603 @@
...
+} from '../utils';
+
+const DRAG_TOSS = 0.05;
+
+const DEFAULT_FRICTION = 1;
+const DEFAULT_OVERSHOOT_FRICTION = 1;
+const DEFAULT_DRAG_OFFSET = 10;
+const DEFAULT_ENABLE_TRACKING_TWO_FINGER_GESTURE = false;
+
+const Swipeable = function Swipeable(props: SwipeableProps) {
```
m-bert: Do we need this `const Swipeable = ...` thing?

packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
```diff
@@ -0,0 +1,603 @@
...
+  RelationPropName,
+  RelationPropType,
+} from '../utils';
+
+const DRAG_TOSS = 0.05;
+
+const DEFAULT_FRICTION = 1;
+const DEFAULT_OVERSHOOT_FRICTION = 1;
+const DEFAULT_DRAG_OFFSET = 10;
+const DEFAULT_ENABLE_TRACKING_TWO_FINGER_GESTURE = false;
```
m-bert, line 33: If we have `ReanimatedSwipeableProps`, wouldn't we also like to keep default values in this file? (cc @j-piasecki)

j-piasecki: I can see the point behind keeping type declarations and all of the implementation (which the defaults are IMO a part of) separate. Unless we already do this pattern of keeping defaults along the type definitions, which I don't remember from the top of my head, I'd be fine with keeping this as-is.

Though, the props file can be renamed to `.ts` since it doesn't use JSX (this is something we're doing already).

latekvo: As far as I see, most instances of `DEFAULT_*` values within our codebase are within the same file where they are used. This includes `Pressable.tsx`, most `*GestureHandler.ts` files and all `*GestureHandler.kt` files. 

> Though, the props file can be renamed to .ts since it doesn't use JSX 

Sure, done in [3fdcabb](https://github.com/software-mansion/react-native-gesture-handler/pull/3579/commits/3fdcabbf4b033bb9de585e5c93104ad24a0f5b62)

packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
latekvo: Fixed in [87eb0f3](https://github.com/software-mansion/react-native-gesture-handler/pull/3579/commits/87eb0f3e0dfa364e44f03f93565e10b177d43e48)

m-bert (approved):

packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeableProps.ts
```diff
@@ -0,0 +1,199 @@
...
+  LEFT = 'left',
+  RIGHT = 'right',
+}
+
+export interface SwipeableProps
+  extends Pick<PanGestureHandlerProps, SwipeableExcludes> {
+  /**
+   *
+   */
+  ref?: React.RefObject<SwipeableMethods | null>;
```
olitynskyi-mtb, line 22: React.Ref<SwipeableMethods | null> ?

latekvo: >`type Ref<T> = RefCallback<T> | RefObject<T | null> | null`

Hey, I'm not sure if `ReanimatedSwipeable` supports `React.RefCallback`. This will require some testing, but if it does turn out to work without any changes needed, feel free to open a PR with this change.

olitynskyi-mtb: Yeah, I checked, and they do support a callback — that’s why I said I was surprised that this type isn’t a Ref =)

litinskii: https://github.com/software-mansion/react-native-gesture-handler/pull/3759/files

PR #3584, opened 2025-06-30
Remove ViewManagerWithGeneratedInterface from old arch interfaces

### LINKED ISSUE #3564: Kotlin Compilation Fails – Unresolved Supertype ViewManagerWithGeneratedInterface and  Cannot access 'ViewManagerWithGeneratedInterface' which is a supertype of ...
### Description

React Native: 0.76.5
React: 18.3.1
react-native-reanimated: 3.16.7
react-native-screens: 4.6.0
Gradle: 8.10.2
Kotlin: 2.0.21.  e: RNGestureHandlerButtonViewManager.kt: Cannot access 'ViewManagerWithGeneratedInterface' 
which is a supertype of 'RNGestureHandlerButtonViewManager'. 
Check your module classpath for missing or conflicting dependencies.

### Steps to reproduce

Try to build the Android project using npm run android.
Compilation fails with Kotlin errors.

https://1drv.ms/i/c/ba34a1b67a6d11bb/EWDlcVaAayxHhxD15VpNqm0BLOv-T82jaRwpHbpm42oIXA?e=PeQdhB

### Gesture Handler version

2.23.1

### React Native version

0.76.5

### Platforms

MacOS

### JavaScript runtime

None

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

geraintwhite:
## Description

Fixes #3564

Remove unneeded `ViewManagerWithGeneratedInterface` extend from old arch to preserve backwards compatibility with RN 0.77

## Test plan

App builds and runs on RN 0.77 with old arch with this patch.

### DIFF
--- a/packages/react-native-gesture-handler/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerButtonManagerInterface.java
+++ b/packages/react-native-gesture-handler/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerButtonManagerInterface.java
@@ -11,9 +11,8 @@ package com.facebook.react.viewmanagers;
 
 import android.view.View;
 import androidx.annotation.Nullable;
-import com.facebook.react.uimanager.ViewManagerWithGeneratedInterface;
 
-public interface RNGestureHandlerButtonManagerInterface<T extends View> extends ViewManagerWithGeneratedInterface {
+public interface RNGestureHandlerButtonManagerInterface<T extends View> {
   void setExclusive(T view, boolean value);
   void setForeground(T view, boolean value);
   void setBorderless(T view, boolean value);
--- a/packages/react-native-gesture-handler/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerRootViewManagerInterface.java
+++ b/packages/react-native-gesture-handler/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerRootViewManagerInterface.java
@@ -10,8 +10,7 @@
 package com.facebook.react.viewmanagers;
 
 import android.view.View;
-import com.facebook.react.uimanager.ViewManagerWithGeneratedInterface;
 
-public interface RNGestureHandlerRootViewManagerInterface<T extends View> extends ViewManagerWithGeneratedInterface {
+public interface RNGestureHandlerRootViewManagerInterface<T extends View> {
   // No props
 }


shashank-bhatotia:
## Fix Confirmed - ViewManagerWithGeneratedInterface Interface Issue

I encountered this exact same issue with React Native 0.74.3 and `react-native-gesture-handler@2.27.2`. After digging deep into the compilation errors, here's what I found and how to fix it:

### Root Cause
The issue occurs because `ViewManagerWithGeneratedInterface` is only available in the new architecture (Fabric), but the Java interface files are trying to extend it even when using the old architecture (`newArchEnabled=false`).

Specifically, these two files have problematic interface definitions:
- `node_modules/react-native-gesture-handler/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerButtonManagerInterface.java`
- `node_modules/react-native-gesture-handler/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerRootViewManagerInterface.java`

### Solution
The fix is to remove the `extends ViewManagerWithGeneratedInterface` clause from both interface files, which is exactly what this [PR #3584](https://github.com/software-mansion/react-native-gesture-handler/pull/3584) addresses (good job, @geraintwhite 🎊).

**Manual Fix:**
1. In `RNGestureHandlerButtonManagerInterface.java`, change:
   ```java
   public interface RNGestureHandlerButtonManagerInterface<T extends View> extends ViewManagerWithGeneratedInterface {
   ```
   to:
   ```java
   public interface RNGestureHandlerButtonManagerInterface<T extends View> {
   ```

2. Do the same for `RNGestureHandlerRootViewManagerInterface.java`

3. Remove the corresponding `import com.facebook.react.uimanager.ViewManagerWithGeneratedInterface;` lines

**Permanent Solution with patch-package:**
```bash
# After making the manual changes above
npm install patch-package --save-dev
npx patch-package react-native-gesture-handler

# Add to package.json scripts
"postinstall": "patch-package"
```

### Why This Works
The `ViewManagerWithGeneratedInterface` is only needed for the new architecture's generated interfaces. For the old architecture, the interfaces work perfectly fine without extending this class, and removing the inheritance resolves the compilation errors.

**Tested with:**
- React Native 0.74.3
- react-native-gesture-handler 2.27.2
- Android Gradle Plugin 8.6.0
- Gradle 8.7
- Old Architecture (`newArchEnabled=false`)

Build now completes successfully with no functional impact on gesture handling.

Commenting here because https://github.com/software-mansion/react-native-gesture-handler/issues/3564 is closed

m-bert:
Hi! Is there a situation where compilation on android crashes without this fix while using compatible versions of both, `react-native` and `react-native-gesture-handler`? We could backport those changes if necessary (but probably down to 2.27), but we need a valid reproduction that it is in fact needed.

Also remember, that starting from 0.82, it is [not possible to use old architecture](https://reactnative.dev/blog/2025/10/08/react-native-0.82#new-architecture-only).

@shashank-bhatotia For **Gesture Handler** 2.27.0 you should use React Native 0.78 or higher. Unfortunately 0.74 is not supported.

m-bert:
Hi! Is there a situation where compilation on android crashes without this fix while using compatible versions of both, `react-native` and `react-native-gesture-handler`? We could backport those changes if necessary (but probably down to 2.27), but we need a valid reproduction that it is in fact needed. I'm closing this PR for now, but if you still believe that those changes are necessary, let me know.

Also remember, that starting from 0.82, it is [not possible to use old architecture](https://reactnative.dev/blog/2025/10/08/react-native-0.82#new-architecture-only).

@shashank-bhatotia For **Gesture Handler** 2.27.0 you should use React Native 0.78 or higher. Unfortunately 0.74 is not supported.

shashank-bhatotia:
@m-bert — following up on your "valid reproduction needed" question with a clean repro on a configuration that's within `react-native-gesture-handler@2.27.x`'s stated "three latest minor releases" support window:

**Repro:** RN **0.77.3** + `react-native-gesture-handler@2.27.2` with `newArchEnabled=false`. `./gradlew app:installStagingdebug` fails at `:react-native-gesture-handler:compileDebugKotlin` with `Cannot access 'ViewManagerWithGeneratedInterface' which is a supertype of 'RNGestureHandlerButtonViewManager'`. Same error on `RNGestureHandlerRootViewManager`.

**Why this is real, not just a "use a newer RN" answer:**

1. RN 0.77 removed `com.facebook.react.uimanager.ViewManagerWithGeneratedInterface` from the runtime AAR. Verified by extracting `react-android-0.77.3-debug.aar`'s `classes.jar` and exhaustively listing every `com/facebook/react/**/*.class` entry — no match. The class lived through 0.76 as a backward-compat marker, then was deleted.

2. RN 0.77's own auto-generated view-manager interfaces are plain interfaces with **no supertype** — confirmed by `javap` against `ActivityIndicatorViewManagerInterface.class` and `ModalHostViewManagerInterface.class` straight out of the AAR. There's no replacement; the supertype is just gone.

3. RN's codegen template `GeneratePropsJavaInterface.js` hasn't emitted `extends ViewManagerWithGeneratedInterface` since at least RN 0.73 — verified by fetching `@react-native/codegen@{0.73,0.74,0.75,0.76,0.77}` from unpkg and grepping the template. So the modern codegen output (which `fabric/src/main/java/` uses at build time) is already in the shape this PR proposes.

4. The reason new-arch users don't see this: gh's `android/build.gradle` line ~213 picks `srcDirs += 'fabric/src/main/java'` when `newArchEnabled=true` and codegen runs at build time. With old arch, gradle picks `srcDirs += 'paper/src/main/java'` — those static files, pre-committed in the npm tarball, were generated against an older RN where the supertype still existed in the runtime, and have never been regenerated.

**So the patch in this PR isn't a workaround — it's the upstream-equivalent regeneration of stale pre-shipped paper sources, matching exactly what RN's current codegen would emit if `paper/` were re-generated against RN 0.77.**

Real-world impact: anyone mid-RN-0.77-upgrade who can't yet flip to New Architecture (e.g. blocked by other native modules whose new-arch readiness isn't ready, or by deadline/scope on a release branch) hits this. Bumping RN to 0.78 to dodge the issue isn't always available as an option on a release timeline. The same root cause is also reported against `react-native-maps` (react-native-maps/react-native-maps#5809), `react-native-iap`, etc. — gh isn't unique, but it's especially load-bearing.

Would you reconsider merging — or shipping a 2.27.3 / 2.32.x patch release with this regeneration applied? Happy to add a CI repro project demonstrating the failure, or to tighten the diff (the original PR is already minimal — just two interface files, four lines removed).

cc @geraintwhite — same change you proposed; reviving with the additional evidence.

m-bert:
HI @shashank-bhatotia! 

> Repro: RN 0.77.3 + react-native-gesture-handler@2.27.2

As I stated earlier, for 2.27.2 you should use **React Native 0.78 or higher**. This one should work as we test both platforms build before release. 

> Would you reconsider merging — or shipping a 2.27.3 / 2.32.x patch release with this regeneration applied?

Unfortunately not, 2.27 is quite old and we no longer support it. Also 2.31.1 was released with React Native 0.85, so we would target changes at most to 0.82, not 0.77.

shashank-bhatotia:
Got it, thanks for the clarification. We'll carry the patch locally until we move to RN 0.78+ in a follow-up upgrade. Appreciate the time.

m-bert:
Thank you for your understanding! Let me know if you need any further assistance.

PR #3586, opened 2025-06-30
Release 2.27.0

latekvo:
## Description

Update package version.

Versions in the documentation remain unchanged.


m-bert:
Also, please update `README` and docs to include **2.27.0** in compatibility table.

latekvo:
@m-bert I checked the compatibility tables in both the docs, and the Readme. The tables currently display `2.26.0+ | 0.78.0+`, which is valid. Are you sure we should be adding a redundant entry? As far as I see, we've never done that before.

PR #3591, opened 2025-07-02
[iOS] Fix `GestureDetector` on `Text`

### LINKED ISSUE #3581: iOS Fabric: Gesture.tap() worked on 0.78.2, no longer at 0.80
### Description

```
import {
    Gesture,
    GestureDetector,
```

/../

```
const doubleTap = Gesture.Tap()
        .maxDuration(1250)
        .numberOfTaps(2)
        .onStart(() => {
            if (prodInfo && typeof prodInfo !== "string") {
            handleConfirmStart(prodInfo.start, false);
            handleConfirmStop(prodInfo.stop, false);
            }
        });
```

/../

```
return (
<View>
 <GestureDetector gesture={Gesture.Exclusive(doubleTap)}>
                    <Text style={styles.txt50}>
                        {" - "}</Text>
                </GestureDetector>
</View>)
```

This used to work on 0.78.2. It no longer does. It still works on Android and Fabric though. I have tried with a simple console.log in onStart() which does not fire either.

Please let me know it anyone else is experiencing the same. If not, it might be that I'm specifically looking for double taps on links in another nav screen. Not sure why this would stop working on 0.80 though.

It is the same in both debug on iPhone 16 Pro simulator and release on real iPhone 14 device. Both on iOS 18 (which worked on 78.2).

### Steps to reproduce

Please see above.

www.unfortunatelynot.com

### Gesture Handler version

2.26.0

### React Native version

0.80.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 16 Pro

### Acknowledgements

Yes

m-bert:
### Problem

Currently the following configuration:

```jsx
<GestureDetector ... >
  <Text> ... </Text>
</GestureDetector>
```

does not work on `iOS`. This is due to change `react-native` introduced in 0.79 - `hitTest` in `RCTParagraphTextView` now returns `nil` by default ([see here](https://github.com/facebook/react-native/blob/dcbbf275cbc4150820691a4fbc254b198cc92bdd/packages/react-native/React/Fabric/Mounting/ComponentViews/Text/RCTParagraphComponentView.mm#L379)). This results in native `UIGestureRecognizer` not responding to touches.

### Solution

We no longer attach native recognizer to `RCTParagraphTextView`, but to its parent - `RCTParagraphComponentView`. The problem with this approach is that `handleGesture` method uses `reactTag` property, which on `RCTParagraphComponentView` is `nil`. This is why we use `reactTag` from `RCTParagraphTextView` when sending event to `JS` side.

Fixes #3581

## Test plan

<details>

```jsx
import React from 'react';
import { StyleSheet, Text } from 'react-native';
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,

  const g = Gesture.Tap().onEnd(() => {
    console.log('Tapped!');
  });

  return (
      <GestureDetector gesture={g}>
        <Text>
          Click me
          <Text> Me too! </Text>
        </Text>
      </GestureDetector>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandler.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandler.mm
@@ -228,6 +228,9 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
 #endif
   self.recognizer.delegate = self;
 
+  // Starting from react-native 0.79 `RCTParagraphTextView` overrides `hitTest` method to return `nil`. This results in
+  // native `UIGestureRecognizer` not responding to gestures. To fix this issue, we attach recognizer to its parent,
+  // i.e. `RCTParagraphComponentView`.
   RNGHUIView *recognizerView = [self isViewParagraphComponent:view.superview] ? view.superview : view;
 
   [recognizerView addGestureRecognizer:self.recognizer];
@@ -257,6 +260,10 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
 #endif
 }
 
+/**
+ This method is used in `handleGesture` to choose appropriate view. `reactTag` in `RCTParagraphComponentView`
+ is `nil`, therefore we want to use `reactTag` from `RCTParagraphTextView`.
+ */
 - (RNGHUIView *)chooseViewForInteraction:(UIGestureRecognizer *)recognizer
 {
   return [self isViewParagraphComponent:recognizer.view] ? recognizer.view.subviews[0] : recognizer.view;
@@ -264,11 +271,11 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer
 {
+  RNGHUIView *view = [self chooseViewForInteraction:recognizer];
+
   // it may happen that the gesture recognizer is reset after it's been unbound from the view,
   // it that recognizer tried to send event, the app would crash because the target of the event
   // would be nil.
-
-  RNGHUIView *view = [self chooseViewForInteraction:recognizer];
   if (view == nil) {
     return;
   }
--- a/packages/react-native-gesture-handler/apple/RNGestureHandler.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandler.mm
@@ -286,7 +286,7 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
   // it may happen that the gesture recognizer is reset after it's been unbound from the view,
   // it that recognizer tried to send event, the app would crash because the target of the event
   // would be nil.
-  if (view == nil) {
+  if (view.reactTag == nil) {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -215,15 +216,31 @@ - (UITouchType)getPointerType
...
+- (BOOL)isViewParagraphComponent:(RNGHUIView *)view
+{
+  return [view isKindOfClass:[RCTParagraphComponentView class]];
+}
+#endif
+
 - (void)bindToView:(RNGHUIView *)view
 {
 #if !TARGET_OS_OSX
   view.userInteractionEnabled = YES;
```
j-piasecki: I'd assume this should be set on the same view as the rest of properties

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -249,12 +266,27 @@ - (RNGestureHandlerEventExtraData *)eventExtraData:(UIGestureRecognizer *)recogn
...
+
 - (void)handleGesture:(UIGestureRecognizer *)recognizer
 {
+  RNGHUIView *view = [self chooseViewForInteraction:recognizer];
+
   // it may happen that the gesture recognizer is reset after it's been unbound from the view,
   // it that recognizer tried to send event, the app would crash because the target of the event
   // would be nil.
-  if (recognizer.view.reactTag == nil) {
+  if (view == nil) {
```
j-piasecki:
```suggestion
  if (view.reactTag == nil) {
```

Are you sure that all other views will have the react tag set?

m-bert: Changed in [36e3441](https://github.com/software-mansion/react-native-gesture-handler/pull/3591/commits/36e3441b92174277f806f3f2fbea28b7f13bf730)

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
m-bert: While it seemed to work fine, I agree that it is a good idea to move it to `recognizerView`. Done in [3d11ef6](https://github.com/software-mansion/react-native-gesture-handler/pull/3591/commits/3d11ef61470eb892113a95b95e84a8b78ea9ab45)

j-piasecki (approved):
I assume you've tested this change on older versions of RN (0.78) as well.

m-bert:
> I assume you've tested this change on older versions of RN (0.78) as well.

If by **_tested_** you mean tested attached example then yes, I did 😅

Daumis102:
Hi, I'm running into this as well. Any idea when this will be released?

m-bert:
Hi @Daumis102! 

> Any idea when this will be released?

I cannot give any ETA since we don't have release plans now. But I will keep that in mind and discuss it with the team 😅

m-bert:
Hi @Daumis102! FYI, I've just released 2.27.2 which includes this PR 😅

Daumis102:
Thank you!

PR #3592, opened 2025-07-03
docs: add Radon IDE banner to TOCItems

kacperkapusciak:
This PR adds a Radon IDE banner to the right-hand side part of the documentation.

Related to https://github.com/software-mansion/react-native-reanimated/pull/7631 and https://github.com/software-mansion/react-native-reanimated/pull/7634

### DIFF
--- a/packages/docs-gesture-handler/src/components/RadonBanner/index.tsx
+++ b/packages/docs-gesture-handler/src/components/RadonBanner/index.tsx
@@ -1,0 +1,79 @@
+import React, { useRef } from 'react';
+import BrowserOnly from '@docusaurus/BrowserOnly';
+import styles from './styles.module.css';
+
+import ArrowRight from '@site/static/img/arrow-right.svg';
+
+const items = [
+  {
+    text: 'An all-in-one IDE for Expo & React Native.',
+    button: 'Try Radon IDE for free',
+  },
+  {
+    text: 'Streamline your React Native & Expo development.',
+    button: 'Install Radon IDE',
+  },
+  {
+    text: 'Set breakpoints and debug React Native apps with ease.',
+    button: 'Try Radon IDE for free',
+  },
+  {
+    text: 'Catch runtime errors right in your editor.',
+    button: 'Install Radon IDE',
+  },
+  {
+    text: 'Preview React Native components in isolation.',
+    button: 'Try Radon IDE',
+  },
+  {
+    text: 'Run, preview, and debug your React Native & Expo apps without leaving VSCode.',
+    button: 'Try Radon IDE for free',
+  },
+  {
+    text: 'Run, preview, and debug apps faster.',
+    button: 'Try our IDE for React Native',
+  },
+  {
+    text: 'Reduce context switching & debug faster.',
+    button: 'Try Radon IDE for free',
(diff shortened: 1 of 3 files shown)


packages/docs-gesture-handler/src/components/RadonBanner/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </div>
+      <span className={styles.ellipseLeft} />
+      <span className={styles.ellipseRight} />
+    </a>
+  );
+}
+
+export default function RadonBanner() {
+  return (
+    <BrowserOnly fallback={<div />}>
```
latekvo:
```suggestion
    <BrowserOnly>
```
`fallback` is optional ([docs](https://docusaurus.io/docs/docusaurus-core#browseronly))

packages/docs-gesture-handler/src/components/RadonBanner/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+      {() => (
+        <>
```
latekvo: Why is the `<>` being used when there is only one child?

packages/docs-gesture-handler/src/theme/TOCItems/index.js
```diff
@@ -1,3 +1,11 @@
 import { TOCItems } from '@swmansion/t-rex-ui';
 
-export default TOCItems;
+import RadonBanner from '../../components/RadonBanner';
+
+export default function TOCItemsWrapper(props) {
+  return (
+    <>
```
latekvo: Same here

packages/docs-gesture-handler/src/components/RadonBanner/index.tsx
kacperkapusciak: i probably had a good reason for that (i have no idea why i added it there)

packages/docs-gesture-handler/src/components/RadonBanner/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </span>
+      </div>
+      <span className={styles.ellipseLeft} />
+      <span className={styles.ellipseRight} />
+    </a>
+  );
+}
+
+export default function RadonBanner() {
+  return <BrowserOnly>{() => <RadonBannerInner />}</BrowserOnly>;
```
latekvo, line 70: Now that I think about it, do we need `BrowserOnly`?
There isn't an app for our documentation, so I'm not sure if it's possible for us not to be on a browser.

I tested it without `BrowserOnly`, and it seems to work fine. Are there any edge-cases which would require said component?

kacperkapusciak: yes, it's needed.

you are running docusaurus in development mode which only runs the client code. in production there are hydration errors because of randomized content in the banner resulting in layout shifts, blinks & rerenders.

latekvo: You're right, my bad. I initially misunderstood what the component is actually for.

latekvo (approved):
Looks good

PR #3597, opened 2025-07-04
Fix Pressable `onPress` callback not working

### LINKED ISSUE #3593: Rerendering Pressable between press-in and press-out causes onPress not to be called
### Description

After upgrading to 2.27.1, I noticed that `onPress` stopped firing for a component that forwards `onPressIn`, `onPress`, and `onPressOut` props to a Pressable. Upon investigating, I found that the component was setting state in `onPressIn`, which caused a rerender and created a new reference for `onPress`. This, in turn, [resets Pressable’s internal state machine](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx#L225).

I understand why changing the onPress handler invalidates the old state, but it's not an obvious behavior. I ended up spending several hours debugging Pressable’s internals before realizing I needed to memoize my event handlers. It might be helpful to mention this requirement in the documentation.

### Steps to reproduce

1. Render a Pressable with **non-memoized** onPress handler
2. Rerender it during a press gesture's lifecycle
3. Pressable's state machine gets reset to the first step, ignoring the now finalized press event

https://gist.github.com/gigobyte/f607dd7cac226ec15a73109451da0e98

### Gesture Handler version

2.27.1

### React Native version

0.79.0

### Platforms

Web, iOS, Android

### JavaScript runtime

None

### Workflow

None

### Architecture

New Architecture (Fabric)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

latekvo:
## Description

Rendering `Pressable` during an active press resulted in `onPress` never being called.
This PR moves `StateMachine` dependencies, such that it is not reset when `handlePressIn` or `handlePressOut` change.

Fixes #3593

## Test plan

- Use the repro provided within #3593
- See how `onPress` is called correctly

### DIFF
--- a/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
+++ b/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
@@ -1,5 +1,6 @@
 import React, {
   useCallback,
+  useEffect,
   useLayoutEffect,
   useMemo,
   useRef,
@@ -35,10 +36,8 @@ import {
   RelationPropName,
   RelationPropType,
 } from '../utils';
-import {
-  getConfiguredStateMachine,
-  StateMachineEvent,
-} from './stateDefinitions';
+import { getConfiguredStates, StateMachineEvent } from './stateDefinitions';
+import { PressableStateMachine } from './StateMachine';
 
 const DEFAULT_LONG_PRESS_DURATION = 500;
 const IS_TEST_ENV = isTestEnv();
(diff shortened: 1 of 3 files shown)


packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -222,10 +221,12 @@ const Pressable = (props: PressableProps) => {
...
 
-  const stateMachine = useMemo(
-    () => getConfiguredStateMachine(handlePressIn, handlePressOut),
-    [handlePressIn, handlePressOut]
-  );
+  const stateMachine = useMemo(() => new PressableStateMachine(), []);
+
+  useEffect(() => {
+    const configuration = getConfiguredStates(handlePressIn, handlePressOut);
+    stateMachine.setStates(configuration);
```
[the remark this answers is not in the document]
latekvo: Invalid suggestion. Closing as resolved.

packages/react-native-gesture-handler/src/components/Pressable/stateDefinitions.ts
```diff
@@ -88,28 +88,28 @@ function getMacosStateMachine(
...
         handlePressIn(event);
         handlePressOut(event);
       },
     },
-  ]);
+  ];
 }
 
-export function getConfiguredStateMachine(
+export function getConfiguredStates(
```
m-bert: If we change this, can we also rename other functions?

latekvo: Thanks for noticing, fixed in [653b6b5](https://github.com/software-mansion/react-native-gesture-handler/pull/3597/commits/653b6b5290e102ef924eef3b33554e6955ea4cfa)

m-bert (approved):

PR #3599, opened 2025-07-04
Implement base for a native detector component

j-piasecki:
## Description

Implements a native `RNGestureHandlerDetector` component based on `display: contents`:
- The view will match the layout of its child
- The lifecycle of gestures is moved to hooks instead of being managed by the detector
- Detector should attach and detach native gesture handlers to itself as it's mounted or unmounted
- Implements sending native events to the detector component
- It's able to work with RN's Animated events

TODO:
- Validate that gestures are correctly attached/detached in all cases (suspense, display: none)
- Validate that the gesture lifecycle is correctly managed in StrictMode
- Correctly handle `NativeViewGestureHandler`, which should be attached to the child (I think)
- Correctly handle the Text component
- Integration with Reanimated

## Test plan

TODO


packages/react-native-gesture-handler/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerRootViewManagerDelegate.java
```diff
@@ -21,6 +21,12 @@ public RNGestureHandlerRootViewManagerDelegate(U viewManager) {
   }
   @Override
   public void setProperty(T view, String propName, @Nullable Object value) {
-    super.setProperty(view, propName, value);
+    switch (propName) {
+      case "moduleId":
+        mViewManager.setModuleId(view, value == null ? 0 : ((Double) value).intValue());
```
WoLewicki, line 26: Wouldn't you want to throw an error here or something similar when `null` is passed? Or is it permitted?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    attachHandlers(newHandlers)
+  }
+
+  fun setModuleId(id: Int) {
+    if (this.moduleId == -1) {
+      this.moduleId = id
+      val handlersToAttach = this.attachedHandlers
+      this.attachedHandlers = emptyList()
+      this.attachHandlers(handlersToAttach)
```
WoLewicki: what is it for? This logic seems weird

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
```diff
@@ -62,14 +71,29 @@ class RNGestureHandlerEvent private constructor() : Event<RNGestureHandlerEvent>
...
       useTopPrefixedName: Boolean = false,
     ): RNGestureHandlerEvent = (EVENTS_POOL.acquire() ?: RNGestureHandlerEvent()).apply {
-      init(handler, dataBuilder, useTopPrefixedName)
+      init(handler, actionType, dataBuilder, useTopPrefixedName)
     }
 
     fun createEventData(dataBuilder: GestureHandlerEventDataBuilder<*>): WritableMap = Arguments.createMap().apply {
       dataBuilder.buildEventData(this)
+      putInt("handlerTag", dataBuilder.handlerTag)
+      putInt("state", dataBuilder.state)
```
WoLewicki, line 87: This is repeated below

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/eventbuilders/TapGestureHandlerEventDataBuilder.kt
```diff
@@ -6,17 +6,10 @@ import com.swmansion.gesturehandler.core.TapGestureHandler
...
-  private val absoluteX: Float
-  private val absoluteY: Float
-
-  init {
-    x = handler.lastRelativePositionX
-    y = handler.lastRelativePositionY
-    absoluteX = handler.lastPositionInWindowX
-    absoluteY = handler.lastPositionInWindowY
-  }
+  private val x: Float = handler.lastRelativePositionX
```
WoLewicki: Aren't those just some prittifying ? Maybe merge it in another PR so this one is smaller?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -0,0 +1,156 @@
...
+
+#if TARGET_OS_OSX
++ (BOOL)shouldBeRecycled
+{
+  return NO;
+}
+#endif
+
+// Needed because of this: https://github.com/facebook/react-native/pull/37274
++ (void)load
```
WoLewicki: For newer RN versions it shouldn't be needed since we have providers like here: https://github.com/software-mansion/react-native-screens/blob/62eb784c6457be366a2daf6d6d074b7a586fa61e/package.json#L162. Or am I messing it up? Not sure.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -0,0 +1,156 @@
...
+    _props = defaultProps;
+    _moduleId = -1;
+  }
+
+  return self;
+}
+
+// TODO: I'm not sure whether this is the correct place for cleanup
+// Possibly allowing recycling and doing this in prepareForRecycle would be better
+- (void)willMoveToWindow:(UIWindow *)newWindow
```
WoLewicki, line 47: Why this method exactly? Do you want it connected to navigating in app?

WoLewicki: What about navigating A -> B -> C and having this view in e.g. B, then going back from C to B, when will it be attached again?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -0,0 +1,156 @@
...
+{
+  return concreteComponentDescriptorProvider<RNGestureHandlerDetectorComponentDescriptor>();
+}
+
+- (void)updateLayoutMetrics:(const facebook::react::LayoutMetrics &)layoutMetrics
+           oldLayoutMetrics:(const facebook::react::LayoutMetrics &)oldLayoutMetrics
+{
+  auto newLayoutMetrics = layoutMetrics;
+  // Override to force hittesting to work outside bounds
+  newLayoutMetrics.overflowInset = {.left = 1, .right = 1, .top = 1, .bottom = 1};
```
WoLewicki, line 96: Why setting it to `1` fixes overflows?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetectorManager.m
```diff
@@ -0,0 +1,14 @@
...
+#import <React/RCTViewManager.h>
+
+@interface RNGestureHandlerDetectorManager : RCTViewManager
+@end
+
+@implementation RNGestureHandlerDetectorManager
+
+RCT_EXPORT_MODULE(RNGestureHandlerDetector)
+
+RCT_EXPORT_VIEW_PROPERTY(handlerTags, NSArray)
```
WoLewicki: I heard rumors that this is no longer needed for new arch

packages/react-native-gesture-handler/apple/RNGestureHandlerEvents.mm
```diff
@@ -243,7 +247,7 @@ - (instancetype)initWithReactTag:(NSNumber *)reactTag
     _viewTag = reactTag;
     _handlerTag = handlerTag;
     _state = state;
-    _prevState = prevState;
+    _previousState = prevState;
```
WoLewicki, line 251: Why this change?

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -13,49 +13,60 @@
...
 #import "RNGestureHandlerButton.h"
-#import "RNGestureHandlerStateManager.h"
 
 #import <React/RCTJSThread.h>
+#import <jsi/jsi.h>
 
-#ifdef RCT_NEW_ARCH_ENABLED
+using namespace gesturehandler;
 using namespace facebook;
+#ifdef RCT_NEW_ARCH_ENABLED
```
WoLewicki: Why using `react` is between `RCT_NEW_ARCH_ENABLED`?

packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerDetectorComponentDescriptor.h
```diff
@@ -0,0 +1,35 @@
...
+class RNGestureHandlerDetectorComponentDescriptor final
+    : public ConcreteComponentDescriptor<RNGestureHandlerDetectorShadowNode> {
+  using ConcreteComponentDescriptor::ConcreteComponentDescriptor;
+  void adopt(ShadowNode &shadowNode) const override {
+    react_native_assert(
+        dynamic_cast<RNGestureHandlerDetectorShadowNode *>(&shadowNode));
+    auto detectorShadowNode =
+        dynamic_cast<RNGestureHandlerDetectorShadowNode *>(&shadowNode);
+
+    detectorShadowNode->setSize({.width = 300, .height = 300});
```
WoLewicki: What are those values?

packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerDetectorState.h
```diff
@@ -0,0 +1,32 @@
...
+  RNGestureHandlerDetectorState(
+      RNGestureHandlerDetectorState const &previousState,
+      folly::dynamic data){};
+  folly::dynamic getDynamic() const {
+    return {};
+  };
+#endif
+};
+
+} // namespace facebook::react
```
WoLewicki: newline

packages/react-native-gesture-handler/src/NativeDetector.tsx
```diff
@@ -0,0 +1,41 @@
...
+      onGestureHandlerStateChange={
+        gesture.config.onGestureHandlerStateChange as any
+      }
+      onGestureHandlerEvent={gesture.config.onGestureHandlerEvent as any}
+      onGestureHandlerTouchEvent={
+        gesture.config.onGestureHandlerTouchEvent as any
+      }
+      animatedEvents={gesture.animatedEvents}
+      // @ts-expect-error _RNGH_MODULE_ID is injected via JSI
+      moduleId={globalThis._RNGH_MODULE_ID}
```
WoLewicki, line 29: you can declare types on global just like reanimated does yeah?

packages/react-native-gesture-handler/src/specs/RNGestureHandlerDetectorNativeComponent.ts
```diff
@@ -0,0 +1,57 @@
...
+  DirectEventHandler,
+  UnsafeMixed,
+  Double,
+} from 'react-native/Libraries/Types/CodegenTypes';
+import type { ViewProps } from 'react-native';
+
+type GestureHandlerEvent = Readonly<{
+  handlerTag: Int32;
+  state: Int32;
+  handlerData: UnsafeMixed;
```
WoLewicki, line 13: You can type it here correctly with hacks similar to the ones from RNSVG and mapbox

packages/react-native-gesture-handler/android/paper/src/main/java/com/facebook/react/viewmanagers/RNGestureHandlerRootViewManagerDelegate.java
j-piasecki: This file is automatically synchronized for the old arch, it doesn't matter.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: https://github.com/facebook/react-native/blob/d2b55ad1bac3196b79a3389270d6dae858d76999/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L647-L649

packages/react-native-gesture-handler/apple/RNGestureHandlerEvents.mm
j-piasecki: It looks better with full word 🤷

packages/react-native-gesture-handler/src/specs/RNGestureHandlerDetectorNativeComponent.ts
j-piasecki: Link plz

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: I want it to correctly attach/detach recognizers when the host view is mounted/unmounted. It's likely that it will break in its current shape but I haven't tested it at all. Added validating navigation to TODO.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: Removed in [47f14d3](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/47f14d37af8b98954f3089abe153067966357be0)

packages/react-native-gesture-handler/src/NativeDetector.tsx
j-piasecki: Done in [097076e](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/097076ed0c40d691083275557c0690f294cf36d8)

packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerDetectorComponentDescriptor.h
j-piasecki: A leftover from before copying the layout from the child node was implemented. Removed.

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
j-piasecki: Removed. I wouldn't pay that much attention to `RCT_NEW_ARCH_ENABLED` blocks at the moment.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetectorManager.m
j-piasecki: Seems like the rumors were correct, removed in [2512608](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/25126082943a052d39628db7877e0bbf274b1526).

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
j-piasecki: Yes

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: Simplified (and added comment) in [a2048fe](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/a2048fef51be0f90d55fd341e941e27085f2c927)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/eventbuilders/TapGestureHandlerEventDataBuilder.kt
j-piasecki: Removed in [e3164c4](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/e3164c4fc2fff990dc118e58b4988ca91dfa1219)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
WoLewicki: Bold move

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
WoLewicki: Are you going to do something about it?

packages/react-native-gesture-handler/src/specs/RNGestureHandlerDetectorNativeComponent.ts
WoLewicki: https://github.com/software-mansion/react-native-svg/blob/15536f737388d66baf0d779dcbc7c2afc7ba1caa/src/fabric/CircleNativeComponent.ts#L41

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
j-piasecki: No

packages/react-native-gesture-handler/src/specs/RNGestureHandlerDetectorNativeComponent.ts
j-piasecki: I don't know if it's worth it - it can be effectively a union of (sometimes complex) objects.

apps/basic-example/src/App.tsx
```diff
@@ -1,59 +1,124 @@
...
+          />
+        </NativeDetector>
+      )}
     </GestureHandlerRootView>
   );
 }
+
+// import * as React from 'react';
+// import { Button } from 'react-native';
+// import {
```
m-bert: Is this another example, or some unused code? 😅

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  fun setModuleId(id: Int) {
+    if (this.moduleId == -1) {
+      this.moduleId = id
+      this.attachHandlers(handlersToAttach ?: return)
+      handlersToAttach = null
+    } else {
+      throw Exception("Tried to change moduleId of a native detector")
+    }
```
m-bert:
```suggestion
    if (this.moduleId == -1) {
      throw Exception("Tried to change moduleId of a native detector")
    }

    this.moduleId = id
    this.attachHandlers(handlersToAttach ?: return)
    handlersToAttach = null

```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -0,0 +1,84 @@
...
+    this.animatedEvents = animatedEvents
+  }
+
+  private fun attachHandlers(newHandlers: List<Int>) {
+    val registry = RNGestureHandlerModule.registries[moduleId]
+      ?: throw Exception("Tried to access a non-existent registry")
+
+    val keep = 0
+    val attach = 1
+    val detach = 2
```
m-bert: What about making it an `enum`, or storing it in `companion object`?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -0,0 +1,84 @@
...
+    for (entry in changes) {
+      if (entry.value == attach) {
+        registry.attachHandlerToView(
+          entry.key,
+          this.id,
+          if (animatedEvents) {
+            GestureHandler.ACTION_TYPE_NATIVE_DETECTOR_ANIMATED_EVENT
+          } else {
+            GestureHandler.ACTION_TYPE_NATIVE_DETECTOR
+          },
```
m-bert: If we use this `if` outside function call, would it fit in one line? I know kotlin doesn't have ternary, but this syntax looks a bit unclear 😅

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -198,6 +211,11 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
   companion object {
     const val NAME = "RNGestureHandlerModule"
 
+    private var nextModuleId = 0
+
+    // TODO: this is likely to break compared to keeping it on module
```
m-bert: 😄

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
```diff
@@ -14,19 +14,21 @@ import com.facebook.react.uimanager.ThemedReactContext
...
   private val orchestrator: GestureHandlerOrchestrator?
   private val jsGestureHandler: GestureHandler?
   val rootView: ViewGroup
   private var shouldIntercept = false
   private var passingTouch = false
 
   init {
+    val registry =
+      RNGestureHandlerModule.registries[moduleId] ?: throw Exception("Tried to access a non-existent registry")
+
```
m-bert, line 27: Is it necessary, if `module` has a `getter` for this use case?

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -287,7 +287,8 @@ - (void)handleGesture:(UIGestureRecognizer *)recognizer
   // it may happen that the gesture recognizer is reset after it's been unbound from the view,
   // it that recognizer tried to send event, the app would crash because the target of the event
   // would be nil.
-  if (view.reactTag == nil) {
+  if (recognizer.view.reactTag == nil && _actionType != RNGestureHandlerActionTypeNativeDetector &&
+      _actionType != RNGestureHandlerActionTypeNativeDetectorAnimatedEvent) {
```
m-bert: What is the reason behind switching from `view` to `recognizer.view`? 🤔

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -298,10 +299,16 @@ - (void)handleGesture:(UIGestureRecognizer *)recognizer
 - (void)handleGesture:(UIGestureRecognizer *)recognizer inState:(RNGestureHandlerState)state
 {
   _state = state;
+
   RNGestureHandlerEventExtraData *eventData = [self eventExtraData:recognizer];
-  RNGHUIView *view = [self chooseViewForInteraction:recognizer];
 
-  [self sendEventsInState:self.state forViewWithTag:view.reactTag withExtraData:eventData];
+  // TODO: do this only for native detector?
```
m-bert: 😄

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.h
```diff
@@ -0,0 +1,28 @@
+#ifdef RCT_NEW_ARCH_ENABLED
```
m-bert: Do we want to use `ifdef`, or switch to `if`? As far as I know `ifdef` checks only whether `RCT_NEW_ARCH_ENABLED` is defined, not for its value (cc @latekvo, he has been looking into this).

Also, since we won't support old architecture, is it necessary?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -0,0 +1,150 @@
...
+#import <React/RCTConversions.h>
+#import <React/RCTFabricComponentsPlugins.h>
+
+#import <react/renderer/components/rngesturehandler_codegen/EventEmitters.h>
+#import <react/renderer/components/rngesturehandler_codegen/Props.h>
+#import <react/renderer/components/rngesturehandler_codegen/RCTComponentViewHelpers.h>
+
+#include <unordered_map>
+
+using namespace facebook::react;
```
m-bert: What about using this namespace in header?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -0,0 +1,150 @@
...
+
+    for (const auto handler : props.handlerTags) {
+      NSNumber *handlerTag = [NSNumber numberWithInt:handler];
+      [handlerManager.registry detachHandlerWithTag:handlerTag];
+    }
+  }
+}
+
+- (void)dispatchStateChangeEvent:
+    (facebook::react::RNGestureHandlerDetectorEventEmitter::OnGestureHandlerStateChange)event
```
```suggestion
    (RNGestureHandlerDetectorEventEmitter::OnGestureHandlerStateChange)event
```

I don't think it is necessary since we use namespace `facebook::react`

m-bert: But don't get me wrong, I don't mind having this - let's just decide whether we should stick to namespaces or not 😅

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -0,0 +1,150 @@
...
+{
+  const auto &newProps = *std::static_pointer_cast<const RNGestureHandlerDetectorProps>(propsBase);
+  const auto &oldProps = *std::static_pointer_cast<const RNGestureHandlerDetectorProps>(oldPropsBase);
+
+  _moduleId = newProps.moduleId;
+  RNGestureHandlerManager *handlerManager = [RNGestureHandlerModule handlerManagerForModuleId:_moduleId];
+  react_native_assert(handlerManager != nullptr && "Tried to access a non-existent handler manager")
+
+      const int KEEP = 0,
+                DROP = 1, ATTACH = 2;
```
m-bert: Same as in kotlin, what about making it an `enum`?

packages/react-native-gesture-handler/apple/RNGestureHandlerEvents.mm
```diff
@@ -193,6 +193,10 @@ - (instancetype)initWithReactTag:(NSNumber *)reactTag
 
 - (NSString *)eventName
 {
+  if (_actionType == RNGestureHandlerActionTypeNativeDetectorAnimatedEvent) {
+    return @"onGestureHandlerAnimatedEvent";
+  }
```
m-bert:
```suggestion
    return _actionType == RNGestureHandlerActionTypeNativeDetectorAnimatedEvent ? @"onGestureHandlerAnimatedEvent" : @"onGestureHandlerEvent";
```

packages/react-native-gesture-handler/apple/RNGestureHandlerNativeEventUtils.mm
```diff
@@ -0,0 +1,84 @@
...
+
+static folly::dynamic rngh_dynamicFromId(id value)
+{
+  if ([value isKindOfClass:[NSNumber class]]) {
+    return [(NSNumber *)value doubleValue];
+  } else if ([value isKindOfClass:[NSArray class]]) {
+    return rngh_dynamicFromArray((NSArray *)value);
+  } else if ([value isKindOfClass:[NSDictionary class]]) {
+    return rngh_dynamicFromDictionary((NSDictionary *)value);
+  }
```
m-bert:
```suggestion
  if ([value isKindOfClass:[NSNumber class]]) {
    return [(NSNumber *)value doubleValue];
  }
  if ([value isKindOfClass:[NSArray class]]) {
    return rngh_dynamicFromArray((NSArray *)value);
  }
  if ([value isKindOfClass:[NSDictionary class]]) {
    return rngh_dynamicFromDictionary((NSDictionary *)value);
  }
```

packages/react-native-gesture-handler/src/specs/NativeRNGestureHandlerModule.ts
```diff
@@ -18,6 +18,8 @@ export interface Spec extends TurboModule {
     newView: Double,
     actionType: Double
   ) => void;
+  // TODO: implement on web (and mac?)
```
m-bert: 😄

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: I replaced this with an assert in [3b1474d](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/3b1474dc934aee532e556120bfd88c979eab62a7), as it shouldn't happen at all.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: Done in [3b1474d](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/3b1474dc934aee532e556120bfd88c979eab62a7)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: It wouldn't - we'd have to duplicate the call though

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: This should be actually solved by keeping the registries in a map.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
j-piasecki: I'm not sure what you mean. We would need an instance of the module to use it.

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
j-piasecki: A merge (fixed now)

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
j-piasecki: Done in [9b7d625](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/9b7d6256b002c8c5067bc16e02dc3463590a948b)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.h
j-piasecki: It's not, removed in [9b7d625](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/9b7d6256b002c8c5067bc16e02dc3463590a948b)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: Done in [9b7d625](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/9b7d6256b002c8c5067bc16e02dc3463590a948b)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: It seems like most of the codegened types were being used without a namespace so I removed it. We can bring it back, though.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: Done in [9b7d625](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/9b7d6256b002c8c5067bc16e02dc3463590a948b)

packages/react-native-gesture-handler/apple/RNGestureHandlerEvents.mm
j-piasecki: [9b7d625](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/9b7d6256b002c8c5067bc16e02dc3463590a948b)

packages/react-native-gesture-handler/apple/RNGestureHandlerNativeEventUtils.mm
j-piasecki: [9b7d625](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/9b7d6256b002c8c5067bc16e02dc3463590a948b)

packages/react-native-gesture-handler/src/specs/NativeRNGestureHandlerModule.ts
j-piasecki: Removed this method entirely in [7443ef2](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/7443ef26ce7ea8991ed312a2ede740f47b2fb413). It shouldn't be necessary when the native detector receives the handler tags as a prop.

apps/basic-example/src/App.tsx
j-piasecki: An older test, removed in [95621a6](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/95621a6cd03ae04f6f00d0e3817421fd83fb838b)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
m-bert: Right, I saw this line below:

```kt
    val module = context.getNativeModule(RNGestureHandlerModule::class.java)!!
```

and when I saw `module` I assumed that we have one 😅

j-piasecki: Ok, so we do have an instance 😅. I don't think it's recommended for views to reference modules, so in the future, it would be best to remove that.

apps/basic-example/src/App.tsx
```diff
@@ -1,59 +1,63 @@
...
-    rightButtonAction: () => {
-      Stack.navigateTo('final');
-    },
-  },
-  final: {
-    component: FinalScreen,
-    title: 'Final Screen',
-  },
-});
+  const tap = useGesture('PanGestureHandler', {
```
m-bert: `tap` or `pan`? 😅 

I know it's an example so probably it will be modified a lot anyway, I just wanted to point it out 😅

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -0,0 +1,86 @@
...
+import com.facebook.react.views.view.ReactViewGroup
+import com.swmansion.gesturehandler.core.GestureHandler
+
+class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
+  private val reactContext: ThemedReactContext
+    get() = context as ThemedReactContext
+  private var handlersToAttach: List<Int>? = null
+  private var attachedHandlers = listOf<Int>()
+  private var moduleId: Int = -1
+  private var animatedEvents: Boolean = false
```
m-bert: What about `shouldBeUseAnimatedEvents`? 😄 

Jokes aside, can we make this more descriptive? Like `shouldSendAnimatedEvents`? Now it looks like array of events and it is quite confusing 😅 Could be shorter, like `usesAnimatedEvents` or `sendsAnimatedEvents`, but let this name reflect what it does 😅

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -0,0 +1,86 @@
...
+
+  fun setHandlerTags(handlerTags: ReadableArray?) {
+    val newHandlers = handlerTags?.toArrayList()?.map { (it as Double).toInt() } ?: emptyList()
```
m-bert, line 20: I assume that we cannot make it `Int` in other way?

```kt
(it as Double).toInt()
```

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -298,10 +299,17 @@ - (void)handleGesture:(UIGestureRecognizer *)recognizer
 - (void)handleGesture:(UIGestureRecognizer *)recognizer inState:(RNGestureHandlerState)state
 {
   _state = state;
+
   RNGestureHandlerEventExtraData *eventData = [self eventExtraData:recognizer];
-  RNGHUIView *view = [self chooseViewForInteraction:recognizer];
 
-  [self sendEventsInState:self.state forViewWithTag:view.reactTag withExtraData:eventData];
+  NSNumber *tag = recognizer.view.reactTag;
```
m-bert: Shouldn't it be `recognizer.view.subviews[0]` in case of `Paragraph`?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -0,0 +1,147 @@
...
+{
+  const auto &newProps = *std::static_pointer_cast<const RNGestureHandlerDetectorProps>(propsBase);
+  const auto &oldProps = *std::static_pointer_cast<const RNGestureHandlerDetectorProps>(oldPropsBase);
+
+  _moduleId = newProps.moduleId;
+  RNGestureHandlerManager *handlerManager = [RNGestureHandlerModule handlerManagerForModuleId:_moduleId];
+  react_native_assert(handlerManager != nullptr && "Tried to access a non-existent handler manager")
+
+      std::unordered_map<int, RNGestureHandlerMutation>
+          changes;
```
m-bert, line 111: Looks weird, have you run `yarn format:apple`?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -0,0 +1,147 @@
...
+
+  for (const auto newHandler : newProps.handlerTags) {
+    changes[newHandler] = changes.contains(newHandler) ? RNGestureHandlerMutationKeep : RNGestureHandlerMutationAttach;
+  }
+
+  for (const auto handlerChange : changes) {
+    NSNumber *handlerTag = [NSNumber numberWithInt:handlerChange.first];
+
+    if (handlerChange.second == RNGestureHandlerMutationAttach) {
+      // TODO: Attach to the child when NativeGestureHandler remounts, track children changes?
```
m-bert: 😄

packages/react-native-gesture-handler/apple/RNGestureHandlerEvents.mm
```diff
@@ -213,21 +214,25 @@ + (NSString *)moduleDotMethod
...
-  NSMutableDictionary *body = [NSMutableDictionary dictionaryWithDictionary:_extraData.data];
-  [body setObject:_viewTag forKey:@"target"];
-  [body setObject:_handlerTag forKey:@"handlerTag"];
-  [body setObject:@(_state) forKey:@"state"];
-  return @[ self.viewTag, @"onGestureHandlerEvent", body ];
+  if (_actionType == RNGestureHandlerActionTypeNativeDetectorAnimatedEvent) {
+    NSMutableDictionary *body = [[NSMutableDictionary alloc] init];
+    [body setObject:_viewTag forKey:@"target"];
+    [body setObject:_handlerTag forKey:@"handlerTag"];
+    [body setObject:@(_state) forKey:@"state"];
```
m-bert, line 221: Not a strong opinion, but we could use ternary to initialize `body` and those 3 `setObject` would be the same for both.

On the other hand, it still requires `if` for returning value, so do as you think is the best 😅

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: We could cast the entire ArrayList, but I don't think it's much better.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: Yes

packages/react-native-gesture-handler/apple/RNGestureHandlerEvents.mm
j-piasecki: Yeah, we'd need two conditionals to do this.

apps/basic-example/src/App.tsx
j-piasecki: [d48b159](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/d48b159b4fe80b15934e57ddebf1eafcae3b38b2) - it will likely be modified so `gesture` should be fine.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: `dispatchesAnimatedEvents` in [d8fe0ef](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/d8fe0ef61290162dc9b5632c1749e4cdc47c5e54)

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
j-piasecki: [55994a0](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/55994a0253687a647d8219f1943cb2b709b7c194)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: Updated the todo in [077f1dc](https://github.com/software-mansion/react-native-gesture-handler/pull/3599/commits/077f1dc2049f9feb1c681194dfe7884be8601557). I think that's best done outside of this PR. Getting native gestures to work with the new detector may be tricky as we will need to attach to the child directly instead of the detector itself.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
m-bert: Yeah, I don't think it changes that much 😅

m-bert (approved):
Let's merge this bad boy 🚀 :shipit: 

This PR got pretty big and given that it targets `next` I think we can merge it and then continue with development 😅

packages/react-native-gesture-handler/android/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+add_library(
+  react_codegen_rngesturehandler_codegen
+  SHARED
+  ${rn_gesture_handler_SRCS}
+  ${rn_gesture_handler_codegen_SRCS}
+)
+
+target_include_directories(react_codegen_rngesturehandler_codegen PUBLIC ../shared/shadowNodes)
+target_include_directories(react_codegen_rngesturehandler_codegen PUBLIC ./build/generated/source/codegen/jni)
```
WoLewicki, line 18: Why do you link codegen files explicitly?

WoLewicki: Oh okay, you use them in your `shared` folder, got it

PR #3606, opened 2025-07-09
Fix `Pressable` requiring `dimensionsAfterResize`

### LINKED ISSUE #3600: [Pressable] dimensionsAfterResize as a prop is not optimal API for setting dimensions
### Description

In our app, we have an animation that scales each new opened window from 0 to full scale. This makes it so that when Pressable is **mounted**, [measure](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx#L118) reports the width and height of the element to be 0 on iOS (dimensions are correct on web). Since Pressable sets the dimensions only once, I had to resort to `dimensionsAfterResize` to set the correct value after `onLayout` on my component.

The problem is that Pressable's `dimensions.current` is only used [within an event handler](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx#L180). It is not a value that needs to be reactive in the context of React i.e. to be a prop. To update dimensionsAfterResize, I currently have to trigger a rerender of my component, which unnecessarily degrades performance, where a simple ref method would suffice.

My feature request is to expose a ref method on Pressable using `useImperativeHandle`, allowing me to imperatively set its dimensions without negatively affecting my app’s performance. Alternatively, a `Pressable` implementation that doesn't depend on `dimensionsAfterResize` for this usecase.

### Steps to reproduce

Example can be seen in the gist

https://gist.github.com/gigobyte/ac9704fb774d67a126768a03f2563bf9

### Gesture Handler version

2.27.1

### React Native version

0.79.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

latekvo:
## Description

The `Pressable` was developed with an incorrect assumption that the `onLayout` prop is not available on the `NativeButton`.

This PR defines the `onLayout` on `RawButtonProps`, and makes use of said prop in `Pressable` to remove the need for the `dimensionsAfterResize` property.

This PR also marks `dimensionsAfterResize` as deprecated.

Fixes #3600

## Test plan

1. Use the provided test code.
2. Notice how the `Pressable`, despite starting out with `0, 0` dimensions, responds correctly to all press events.
3. Use the `Pressable` examples in our example app to confirm there are no new issues with the component.

## Test code

<details>

```tsx
import React, { useEffect } from 'react';
import { StyleSheet } from 'react-native';
import {
  Pressable,
  GestureHandlerRootView,
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withTiming,

  const opacity = useSharedValue(0);

  const containerAnimatedStyle = useAnimatedStyle(() => {
    'worklet';
    return {
      opacity: opacity.value,
      transform: [{ scale: opacity.value }],
    };
  });

  useEffect(() => {
    opacity.value = withTiming(1, { duration: 200 });
  }, [opacity]);

  return (
      <Animated.View style={containerAnimatedStyle}>
        <Pressable style={styles.box} onPress={() => console.log('press')} />
      </Animated.View>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    gap: 10,
  },
  box: {
    width: 150,
    height: 150,
    backgroundColor: 'red',
    alignItems: 'center',
    justifyContent: 'center',
  },
});


```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
+++ b/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
@@ -14,6 +14,7 @@ import {
 } from './PressableProps';
 import {
   Insets,
+  LayoutChangeEvent,
   Platform,
   StyleProp,
   ViewStyle,
@@ -355,11 +356,15 @@ const Pressable = (props: PressableProps) => {
       : processColor(unprocessedRippleColor);
   }, [android_ripple]);
 
+  const setDimensions = useCallback((event: LayoutChangeEvent) => {
+    dimensions.current = event.nativeEvent.layout;
+  }, []);
+
   return (
     <GestureDetector gesture={gesture}>
       <NativeButton
         {...remainingProps}
-        onLayout={(e) => (dimensions.current = e.nativeEvent.layout)}
+        onLayout={setDimensions}
         accessible={accessible !== false}
         hitSlop={appliedHitSlop}
         enabled={isPressableEnabled}
--- a/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
+++ b/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
@@ -364,7 +364,6 @@ const Pressable = (props: PressableProps) => {
     <GestureDetector gesture={gesture}>
       <NativeButton
         {...remainingProps}
-        // @ts-ignore Prop `onLayout` always works. Todo: Define it's availability on NativeButton
         onLayout={(e) => (dimensions.current = e.nativeEvent.layout)}
         ref={ref ?? fallbackRef}
         accessible={accessible !== false}
--- a/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
+++ b/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
@@ -16,7 +16,6 @@ import {
   Insets,
   Platform,
   StyleProp,
-  View,
   ViewStyle,
   processColor,
 } from 'react-native';
@@ -45,7 +44,6 @@ let IS_FABRIC: null | boolean = null;
 
 const Pressable = (props: PressableProps) => {
   const {
-    ref,
     testOnly_pressed,
     hitSlop,
     pressRetentionOffset,
@@ -76,9 +74,6 @@ const Pressable = (props: PressableProps) => {
     requireExternalGestureToFail,
     blocksExternalGesture,
   };
-
-  // used only if `ref` is undefined
-  const fallbackRef = useRef<View>(null);
 
   const [pressedState, setPressedState] = useState(testOnly_pressed ?? false);
 
@@ -365,7 +360,6 @@ const Pressable = (props: PressableProps) => {
       <NativeButton
         {...remainingProps}
         onLayout={(e) => (dimensions.current = e.nativeEvent.layout)}
-        ref={ref ?? fallbackRef}
         accessible={accessible !== false}
         hitSlop={appliedHitSlop}
         enabled={isPressableEnabled}
--- a/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
+++ b/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
@@ -1,7 +1,6 @@
 import React, {
   useCallback,
   useEffect,
-  useLayoutEffect,
   useMemo,
   useRef,
   useState,
@@ -69,7 +68,6 @@ const Pressable = (props: PressableProps) => {
     simultaneousWithExternalGesture,
     requireExternalGestureToFail,
     blocksExternalGesture,
-    dimensionsAfterResize,
     ...remainingProps
   } = props;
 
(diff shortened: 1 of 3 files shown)


packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -46,7 +45,6 @@ let IS_FABRIC: null | boolean = null;
 
 const Pressable = (props: PressableProps) => {
   const {
-    ref,
```
m-bert, line 49: Should we remove this? I'm not sure how many of our users use this, but some of them may (check #3427)

gigobyte: I don't think it should be removed, we started using the ref after it was introduced in a recent version, also it would be a bad practice to remove public API without a valid reason.

m-bert: Yes, I agree 😄

latekvo: The `ref` isn't removed, it is now contained within `remainingProps`.
It is still being passed to `NativeButton` via `{...remainingProps}` exactly how it used to be before this PR.

m-bert: > It is still being passed to NativeButton via {...remainingProps} exactly how it used to be before this PR.

m-bert: > It is still being passed to NativeButton via {...remainingProps} exactly how it used to be before this PR.

My bad, I've just seen that you removed `ref` and missed that it will be in `remainingProps`. But just to be sure, have you tested that? 😅

latekvo: Yes, I tested it, and the `ref` is being passed, and is set correctly on layout. 

Although I just noticed, the `ref` exposed by `Pressable` is typed incorrectly.
It exposes `ref?: React.RefObject<View>;` when it should be exposing `ref?: React.Ref<View>;`.

latekvo: Also found another issue right now, since we now added `onLayout` to the `NativeButton` within `Pressable`, it is overwriting any `onLayout` that is passed to the `Pressable` as a prop.

latekvo: Fixed both issues in [7342966](https://github.com/software-mansion/react-native-gesture-handler/pull/3606/commits/734296672dcc521d422bc3cf6f3763f79a1f715f) and [6f3da5d](https://github.com/software-mansion/react-native-gesture-handler/pull/3606/commits/6f3da5dc3f9fc76f5023b19935f773ffbff04017).

packages/react-native-gesture-handler/src/components/Pressable/PressableProps.tsx
```diff
@@ -171,6 +171,8 @@ export interface PressableProps
    * Defines the dimensions of the Pressable after it's been resized.
    * This property does not affect Pressable's physical appearance.
    * Required when the Pressable is resized **and** uses pressRetentionOffset.
+   *
+   * @deprecated This property is no longer used. It can be safely removed.
```
m-bert: Do we plan to remove it? If so, let's state it right away as this is fairly new prop and I guess most of users have not adapted it yet. Otherwise we could add entry in `docs`.

m-bert (approved):

packages/react-native-gesture-handler/src/components/Pressable/PressableProps.tsx
latekvo: We do. I think `This property is no longer used.` paints a clear enough picture, but i think it would make sense to remove rest of the prop's description, as the prop is not used for anything anymore.
What do you think?

m-bert: In that case I'd explicitly state that it will be removed - if we don't there's a chance that people will continue to use it.

latekvo: Updated the comment as discussed in-person in [8bd8607](https://github.com/software-mansion/react-native-gesture-handler/pull/3606/commits/8bd860771fc8e4673849298ef35546ba2655dcaa).

packages/react-native-gesture-handler/src/components/Pressable/PressableProps.tsx
```diff
@@ -168,11 +168,7 @@ export interface PressableProps
   blocksExternalGesture?: RelationPropType;
 
   /**
-   * Defines the dimensions of the Pressable after it's been resized.
-   * This property does not affect Pressable's physical appearance.
-   * Required when the Pressable is resized **and** uses pressRetentionOffset.
-   *
-   * @deprecated This property is no longer used. It can be safely removed.
+   * @deprecated This property is no longer used, and will be removed in the future. It can be safely removed.
```
m-bert: I think it is not necessary, but it is not a strong opinion.

```suggestion
   * @deprecated This property is no longer used, and will be removed in the future.
```

m-bert (approved):

packages/react-native-gesture-handler/src/components/Pressable/PressableProps.tsx
latekvo: Yeah alright, done in [e4cd9bc](https://github.com/software-mansion/react-native-gesture-handler/pull/3606/commits/e4cd9bc60a5d38f21953d9fcff9fc7ce5623740f)

Issue #2385, opened 2023-01-20
enabled = false has no effect during tests


Andarius:
### Description

Passing `enable={false}` to a `RectButton` will have no effect when running jest tests. 
My guess is that since `RectButton` is using `TouchableNativeFeedback` (https://github.com/software-mansion/react-native-gesture-handler/blob/f0868f7ccf678c947ef65519ebf97ae149a10289/src/mocks.ts#L34) that uses `disabled` instead, it has no effect.

A potential fix could be:

```ts
import React from 'react';
import { Pressable } from 'react-native';
import type { RectButtonProperties } from 'react-native-gesture-handler';

const RectButtonMock = (props: RectButtonProperties) => {
  const { onPress, enabled, ...rest } = props;
  return (
    <Pressable
      disabled={!enabled}
      onPress={() => onPress && onPress(true)}
      {...rest}
    >
      {props.children}
    </Pressable>
  );
};

jest.mock('react-native-gesture-handler', () => {
  const actual = jest.requireActual('react-native-gesture-handler');
  return {
    ...actual,
    RectButton: RectButtonMock,
  };
});
```

### Steps to reproduce

```ts
describe.only('Testing disabled Button', () => {
  it('onPress does not trigger', function () {
    const onPress = jest.fn();
    const { getByTestId } = render(
      <RectButton testID="btn" onPress={onPress} enabled={false} />
    );
    const btn = getByTestId('btn');

    expect(onPress).not.toHaveBeenCalled();
    fireEvent.press(btn);
    expect(onPress).not.toHaveBeenCalled();   // Will fail
  });
});
```

### Snack or a link to a repository

NA

### Gesture Handler version

2.8.0

### React Native version

0.70.5

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

versten1uk:
Same issue here

m-bert:
Hi! Could you please check if #3062 fixes this?

sidferreira:
@m-bert @versten1uk @Andarius I did some digging and I believe the problem is some weird interaction on how `fireEvent` works.

https://github.com/callstack/react-native-testing-library/blob/3545a94d080687bbf96c0d4426b3e55201a3426f/src/fire-event.ts#L50

To fix my tests I just set `pointerEvents: none`, but it is a sub-optimal solution.

[...]

PR #3612, opened 2025-07-14
[jest] Fixing enabled prop in buttons

akwasniewski:
## Description

Passing `enabled={false}` to buttons in jest tests now has effect. I properly configured the corresponding mocks. Change was motivated by issue #2385. 

## Test plan
Tested using React Native 80.1
```ts
import { render, userEvent } from '@testing-library/react-native';


export const user = userEvent.setup()


describe('Testing disabled Button', () => {
  it('onPress does not trigger', async () => {
    const onPress = jest.fn();
    const { getByTestId } = render(
      <RectButton testID="btn" onPress={onPress} enabled={false} />
    );
    const btn = getByTestId('btn');

    expect(onPress).not.toHaveBeenCalled();
    await user.press(btn);
    expect(onPress).not.toHaveBeenCalled(); 
  });
});
```

### DIFF
--- a/packages/react-native-gesture-handler/src/components/GestureButtonsProps.ts
+++ b/packages/react-native-gesture-handler/src/components/GestureButtonsProps.ts
@@ -58,24 +58,28 @@ export interface RawButtonProps
 
   /**
    * Used for testing-library compatibility, not passed to the native component.
+   * @deprecated test-only props are deprecated and will be removed in the future.
    */
   // eslint-disable-next-line @typescript-eslint/ban-types
   testOnly_onPress?: Function | null;
 
   /**
    * Used for testing-library compatibility, not passed to the native component.
+   * @deprecated test-only props are deprecated and will be removed in the future.
    */
   // eslint-disable-next-line @typescript-eslint/ban-types
   testOnly_onPressIn?: Function | null;
 
   /**
    * Used for testing-library compatibility, not passed to the native component.
+   * @deprecated test-only props are deprecated and will be removed in the future.
    */
   // eslint-disable-next-line @typescript-eslint/ban-types
   testOnly_onPressOut?: Function | null;
 
   /**
    * Used for testing-library compatibility, not passed to the native component.
+   * @deprecated test-only props are deprecated and will be removed in the future.
    */
   // eslint-disable-next-line @typescript-eslint/ban-types
   testOnly_onLongPress?: Function | null;


packages/react-native-gesture-handler/jest.config.js
```diff
@@ -8,5 +8,5 @@ module.exports = {
   modulePathIgnorePatterns: [
     "<rootDir>/lib/"
   ],
-  roots: ["<rootDir>/src/"]
+  roots: ["<rootDir>/src/"],
```
m-bert: Is this added by `prettier` or some other formatter? I don't think it is necessary 😅 

```suggestion
  roots: ["<rootDir>/src/"]
```

packages/react-native-gesture-handler/jestSetup.js
```diff
@@ -1,7 +1,16 @@
 jest.mock('./src/RNGestureHandlerModule', () => require('./src/mocks'));
+jest.mock('./src/components/GestureButtons', () => require('./src/__mocks__/GestureButtons'));
```
m-bert: As I read in [jest docs](https://jestjs.io/docs/manual-mocks#mocking-node-modules), `__mocks__` directory should be used to mock node modules. User modules can be mocked, but should be placed in the same directory ([docs](https://jestjs.io/docs/manual-mocks#mocking-node-modules)).

Could you check if mocking `GestureButtons` like this works? Something tells me that we don't even need  this directory for the current use case, but I wouldn't change that. 

So to summarize, if we need to create separate file with button mocks, then lets either make it in `components/__mocks__`, or create our own `mocks` directory and use them for both, current `mocks.tsx` and `GestureButtons` mock.

packages/react-native-gesture-handler/jestSetup.js
```diff
@@ -1,7 +1,16 @@
...
+
 jest.mock('./lib/commonjs/RNGestureHandlerModule', () =>
   require('./lib/commonjs/mocks')
 );
+jest.mock('./lib/commonjs/components/GestureButtons', () =>
+  require('./lib/commonjs/mocks')
```
m-bert: As far as I can see `bob` doesn't generate `__mocks__` directory when running `yarn build`. That's ok, but does this actually work, or does it work by accident because this file already contains these mocks?

What I want to say - do we need `GestureButtons` mock file?

packages/react-native-gesture-handler/jestSetup.js
```diff
@@ -1,7 +1,16 @@
...
+);
 jest.mock('./lib/module/RNGestureHandlerModule', () =>
   require('./lib/module/mocks')
 );
+jest.mock('./lib/module/components/GestureButtons', () =>
+  require('./lib/module/mocks')
```
m-bert: Same as above

m-bert (changes requested):
Have you checked if `Pressable` works? 😅

packages/react-native-gesture-handler/jestSetup.js
akwasniewski: I checked it, it's not by accident. I managed to find an easier solution, we don't need seperate `GestureButtons` mock file. However, in the jestSetup we still must mock source files separetely to `mocks`

packages/react-native-gesture-handler/src/mocks.tsx
akwasniewski, line 1: Due to how they are typically imported, it's not enough to export the buttons as default

packages/react-native-gesture-handler/jest.config.js
akwasniewski: my bad, `prettier` betrayed me, fixed

packages/react-native-gesture-handler/jestSetup.js
akwasniewski: fixed, I managed to find an easier solution, we don't need seperate `GestureButtons` mock file. However, in the jestSetup we still must mock source files separetely to `mocks`

packages/react-native-gesture-handler/jest.config.js
akwasniewski: Yeah, I forgot about pressable, fixed in the same way as others

packages/react-native-gesture-handler/src/mocks.tsx
```diff
@@ -31,14 +31,15 @@ const LongPressGestureHandler = View;
...
   <TouchableNativeFeedback disabled={!enabled} {...rest}>
     <View />
   </TouchableNativeFeedback>
 );
-const BaseButton = RawButton;
-const RectButton = RawButton;
-const BorderlessButton = TouchableNativeFeedback;
+export const BaseButton = RawButton;
+export const RectButton = RawButton;
+export const Pressable = RawButton;
```
latekvo: The real `Pressable` accepts a `disabled` prop, while the mock version expects `enabled` for the `disabled={!enabled}` toggle to work.

packages/react-native-gesture-handler/src/mocks.tsx
```diff
@@ -31,14 +31,15 @@ const LongPressGestureHandler = View;
 const PinchGestureHandler = View;
 const RotationGestureHandler = View;
 const FlingGestureHandler = View;
-const RawButton = ({ enabled, ...rest }: any) => (
+export const RawButton = ({ enabled, ...rest }: any) => (
   <TouchableNativeFeedback disabled={!enabled} {...rest}>
```
latekvo: This isn't really an issue with this PR, but i just noticed, won't the `RawButton` by default be disabled?
When `enabled` is `undefined`, the `disabled` is set to `true`.
Rather, the `enabled` should be checked against strict `false` - `disabled={enabled === false}`.

image

packages/react-native-gesture-handler/src/mocks.tsx
akwasniewski: fixed, I now mock `GestureHandler`'s `Pressable` using the one from `react-native`

packages/react-native-gesture-handler/src/mocks/Pressable.tsx
```diff
@@ -0,0 +1,3 @@
+import { Pressable as RnPressable } from 'react-native';
+export const Pressable = RnPressable;
+export default Pressable;
```
m-bert: Do we need both exports?

akwasniewski: not really, residue from fixing. I only left the necessary one

packages/react-native-gesture-handler/jestSetup.js
```diff
@@ -1,7 +1,27 @@
...
+  require('./lib/module/mocks/mocks')
+);
+jest.mock('./lib/module/components/GestureButtons', () =>
+  require('./lib/module/mocks/mocks')
+);
+jest.mock('./lib/module/components/Pressable', () =>
+  require('./lib/module/mocks/Pressable')
 );
+
+
```
m-bert: Let's leave only one new line
```suggestion
```

packages/react-native-gesture-handler/src/mocks/Pressable.tsx
```diff
@@ -0,0 +1,3 @@
+import { Pressable as RnPressable } from 'react-native';
+const Pressable = RnPressable;
+export default Pressable;
```
m-bert: Have you checked if this would work?

```ts
export { Pressable } from 'react-native';
```

packages/react-native-gesture-handler/src/mocks.tsx
m-bert: > This isn't really an issue with this PR

It is not, but it would be great to fix it in this one (cc @akwasniewski)

packages/react-native-gesture-handler/src/mocks/Pressable.tsx
m-bert: Oh, that's `default` export, so it probably won't, my bad 😅

akwasniewski: i checked, it won't work. I can maybe make it more compact by doing this.
```ts
import { Pressable as RnPressable } from 'react-native';
export default RnPressable;
```

m-bert: In that case I think you don't really need this alias, couldn't it be just `Pressable`?

akwasniewski: hah sorry, yes, fixed

packages/react-native-gesture-handler/src/mocks.tsx
akwasniewski: Fixed, and tested. Now it works as intended

m-bert:
What about deprecating `testOnly` props form `Pressable`? (cc @akwasniewski, @latekvo)

packages/react-native-gesture-handler/src/mocks/Pressable.tsx
latekvo: You can do:

```tsx
export { Pressable as default } from 'react-native';
```

It seems to work fine.

latekvo:
> What about deprecating testOnly props form Pressable?

I have nothing against it, since these won't be used anymore once the mocks are fixed.

Although it'd be best to notify `testing-library` maintainers of this change, since `testOnly_*` props are there only for tight integration with the `testing-library` and for nothing else.

(c.c. @mdjastrzebski)

packages/react-native-gesture-handler/src/mocks/Pressable.tsx
akwasniewski: even better, i'll amend it, thanks

m-bert:
> Although it'd be best to notify testing-library maintainers of this change

Yes I agree, I wanted to know what do you think about deprecation 😅

PR #3617, opened 2025-07-16
`NativeDetector` <-> `Reanimated` integration 

m-bert:
## Description

This PR introduces integration with `Reanimated` to `NativeDetector`. When `Reanimated` is installed callbacks are handled via `useEvent` hook. Otherwise, they run on JS.

It also creates `v3` directory (name can be changed later) that will be used in further development to separate new code from the old one. 

## Test plan

Tested on basic-example app.


packages/react-native-gesture-handler/src/v3/hooks/useGestureEvent.ts
```diff
@@ -0,0 +1,47 @@
...
+    'worklet';
+
+    if (!isTouchEvent(event)) {
+      return;
+    }
+
+    if (event.eventType !== TouchEventType.UNDETERMINED) {
+      runWorklet(touchEventTypeToCallbackType(event.eventType), config, event);
+    }
+  };
```
j-piasecki: I believe this also needs to be wrapped with `useEvent` & `useHandler`.

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  if (shouldUseReanimated) {
+    const handlers = {
+      onChange: config.onChange,
+      onUpdate: config.onUpdate,
+    };
+
+    const { doDependenciesDiffer } = Reanimated!.useHandler(handlers);
+
+    return Reanimated!.useEvent(
```
j-piasecki: We may want to investigate whether there's a difference (performance-wise) of calling `useEvent` separately for every event vs. calling it once and handling all events there.

packages/react-native-gesture-handler/src/v3/hooks/useGestureEvent.ts
m-bert: You're right!

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: I discussed it with Reanimated team and they said that it shouldn't be a big deal. I think we can leave it as it is for now and later we can check whether it is a big performance issue or not 😅

packages/react-native-gesture-handler/src/v3/interfaces.ts
```diff
@@ -0,0 +1,23 @@
+import {
```
j-piasecki: I find it funny that the file is named `interfaces` and there's not a single interface inside

packages/react-native-gesture-handler/src/v3/interfaces.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export type StateChangeEvent =
+  | GestureStateChangeEvent
+  | EventWithNativeEvent<GestureStateChangeEvent>;
+
+export type TouchEvent =
+  | GestureTouchEvent
+  | EventWithNativeEvent<GestureTouchEvent>;
+
+export type EventWithNativeEvent<T> = {
```
j-piasecki: React Native exposes the type for it: `NativeSyntheticEvent<T>`

packages/react-native-gesture-handler/src/v3/interfaces.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  GestureStateChangeEvent,
+  GestureTouchEvent,
+  GestureUpdateEvent,
```
j-piasecki: Update and change types will be incorrect in this case - there's a `handlerData` field holding the handler-specific info (yay codegen).

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
```diff
@@ -18,13 +15,17 @@ let Reanimated:
...
-        callback: (event: GestureUpdateEvent | GestureStateChangeEvent) => void,
+      useHandler: (handlers: Record<string, unknown>) => {
+        doDependenciesDiffer: boolean;
+      };
+      useEvent: <T>(
+        callback: (event: T) => void,
         events: string[],
         rebuild: boolean
-      ) => unknown;
+      ) => (event: GestureHandlerEvent, context?: Context<unknown>) => void;
```
j-piasecki:
```suggestion
      ) => (event: unknown, context?: Context<unknown>) => void;
```

Wouldn't something like this work?

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
```diff
@@ -18,13 +15,17 @@ let Reanimated:
...
       useSharedValue: <T>(value: T) => SharedValue<T>;
       setGestureState: (handlerTag: number, newState: number) => void;
+      isSharedValue: (value: unknown) => value is SharedValue<unknown>;
```
j-piasecki, line 33: Is this new in Reanimated?

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
j-piasecki: I don't think this is the right Context type

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
```diff
@@ -0,0 +1,55 @@
...
+  RNGestureHandlerDetectorNativeComponent
+);
+
+export function NativeDetector({ gesture, children }: NativeDetectorProps) {
+  const NativeDetectorComponent = gesture.dispatchesAnimatedEvents
+    ? AnimatedNativeDetector
+    : gesture.shouldUseReanimated
+      ? ReanimatedNativeDetector!
+      : RNGestureHandlerDetectorNativeComponent;
+
```
j-piasecki, line 28: Let's throw when `NativeDetectorComponent` is undefined (no Reanimated and shouldUseReanimated is true, which shouldn't happen)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface NativeGesture {
+  tag: number;
+  name: GestureType;
+  config: Record<string, unknown>;
+  gestureEvents: GestureEvents;
+  shouldUseReanimated: boolean;
+  dispatchesAnimatedEvents: boolean;
+}
+
+function hasWorklets(config: Record<string, unknown>) {
```
```suggestion
function hasWorkletEventHandlers(config: Record<string, unknown>) {
```

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+      return CALLBACK_TYPE.TOUCHES_MOVE;
+    case TouchEventType.TOUCHES_UP:
+      return CALLBACK_TYPE.TOUCHES_UP;
+    case TouchEventType.TOUCHES_CANCELLED:
+      return CALLBACK_TYPE.TOUCHES_CANCELLED;
+  }
+  return CALLBACK_TYPE.UNDEFINED;
+}
+
+export function runWorklet(
```
```suggestion
export function runWorkletCallback(
```

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
```diff
@@ -0,0 +1,83 @@
+import { CALLBACK_TYPE } from '../../../handlers/gestures/gesture';
+import { compareTags, runWorklet } from '../utils';
+import { Reanimated } from '../../../handlers/gestures/reanimatedWrapper';
+import { UpdateEvent } from '../../interfaces';
+
+function useReanimatedEvent(handlerTag: number, config: any) {
+  const lastUpdateEvents = Reanimated!.useSharedValue<
```
j-piasecki: This can be replaced with context from `useEvent`

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
```diff
@@ -0,0 +1,83 @@
...
+        config,
+        config.changeEventCalculator?.(
+          event,
+          lastUpdateEvents.value[handlerTag]
+        )
+      );
+
+      lastUpdateEvents.value[handlerTag] = event;
+    }
+  };
```
j-piasecki: Can this be shared between JS/Reanimated handlers?

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export function useGestureHandlerEvent(
+  handlerTag: number,
+  config: any,
+  shouldUseReanimated: boolean
+) {
+  return shouldUseReanimated
+    ? // eslint-disable-next-line react-hooks/rules-of-hooks
+      useReanimatedEvent(handlerTag, config)
+    : gestureHandlerEvent(handlerTag, config);
```
j-piasecki: Can we do this without calling hooks conditionally?

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  // @ts-ignore It works, duh -_-
+  handler?.(event, ...args);
+}
+
+export function compareTags(handlerTag: number, event: GestureHandlerEvent) {
+  'worklet';
+
+  if ('nativeEvent' in event) {
+    return (
+      (event as EventWithNativeEvent<any>).nativeEvent.handlerTag === handlerTag
```
latekvo: `any` can be typed as `GestureTouchEvent`

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureStateChangeEvent.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  };
+
+  if (shouldUseReanimated) {
+    const handlers = {
+      onBegin: config.onBegin,
+      onStart: config.onStart,
+      onEnd: config.onEnd,
+      onFinalize: config.onFinalize,
+    };
```
j-piasecki: Can we use this instead of `config` in the event handler? Feels illegal to pass this to `useHandler` and then use a different object to call methods.

(the same in other files)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  type: CALLBACK_TYPE,
+  config: HandlerCallbacks<Record<string, unknown>>,
+  event: GestureHandlerEvent,
+  ...args: unknown[]
+) {
+  'worklet';
+  const handler = getHandler(type, config);
+
+  // @ts-ignore It works, duh -_-
+  handler?.(event, ...args);
```
latekvo, line 66: This is really weirdly typed. 

The `handler` is never supposed to accept more than `2` arguments, and `runWorklet` is always called with `args` being set to `boolean | undefined`. 

The second arg of `handler` should either be `success` or `GestureStateManagerType` depending on the underlying type of the `handler`.

I'm fine with the `@ts-ignore`, but i think we should at least try typing `args` correctly.

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
m-bert: Approx 2 years 😅 ([PR](https://github.com/software-mansion/react-native-reanimated/pull/5019))

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureStateChangeEvent.ts
```diff
@@ -0,0 +1,74 @@
...
+      runWorklet(CALLBACK_TYPE.FINALIZE, config, event, true);
+    } else if (
+      (state === State.FAILED || state === State.CANCELLED) &&
+      state !== oldState
+    ) {
+      if (oldState === State.ACTIVE) {
+        runWorklet(CALLBACK_TYPE.END, config, event, false);
+      }
+      runWorklet(CALLBACK_TYPE.FINALIZE, config, event, false);
+    }
```
latekvo: This is a subject for another PR, but is there a point in verifying the `oldState` in transitions to states `ACTIVE`, `BEGAN`, `FAILED` or `CANCELLED`? (`END` requires the `oldState === State.ACTIVE` check)

Even in an (impossible?) edge case, where the `CANCELLED` -> `ACTIVE` transition occurs, do we really not want to call the `START` callback? In such scenario, the gesture did activate, it just had an unusual state transition.

That is of course, if we even consider such transitions possible.

(c.c. @j-piasecki)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: This is exactly the logic that we already have. I agree that it could use better typing, but I don't think it is the scope of this PR (cc @j-piasecki)

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: I've split it on purpose since there were a lot of `if` about `Reanimated`. I can bring it back, but I believe this one looked a bit cleaner.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
latekvo: The `// @ts-ignore It works, duh -_-` is a new addition in this PR, which is why i commented on this. Something must have changed

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: As above, I can bring back shared implementation with `if` statements. 

This is very specific conditional call as it depends on whether gesture uses `Reanimated` or not and it shuoldn't change in runtime. We already have conditional calls for `useEvent`. Also our existing code disables this rule in some places.

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
m-bert: `shouldUseReanimated` will not be `true` if `Reanimated` is not installed

latekvo: `shouldUseReanimated` check is based on heuristics which are not directly related to `react-native-reanimated`. I think it'd make sense to have a clear error when the edge case @j-piasecki mention occurs, instead of wondering why some setups are randomly crashing.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Why `GestureTouchEvent` and not for example `GestureStateChangeEvent`?

latekvo: Either will work, all possible arguments containing `nativeEvent` passed to `compareTags` (`EventWithNativeEvent<GestureUpdateEvent | GestureTouchEvent | GestureStateChangeEvent>`) share the `handlerTag` which is used within `compareTags`.

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
m-bert: Interesting suggestion 😄

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureStateChangeEvent.ts
j-piasecki: IIRC, there are edge cases we don't want to send: undetermined -> failed/canceled, canceled -> failed, ...

We should not be handling state changes that are erroneous (like your example). I'd rather throw in asserts, making sure each transition is correct. The state machine will likely be reimplemented in a better way, so we can decide how to handle it when we get to it.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
latekvo: Could use smart casting:

```tsx

function isNativeEvent(
  event: GestureHandlerEvent | EventWithNativeEvent<unknown>
): event is EventWithNativeEvent<unknown> {
  'worklet';

  return 'nativeEvent' in event;
}

export function compareTags(
  handlerTag: number,
  event:
    | GestureHandlerEvent
    | EventWithNativeEvent<
        GestureUpdateEvent | GestureTouchEvent | GestureStateChangeEvent
      >
) {
  'worklet';

  if (isNativeEvent(event)) {
    return event.nativeEvent.handlerTag === handlerTag;
  }

  return event.handlerTag === handlerTag;
}
```

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: It's not really a new addition, the text has changed though 😅

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
m-bert: I think you looked at wrong `shouldUseReanimated` @latekvo 😅

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: Not really from `useEvent`, but from `useHandler`. We looked at it with Reanimated guys, I'll try to make it work

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
latekvo: My bad, i looked at the `gesture.ts` instead of `useGesture.ts` :P

The latter looks fine.

j-piasecki: > shouldUseReanimated will not be true if Reanimated is not installed

Yeah, in its current form, it won't. It's possible that it will change in the future, and adding a dev-only check won't hurt.

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
j-piasecki: > This is very specific conditional call as it depends on whether gesture uses Reanimated or not and it shuoldn't change in runtime.

If by shouldn't you mean it won't - that's only because `runOnJS` isn't implemented yet. If you mean that it shouldn't, then I agree but we should throw with a sensible message then.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Done in [0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243)

packages/react-native-gesture-handler/src/v3/interfaces.ts
m-bert: **_For Entertainment Purposes Only_**

Changed in [0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243) 😄

packages/react-native-gesture-handler/src/v3/interfaces.ts
m-bert: Changed in [0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243)

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
m-bert: Changed in [0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243)

m-bert: Also changed in [0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: Changed in [0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Changed in [0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243)

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: Let me know what you think about current approach ([0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243))

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: Changed in [0da15d7](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0da15d7f469fbbdfe1ecc6d275a36acb67ea4243)

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
m-bert: Added error in [650103c](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/650103ccf297a352302f0edfdbf551abd0c6ea8e)

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: Changed in [ad410e9](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/ad410e9daa95e2f981c9ba7943cc2634d26773e3), let me know what you think.

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureStateChangeEvent.ts
m-bert: Though I discussed it with Reanimated guys and they said that it should work, I changed it in [59e9c18](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/59e9c188aaa2d2bdcbeb50d05b9b86614ecd9ea9). Let me know what you think.

packages/react-native-gesture-handler/src/v3/interfaces.ts
m-bert: Added `handlerData` in [8b6e762](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/8b6e7629fd7886245bb980fd98134404243489f0)

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
```diff
@@ -0,0 +1,60 @@
...
+      {children}
+    </NativeDetectorComponent>
+  );
+}
+
+const styles = StyleSheet.create({
+  detector: {
+    display: 'contents',
+    // TODO: remove, debug info only
+    backgroundColor: 'red',
```
latekvo, line 63:
```suggestion
```

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -0,0 +1,116 @@
...
+  useEffect(() => {
+    return () => {
+      RNGestureHandlerModule.dropGestureHandler(tag);
+      RNGestureHandlerModule.flushOperations();
+    };
+  }, [type, tag]);
+
+  useEffect(() => {
+    // TODO: filter changes - passing functions (and possibly other types)
+    // causes a native crash
```
latekvo, line 100: Do you plan fixing it in this PR or a different one?

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -0,0 +1,116 @@
...
+    config,
+    gestureEvents: {
+      onGestureHandlerStateChange,
+      onGestureHandlerEvent,
+      onGestureHandlerTouchEvent,
+      onGestureHandlerAnimatedEvent,
+    },
+    shouldUseReanimated,
+    dispatchesAnimatedEvents:
+      !!onGestureHandlerAnimatedEvent &&
```
latekvo: Is the bool casting (`!!`) required here?

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: I'd do it in different (cc @j-piasecki)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: Not really (but correct me if I'm wrong @j-piasecki)

Removed in [a548fce](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/a548fce24084e5cc2e82edf6815812b93571627a)

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
m-bert: I'd leave it for now as we are still working on these components (cc @j-piasecki)

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -6,15 +6,26 @@ import {
 } from '../handlers/gestureHandlerCommon';
 import { HandlerCallbacks } from '../handlers/gestures/gesture';
 
-export type GestureHandlerEvent = UpdateEvent | StateChangeEvent | TouchEvent;
+type GestureUpdateEventWithData<T> = GestureUpdateEvent & {
```
j-piasecki: Shouldn't it be `GestureEventPayload`?

```suggestion
type GestureUpdateEventWithData<T> = GestureEventPayload & {
```

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -6,15 +6,26 @@ import {
...
+  handlerData: T;
+};
 
-export type UpdateEvent =
-  | GestureUpdateEvent
-  | NativeSyntheticEvent<GestureUpdateEvent>;
+type GestureStateChangeEventWithData<T> = GestureStateChangeEvent & {
```
j-piasecki: Shouldn't it be `HandlerStateChangeEventPayload`?

```suggestion
type GestureStateChangeEventWithData<T> = HandlerStateChangeEventPayload & {
```

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
j-piasecki: Can the error message be more descriptive? Something along the lines of

```
Gesture expects to run on the UI thread, but failed to create the Reanimated NativeDetector.
```

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
```diff
@@ -1,83 +1,60 @@
...
         CALLBACK_TYPE.CHANGE,
         config,
-        config.changeEventCalculator?.(event, lastUpdateEvents[handlerTag])
+        config.changeEventCalculator?.(event, context.lastUpdateEvent)
       );
 
-      lastUpdateEvents[handlerTag] = event;
+      context.lastUpdateEvent = event;
     }
   };
```
j-piasecki: I think we can merge the two together so the `change...` values are always present. The only issue here I can see is Animated, which would intercept the events on the native side where those fields wouldn't be available.

We should be able to detect that (by reading internals of `Animated.Event`) and throw in that case.

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
j-piasecki: Now that I think about it we can do it both ways - add a new param (`disableReanimated` or whatever, rename `shouldUseReanimated` to `runOnUI`). When `disableReanimated` is false (default) do what it's currently doing, if it's true, return before calling Reanimated hooks. Changing `disableReanimated` after the gesture has been created should throw.

This would keep the default working as it does in v2 but allow to opt-out of calling Reanimated hooks at all.

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
j-piasecki: No strong opinion, but leaving it for now makes sure that we'd notice if the layout logic is wrong.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+      >
+): event is NativeSyntheticEvent<
+  GestureUpdateEvent | GestureStateChangeEvent | GestureTouchEvent
+> {
+  'worklet';
+
+  return 'nativeEvent' in event;
+}
+
+export function compareTags(
```
```suggestion
export function isEventForHandlerWithTag(
```

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
j-piasecki: > This is exactly the logic that we already have. I agree that it could use better typing, but I don't think it is the scope of this PR

I can see the argument, but please add a TODO if you decide not to update it in this PR.

packages/react-native-gesture-handler/src/v3/hooks/useGestureEvent.ts
```diff
@@ -0,0 +1,36 @@
...
+  );
+
+  const onGestureHandlerTouchEvent = useTouchEvent(
+    handlerTag,
+    config,
+    shouldUseReanimated
+  );
+
+  const onGestureHandlerAnimatedEvent =
+    config.onGestureHandlerAnimatedEvent as (...args: any[]) => void;
```
j-piasecki, line 30: In my head, `onGestureHandlerAnimatedEvent` is set automatically when `onUpdate` is set to an `Animated.Event`. Please add a TODO if you don't have any objections.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -0,0 +1,116 @@
...
+  config: Record<string, unknown>;
+  gestureEvents: GestureEvents;
+  shouldUseReanimated: boolean;
+  dispatchesAnimatedEvents: boolean;
+}
+
+function hasWorkletEventHandlers(config: Record<string, unknown>) {
+  return Object.values(config).some(
+    // @ts-ignore `__workletHash` comes from Reanimated and it does exist (on JS thread) if function is worklet.
+    (prop) => typeof prop === 'function' && prop.__workletHash !== undefined
```
latekvo:
```suggestion
    (prop) => typeof prop === 'function' && '__workletHash' in prop
```

This seems to work, not sure if it introduces new edge-cases.

packages/react-native-gesture-handler/src/v3/hooks/events/useTouchEvent.ts
```diff
@@ -0,0 +1,65 @@
...
+        .eventType !== TouchEventType.UNDETERMINED
+    ) {
+      runWorkletCallback(
+        touchEventTypeToCallbackType(
+          (event as NativeSyntheticEvent<GestureTouchEvent>).nativeEvent
+            .eventType
+        ),
+        handlers,
+        event
+      );
```
latekvo:
```suggestion
    if (
      isNativeEvent(event) &&
      event.nativeEvent.eventType !== TouchEventType.UNDETERMINED
    ) {
      runWorkletCallback(
        touchEventTypeToCallbackType(event.nativeEvent.eventType),
        handlers,
        event
      );
```

We can remove some of these casts and the `@ts-ignore` if we use `isNativeEvent`.

latekvo: If you don't want to make a function call, which is understandable, `'nativeEvent' in event` seems to work perfectly fine too.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
latekvo: Note that `__workletHash` is typed as a `number`, so it shouldn't be possible for `__workletHash` to be `undefined`

packages/react-native-gesture-handler/src/v3/interfaces.ts
latekvo: [8b6e762](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/8b6e7629fd7886245bb980fd98134404243489f0) adds `Record<string, unknown>` into a lot of places. This is a nitpick, but most of these `Record<string, unknown>` could be replaced with just `unknown`.

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureStateChangeEvent.ts
```diff
@@ -18,9 +18,9 @@ export function useGestureStateChangeEvent(
     onFinalize: config.onFinalize,
   };
 
-  const onGestureHandlerStateChange = (event: StateChangeEvent) => {
-    'worklet';
-
+  const onGestureHandlerStateChange = (
```
latekvo, line 23: As per the [docs](https://docs.swmansion.com/react-native-reanimated/docs/advanced/useEvent), `onGestureHandlerStateChange` being a callback called by the `Reanimated.useEvent` has to be a worklet.

latekvo (changes requested):

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
```diff
@@ -18,13 +19,18 @@ let Reanimated:
...
+      useHandler: (handlers: Record<string, unknown>) => {
+        doDependenciesDiffer: boolean;
+        context: ReanimatedContext;
+      };
+      useEvent: <T>(
+        callback: (event: T) => void,
         events: string[],
         rebuild: boolean
-      ) => unknown;
+      ) => (event: unknown) => void;
```
latekvo, line 30: I think this should be typed as:

```suggestion
      ) => EventHandlerProcessed<Event, Record<string, unknown>>;
```

With:

```tsx
type EventHandlerProcessed<
  Event extends object,
  Context extends Record<string, unknown> = never,
> = (event: Event, context?: Context) => void;
```

But I'm not sure if this is all that important.

latekvo: More details: [link](https://github.com/software-mansion/react-native-reanimated/blob/bdf9f39ee1d63e784cd7b67911ee2c1442b65d04/packages/react-native-reanimated/src/hook/useEvent.ts#L47)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: > so it shouldn't be possible for __workletHash to be undefined

Well, if function is not marked as `worklet` it won't have `__workletHash` 🤷 

Also, I know it is not our use-case, but it might be worth mentioning that these 2 [are not the same](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#description). Though I'm more used to `!== undefined`, I can see that this one looks better.

latekvo: >it might be worth mentioning that that these 2 [are not the same](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#description)

I know, that's why i added the "Note", to assure you that in this specific use-case, both `in` and `!== undef` will behave the same.

>Well, if function is not marked as worklet it won't have __workletHash 🤷

That's not oblivious at the first glance. Before I checked Reanimated source code, I considered a case where Reanimated would be setting `__workletHash` to `undefined` under some circumstances, which would make usage of `in` invalid.

latekvo: This is the exact case I was referring to in the "Note":
image

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -0,0 +1,116 @@
...
+    !!config.onTouchesDown ||
+    !!config.onTouchesMove ||
+    !!config.onTouchesUp ||
+    !!config.onTouchesCancelled
+  );
+}
+
+export function useGesture(
+  type: GestureType,
+  config: Record<string, unknown>
```
latekvo, line 52: Could we add a `TODO` to type `config` properly in the future?

As a side note, all instances of `useGesture*()` besides `useGesture()` could have their `config` argument typed as `Partial<CallbackHandlers & GestureEvents>`, but I don't think it's worth adding a temporary solution like that.

packages/react-native-gesture-handler/src/v3/interfaces.ts
m-bert: We discussed it before and decided to use `Record`

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
m-bert: Yes, I know these docs but I believe it doesn't make much sense to copy exact types from Reanimated, especially that `context` is wrong and is not internally passed into this handler (cc @j-piasecki)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: I think it is quite obvious and doesn't need `TODO`, we will take care of that later when working on stable API. I can do it if you insist, but I don't think it is necessary in that case.

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: `GestureUpdateEvent` is defined as follows: 

```tsx
export type GestureUpdateEvent<GestureEventPayloadT = Record<string, unknown>> =
  GestureEventPayload & GestureEventPayloadT;
```

so it is almost the same. Also current implementation uses `GestureUpdateEvent` instead of `Payload` (e.g. [here](https://github.com/software-mansion/react-native-gesture-handler/blob/2d19b817ea60e822a73161a29f5e8f26dd3014ae/packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/useAnimatedGesture.ts#L81)). I can change that if you want, but I guess it was there on purpose (though I agree that it doesn't have to be here).

I think that we will also have to improve types in the future, so maybe that will be good moment to look at it 😅

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: Same as above

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Added in [e9df982](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/e9df9823cf4eccdae4408b37ea99679faf090751)

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: Done in [0debc45](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/0debc45fbd2b6a0f48c78201fee96c334e928710), let me know what you think.

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: Done in [386641f](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/386641f4c792ce66aa167957d4b14dec868ed307), although it will require follow-up. Let me know what you think.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Done in [8573979](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/857397946a0648396bee611cc6ce7471e68619b7)

packages/react-native-gesture-handler/src/v3/hooks/useGestureEvent.ts
m-bert: Done in [e9df982](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/e9df9823cf4eccdae4408b37ea99679faf090751)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: Done in [f9b8e1c](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/f9b8e1c198cdebfabfbbdef702f1de3876dcc9c5)

packages/react-native-gesture-handler/src/v3/hooks/events/useTouchEvent.ts
m-bert: Changed in [da513f8](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/da513f8339633e3cd4289d5a58b3ca6c080a534c)

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureStateChangeEvent.ts
m-bert: Yep, I've caught that during testing, must have changed that during one of the "refactors". Fixed in [700d14d](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/700d14d70a1d25ef3a723688d5522ac9857cea50)

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
m-bert: Sure, changed in [4f90c6a](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/4f90c6abbdbec9ba6a8af953ffb0a1e345f27962)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -0,0 +1,125 @@
...
+  name: GestureType;
+  config: Record<string, unknown>;
+  gestureEvents: GestureEvents;
+  shouldUseReanimated: boolean;
+  dispatchesAnimatedEvents: boolean;
+}
+
+function hasWorkletEventHandlers(config: Record<string, unknown>) {
+  return Object.values(config).some(
+    // @ts-ignore `__workletHash` comes from Reanimated and it does exist (on JS thread) if function is worklet.
```
latekvo:
```suggestion
```

You can remove the `@ts-ignore` since you've removed the need for it in [f9b8e1c](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/f9b8e1c198cdebfabfbbdef702f1de3876dcc9c5).

Or you can convert it to a regular comment if you'd prefer that.

m-bert: You're right, done in [d4c5fd2](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/d4c5fd2ee5de47a470eba820bccb6d2712a38082)

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
j-piasecki: Looks good. One thing I'd maybe change is renaming `disableReanimated` to `unsafeDisableReanimated`. WDYT?

packages/react-native-gesture-handler/src/v3/types.ts
j-piasecki: You're making an intersection with `Record<string, unknown>`, which makes things like `event.randomFieldThatDoesntExist` not a TS error.

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
```diff
@@ -27,24 +27,22 @@ export function useGestureHandlerEvent(
...
-        CALLBACK_TYPE.CHANGE,
-        handlers,
-        // @ts-ignore This type comes from old API and should be fixed in the future
-        handlers.changeEventCalculator?.(event, context.lastUpdateEvent)
-      );
+    runWorkletCallback(
+      CALLBACK_TYPE.UPDATE,
+      handlers,
+      config.changeEventCalculator
+        ? config.changeEventCalculator?.(event, context.lastUpdateEvent)
```
j-piasecki:
```suggestion
        ? config.changeEventCalculator(event, context.lastUpdateEvent)
```

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: I have mixed feelings about it. Right now it seems like good option if you want to opt out from using Reanimated. Adding `unsafe` prefix makes it look like it will blow all your code 💥 

Jokes aside, I think that adding `unsafe` may make it look more like experimental feature, or something that probably shouldn't be used, while in reality it shouldn't be harmful unless you use it wrong.

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: Yep, that's true. I meant that this one is used across our codebase instead of `GestureEventPayload`, that's why I asked.

m-bert: But it could be that they have parameter, so it would make sense in these cases.

m-bert: Changed in [665cb83](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/665cb83b711773e3ea6ff411b341743f7ccf4359)

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: Changed in [665cb83](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/665cb83b711773e3ea6ff411b341743f7ccf4359)

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: Done in [81c1714](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/81c17149c593af77effee0b6ee63fef9d9d23f7b)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -69,10 +69,10 @@ export function isNativeEvent(
   event:
     | GestureHandlerEvent<Record<string, unknown>>
     | NativeSyntheticEvent<
-        GestureUpdateEvent | GestureStateChangeEvent | GestureTouchEvent
+        GestureEventPayload | HandlerStateChangeEventPayload | GestureTouchEvent
       >
```
j-piasecki: Isnt this basically `GestureHandlerEvent<T>`?

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -69,10 +69,10 @@ export function isNativeEvent(
...
 ): event is NativeSyntheticEvent<
-  GestureUpdateEvent | GestureStateChangeEvent | GestureTouchEvent
+  GestureEventPayload | HandlerStateChangeEventPayload | GestureTouchEvent
```
j-piasecki: I think we can use the new types here

```suggestion
  GestureUpdateEventWithData<unknown> | GestureStateChangeEventWithData<unknown> | GestureTouchEvent
```

m-bert: Done in [5b11259](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/5b112595c293846fcfd84baac360b51f3b888bb0)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Also done in [5b11259](https://github.com/software-mansion/react-native-gesture-handler/pull/3617/commits/5b112595c293846fcfd84baac360b51f3b888bb0)

PR #3619, opened 2025-07-17
Dropping handlers when components are dropped

### LINKED ISSUE #3608: [Pressable] Gesture is not cleaned up if component becomes unmounted while the gesture is active on Web
### Description

So far I've only observed this issue on Web. If a Pressable gets unmounted immediately after its `buttonGesture` enters active state, then it becomes stuck and continues cancelling other gesture handlers even though its DOM node doesn't exist anymore.

### Steps to reproduce

1. Render the example component from the gist
2. Click the TextInput to focus it
3. Press Enter
4. Try pressing the newly appeared button and observe no logs in the console

https://github.com/user-attachments/assets/e5584e59-f890-4a4d-894f-cd0e9317b707

https://gist.github.com/gigobyte/70903354799073ba0f9f657d0ffc3f48

### Gesture Handler version

2.27.1

### React Native version

0.79.0

### Platforms

Web

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

akwasniewski:
## Description

Fixes #3608. On web, when gesture was active and its component had been dropped, the gesture was not dropped. This caused conflicts with other handlers, and prevented registering other gestures. 
I added a method to drop handlers in the `GestureOchestrator` and call it when component is dropped.

-->

## Test plan
```ts
import React from 'react';
import { Pressable, TextInput } from 'react-native-gesture-handler';
  const [shown, setShown] = useState(true)

  if (!shown) {
    return (
      <Pressable
        key="1"
        testID="other-pressable"
        style={{ width: 30, height: 30, backgroundColor: 'red' }}
        onPress={() => console.log('pressed')}
      />
    )
  }

  return (
    <Pressable key="2" testID="bad-pressable" onPress={() => { }}>
      <TextInput
        style={{ backgroundColor: 'green', width: 100, height: 30 }}
        onSubmitEditing={() => setShown(false)}
      />
    </Pressable>
  )
};
```

### DIFF
--- a/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
@@ -727,9 +727,7 @@ export default abstract class GestureHandler implements IGestureHandler {
   protected resetConfig(): void {}
 
   public onDestroy(): void {
-    GestureHandlerOrchestrator.instance.removeHandlerFromOrchestratorByTag(
-      this.handlerTag
-    );
+    GestureHandlerOrchestrator.instance.removeHandlerFromOrchestrator(this);
     this.delegate.destroy(this.config);
   }
 
--- a/packages/react-native-gesture-handler/src/web/tools/GestureHandlerOrchestrator.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/GestureHandlerOrchestrator.ts
@@ -45,16 +45,6 @@ export default class GestureHandlerOrchestrator {
     }
   }
 
-  // Called in handler's onDestroy
-  public removeHandlerFromOrchestratorByTag(handlerTag: number): void {
-    const handler = this.gestureHandlers.find(
-      (item) => item.handlerTag === handlerTag
-    );
-    if (handler !== undefined) {
-      this.removeHandlerFromOrchestrator(handler);
-    }
-  }
-
   private cleanupFinishedHandlers(): void {
     const handlersToRemove = new Set<IGestureHandler>();
 
--- a/packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
@@ -1,6 +1,7 @@
 import { ValueOf } from '../../typeUtils';
 import { Gestures } from '../Gestures';
 import type IGestureHandler from '../handlers/IGestureHandler';
+
 // eslint-disable-next-line @typescript-eslint/no-extraneous-class
 export default abstract class NodeManager {
   private static gestures: Record<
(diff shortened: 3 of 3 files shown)


packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
```diff
@@ -35,9 +35,8 @@ export default abstract class NodeManager {
     if (!(handlerTag in this.gestures)) {
       return;
     }
-
+    GestureHandlerOrchestrator.instance.deleteGesture(handlerTag);
```
m-bert: I don't think it is the best place to call this, wouldn't it work if we do this in `onDestroy` method?

packages/react-native-gesture-handler/src/web/tools/GestureHandlerOrchestrator.ts
```diff
@@ -382,6 +382,13 @@ export default class GestureHandlerOrchestrator {
     });
   }
 
+  // Manually deleting gesture handler from the array, called in NodeManager whenever component has been dropped
+  public deleteGesture(handlerTag: number): void {
+    this.gestureHandlers = this.gestureHandlers.filter((item) => {
+      return item.handlerTag !== handlerTag;
+    });
+  }
+
```
m-bert: As far as I can see there's already [removeHandlerFromOrchestrator](https://github.com/software-mansion/react-native-gesture-handler/blob/b70def41e64706f20f317a897b6f9d7efd127998/packages/react-native-gesture-handler/src/web/tools/GestureHandlerOrchestrator.ts#L34) function, can't we use that?

m-bert (changes requested):

packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
akwasniewski: good idea, fixed

packages/react-native-gesture-handler/src/web/tools/GestureHandlerOrchestrator.ts
akwasniewski: we cannot call it directly as we only now the handlerTag. Though, it may be better to call the existing function indirectly, I'll fix it.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -727,6 +727,9 @@ export default abstract class GestureHandler implements IGestureHandler {
   protected resetConfig(): void {}
 
   public onDestroy(): void {
+    GestureHandlerOrchestrator.instance.removeHandlerFromOrchestratorByTag(
+      this.handlerTag
+    );
```
m-bert: Wouldn't

```jsx
    GestureHandlerOrchestrator.instance.removeHandlerFromOrchestrator(this);
```

be enough?

akwasniewski: it would, i didn't notice that IGestureHandler was an interface, now i know where the "I" in the name came from, fixed

m-bert (approved):

Issue #3622, opened 2025-07-18
Submiting a `TextInput` inside pressable on web cancels all subsequent gestures


akwasniewski:
### Description

On web, when we submit `TextInput` using enter, which is inside `Pressable`, then Gesture Handler doesn't recognize any subsqeuent taps or gestures. `TextInput` tap gesture is kept in active state, and causes or other gestures to fail.

### Steps to reproduce

1. Add a `TextInput` inside of a `Pressable` or `Button`.
2. Submit the TextInput, using enter
3. All subsequent gestures fail

### A link to a [Gist](https://gist.github.com/), an [Expo Snack](https://snack.expo.io/) or a link to a repository based on [this template](https://github.com/react-native-community/reproducer-react-native) that reproduces the bug.

https://gist.github.com/akwasniewski/78b971da9c04631abe40f3f398024576

### Gesture Handler version

2.27

### React Native version

0.79

### Platforms

Web

### JavaScript runtime

None

### Workflow

Using Expo Prebuild or an Expo development build

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

Real device

### Device model

Web, Chrome

### Acknowledgements

Yes

gigobyte:
Hey! Would it be possible to publish the fix for this issue, it's sitting in `master`?

m-bert:
Hi @gigobyte! While I can't give you ETA for the release, it will be coming soon since we want to release support for 0.81 😅

PR #3623, opened 2025-07-21
[Web] fixed texinput submition when inside pressable

akwasniewski:
## Description

Fixes issue #3622. On web, when we submitted `TextInput` using enter, which was inside `Pressable`, then Gesture Handler didn't recognize any subsqeuent taps or gestures. Tap gesture was kept in active state, and caused all other gestures to fail.

## Test plan
```ts
import React from 'react';
import { Pressable, TextInput } from 'react-native-gesture-handler';
  const [shown, setShown] = useState(true)
  return (
    <View>
      <Pressable key="2" style={{ backgroundColor: 'red', width: 100, height: 60 }} testID="bad-pressable" onPress={() => { console.log("pressed") }}>
        {shown ? <TextInput
          style={{ backgroundColor: 'green', width: 100, height: 60 }}
          onSubmitEditing={() => { console.log("submit"); setShown(false); }}
        /> : <View ></View>}
      </Pressable >
    </View >
  )
};
```

### DIFF
--- a/packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
@@ -82,7 +82,8 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
       KeyboardEventManager.registeredStaticListeners = true;
       document.addEventListener(
         'keyup',
-        KeyboardEventManager.keyUpStaticCallback
+        KeyboardEventManager.keyUpStaticCallback,
+        { capture: true }
       );
     }
   }
@@ -95,7 +96,8 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
     if (KeyboardEventManager.instances.size === 0) {
       document.removeEventListener(
         'keyup',
-        KeyboardEventManager.keyUpStaticCallback
+        KeyboardEventManager.keyUpStaticCallback,
+        { capture: true }
       );
       KeyboardEventManager.registeredStaticListeners = false;
     }


packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -6,6 +6,14 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
   private activationKeys = ['Enter', ' '];
   private cancelationKeys = ['Tab'];
   private isPressed = false;
+  private static registeredStaticListeners = false;
```
m-bert, line 9: I think you forgot to change it to `true` in callback

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -53,12 +61,15 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
 
   public registerListeners(): void {
     this.view.addEventListener('keydown', this.keyDownCallback);
-    this.view.addEventListener('keyup', this.keyUpCallback);
+    if (!KeyboardEventManager.registeredStaticListeners) {
+      document.addEventListener('keyup', KeyboardEventManager.keyUpCallback);
+    }
+    KeyboardEventManager.instances.add(this);
```
m-bert: Could we make it so that we still have listener on `view` and one global listener that will trigger callback only if the original one was not called? As far as I know callback on `document` will be called after callback on `view`, but let me know if that is not the case. I think it would be better solution (cc @latekvo)

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -53,12 +61,15 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
...
   }
 
   public unregisterListeners(): void {
     this.view.removeEventListener('keydown', this.keyDownCallback);
-    this.view.removeEventListener('keyup', this.keyUpCallback);
+    KeyboardEventManager.instances.delete(this);
```
m-bert, line 94: Seems like we will have small memleak since we do not clear global listener 😅

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
akwasniewski: fixed

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
akwasniewski: i figured it will never really be cleared as long as any object exist -> thus probably never, therefore there is no point in clearing it. But i agree, it may be better to clear it anyway, it should be fixed now

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
akwasniewski: unfortunetly, global callback is registered first.

akwasniewski: fixed, I'm stopping event propagation when handled locally

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -54,11 +67,27 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
...
     this.view.addEventListener('keydown', this.keyDownCallback);
     this.view.addEventListener('keyup', this.keyUpCallback);
+    KeyboardEventManager.instances.add(this);
+    if (!KeyboardEventManager.registeredStaticListeners) {
+      KeyboardEventManager.registeredStaticListeners = true;
+      document.addEventListener(
+        'keyup',
+        KeyboardEventManager.keyUpStaticCallback
+      );
+    }
```
m-bert:
```suggestion

    KeyboardEventManager.instances.add(this);
    
    if (!KeyboardEventManager.registeredStaticListeners) {
      KeyboardEventManager.registeredStaticListeners = true;
      document.addEventListener(
        'keyup',
        KeyboardEventManager.keyUpStaticCallback
      );
    }
```

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -6,6 +6,19 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
...
+  private static instances: Set<KeyboardEventManager> = new Set();
+
+  private static keyUpStaticCallback = (event: KeyboardEvent): void => {
+    this.instances.forEach((item) => {
+      item.keyUp(event);
+    });
+  };
+
+  private keyUpCallback = (event: KeyboardEvent): void => {
+    event.stopPropagation();
```
m-bert: I'm not sure if we want to call `stopPropagation` as it may interfere with other libraries.

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -54,11 +67,27 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
...
     this.view.removeEventListener('keydown', this.keyDownCallback);
     this.view.removeEventListener('keyup', this.keyUpCallback);
+    KeyboardEventManager.instances.delete(this);
+    if (KeyboardEventManager.instances.size === 0) {
+      document.removeEventListener(
+        'keyup',
+        KeyboardEventManager.keyUpStaticCallback
+      );
+      KeyboardEventManager.registeredStaticListeners = false;
+    }
```
m-bert:
```suggestion

    KeyboardEventManager.instances.delete(this);
    
    if (KeyboardEventManager.instances.size === 0) {
      document.removeEventListener(
        'keyup',
        KeyboardEventManager.keyUpStaticCallback
      );
      KeyboardEventManager.registeredStaticListeners = false;
    }
```

akwasniewski: fixed, in both register and unregister

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
akwasniewski: I fixed it in less clean, but probably less problematic way

m-bert: I've looked into our existing code and as far as I can see this is already handled by `isPressed` flag. If manager receives `keyUp` event, it sets `isPressed` to `false`. Then in `keyUp` we have this:

```tsx
    if (this.activationKeys.indexOf(event.key) === -1 || !this.isPressed) {
      return;
    }
```

so it should handle the case of double up callback. Could you verify it?

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -20,7 +40,7 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
     this.dispatchEvent(event, EventTypes.DOWN);
   };
 
-  private keyUpCallback = (event: KeyboardEvent): void => {
+  private keyUp = (event: KeyboardEvent): void => {
```
m-bert: Could we use something like `onKeyUp`?

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
m-bert: And if so, please add a comment about it 😅

akwasniewski: Yeah, you are right, keyUp will return immediately for all event managers, apart from the pressed one. But, without the timestamp check the static handler will loop through all eventManagers and call their keyUp anyway, it may be ineffective to do it when it is not necessary.

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
akwasniewski: ok, fixed

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -54,11 +74,31 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
   public registerListeners(): void {
     this.view.addEventListener('keydown', this.keyDownCallback);
     this.view.addEventListener('keyup', this.keyUpCallback);
```
j-piasecki: I was confused at first about why we need both: a document-level listener and an element-level listener. If that's to avoid missing events in case they get stop-propagated while bubbling, can you add a comment about that?

And if it's not that, then why 😅?

m-bert: IMO it will be better to handle events on the handler-specific view and use global callback in case if that fails. Behavior might be the same in the end, but it looks better. Do you think that only global listener fits better?

m-bert: Also cc @latekvo

j-piasecki: > IMO it will be better to handle events on the handler-specific view and use global callback in case if that fails.

Isn't that what's happening currently? The view should receive the event before the window (bubbling up), and if the handler is invoked, then the global handler will ignore it due to the matching timestamps. If the listener on the view isn't called, then the event will be handled by the global listener.

One thing that could cause problems is two views where the first one handles the event by itself (updating the latest timestamp) and the other doesn't. Then the global handler will ignore the event, and the second view won't receive the up event. This is easily fixable by storing a separate timestamp for each handler.

Am I missing something?

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -6,6 +6,26 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
   private activationKeys = ['Enter', ' '];
   private cancelationKeys = ['Tab'];
   private isPressed = false;
+  private static registeredStaticListeners = false;
+  private static instances: Set<KeyboardEventManager> = new Set();
+  private static lastKeyUpTimeStamp: number;
+
+  private static keyUpStaticCallback = (event: KeyboardEvent): void => {
+    if (event.key !== 'Enter' || this.lastKeyUpTimeStamp === event.timeStamp) {
```
latekvo: Please check if `key` is in `activationKeys`, instead of directly checking for `Enter`.

latekvo (changes requested):

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
akwasniewski: I don't see how how the issue we fix could result from a space, but I agree - it may be better to check anyway. Fixed.

latekvo: My bad, you're right.

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
akwasniewski: The `timestamp` was only really meant to prevent looping through all handlers when not necessary - when the event had been handled locally. Even without timestamps it won't dispatch two events for each handler, as the `isPressed` flag is set to false while handling the first one, therefore the second time `onKeyUp` executes it will return immediately. 
Thus, I don't think there is a point in storing separate timestamps for each handler. 


> One thing that could cause problems is two views where the first one handles the event by itself (updating the latest timestamp) and the other doesn't. 

This means that it is not safe to use a static timestamp. 

If we still have to loop everytime, is there a point in having a separate element-level listener? @m-bert

m-bert: >One thing that could cause problems is two views where the first one handles the event by itself (updating the latest timestamp) and the other doesn't. 

As far as I know `events` are read-only so you should not be able to modify `timeStamp`. Though I'm not sure if it doesn't happen under the hood when you handle events (but based on tests I don't think so)

m-bert: > is there a point in having a separate element-level listener? @m-bert

Both options seem ok 🤷 For me it looks a bit cleaner (or let's say, easier to understand at first glance) that events are handled by local handlers, but I can see the point why we don't want to double this behavior.

I'll leave that up to you to decide (unless @j-piasecki has strong opinion on this)

akwasniewski: Ok, I don't think it's necessary to double the behaviour, I updated it to only use global listener and added a justification comment

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -3,25 +3,49 @@ import EventManager from './EventManager';
...
   private isPressed = false;
+  private static registeredStaticListeners = false;
+  private static instances: Set<KeyboardEventManager> = new Set();
+
+  private static keyUpStaticCallback = (event: KeyboardEvent): void => {
+    /*
+     We need a global listener, as in some cases, keyUp event gets stop-propagated.
+     Then, if we used only component-level listeners the gesture would never end,
+     causing other gestues to fail.
+     */
```
m-bert: Just one more thing, can we stick to one convention of writing comments? 😅 We tend to use `//` instead of multiline comments (e.g. [here](https://github.com/software-mansion/react-native-gesture-handler/blob/2d19b817ea60e822a73161a29f5e8f26dd3014ae/packages/react-native-gesture-handler/src/web/tools/PointerEventManager.ts#L88))

m-bert (approved):
Please take a look at the comment below before merging 😅

packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts
```diff
@@ -53,12 +77,30 @@ export default class KeyboardEventManager extends EventManager<HTMLElement> {
 
   public registerListeners(): void {
     this.view.addEventListener('keydown', this.keyDownCallback);
-    this.view.addEventListener('keyup', this.keyUpCallback);
+
+    KeyboardEventManager.instances.add(this);
```
j-piasecki, line 79: Isn't the global-only approach susceptible to events not bubbling due to some in-between element calling `stopPropagation` (https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Event_bubbling)?

latekvo: Yes, I was looking into that just now, and it seems like that's the case.

If the propagation is stopped anywhere between the `Pressable` or the text input and the `document`, the `key*` events will not be received.

<details>
<summary>
Collapsed test code
</summary>


```tsx

import React, { useLayoutEffect, useRef, useState } from 'react';
import { Pressable, TextInput } from 'react-native-gesture-handler';
  const [shown, setShown] = useState(true);
  const view = useRef<View>(null);

  useLayoutEffect(() => {
    console.log(view);
    view.current.addEventListener('keyup', (e) => {
      console.log('foo');
      e.stopPropagation();
    });
    view.current.addEventListener('keydown', (e) => {
      console.log('bar');
      e.stopPropagation();
    });
  }, []);

  return (
    <View ref={view}>
      <Pressable
        key="2"
        style={{ backgroundColor: 'red', width: 100, height: 120 }}
        testID="bad-pressable"
        onPress={() => {
          console.log('pressed');
        }}>
        {shown ? (
          <TextInput
            style={{ backgroundColor: 'green', width: 100, height: 60 }}
            onSubmitEditing={() => {
              console.log('submit');
              setShown(false);
            }}
          />
        ) : (
          <View></View>
        )}
      </Pressable>
    </View>
  );
}

```

</details>

latekvo: Switching the global listener to [`capture`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#capture) mode (`{ capture: true }`) fixes the issue with `stopPropagation` intercepting our events, but I have no idea if this is a good solution, or if it causes some new edge-cases.
(c.c. @m-bert )

m-bert: If you switch to `capture` it should work fine as it is `document`-level listener, so it should be handled before other event handlers. Not sure how it will interact with other listeners on the same level though (could break if someone calls [stopImmediatePropagation](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation)), but I think it will be very rare edge-case if something breaks in that setup.

akwasniewski: If we change key up listener in @latekvo's example to have `stopImmediatePropagation` it does indeed break it. Do you think it justifies ditching the `capture: true`?

akwasniewski: @latekvo's example works even with `stopImmediatePropagation`. Though it may be that it is only because our listener is mounted first.

m-bert: > Though it may be that it is only because our listener is mounted first. 

Maybe, the order doesn't have to be the same in every case though.

latekvo: The capture phase occurs first (before the bubbling phase), so if anyone really placed a `stopImmediatePropagation` on the `document` during the capture phase, then they are purposefully blocking every single event coming through, and there isn't much we can do about that.

In all other cases `stopImmediatePropagation` should not be a problem.

Perhaps the only other issue could be that someone **wants** to block these events, but is unable to, due to us being the first in line for receiving these events.

m-bert: I meant that order of adding listeners might not be the same, so we don't have to be first in each case.

akwasniewski: Ok, `capture: true` should be enought to fix it, I added it. What do you think @j-piasecki?

PR #3626, opened 2025-07-21
[Web] fixed repeated listener registration

akwasniewski:
## Description

On web, In `EventManagers`, the `registerListeners` was called twice for each handler of a `Button`

## Test plan

* Add a console.log to both `registerListners()` and `unregisterListeners()` in `web/tools/KeyboardEventManager` (or any other event manager)
* Observe logs in the `Web styles reset` example in the example app

### DIFF
--- a/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
@@ -570,6 +570,7 @@ export default abstract class GestureHandler implements IGestureHandler {
     if (this.enabled !== enabled) {
       this.delegate.onEnabledChange(enabled);
     }
+
     this.enabled = enabled;
 
     if (this.config.shouldCancelWhenOutside !== undefined) {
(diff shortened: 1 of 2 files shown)


packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -566,9 +566,11 @@ export default abstract class GestureHandler implements IGestureHandler {
 
   public updateGestureConfig({ enabled = true, ...props }: Config): void {
     this._config = { enabled: enabled, ...props };
-    this.enabled = enabled;
 
-    this.delegate.onEnabledChange(enabled);
+    if (this.enabled !== enabled) {
+      this.delegate.onEnabledChange(enabled);
+    }
+    this.enabled = enabled;
```
m-bert:
```suggestion
    if (this.enabled !== enabled) {
      this.delegate.onEnabledChange(enabled);
    }
    
    this.enabled = enabled;
```

akwasniewski: fixed

PR #3630, opened 2025-07-23
Handle `Animated.Event` in `onUpdate` callback

m-bert:
## Description

This PR automatically assigns `onUpdate` callback to `onGestureHandlerAnimatedEvent` when `Animated.Event` is passed. It also checks whether `change*` fields were used in `Animated.Event` - if so, we throw an error.

## Test plan

Since basic example is already "under construction", I'll copy whole code. Forgive me for I have sinned 🙏 

Basically it covers the case where we switch between `Animated.Event` and standard callback, along with using `change*` in `Animated.Event`.

<details>
<summary> Test code </summary>

```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,

  const [shouldUseAnimated, setShouldUseAnimated] = React.useState(true);

  const animatedEvent = Animated.event(
    [
      { nativeEvent: { handlerData: { translationX: value } } },
      // { nativeEvent: { handlerData: { changeX: 10 } } },
    ],
    {
      useNativeDriver: true,
    }
  );
  const jsCallback = (e: any) => console.log(e.nativeEvent);

    // onGestureHandlerAnimatedEvent: event,
    // onGestureHandlerEvent: (e: any) =>
    //   console.log('onGestureHandlerEvent', e.nativeEvent),
    onUpdate: shouldUseAnimated ? animatedEvent : jsCallback,
    onEnd: (_e: any) => {
      setShouldUseAnimated((prev) => !prev);
    },

    changeEventCalculator: (event: any, _lastUpdateEvent: any) => {
      return { ...event.nativeEvent, changeX: 10 };
    },
  });

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}

```

</details>


packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -87,3 +89,25 @@ export function isEventForHandlerWithTag(
...
+
+export function checkMappingForChangeProperties(
+  obj: NativeSyntheticEvent<GestureUpdateEventWithData<object>>
+) {
+  const payload = obj.nativeEvent.handlerData;
+
+  for (const key in payload) {
+    if (key.startsWith('change')) {
+      throw new Error(
+        tagMessage(`change properties are not supported in Animated.Event.`)
```
```suggestion
        tagMessage(`${key} is not available when using Animated.Event.`)
```

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -87,3 +89,25 @@ export function isEventForHandlerWithTag(
...
+): callback is AnimatedEvent {
+  'worklet';
+
+  return '_argMapping' in callback;
+}
+
+export function checkMappingForChangeProperties(
+  obj: NativeSyntheticEvent<GestureUpdateEventWithData<object>>
+) {
+  const payload = obj.nativeEvent.handlerData;
```
```suggestion
  const payload = obj?.nativeEvent?.handlerData;
  
  if (!payload) {
    return;
  }
```

Let Animated throw when wrong mapping is used

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -87,3 +89,25 @@ export function isEventForHandlerWithTag(
...
+export function isAnimatedEvent(
+  callback: ((event: any) => void) | AnimatedEvent
+): callback is AnimatedEvent {
+  'worklet';
+
+  return '_argMapping' in callback;
+}
+
+export function checkMappingForChangeProperties(
+  obj: NativeSyntheticEvent<GestureUpdateEventWithData<object>>
```
j-piasecki: Please type it differently. I was really confused for a moment about why we're checking the event there.

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
```diff
@@ -50,8 +54,11 @@ export function useGestureHandlerEvent(
   };
 
   if (config.disableReanimated) {
-    return (event: UpdateEvent<Record<string, unknown>>) =>
-      onGestureHandlerEvent(event, jsContext);
+    return isAnimatedEvent(config.onUpdate)
+      ? // eslint-disable-next-line @typescript-eslint/no-empty-function
+        () => {}
```
j-piasecki: Can we return `undefined` here?

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: I'm not sure 🤷 I like the idea of pointing that specific property is not accessible, but on the other hand first version clearly states which one should be avoided. (Or maybe it was not clear hence the suggestion? Let me know)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Done in [ed1f063](https://github.com/software-mansion/react-native-gesture-handler/pull/3630/commits/ed1f06374ab6ac66a4c42cc8dcb0b62847bd1721). However, I think it will change given that I'll change types in this function 😅

packages/react-native-gesture-handler/src/v3/hooks/events/useGestureHandlerEvent.ts
m-bert: Yep, that's better. Done in [78c1903](https://github.com/software-mansion/react-native-gesture-handler/pull/3630/commits/78c19033860590030bc1206a214c184bb027b2b9).

At first I have not used `undefined` since it was throwing `Failed to create event handlers` error... but I was the one who added it and I'll change it to our needs 👑

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Well, it requires some type gymnastics but I can see the point. Done in [78c1903](https://github.com/software-mansion/react-native-gesture-handler/pull/3630/commits/78c19033860590030bc1206a214c184bb027b2b9)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
j-piasecki: I feel like it's the job for the docs, while the error should tell you exactly what you did wrong.

m-bert: I'm leaning towards first version as it also states what went wrong, it is generic message but you can easily figure out what happened. 

I get the idea so I can change it, but I'd also like to know what others think (though I think this error will be rarely thrown).

cc @latekvo @akwasniewski

akwasniewski: I think the first version is cleaner and achieves the same goal. The key name always contains `change` so it's easy to figure out what is wrong

latekvo: I prefer the version recommended by @j-piasecki, it's more precise while conveying the same meaning.

The original version could also be clearer, I initially thought `change` was a verb, and not a noun. The sentence could definietely cause confusion in non-native speakers.
This would work better:

> `'change*' properties are not supported by Animated.Event`

latekvo: We've got a tie. Could do:

```ts

tagMessage(`'${key}' is not supported by Animated.Event. 'change*' props are not supported.`);
```

I believe excessive clarity is better than confusion.
It's still brief enough for an error.

m-bert: Seems like we got a draw 😬 But I've asked an expert and `cmd + f` convinced me. Changed in [0829eb0](https://github.com/software-mansion/react-native-gesture-handler/pull/3630/commits/0829eb00e9cc4c6f19cba6892e9384a539c93d50).

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -98,10 +98,17 @@ export function isAnimatedEvent(
...
+  if (!('nativeEvent' in obj)) {
+    return;
+  }
+  const nativeEvent = obj.nativeEvent;
+
+  if (!('handlerData' in nativeEvent)) {
+    return;
+  }
+
+  const payload = nativeEvent.handlerData;
```
```suggestion
  if (!('nativeEvent' in obj) || !('handlerData' in obj.nativeEvent)) {
    return;
  }

  const payload = obj.nativeEvent.handlerData;
```

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -18,9 +19,9 @@ type GestureType =
 
 type GestureEvents = {
   onGestureHandlerStateChange: (event: any) => void;
-  onGestureHandlerEvent: (event: any) => void;
+  onGestureHandlerEvent: undefined | ((event: any) => void);
   onGestureHandlerTouchEvent: (event: any) => void;
-  onGestureHandlerAnimatedEvent: (event: any) => void;
+  onGestureHandlerAnimatedEvent: undefined | AnimatedEvent;
```
latekvo, line 24:
```suggestion
  onGestureHandlerAnimatedEvent?: AnimatedEvent;
```

We usually use `?:` instead of `undefined` for optional fields.

Same for 2 lines above.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Good catch, done in [f0b6e82](https://github.com/software-mansion/react-native-gesture-handler/pull/3630/commits/f0b6e82af958f023d84689e47b39955bda7b6456)

Also it made me think, if we are sure that `handlerData` exists, do we still need 

```js
if (!payload) {
  return;
}
```

? The only situation where it would fail is when `handlerData` is explicitly marked as `undefined`, i.e. `{handlerData: undefined}`.

j-piasecki: Yep, you're right. I think we can remove that check

m-bert: Removed in [2010f62](https://github.com/software-mansion/react-native-gesture-handler/pull/3630/commits/2010f62fbe5bf51b5f723c5e1c4c28fa128895fb)

packages/react-native-gesture-handler/src/v3/hooks/useGestureEvent.ts
```diff
@@ -24,10 +26,15 @@ export function useGestureEvent(
...
-  const onGestureHandlerAnimatedEvent =
-    config.onGestureHandlerAnimatedEvent as (...args: any[]) => void;
+  let onGestureHandlerAnimatedEvent: AnimatedEvent | undefined;
+
+  if (isAnimatedEvent(config.onUpdate)) {
+    for (const mapping of config.onUpdate._argMapping) {
+      checkMappingForChangeProperties(mapping);
+    }
+
+    onGestureHandlerAnimatedEvent = config.onUpdate as AnimatedEvent;
```
latekvo, line 37: This typecast doesn't seem necessary. `isAnimatedEvent` already smart-casts `config.onUpdate` to `AnimatedEvent`.

Removing it will cause a lint warning, but that's only because `config` is typed as `any`, which won't be the case in the future.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: This is not optional field, it may have a value of `undefined`, but it is not optional in a way that you don't have to pass it.

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -41,3 +41,10 @@ export type CallbackHandlers = Omit<
   | 'changeEventCalculator'
   | 'onChange'
 >;
+
+// This is almost how Animated.event is typed in React Native. We add _argMapping in order to:
+// 1. Distinguish it from a regular function,
+// 2. Have access to the _argMapping property to check for usage of `change*` callbacks.
+export type AnimatedEvent = ((...args: any[]) => void) & {
+  _argMapping?: unknown;
```
latekvo, line 49:
```suggestion
  _argMapping?: Animated.Mapping[];
```

`_argMapping` is already typed as `Animated.Mapping[]` in `checkMappingForChangeProperties`

packages/react-native-gesture-handler/src/v3/hooks/useGestureEvent.ts
m-bert: > Removing it will cause a lint warning, but that's only because config is typed as any, which won't be the case in the future.

So we might as well do it in the future. But I can change that if you insist.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
latekvo: Alright, understood.

packages/react-native-gesture-handler/src/v3/hooks/useGestureEvent.ts
latekvo: Sure, you may fix this in future if you insist. Just note that this typecast is redundant, duplicating a cast that already occurs a few lines before.

j-piasecki: > So we might as well do it in the future.

Please leave TODOs for things to do in the future. I can guarantee we won't remember about this cast in a few weeks.

m-bert: Fine, added in [8c074ce](https://github.com/software-mansion/react-native-gesture-handler/pull/3630/commits/8c074ce1e2f7b617e1ded618e90541745feba554)

akwasniewski (approved):

latekvo (approved):
:shipit:

PR #3631, opened 2025-07-23
Remove `onChange` callback references

m-bert:
## Description

In #3617 we've merged `onUpdate` and `onChange` callbacks so `onUpdate` now contains information about changes (or better to say, it will when we implement it 😄). However, there are still places in code where you can see `onChange` being referenced. This PR cleans that up.

## Test plan

Tested on basic-example.

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/types.ts
+++ b/packages/react-native-gesture-handler/src/v3/types.ts
@@ -34,5 +34,9 @@ export type TouchEvent =
 
 export type CallbackHandlers = Omit<
   HandlerCallbacks<Record<string, unknown>>,
-  'gestureId' | 'handlerTag' | 'isWorklet'
+  | 'gestureId'
+  | 'handlerTag'
+  | 'isWorklet'
+  | 'changeEventCalculator'
+  | 'onChange'
 >;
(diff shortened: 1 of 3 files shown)


packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -34,5 +34,9 @@ export type TouchEvent =
 
 export type CallbackHandlers = Omit<
   HandlerCallbacks<Record<string, unknown>>,
-  'gestureId' | 'handlerTag' | 'isWorklet'
+  | 'gestureId'
+  | 'handlerTag'
+  | 'isWorklet'
+  | 'changeEventCalculator'
+  | 'onChange'
 >;
```
j-piasecki, line 43: Can we get a todo to replace this type with a v3-specific one?

m-bert: Sure, I assumed that we will remember, but it won't hurt to add `todo`.

m-bert: Added in [35bd2ed](https://github.com/software-mansion/react-native-gesture-handler/pull/3631/commits/35bd2ede11102ae019f6866e83410d5b105fc664)

PR #3632, opened 2025-07-23
Deprecate `gestureHandlerRootHOC`

m-bert:
## Description

This PR deprecates usage of `gestureHandlerRootHOC` function. It also removes it from docs.

## Test plan

Run docs.

### DIFF
--- a/packages/react-native-gesture-handler/src/components/gestureHandlerRootHOC.tsx
+++ b/packages/react-native-gesture-handler/src/components/gestureHandlerRootHOC.tsx
@@ -5,7 +5,7 @@ import GestureHandlerRootView from './GestureHandlerRootView';
 
 /**
  * @deprecated `gestureHandlerRootHOC` is deprecated and will be removed in the future version of Gesture Handler.
- * Use `GestureHandlerRootView` instead.
+ * Use `GestureHandlerRootView` directly instead.
  */
 export default function gestureHandlerRootHOC<P extends object>(
   Component: React.ComponentType<P>,


packages/docs-gesture-handler/docs/fundamentals/installation.md
```diff
@@ -145,22 +141,40 @@ import PushedScreen from './PushedScreen';
...
   Navigation.registerComponent(
     'example.FirstTabScreen',
-    () => gestureHandlerRootHOC(FirstTabScreen),
+    () => {
+      return (
+        <GestureHandlerRootView>
+          <FirstTabScreen />
+        </GestureHandlerRootView>
+      );
+    },
```
m-bert, line 150: We could also add inline version of `gestureHandlerRootHOC` above in these examples, let me know what you think.

j-piasecki: I don't see the reason - besides adding a display name and calling `hoistNonReactStatics`, it doesn't do anything more compared to this snippet.

j-piasecki (approved):

packages/docs-gesture-handler/docs/fundamentals/installation.md
m-bert: I was thinking about cosmetic reasons, to make this example look a bit cleaner. But yes, I agree that it is not necessary (and probably better to explicitly use `GestureHandlerRootView`) 😅

PR #3634, opened 2025-07-24
`display: contents` based button styling

j-piasecki:
## Description

Changes the button implementation of the native `GestureHandlerButton` component, not to re-export the button component. Instead, it provides a "sandwich" implementation based on `display: contents`.

The issues we had with the styling of the buttons in the past stemmed from it extending `ViewGroup` instead of `ReactViewGroup`. This cannot be changed due to the difference in handling of `onTouchEvent`. We experimented with wrapping the button with a `View` in the past, but properly handling layout turned out to be quite a challenge. Hopefully, with the addition of `display: contents`, this will be easier now.

This PR changes the structure of the button to be:
- `RNGestureHandlerButtonWrapperNativeComponent`
  - `View`
    - `ButtonComponent`

The styles passed from the user are split into two categories:
- visual only
- affecting layout

Those affecting the layout need to be set on the button itself, while visual changes need to be set on the View wrapping it. Exceptions are: `zIndex`, `transform`, and `transformOrigin`, which need to be on the View. `zIndex` because the View is being rendered in the hierarchy at the level the button would be expected, and transforms to properly calculate touch coordinate transforms.

### Why not just `ButtonComponent`?

We need a `View` component as its parent to handle all styles that React Native supports properly.

### Why not just `ButtonComponent` wrapped with `View`?

In order for the view to always match the layout of the button, we need to set it at the shadow node level. Shadow nodes only have references to their children, not their parents. To have access to the layout of the button and the ability to copy it to the view wrapping it, there needs to be one more node as the parent of that subtree. That node is responsible for making sure that the size and positioning of the View match the Button.

The native components for the wrapper serve no purpose, since it's being rendered with `display: contents`, but RN doesn't know that - there needs to be something it can link to.

## Test plan

See the `ContentsButton` file

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
@@ -332,13 +332,17 @@
 
 - (RNGHUIView *)tryFindGestureHandlerButton:(RNGHUIView *)inView
 {
-  if (inView.subviews.count > 0) {
-    auto view = inView.subviews[0];
-    if ([view isKindOfClass:[RNGestureHandlerButtonComponentView class]]) {
-      RCTViewComponentView *componentView = (RCTViewComponentView *)view;
-      if (componentView.contentView != nil) {
-        return componentView.contentView;
-      }
+  if (inView.subviews.count == 0) {
+    return nil;
+  }
+  
+  auto view = inView.subviews[0];
+  
+  if ([view isKindOfClass:[RNGestureHandlerButtonComponentView class]]) {
+    RCTViewComponentView *componentView = (RCTViewComponentView *)view;
+    
+    if (componentView.contentView != nil) {
+      return componentView.contentView;
     }
   }
 


apps/basic-example/src/ContentsButton.tsx
m-bert, line 1: `MyButton` ❤️

packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerButtonWrapperShadowNode.cpp
```diff
@@ -0,0 +1,90 @@
+
+/**
+ * This code was generated by
+ * [react-native-codegen](https://www.npmjs.com/package/react-native-codegen).
+ *
+ * Do not edit this file as changes may cause incorrect behavior and will be
+ * lost once the code is regenerated.
+ *
+ * @generated by codegen project: GenerateShadowNodeCpp.js
+ */
```
m-bert: Maybe we can remove this header 😅

packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerButtonWrapperShadowNode.h
```diff
@@ -0,0 +1,59 @@
+
+/**
+ * This code was generated by
+ * [react-native-codegen](https://www.npmjs.com/package/react-native-codegen).
+ *
+ * Do not edit this file as changes may cause incorrect behavior and will be
+ * lost once the code is regenerated.
+ *
+ * @generated by codegen project: GenerateShadowNodeH.js
+ */
```
m-bert: Same

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -160,6 +162,14 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
     val id = if (child is ReactSwipeRefreshLayout) {
       child.getChildAt(0).id
+      // TODO: figure out how to do it correctly
+    } else if (child is ViewGroup && child.isNotEmpty()) {
+      val grandChild = child.getChildAt(0)
+      if (grandChild is RNGestureHandlerButtonViewManager.ButtonViewGroup) {
+        grandChild.id
+      } else {
+        child.id
+      }
```
m-bert: Okay, for now I don't have better way of doing this. This one seems ok though. 

What I'd change is to extract this into helper (like `isGestureHandlerButton` or something along this).

If you prefer not to do this, I'd combine it into one condition to avoid nesting, i.e.

```kt
else if (child is ViewGroup && child.isNotEmpty() && child.getChildAt(0) is RNGestureHandlerButtonViewManager.ButtonViewGroup) {
  child.getChildAt(0).id
} else {
  child.id
}
```

(but as I think about it now, it still doesn't look great)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -295,10 +296,19 @@ - (void)tryAttachNativeHandlersToChildView
...
   RNGHUIView *view = self.subviews[0];
 
+  // TODO: figure out how to do it correctly
   if ([view isKindOfClass:[RCTViewComponentView class]]) {
     RCTViewComponentView *componentView = (RCTViewComponentView *)view;
     if (componentView.contentView != nil) {
       view = componentView.contentView;
+    } else if (componentView.subviews.count > 0) {
+      view = componentView.subviews[0];
+      if ([view isKindOfClass:[RCTViewComponentView class]]) {
```
m-bert: Similar, to android, probably we could somehow try to extract this button specific logic. Also, wouldn't it be possible to check directly for `RNGestureHandlerButtonComponentView`? Or is there a specific reason why you don't want to do this?

```obj-c
 if ([view isKindOfClass:[RNGestureHandlerButtonComponentView class]]) {
  RNGestureHandlerButtonComponentView *componentView = (RNGestureHandlerButtonComponentView *)view;
  if (componentView.contentView != nil) {
    view = componentView.contentView;
  }
}
```

j-piasecki: Done in [4b52138](https://github.com/software-mansion/react-native-gesture-handler/pull/3634/commits/4b521387759786bc585db7949afc3c7500c96493)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: Done in [4b52138](https://github.com/software-mansion/react-native-gesture-handler/pull/3634/commits/4b521387759786bc585db7949afc3c7500c96493)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -322,6 +330,21 @@ - (void)detachNativeGestureHandlers
...
+{
+  if (inView.subviews.count > 0) {
+    auto view = inView.subviews[0];
+    if ([view isKindOfClass:[RNGestureHandlerButtonComponentView class]]) {
+      RCTViewComponentView *componentView = (RCTViewComponentView *)view;
+      if (componentView.contentView != nil) {
+        return componentView.contentView;
+      }
+    }
+  }
```
```suggestion
  if (inView.subviews.count == 0) {
    return nil;
  }
  
  auto view = inView.subviews[0];
  
  if ([view isKindOfClass:[RNGestureHandlerButtonComponentView class]]) {
    RCTViewComponentView *componentView = (RCTViewComponentView *)view;
    
    if (componentView.contentView != nil) {
      return componentView.contentView;
    }
  }
```

In `android` it looks ok, but Obj-c is not the prettiest language and I think we can avoid this one level 😅

PR #3636, opened 2025-07-25
Fix `onUpdate` crash 

m-bert:
## Description

#3630 introduced automatic support for `Animated.Event` in `onUpdate` callback. However, I've missed the fact that this callback may not be defined. This PR fixes this crash.

## Test plan

Tested on current basic-example without specifying `onUpdate`

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/utils.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/utils.ts
@@ -93,7 +93,7 @@ export function isAnimatedEvent(
 ): callback is AnimatedEvent {
   'worklet';
 
-  return '_argMapping' in callback;
+  return !!callback && '_argMapping' in callback;
 }
 
 export function checkMappingForChangeProperties(obj: Animated.Mapping) {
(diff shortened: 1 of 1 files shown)


packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -93,7 +93,7 @@ export function isAnimatedEvent(
 ): callback is AnimatedEvent {
   'worklet';
 
-  return '_argMapping' in callback;
+  return !!callback && '_argMapping' in callback;
```
j-piasecki, line 96: Why isn't this typed as `undefined` then?

m-bert: Good catch, seems like I've removed this type when experimenting with @coado 😅 

Added in [003ae65](https://github.com/software-mansion/react-native-gesture-handler/pull/3636/commits/003ae657f4f62338079e2f6e8e1cea6266fa9aab)

j-piasecki (approved):

PR #3638, opened 2025-07-25
Adapt `NativeViewGestureHandler` to `NativeDetector`

m-bert:
## Description

This PR brings support for `NativeViewGestureHandler` into `NativeDetector`. It does so, by attaching handler into child instead of detector view. 

## Status

- ### Android ✅ 
- ### iOS ✅ 

## Test plan

<details>

```jsx
import { StyleSheet, Text, View, ScrollView } from 'react-native';
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,

  const items = Array.from({ length: 30 }, (_, index) => `Item ${index + 1}`);

    onBegin: (e) => {
      console.log('onBegin', e);
    },
    onStart: (e) => {
      console.log('onStart', e);
    },
    onEnd: (e) => {
      console.log('onEnd', e);
    },
    onFinalize: (e) => {
      console.log('onFinalize', e);
    },
    onTouchesDown: (e) => {
      console.log('onTouchesDown', e);
    },
    onTouchesMove: (e) => {
      console.log('onTouchesMove', e);
    },
    onTouchesCancelled: (e) => {
      console.log('onTouchesCancelled', e);
    },
    onTouchesUp: (e) => {
      console.log('onTouchesUp', e);
    },
    onUpdate: (e) => {
      console.log('onUpdate', e);
    },
    onChange: (e) => {
      console.log('onChange', e);
    },
    onCancel: (e) => {
      console.log('onCancel', e);
    },
  });

  return (
    <GestureHandlerRootView
        <ScrollView style={styles.scrollView}>
          {items.map((item, index) => (
            </View>
          ))}
        </ScrollView>
      </NativeDetector>
    </GestureHandlerRootView>
  );
}

  scrollView: {
    backgroundColor: 'lightgrey',
    marginHorizontal: 20,
  },
  item: {
    flexDirection: 'row',
    justifyContent: 'center',
    alignItems: 'center',
    padding: 20,
    margin: 2,
    backgroundColor: 'white',
    borderRadius: 10,
  },
  text: {
    fontSize: 20,
    color: 'black',
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
@@ -103,10 +103,6 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
       ?: throw Exception("Tried to access a non-existent registry")
 
     for (tag in nativeHandlersToAttach) {
-      if (tag in attachedHandlers) {
-        continue
-      }
-
       registry.attachHandlerToView(tag, childId, GestureHandler.ACTION_TYPE_NATIVE_DETECTOR)
 
       attachedHandlers.add(tag)
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
@@ -178,10 +178,6 @@ typedef NS_ENUM(NSInteger, RNGestureHandlerMutation) {
   RNGestureHandlerManager *handlerManager = [RNGestureHandlerModule handlerManagerForModuleId:_moduleId];
 
   for (NSNumber *handlerTag in _nativeHandlersToAttach) {
-    if ([_attachedHandlers containsObject:handlerTag]) {
-      continue;
-    }
-
     [handlerManager.registry attachHandlerWithTag:handlerTag
                                            toView:self.subviews[0]
                                    withActionType:RNGestureHandlerActionTypeNativeDetector];


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEventDispatcher.kt
```diff
@@ -78,35 +79,43 @@ class RNGestureHandlerEventDispatcher(private val reactApplicationContext: React
...
-          val event = RNGestureHandlerEvent.obtain(
-            handler,
-            handler.actionType,
-            handlerFactory.createEventBuilder(handler),
-          )
-          view.dispatchEvent(event)
+        val view = if (handler is NativeViewGestureHandler) handler.view?.parent else handler.view
+
+        assert(view is RNGestureHandlerDetectorView) {
+          "[react-native-gesture-handler] Expected RNGestureHandlerDetectorView."
```
m-bert: I'm open to suggestions how to enhance this error message 😅

packages/react-native-gesture-handler/apple/RNGestureHandlerPointerTracker.mm
```diff
@@ -237,12 +237,14 @@ - (void)cancelPointers
 
 - (void)sendEvent
 {
+  if (!_gestureHandler.needsPointerData) {
+    return;
+  }
+
```
m-bert, line 243: I think it looks cleaner when it is split. Let me know if you disagree.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -124,6 +124,9 @@ open class GestureHandler {
     return false
   }
 
+  fun hasNativeDetectorActionType() =
```
akwasniewski: This function can be inline 
```suggestion
  inline fun hasNativeDetectorActionType() =
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -56,22 +72,44 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
-        registry.attachHandlerToView(
-          entry.key,
-          this.id,
-          if (dispatchesAnimatedEvents) {
-            GestureHandler.ACTION_TYPE_NATIVE_DETECTOR_ANIMATED_EVENT
-          } else {
-            GestureHandler.ACTION_TYPE_NATIVE_DETECTOR
-          },
-        )
+        if (registry.getHandler(entry.key) is NativeViewGestureHandler) {
```
j-piasecki: You're not calling `attachNativeGestureHandlers` after populating `nativeHandlersToAttach`. If the gesture is attached after the view hierarchy is ready, it won't attach.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -40,6 +43,19 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
+  override fun addView(child: View?, index: Int, params: LayoutParams?) {
+    super.addView(child, index, params)
+
+    if (child == null) {
+      throw Exception(
+        "[react-native-gesture-handler] Cannot attach gesture handler. NativeGestureDetector got null as child.",
+      )
+    }
+
+    attachNativeGestureHandlers(child.id)
```
j-piasecki: I think you should also detach them in `removeChild`.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEventDispatcher.kt
j-piasecki: `Expected RNGestureHandlerDetectorView to be the target for the event`.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki:
```suggestion
        if (registry.getHandler(entry.key).wantsToAttachDirectlyToView()) {
```

What do you think about moving that to a method instead of checking for Native handler all the time?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
j-piasecki:
```suggestion
  fun isSendingEventsToNativeDetector() =
```

WDYT?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
```diff
@@ -33,7 +34,16 @@ class RNGestureHandlerEvent private constructor() : Event<RNGestureHandlerEvent>
     useNativeAnimatedName: Boolean,
   ) {
     val view = handler.view!!
-    super.init(UIManagerHelper.getSurfaceId(view), view.id)
+
+    if (handler is NativeViewGestureHandler && handler.hasNativeDetectorActionType()) {
+      super.init(
+        UIManagerHelper.getSurfaceId(view.parent as RNGestureHandlerDetectorView),
+        (view.parent as RNGestureHandlerDetectorView).id,
+      )
```
j-piasecki:
```suggestion
      val detector = view.parent as RNGestureHandlerDetectorView
      super.init(
        UIManagerHelper.getSurfaceId(detector),
        detector.id,
      )
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEventDispatcher.kt
```diff
@@ -78,35 +79,43 @@ class RNGestureHandlerEventDispatcher(private val reactApplicationContext: React
...
       GestureHandler.ACTION_TYPE_NATIVE_DETECTOR -> {
-        val view = handler.view
-        if (view is RNGestureHandlerDetectorView) {
-          val event = RNGestureHandlerEvent.obtain(
-            handler,
-            handler.actionType,
-            handlerFactory.createEventBuilder(handler),
-          )
-          view.dispatchEvent(event)
+        val view = if (handler is NativeViewGestureHandler) handler.view?.parent else handler.view
```
j-piasecki:
```suggestion
        val view = handler.viewForEvents
```

We can do it the same way as on iOS. The assert could then be moved into the getter.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerStateChangeEvent.kt
```diff
@@ -28,7 +29,15 @@ class RNGestureHandlerStateChangeEvent private constructor() : Event<RNGestureHa
     dataBuilder: GestureHandlerEventDataBuilder<T>,
   ) {
     val view = handler.view!!
-    super.init(UIManagerHelper.getSurfaceId(view), view.id)
+
+    if (handler is NativeViewGestureHandler && handler.hasNativeDetectorActionType()) {
+      super.init(
+        UIManagerHelper.getSurfaceId(view.parent as RNGestureHandlerDetectorView),
+        (view.parent as RNGestureHandlerDetectorView).id,
+      )
```
j-piasecki:
```suggestion
      val detector = view.parent as RNGestureHandlerDetectorView
      super.init(
        UIManagerHelper.getSurfaceId(detector),
        detector.id,
      )
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerTouchEvent.kt
```diff
@@ -14,7 +15,15 @@ class RNGestureHandlerTouchEvent private constructor() : Event<RNGestureHandlerT
 
   private fun <T : GestureHandler> init(handler: T, actionType: Int) {
     val view = handler.view!!
-    super.init(UIManagerHelper.getSurfaceId(view), view.id)
+
+    if (handler is NativeViewGestureHandler && handler.hasNativeDetectorActionType()) {
+      super.init(
+        UIManagerHelper.getSurfaceId(view.parent as RNGestureHandlerDetectorView),
+        (view.parent as RNGestureHandlerDetectorView).id,
+      )
```
j-piasecki:
```suggestion
      val detector = view.parent as RNGestureHandlerDetectorView
      super.init(
        UIManagerHelper.getSurfaceId(detector),
        detector.id,
      )
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
m-bert: Yes and no 😅 This file has only one such check so it is not a big deal. But we also have checks for `NativeViewGestureHandler` **and** `actionType`. We could combine these 2. In that case I think we should come up with something else than `wantsToAttachDirectlyToView`. Also, it would be a bit redundant check here, but I don't think that it will have a big impact on performance. Let me know what you think.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEventDispatcher.kt
m-bert: Done [3b783d2](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/3b783d2ab7e9cab27f6a853a7d4d07b9f54743d4)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
m-bert: I agree 👍 Done in [ad313ae](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/ad313ae69ba0c67f57d985ba0288dcfbd41b4f01)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
m-bert: Done in [5ffc4e8](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/5ffc4e8935a6d3052a92f0569958421ff1cee67a)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerStateChangeEvent.kt
m-bert: Also done in [5ffc4e8](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/5ffc4e8935a6d3052a92f0569958421ff1cee67a)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerTouchEvent.kt
m-bert: Also done in [5ffc4e8](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/5ffc4e8935a6d3052a92f0569958421ff1cee67a)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerTouchEvent.kt
```diff
@@ -14,7 +15,13 @@ class RNGestureHandlerTouchEvent private constructor() : Event<RNGestureHandlerT
...
   private fun <T : GestureHandler> init(handler: T, actionType: Int) {
     val view = handler.view!!
-    super.init(UIManagerHelper.getSurfaceId(view), view.id)
+
+    if (handler is NativeViewGestureHandler && handler.isSendingEventsToNativeDetector()) {
+      val detector = view.parent as RNGestureHandlerDetectorView
+      super.init(UIManagerHelper.getSurfaceId(detector), (detector).id)
+    } else {
+      super.init(UIManagerHelper.getSurfaceId(view), view.id)
+    }
```
latekvo: While not a strong preference, could simplify this a bit:

```suggestion
    val view = if (handler is NativeViewGestureHandler && handler.isSendingEventsToNativeDetector()) {
      handler.view!!.parent as RNGestureHandlerDetectorView
    } else {
      handler.view!!
    }

    super.init(UIManagerHelper.getSurfaceId(view), view.id)
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerStateChangeEvent.kt
```diff
@@ -28,7 +29,13 @@ class RNGestureHandlerStateChangeEvent private constructor() : Event<RNGestureHa
...
   ) {
     val view = handler.view!!
-    super.init(UIManagerHelper.getSurfaceId(view), view.id)
+
+    if (handler is NativeViewGestureHandler && handler.isSendingEventsToNativeDetector()) {
+      val detector = view.parent as RNGestureHandlerDetectorView
+      super.init(UIManagerHelper.getSurfaceId(detector), detector.id)
+    } else {
+      super.init(UIManagerHelper.getSurfaceId(view), view.id)
+    }
```
latekvo: Same here

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
```diff
@@ -33,7 +34,14 @@ class RNGestureHandlerEvent private constructor() : Event<RNGestureHandlerEvent>
...
   ) {
     val view = handler.view!!
-    super.init(UIManagerHelper.getSurfaceId(view), view.id)
+
+    if (handler is NativeViewGestureHandler && handler.isSendingEventsToNativeDetector()) {
+      val detector = view.parent as RNGestureHandlerDetectorView
+      super.init(UIManagerHelper.getSurfaceId(detector), detector.id)
+    } else {
+      super.init(UIManagerHelper.getSurfaceId(view), view.id)
+    }
```
latekvo: Same here, although seeing how this same piece of code is duplicated 3 times, perhaps we could extract it to a utility function?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEventDispatcher.kt
```diff
@@ -208,11 +221,15 @@ class RNGestureHandlerEventDispatcher(private val reactApplicationContext: React
...
+
+        assert(view is RNGestureHandlerDetectorView) {
+          "[react-native-gesture-handler] Expected RNGestureHandlerDetectorView."
         }
+
+        val event = RNGestureHandlerTouchEvent.obtain(handler, handler.actionType)
+
+        (view as RNGestureHandlerDetectorView).dispatchEvent(event)
```
latekvo: If you were to throw error conditionally instead of asserting, then the `as RNGestureHandlerDetectorView` typecast would not be necessary due to smart-casting.

But I'm completely neutral on this, just wanted to hear your opinion.

packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
```diff
@@ -354,23 +354,23 @@ - (void)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer
 
 - (void)sendEvent:(RNGestureHandlerStateChange *)event
     withActionType:(RNGestureHandlerActionType)actionType
-     forRecognizer:(UIGestureRecognizer *)recognizer
+           forView:(RNGHUIView *)detectorView // I'd love to make it RNGestureHandlerDetector, however it results in
+                                              // compilation errors because of importing C++
```
latekvo: Could we make this comment a little bit more straight to the point?

>Typing as `RNGestureHandlerDetector` is preferable, but results in a compilation error.

latekvo (changes requested):

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
m-bert: >although seeing how this same large piece of code is duplicated 3 times, perhaps we could extract it to a utility function?

This would probably require creating our own super class for these events. I don't say it is bad, but also I don't think it is something to be done in this PR, and if we want it at all (cc @j-piasecki)

packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
m-bert: What change do you expect in this comment?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerTouchEvent.kt
m-bert: Done in [888964b](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/888964b0cafe3c8fad27185e968f65163519bf03)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerStateChangeEvent.kt
m-bert: Done in [888964b](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/888964b0cafe3c8fad27185e968f65163519bf03)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
m-bert: Reformatted in [888964b](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/888964b0cafe3c8fad27185e968f65163519bf03)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEventDispatcher.kt
m-bert: `assert` looks better, but I think if we can get rid of this cast than yes, we can change it. Done in [4a834b2](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/4a834b2c1e60645f05017237f62f95e5fdb85816)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
m-bert: I thought about it for a while and I don't think it really matters. JIT should perform inlining anyway and if you explicitly mark it as `inline` (which is only a suggestion for the compiler) it will warn you that increase in performance will not be that big because there's no lambda passed as parameter.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEventDispatcher.kt
m-bert: Actually in this commit I've missed other events, but it doesn't really matter as it was replaced by `viewForEvents` in [8f8b708](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/8f8b7082763b279396a0a876e19c72b9c905fec5)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEventDispatcher.kt
m-bert: Let me know what you think about [8f8b708](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/8f8b7082763b279396a0a876e19c72b9c905fec5)

packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
latekvo: My suggestion was in the quote block in my previous comment. Here it is again, but in a `suggestion` block:

```suggestion
           forView:(RNGHUIView *)detectorView // Typing as RNGestureHandlerDetector is preferable, but results in a compilation error
```

m-bert: Oh, that was a suggestion. I thought you wanted some detailed explanation

m-bert: Changed in [d118d6a](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/d118d6ac02226304393411af20090f20dbb9d5ac)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
m-bert: This should be fixed, along with detaching children in [78ceffa](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/78ceffa948c97b1afd7ab1dc3c3db7bd82a4a378)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
m-bert: Fixed:

- Android [78ceffa](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/78ceffa948c97b1afd7ab1dc3c3db7bd82a4a378)
- iOS [97049a4](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/97049a498d358c6951e923241102afed389cd03d)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -124,21 +141,65 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
...
+  RNGestureHandlerManager *handlerManager = [RNGestureHandlerModule handlerManagerForModuleId:_moduleId];
+
+  for (NSNumber *handlerTag in _nativeHandlersToAttach) {
+    if ([_attachedHandlers containsObject:handlerTag]) {
+      continue;
+    }
+
+    [handlerManager.registry attachHandlerWithTag:handlerTag
+                                           toView:self.subviews[0]
+                                   withActionType:RNGestureHandlerActionTypeNativeDetector];
```
m-bert, line 183: We will remove animated detector type anyway so we don't need that here

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
```diff
@@ -33,8 +33,14 @@ class RNGestureHandlerEvent private constructor() : Event<RNGestureHandlerEvent>
     dataBuilder: GestureHandlerEventDataBuilder<T>,
     useNativeAnimatedName: Boolean,
   ) {
-    val view = handler.view!!
+    val view = if (handler.actionType == GestureHandler.ACTION_TYPE_NATIVE_DETECTOR) {
+      handler.viewForEvents!!
+    } else {
+      handler.view!!
+    }
```
m-bert, line 40: We probably could use `handler.viewForEvents` directly, but this would require handling old APIs in this getter and I'm not sure if we want that (cc @j-piasecki)

j-piasecki: IMO, it's fine as is.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
j-piasecki: Not a strong opinion, but it's fine-ish? This is not a part of code that's changed often, and generalizing events may not be obvious due to their structure, especially now with `handlerData`. I'd add a TODO to clean it up once v3 is the only one left. I'm open to ideas, though.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -30,6 +31,19 @@ open class GestureHandler {
...
+
+      if (detector !is RNGestureHandlerDetectorView) {
+        throw Error(
+          "[react-native-gesture-handler] Expected RNGestureHandlerDetectorView to be the target for the event.",
+        )
+      }
+
+      return detector
+    }
+    private set
```
j-piasecki:
```suggestion
  val viewForEvents: RNGestureHandlerDetectorView?
    get() {
      val detector = if (this is NativeViewGestureHandler) this.view?.parent else view

      if (detector !is RNGestureHandlerDetectorView) {
        throw Error(
          "[react-native-gesture-handler] Expected RNGestureHandlerDetectorView to be the target for the event.",
        )
      }

      return detector
    }
```

You're not assigning it anywhere, nor using the backing field.

Also, what do you think about adding an assert that action type must be native detector when accessing this field?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -40,6 +43,32 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
     this.dispatchesAnimatedEvents = dispatchesAnimatedEvents
   }
 
+  private fun shouldAttachGestureToSubview(tag: Int): Boolean {
```
j-piasecki:
```suggestion
  private fun shouldAttachGestureToChildView(tag: Int): Boolean {
```

I don't like ios naming scheme and I'll die on this hill.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -40,6 +43,32 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
+
+  // We override this `addView` because it is called inside `addView(child: View?, index: Int)`
+  override fun addView(child: View?, index: Int, params: LayoutParams?) {
+    super.addView(child, index, params)
+
+    if (child == null) {
+      throw Exception(
+        "[react-native-gesture-handler] Cannot attach gesture handler. NativeGestureDetector got null as child.",
+      )
+    }
```
j-piasecki: Can this even happen? Have you tried changing the signature not to be nullable?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -55,15 +84,57 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
+          attachedHandlers.add(tag)
+        }
       } else if (entry.value == GestureHandlerMutation.Detach) {
-        registry.detachHandler(entry.key)
+        registry.detachHandler(tag)
+        attachedHandlers.remove(tag)
+      }
+    }
+
+    // This covers the case where `NativeViewGestureHandlers` are attached after children views were created.
```
j-piasecki:
```suggestion
    // This covers the case where `NativeViewGestureHandlers` are attached after child views were created.
```

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.h
```diff
@@ -14,12 +14,21 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface RNGestureHandlerDetector : RCTViewComponentView
 
+@property (nonatomic, nonnull) NSMutableSet *nativeHandlersToAttach;
+@property (nonatomic, nonnull) NSMutableSet *attachedHandlers;
```
j-piasecki: Those shouldn't be exposed publicly, no?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: It would allow us to get rid of the dependency between the detector and the native gesture handler if every gesture had `wantsToAttachDirectlyToView` returning false and native returning true.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -55,15 +84,57 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
+
+    // This covers the case where `NativeViewGestureHandlers` are attached after child views were created.
+    val child = getChildAt(0)
+
+    if (child != null) {
+      maybeAttachNativeGestureHandlers(child.id)
+    }
+  }
+
+  private fun maybeAttachNativeGestureHandlers(childId: Int) {
```
akwasniewski:
```suggestion
  private fun attachUnattachedNativeGestureHandlers(childId: Int) {
```

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -124,19 +148,59 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
...
+
+    [self maybeAttachNativeGestureHandlers];
   }
 
   [super updateProps:propsBase oldProps:oldPropsBase];
   // Override to force hittesting to work outside bounds
   self.clipsToBounds = NO;
 }
+
+- (void)maybeAttachNativeGestureHandlers
```
akwasniewski:
```suggestion
- (void)attachUnattachedNativeGestureHandlers
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
m-bert: I know that `maybe` prefix is more common in _**Reanimated**_, but I like it because when you look at method call you can clearly see that it may not attach any handler. `attachUnattached` suggests that there are some handlers that are being attached after others, which is not really the case as there are 2 possible ways for **_all_** handlers to be attached (but if you can see that it might not be the case let me know). 

Since we got a draw we need someone with strong opinion cc @j-piasecki 👑

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
m-bert: Same as above

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -81,6 +88,27 @@ - (void)dispatchTouchEvent:(RNGestureHandlerDetectorEventEmitter::OnGestureHandl
   }
 }
 
+- (BOOL)shouldAttachGestureToSubview:(NSNumber *)handlerTag
+{
+  RNGestureHandlerManager *handlerManager = [RNGestureHandlerModule handlerManagerForModuleId:_moduleId];
+
+  return [[[handlerManager registry] handlerWithTag:handlerTag] isKindOfClass:[RNNativeViewGestureHandler class]];
+}
```
j-piasecki: Same as android, this can be a method on `RNGestureHandler` and then you don't have a dependency on `RNNativeViewGestureHandler`.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: I don't have a strong opinion, but if this is the pattern that Reanimated follows, then we can stick to it.

Though, now that this can be more generic than just native gestures (with `wantsToAttachDirectlyToView`) I'd rename it to something like `maybeAttachHandlersToChildView` or something like that.

latekvo: I'll chip-in uninvited 😅 

> when you look at method call you can clearly see that it may not attach any handler

I think `try`, as in `tryAttach...` does that job better, as it is frequently used in our codebase for this purpose, while `maybe` isn't used anywhere in our codebase. Unless I'm missing something. 

On the other hand `attachUnattached` is a bit unnatural, but i think it pretty accurately describes what this method does.

akwasniewski: Ok, if it is used in reanimated than I guess we may leave it as is, though it still seems undescriptive. I agree with @latekvo `try..` does the job the best.

m-bert: > I'll chip-in uninvited 😅

Feel free to join 😅 

> I think try, as in tryAttach... does that job better

I've also thought about `try` and it might be the best.

m-bert: Okay, renamed in [e8966b4](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/e8966b4f7890c415dd3f8e98095462a4267bab1c)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
m-bert: Sure, added in [9d40768](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/9d40768fe46a9513e5f0b6ce6f4b079c590caafe)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -124,19 +151,59 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
     NSNumber *handlerTag = [NSNumber numberWithInt:handlerChange.first];
 
     if (handlerChange.second == RNGestureHandlerMutationAttach) {
-      // TODO: Attach to the child when attached gesture is a NativeGestureHandler, track children changes then
-      [handlerManager.registry attachHandlerWithTag:handlerTag
-                                             toView:self
-                                     withActionType:RNGestureHandlerActionTypeNativeDetector];
+      if ([self shouldAttachGestureToSubview:handlerTag]) {
+        [_nativeHandlersToAttach addObject:handlerTag];
```
akwasniewski, line 155: Why only native handlers are atteched in a dedicated function? Apart from a different prop it is the same as attaching normal handlers.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -124,19 +151,59 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
...
   // Override to force hittesting to work outside bounds
   self.clipsToBounds = NO;
 }
+
+- (void)maybeAttachNativeGestureHandlers
+{
+  RNGestureHandlerManager *handlerManager = [RNGestureHandlerModule handlerManagerForModuleId:_moduleId];
+
+  for (NSNumber *handlerTag in _nativeHandlersToAttach) {
+    if ([_attachedHandlers containsObject:handlerTag]) {
```
akwasniewski: This should never happen, the function is only called on new handlers.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -124,19 +151,59 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
...
+      continue;
+    }
+
+    [handlerManager.registry attachHandlerWithTag:handlerTag
```
akwasniewski, line 181: What if the first child changes? Shouldn't we reattach handlers?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
m-bert: There are two places where those handlers can be attached - here in `updateProps` and in `addSubview`, so it's better not to duplicate this logic but keep it in one place instead.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
m-bert: What about situation where we have composed gestures? For example, we have `NativeViewGestureHandler` and then we change it to `Race` with `Pan`. This way `Native` probably won't be detached. One question here is here is, will it be passed into `updateProps` (and `attachHandlers` on android).

cc @j-piasecki

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
m-bert: Correct me if I'm wrong, but if the child changes then handler will be detached in `willRemoveSubview` method and then attached here.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: All attached handler tags will be passed to `updateProps`, but after diffing, the ones that were already attached should have `RNGestureHandlerMutationKeep` status. They shouldn't go through this path at all.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerEvent.kt
m-bert: Well, this has changed a bit since the last time, let me know if you still want to add this `TODO`.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
m-bert: Then I assume the same holds for android. Changed in [68474f3](https://github.com/software-mansion/react-native-gesture-handler/pull/3638/commits/68474f3236a63b4022d4ef3af4c44a5d2eb04af2)

latekvo (approved):

PR #3646, opened 2025-07-30
Remove `AnimatedEvent` action type from `NativeDetector`

m-bert:
## Description

We came to a conclusion that current event handling may conflict with gesture relations, such as `simultaneous` when one of the gestures is using `Animated`. This PR removes `AnimatedEvent` action type from `NativeDetector` and adds flag to `GestureHandler`, which represents if handler should send events for `Animated`.

## Status 

- ### Android ✅ 
- ### iOS ✅ 

## Test plan

Test `pan1`, `pan2` and `gesture` from the code below.

>[!NOTE]
> Since `simultaneous` relation is not yet implemented, only first pan works when `gesture` is used. On Android you can solve that ba commenting out `otherHandler.cancel()` in `makeActive` function ([this line](https://github.com/software-mansion/react-native-gesture-handler/blob/0e5d58efcb2bd0ee2bd4eefa78fd366e003eea41/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt#L205))

<details>
<summary>Tested on the following code:</summary>

```tsx
import * as React from 'react';
import { Animated, Button, useAnimatedValue } from 'react-native';
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,
} from 'react-native-gesture-handler';

export default function App() {
  const [visible, setVisible] = React.useState(true);

  const value = useAnimatedValue(0);
  const event = Animated.event(
    [{ nativeEvent: { handlerData: { translationX: value } } }],
    {
      useNativeDriver: true,
    }
  );

  const pan1 = useGesture('PanGestureHandler', {
    onUpdate: (e: any) => {
      console.log('Pan1 update:', e);
    },
  });

  const pan2 = useGesture('PanGestureHandler', {
    onUpdate: event,
    dispatchesAnimatedEvents: true,
  });

  const gesture = {
    tag: [pan1.tag, pan2.tag],
    gestureEvents: {
      onGestureHandlerStateChange: (e) => {
        pan1.gestureEvents.onGestureHandlerStateChange(e);
        pan2.gestureEvents.onGestureHandlerStateChange(e);
      },
      onGestureHandlerEvent: pan1.gestureEvents.onGestureHandlerEvent,
      onGestureHandlerAnimatedEvent:
        pan2.gestureEvents.onGestureHandlerAnimatedEvent,
      onGestureHandlerTouchEvent: (e) => {
[...]

PR #3650, opened 2025-08-01
Automatically assign `dispatchesAnimatedEvents`

m-bert:
## Description

#3646 removed _animated native detector_ action type. However, now we have to pass `dispatchesAnimatedEvents` to gesture config, what was not done automatically. This PR fixes it.

## Test plan

Tested on `basic-example`

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
@@ -20,7 +20,7 @@ const ReanimatedNativeDetector = Reanimated?.default.createAnimatedComponent(
 );
 
 export function NativeDetector({ gesture, children }: NativeDetectorProps) {
-  const NativeDetectorComponent = gesture.dispatchesAnimatedEvents
+  const NativeDetectorComponent = gesture.config.dispatchesAnimatedEvents
     ? AnimatedNativeDetector
     : gesture.shouldUseReanimated
       ? ReanimatedNativeDetector
@@ -47,7 +47,9 @@ export function NativeDetector({ gesture, children }: NativeDetectorProps) {
       onGestureHandlerTouchEvent={
         gesture.gestureEvents.onGestureHandlerTouchEvent
       }
-      dispatchesAnimatedEvents={gesture.dispatchesAnimatedEvents}
+      dispatchesAnimatedEvents={
+        gesture.config.dispatchesAnimatedEvents as boolean
+      }
       moduleId={globalThis._RNGH_MODULE_ID}
       handlerTags={[gesture.tag]}
       style={styles.detector}>
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
@@ -30,7 +30,6 @@ export interface NativeGesture {
   config: Record<string, unknown>;
   gestureEvents: GestureEvents;
   shouldUseReanimated: boolean;
-  dispatchesAnimatedEvents: boolean;
 }
 
 function hasWorkletEventHandlers(config: Record<string, unknown>) {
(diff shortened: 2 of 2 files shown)


packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -119,8 +122,5 @@ export function useGesture(
       onGestureHandlerAnimatedEvent,
     },
     shouldUseReanimated,
-    dispatchesAnimatedEvents:
-      !!onGestureHandlerAnimatedEvent &&
-      '__isNative' in onGestureHandlerAnimatedEvent,
```
m-bert, line 124: I'm not sure if we want to remove it from `gesture` itself or not, let me know what you think @j-piasecki.

j-piasecki: We can remove it since it's not used anymore, though I don't like that `dispatchesAnimatedEvents` and `shouldUseReanimated` are passed down a different paths now.

m-bert: I've moved `shouldUseReanimated` to gesture `config` in [c3c4e75](https://github.com/software-mansion/react-native-gesture-handler/pull/3650/commits/c3c4e758b9890f7d3d0a35fc6a032bf7bff6d731)

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
```diff
@@ -22,7 +22,7 @@ const ReanimatedNativeDetector = Reanimated?.default.createAnimatedComponent(
 export function NativeDetector({ gesture, children }: NativeDetectorProps) {
   const NativeDetectorComponent = gesture.config.dispatchesAnimatedEvents
     ? AnimatedNativeDetector
-    : gesture.shouldUseReanimated
+    : (gesture.config.shouldUseReanimated as boolean)
```
j-piasecki: Why is this cast needed?

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -62,15 +61,16 @@ export function useGesture(
     );
   }
 
-  const shouldUseReanimated =
+  config.shouldUseReanimated =
```
j-piasecki, line 65: Can you add a comment that this needs to be don ASAP since other things may depend on `shouldUseReanimated`?

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
m-bert: Because config is not properly typed yet. Without this cast there's a warning that `shouldUseReanimated` is `unknown`

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: Sure, done in [018f93f](https://github.com/software-mansion/react-native-gesture-handler/pull/3650/commits/018f93fbbcd17f8899fdc686c043aa76d8e91b0e)

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
j-piasecki: I'd either leave it without a cast or add a todo to remove it once typed

m-bert: Added todo in [f25a516](https://github.com/software-mansion/react-native-gesture-handler/pull/3650/commits/f25a516d64463e0fec5803707437cf7ca81f6fcc)

apps/basic-example/src/NativeDetector.tsx
```diff
@@ -18,9 +18,7 @@ export default function App() {
   );
 
   const gesture = useGesture('PanGestureHandler', {
-    onGestureHandlerAnimatedEvent: event,
-    onGestureHandlerEvent: (e: any) =>
-      console.log('onGestureHandlerEvent', e.nativeEvent),
+    onUpdate: event,
```
m-bert, line 21: Not the best, but still better than example crashing on passing `onGestureHandlerAnimatedEvent`

PR #3637, opened 2025-07-25
[Web] Native detector component

akwasniewski:
## Description

This PR adds a web version of `NativeDetector` component following its IOS/android implementations. 

## Test plan
* copy all react source files from basic-example to expo-example
* add a console log, in the `NativeDetector` pan Gesture
* check whether the gesture is properly recognised

PR #3651, opened 2025-08-01
[Web] v3 typing and using handler data

akwasniewski:
## Description

Changing web implementation to use v3 types with a proper `handlerData` field.

## Test plan

check whether handledData is a proper field
```ts
    onBegin: (e: any) => {
      console.log('onGestureHandlerEvent', e.nativeEvent.handlerData);
    },
  });
 ```

### DIFF
--- a/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
@@ -6,7 +6,6 @@ import {
   PropsRef,
   ResultEvent,
   PointerData,
-  ResultTouchEvent,
   TouchEventType,
   EventTypes,
 } from '../interfaces';
@@ -20,6 +19,7 @@ import { PointerType } from '../../PointerType';
 import { GestureHandlerDelegate } from '../tools/GestureHandlerDelegate';
 import { ActionType } from '../../ActionType';
 import { tagMessage } from '../../utils';
+import { StateChangeEvent, UpdateEvent } from '../../v3/types';
 
 export default abstract class GestureHandler implements IGestureHandler {
   private lastSentState: State | null = null;
@@ -367,8 +367,7 @@ export default abstract class GestureHandler implements IGestureHandler {
     const { onGestureHandlerEvent, onGestureHandlerTouchEvent }: PropsRef =
       this.propsRef!.current;
 
-    const touchEvent: ResultTouchEvent | undefined =
-      this.transformTouchEvent(event);
+    const touchEvent: ResultEvent | undefined = this.transformTouchEvent(event);
 
     if (touchEvent) {
       if (
@@ -408,7 +407,6 @@ export default abstract class GestureHandler implements IGestureHandler {
       invokeNullableMethod(onGestureHandlerStateChange, resultEvent);
     }
     if (this.state === State.ACTIVE) {
-      resultEvent.nativeEvent.oldState = undefined;
       if (
         onGestureHandlerAnimatedEvent &&
         (this.actionType === ActionType.NATIVE_ANIMATED_EVENT ||
@@ -426,24 +424,24 @@ export default abstract class GestureHandler implements IGestureHandler {
     }
     return {
       nativeEvent: {
-        numberOfPointers: this.tracker.trackedPointersCount,
         state: newState,
-        pointerInside: this.delegate.isPointerInBounds(
-          this.tracker.getAbsoluteCoordsAverage()
-        ),
-        ...this.transformNativeEvent(),
         handlerTag: this.handlerTag,
-        target: this.viewRef,
-        oldState: newState !== oldState ? oldState : undefined,
-        pointerType: this.pointerType,
-      },
+        oldState: oldState,
+        handlerData: {
+          pointerType: this.pointerType,
+          numberOfPointers: this.tracker.trackedPointersCount,
+          pointerInside: this.delegate.isPointerInBounds(
+            this.tracker.getAbsoluteCoordsAverage()
+          ),
+          ...this.transformNativeEvent(),
+          target: this.viewRef,
+        },
+      } as StateChangeEvent<unknown> | UpdateEvent<unknown>,
       timeStamp: Date.now(),
     };
   }
 
-  private transformTouchEvent(
-    event: AdaptedEvent
-  ): ResultTouchEvent | undefined {
+  private transformTouchEvent(event: AdaptedEvent): ResultEvent | undefined {
     const rect = this.delegate.measureView();
 
     const all: PointerData[] = [];
@@ -573,7 +571,7 @@ export default abstract class GestureHandler implements IGestureHandler {
       });
     });
 
-    const cancelEvent: ResultTouchEvent = {
+    const cancelEvent: ResultEvent = {
       nativeEvent: {
         handlerTag: this.handlerTag,
         state: this.state,
@@ -880,10 +878,10 @@ export default abstract class GestureHandler implements IGestureHandler {
 
 function invokeNullableMethod(
   method:
-    | ((event: ResultEvent | ResultTouchEvent) => void)
-    | { __getHandler: () => (event: ResultEvent | ResultTouchEvent) => void }
+    | ((event: ResultEvent) => void)
+    | { __getHandler: () => (event: ResultEvent) => void }
     | { __nodeConfig: { argMapping: unknown[] } },
-  event: ResultEvent | ResultTouchEvent
+  event: ResultEvent
 ): void {
   if (!method) {
     return;
@@ -915,7 +913,7 @@ function invokeNullableMethod(
     }
 
     // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-    const nativeValue = event.nativeEvent[key];
+    const nativeValue = (event.nativeEvent as any)[key];
 
     // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
     if (value?.setValue) {
--- a/packages/react-native-gesture-handler/src/web/interfaces.ts
+++ b/packages/react-native-gesture-handler/src/web/interfaces.ts
@@ -5,8 +5,8 @@ import {
   TouchAction,
 } from '../handlers/gestureHandlerCommon';
 import { Directions } from '../Directions';
-import { State } from '../State';
 import { PointerType } from '../PointerType';
+import { GestureHandlerEvent } from '../v3/types';
 
 export interface HitSlop {
   left?: number;
@@ -81,17 +81,6 @@ export interface Config extends Record<string, ConfigArgs> {
   enableTrackpadTwoFingerGesture?: boolean;
 }
 
-type NativeEventArgs = number | State | boolean | undefined;
-interface NativeEvent extends Record<string, NativeEventArgs> {
-  numberOfPointers: number;
-  state: State;
-  pointerInside: boolean | undefined;
-  handlerTag: number;
-  target: number;
-  oldState?: State;
-  pointerType: PointerType;
-}
-
 export interface Point {
   x: number;
   y: number;
@@ -105,26 +94,9 @@ export interface PointerData {
   absoluteY: number;
 }
 
-type TouchNativeArgs = number | State | TouchEventType | PointerData[];
-
-interface NativeTouchEvent extends Record<string, TouchNativeArgs> {
-  handlerTag: number;
-  state: State;
-  eventType: TouchEventType;
-  changedTouches: PointerData[];
-  allTouches: PointerData[];
-  numberOfTouches: number;
-  pointerType: PointerType;
-}
-
-export interface ResultEvent extends Record<string, NativeEvent | number> {
-  nativeEvent: NativeEvent;
-  timeStamp: number;
-}
-
-export interface ResultTouchEvent
-  extends Record<string, NativeTouchEvent | number> {
-  nativeEvent: NativeTouchEvent;
+export interface ResultEvent
+  extends Record<string, GestureHandlerEvent<unknown> | number> {
+  nativeEvent: GestureHandlerEvent<unknown>;
   timeStamp: number;
 }
 

### CODE REVIEW

File: packages/react-native-gesture-handler/src/handlers/gestureHandlerCommon.ts [hunk 1]
```diff
@@ -165,6 +165,14 @@ export type GestureTouchEvent = {
   pointerType: PointerType;
 };
 
+export interface StylusData {
+  tiltX: number;
+  tiltY: number;
+  azimuthAngle: number;
+  altitudeAngle: number;
+  pressure: number;
```
akwasniewski, line 173: I'm not sure whether this is the right place for it,  but I had to put it somewhere else than web/interfaces , and this looked like the most promising spot.

m-bert: Why is that?

akwasniewski: Without moving `StylusData` there is a circular dependency. Moving this interface seemed the most sensible option to resolve it. `gestureHandlerCommon` contains common interfaces for all platforms.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -406,7 +405,6 @@ export default abstract class GestureHandler implements IGestureHandler {
       invokeNullableMethod(onGestureHandlerStateChange, resultEvent);
     }
     if (this.state === State.ACTIVE) {
-      resultEvent.nativeEvent.oldState = undefined;
```
m-bert, line 413: As far as I remember this was necessary in order to keep compatibility with old API. Right, @j-piasecki?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -420,24 +418,24 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+        handlerData: {
+          pointerType: this.pointerType,
+          numberOfPointers: this.tracker.trackedPointersCount,
+          pointerInside: this.delegate.isPointerInBounds(
+            this.tracker.getAbsoluteCoordsAverage()
+          ),
+          ...this.transformNativeEvent(),
+          target: this.viewRef,
+        },
+      } as StateChangeEvent<unknown> | UpdateEvent<unknown>,
```
m-bert: Correct me if I'm wrong, but this will break compatibility with V2 as all events will have `handlerData` field, right?

akwasniewski: My bad, should be fixed in [04ec35f](https://github.com/software-mansion/react-native-gesture-handler/pull/3651/commits/04ec35f299a8b3fcd128b1d4052e6f58d0f26af1)

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -416,23 +416,41 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+            pointerInside: this.delegate.isPointerInBounds(
+              this.tracker.getAbsoluteCoordsAverage()
+            ),
+            ...this.transformNativeEvent(),
+            target: this.viewRef,
+          },
+        } as StateChangeEvent<unknown> | UpdateEvent<unknown>,
+        timeStamp: Date.now(),
+      };
+    } else {
```
m-bert: You can either use ternary operator for conditional return `?:`, or remove this `else` statement as it doesn't do anything 😅

akwasniewski: true hah, fixed in [0c9692](https://github.com/software-mansion/react-native-gesture-handler/pull/3651/commits/0c96928df93b5acd690033fe0afdfadc465a4a3d)

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: I think this is the case referred in the comment above as 
> this may cause trouble in the future

Due to how new events are structures (no `oldState` field in update) we could either do a weird conditionality check as in [0c9692](https://github.com/software-mansion/react-native-gesture-handler/pull/3651/commits/0c96928df93b5acd690033fe0afdfadc) or change the new types so they all include 'oldState'. Both solutions seem off, do you maybe have any better ideas?

packages/react-native-gesture-handler/src/web/interfaces.ts
```diff
@@ -106,42 +109,20 @@ export interface PointerData {
...
-  nativeEvent: NativeTouchEvent;
+// Native event has to stay for v2 compatibility
+type ResultEventType = GestureHandlerEvent<unknown> | GestureHandlerNativeEvent;
+export interface ResultEvent<T extends ResultEventType = ResultEventType>
+  extends Record<string, T | number> {
+  nativeEvent: T;
   timeStamp: number;
 }
 
+// We need to leave any for v2 compatibility
```
m-bert: Which `any`?

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -262,4 +262,8 @@ export class GestureHandlerWebDelegate
   public set view(value: HTMLElement) {
     this._view = value;
   }
+
+  get initialized(): boolean {
+    return this.isInitialized;
+  }
```
m-bert: Wasn't it added in other PR?

packages/react-native-gesture-handler/src/web/interfaces.ts
```diff
@@ -106,42 +109,20 @@ export interface PointerData {
...
-}
-
-export interface ResultTouchEvent
-  extends Record<string, NativeTouchEvent | number> {
-  nativeEvent: NativeTouchEvent;
+// Native event has to stay for v2 compatibility
+type ResultEventType = GestureHandlerEvent<unknown> | GestureHandlerNativeEvent;
+export interface ResultEvent<T extends ResultEventType = ResultEventType>
+  extends Record<string, T | number> {
+  nativeEvent: T;
```
m-bert: Correct me if I'm wrong, but `GestureHandlerEvent` already has `nativeEvent` field (look [here](https://github.com/software-mansion/react-native-gesture-handler/blob/8e241748a1fa4017106d62335bfc94a86f3326ec/packages/react-native-gesture-handler/src/v3/types.ts#L18) for context). In that case, we will have event type where it is possible to have `nativeEvent.nativeEvent`

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -597,6 +650,12 @@ export default abstract class GestureHandler implements IGestureHandler {
     }
   }
 
+  private ensureViewRef(viewRef: any): asserts viewRef is number {
+    if (!viewRef) {
+      throw new Error(tagMessage('Cannot handle event when target is null'));
+    }
+  }
```
m-bert, line 662: Same here, little deja vu - wasn't it added in other PR?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -369,7 +370,7 @@ export default abstract class GestureHandler implements IGestureHandler {
     const { onGestureHandlerEvent, onGestureHandlerTouchEvent }: PropsRef =
       this.propsRef!.current;
 
-    const touchEvent: ResultTouchEvent | undefined =
+    const touchEvent: ResultEvent<TouchEvent> | undefined =
```
m-bert: Are you sure that `TouchEvent` will have correct fields and won't expose something that we do not provide?

packages/react-native-gesture-handler/src/web/interfaces.ts
m-bert: Also I find it quite confusing with this default value, wouldn't it be easier to read if we had just `ResultEvent<T>` and then pass either `GestureHandlerEvent`, `GestureHandlerNativeEvent` or `TouchEvent`?

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
akwasniewski: Yes, but it was no longer necessary so it was removed. I didn't notice it during manual merge so it remained here. Fixed [1cd5eef](https://github.com/software-mansion/react-native-gesture-handler/pull/3651/commits/1cd5eef70830978daedb0968a24258ca7105aaa1)

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Similar function, `ensurePropsRef` was added [here](https://github.com/software-mansion/react-native-gesture-handler/pull/3637/files#diff-5a8a40e22f71f0b86f855895a4b343fc54c10d3063bc268c86b79d206df44cbfR588)

packages/react-native-gesture-handler/src/web/interfaces.ts
akwasniewski: >Correct me if I'm wrong, but GestureHandlerEvent already has nativeEvent

No,  types in GestureHandlerEvent are wrappers on payload types defined [here](https://github.com/software-mansion/react-native-gesture-handler/blob/8e241748a1fa4017106d62335bfc94a86f3326ec/packages/react-native-gesture-handler/src/handlers/gestureHandlerCommon.ts#L49C1-L57C2), and only add `handlerData`. Those types do not have `nativeEvent` field

> Also I find it quite confusing with this default value

Ok, it was inspired by [this](https://github.com/software-mansion/react-native-gesture-handler/blob/8e241748a1fa4017106d62335bfc94a86f3326ec/packages/react-native-gesture-handler/src/handlers/gestureHandlerCommon.ts#L141C1-L143C2). I thought that it would simplify the code as we would not have to write `<unknown>` everywhere. I don't have a strong preference, if you think it's better to add `<unknown>` i can change it.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Yes, `gestureTouchEvent` defined [here](https://github.com/software-mansion/react-native-gesture-handler/blob/8e241748a1fa4017106d62335bfc94a86f3326ec/packages/react-native-gesture-handler/src/handlers/gestureHandlerCommon.ts#L158), on which `TouchEvent` is a union has the same fields as the old `nativeTouchEvent` defined within [web/interfaces](https://github.com/software-mansion/react-native-gesture-handler/blob/8e241748a1fa4017106d62335bfc94a86f3326ec/packages/react-native-gesture-handler/src/web/interfaces.ts#L111).

packages/react-native-gesture-handler/src/web/interfaces.ts
akwasniewski: I managed to fix the underlying issue, thus the comment became obsolete. Removed in [2e47b1ae](https://github.com/software-mansion/react-native-gesture-handler/pull/3651/commits/2e47b1ae42e71f5f368a3ee5b16159b6516b1b80).

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Now when I think about it we could probably do the same for `GestureHandlerNativeEvent`, is there a reason why v2 redeclared those types for the web, when we already have similar types in commons? 
EDIT: ok, nevermind they have conflicting type

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -439,9 +451,55 @@ export default abstract class GestureHandler implements IGestureHandler {
     };
   }
 
+  private transformStateChangeEvent(
+    newState: State,
+    oldState: State
+  ): ResultEvent<GestureStateChangeEventWithData<unknown>> {
+    this.ensureViewRef(this.viewRef);
```
m-bert, line 458: Just a cosmetic thing, but shouldn't it be checked before entering this function? I think semantically it would make more sense.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -439,9 +451,55 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+        },
+      },
+      timeStamp: Date.now(),
+    };
+  }
+
+  private transformUpdateEvent(
+    newState: State
+  ): ResultEvent<GestureUpdateEventWithData<unknown>> {
+    this.ensureViewRef(this.viewRef);
```
m-bert, line 481: Same as above

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: It's here to satisfy eslint. I guess I can move it to before the function and use `this.viewRef!`, it would work the same. But, I think we should avoid the ! operator if we can. I don't have a strong opinion, if you think it's better I will move it.

packages/react-native-gesture-handler/src/web/interfaces.ts
akwasniewski: >if you think it's better to add <unknown> i can change it.

What do you think I should do? I feel that current solution is stronger typed, what I think is better

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
m-bert: Not a strong opinion, so we can leave it.

packages/react-native-gesture-handler/src/web/interfaces.ts
```diff
@@ -83,7 +89,8 @@ export interface Config extends Record<string, ConfigArgs> {
 }
 
 type NativeEventArgs = number | State | boolean | undefined;
-interface NativeEvent extends Record<string, NativeEventArgs> {
+export interface GestureHandlerNativeEvent
+  extends Record<string, NativeEventArgs> {
   numberOfPointers: number;
   state: State;
```
akwasniewski, line 95: Why was it `State` instead of `ValueOf<typeof State>` as in types in `gestureHandlerCommons`? We might want to unify it and extend [`GestureEventPayload`](https://github.com/software-mansion/react-native-gesture-handler/blob/7ee21d97b8921e2406f06f08ca70dd06abeb0cf1/packages/react-native-gesture-handler/src/handlers/gestureHandlerCommon.ts#L49) @m-bert

m-bert: `ValueOf` is defined as follows:

```ts
export type ValueOf<T> = T[keyof T];
```

and type `State` like this:

```ts
export type State = (typeof State)[keyof typeof State];
```

So it is basically the same if you pass `ValueOf<typeof State>`

m-bert (approved):
LGTM

PR #3653, opened 2025-08-04
[Web] Adapt `NativeViewGestureHandler` to `NativeDetector`

akwasniewski:
## Description

Following #3638 this PR brings support for `NativeViewGestureHandler` into `NativeDetector` for web. It does so, by attaching handler into child instead of detector.

## Test plan

Tested on the following code 
```ts
import { StyleSheet, Text, View, ScrollView, Button } from 'react-native';
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,

  const items = Array.from({ length: 5 }, (_, index) => `Item ${index + 1}`);
  const [enabled, setEnabled] = React.useState(true);
    onBegin: (e) => {
      console.log('onBegin', e);
    },
    onStart: (e) => {
      console.log('onStart', e);
    }
  });

  const SV1 = () => (
    <ScrollView style={styles.scrollView1}>
      {items.map((item, index) => (
        </View>
      ))}
    </ScrollView>
  );

  const SV2 = () => (
    <ScrollView style={styles.scrollView2}>
      {items.map((item, index) => (
        </View>
      ))}
    </ScrollView>
  );

  return (
    <GestureHandlerRootView
      <Button
        title="Swap the child"
        onPress={() => {
          setEnabled(!enabled);
        }}
      />
        {enabled ? <SV1 /> : <SV2 />}
      </NativeDetector >
    </GestureHandlerRootView >
  );
}

  scrollView1: {
    backgroundColor: 'pink',
    marginHorizontal: 20,
  },
  scrollView2: {
    backgroundColor: 'lightblue',
    marginHorizontal: 20,
  },

  item: {
    flexDirection: 'row',
    justifyContent: 'center',
    alignItems: 'center',
    padding: 20,
    margin: 2,
    backgroundColor: 'white',
    borderRadius: 10,
  },
  text: {
    fontSize: 20,
    color: 'black',
  },
});
```


packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -14,12 +15,25 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
...
+  const attachedNativeHandlerTags = useRef<Set<number>>(new Set<number>());
+
+  const shouldAttachGestureToChildView = useCallback(
+    (handlerTag: number): boolean => {
+      return RNGestureHandlerModule.getGestureHandlerNode(
+        handlerTag
+      ).shouldAttachGestureToChildView();
+    },
+    []
+  );
```
m-bert: What is the benefit of using `useCallback` in this case?

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -3,6 +3,7 @@ import { View } from 'react-native';
 import RNGestureHandlerModule from '../RNGestureHandlerModule.web';
 import { ActionType } from '../ActionType';
 import { PropsRef } from '../web/interfaces';
+import { Wrap } from '../handlers/gestures/GestureDetector/Wrap';
```
m-bert: I don't like using `Wrap` from `GestureDetector` in v3. If the only reason is to pass `ref` to child, couldn't we clone it in-place?

Theoretically we could simply use `.ref` as it is standard prop, but I don't think I like this idea.

cc @j-piasecki

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -14,12 +15,25 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
...
+      ).shouldAttachGestureToChildView();
+    },
+    []
+  );
 
   const detachHandlers = useCallback((oldHandlerTags: Set<number>) => {
     oldHandlerTags.forEach((tag) => {
       RNGestureHandlerModule.detachGestureHandler(tag);
+      attachedNativeHandlerTags.current.delete(tag);
+      attachedHandlerTags.current.delete(tag);
```
m-bert, line 26: Given that `attachedHandlerTags` is not added in this PR, I think it should be included in the parent PR.

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -234,11 +232,7 @@ export class GestureHandlerWebDelegate
   }
 
   public get view() {
-    if (!this._view) {
-      throw new Error(tagMessage("component's view is null"));
-    }
-
-    return this._view;
+    return this._view!;
```
m-bert: Why is that changed?

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -33,19 +47,32 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
     detachHandlers(oldHandlerTags);
 
     newHandlerTags.forEach((tag) => {
+      const shouldAttachToChild = shouldAttachGestureToChildView(tag);
+      const attachTarget = shouldAttachToChild
+        ? childRef.current
+        : viewRef.current;
```
j-piasecki:
```suggestion
      const attachTarget = shouldAttachToChild
        ? viewRef.current.firstChild
        : viewRef.current;
```

I think `viewRef.current` is of type `HTMLDivElement` in web in this case, so we should be able to read the child (we may want to assert there is one).

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
j-piasecki: https://github.com/software-mansion/react-native-gesture-handler/pull/3653/files#r2254325526

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
akwasniewski: Native gesture broke without this change due to a bug, I patched this bug in the parent but didn't revert the quickfix, Fixed [13ab34cf](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/13ab34cf60b5ed73bc50819b737733f4091086d1)

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: It's not really needed in the parent PR. In the parent PR wesimply override the set with a new one, thus we don't have to manually track the changes. Manually deleting from the `attachedHandlerTags` set is only needed for reattaching native views.

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: Thanks! Tried to figure out the solution without wrapping, but I did not know that on web we could just specify the view to be a div. Fixed in https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/1f473cd0daa4c82d1057934d9697951bf7cea7ad

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: Ok, I figured to use `useCallback` just in case, but I dig into the docs and it is not the best practice. After I removed the dependencies it became redundant (I think). Therefore I deleted it from all functions in the component, both [here](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/cdfc5d5f2117e6809fff84669277f7f961f9649a) and in the [parent](https://github.com/software-mansion/react-native-gesture-handler/pull/3637/commits/b738254007553bdc866b683d9ddc46a989ab6d69).

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -13,13 +12,22 @@ export interface GestureHandlerDetectorProps extends PropsRef {
 const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
   const { handlerTags, children } = props;
 
-  const viewRef = useRef(null);
+  const viewRef = useRef<HTMLDivElement>(null);
```
m-bert: In theory it can also be `SVGElement`. Would it be ok to use just `Element` as type?

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -54,9 +75,9 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
   }, []);
 
   return (
-    <View style={{ display: 'contents' }} ref={viewRef}>
+    <div style={{ display: 'contents' }} ref={viewRef}>
```
m-bert: I know it works, but why do you switch to `div`?

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: As @j-piasecki suggested I used HTMLDivElement as only it has `firstChild` attribute and view on web is a div anyway. It allowed me to get rid of the wrapping. If it really can be somethng different than a div I might have to go back to wrapping. Is there a use case for an SVG element to have a native gesture? Maybe I can only typecast there, when accessing firstChild.

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: Explained in https://github.com/software-mansion/react-native-gesture-handler/pull/3653#discussion_r2256807763

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
m-bert: > Maybe I can only typecast there, when accessing firstChild.

If `firstChild` doesn't exist then type cast won't help - it won't work anyway 🤷 

> If it really can be something different than a div I might have to go back to wrapping.

As far as I remember `svg` tags inherit from `SVGElement`, you can check that 😅 

> Is there a use case for an SVG element to have a native gesture?

Probably not, this configuration doesn't sound reasonable to me (cc @j-piasecki)

akwasniewski: > If firstChild doesn't exist then type cast won't help - it won't work anyway

Yes, but if native gesture could only be on divs  I could possibly check the view type in 
```ts
  const shouldAttachGestureToChildView = (handlerTag: number): boolean => {
    return RNGestureHandlerModule.getGestureHandlerNode(
      handlerTag
    ).shouldAttachGestureToChildView() && viewRef.current instanceof HTMLDivElement;
  };
  ```
  and then the type cast would be safe

Comment (m-bert) [reply packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx]: Then you could probably use type guard and type cast wouldn't be necessary

Comment (akwasniewski) [reply packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx]: Indeed, after quite a bit of hassling with Eslint I added it in [b0ca85d9](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/b0ca85d9252ba3f69a310f2fa59a9c8111a09f54)

Comment (m-bert) [reply packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx]: I don't think that this method is the right place to make it a type guard. I was thinking more about 

```ts
  const isDivElement = (view: unknown): view is HTMLDivElement =>
    view instanceof HTMLDivElement;
```

Also

> I used HTMLDivElement as only it has firstChild attribute 

That's not true, `Element` inherits from `Node` and `firstChild` is property of `Node`, therefore `SVGElement` also has it.

akwasniewski: But now, if SVGElement also has this property then do we have to check it? All instances will have this property so we could simply cast to `Node`.

akwasniewski: Let me know what you think about this solution - [1e079d2](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/1e079d23f14cd64ccaf4f4e3f268d5a8bf94186a).

m-bert: I'd type it as `Element`, not `Node`. I think `Node` may be too wide. The rest is fine though 😅

akwasniewski: Great! Changed to `Element`  [16b2e5ef](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/16b2e5ef7aa928e116bf188defdcce551f6d4871)

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -13,13 +13,24 @@ export interface GestureHandlerDetectorProps extends PropsRef {
...
+  const viewRef = useRef<Element>(null);
   const propsRef = useRef<PropsRef>(props);
   const attachedHandlerTags = useRef<Set<number>>(new Set<number>());
+  const attachedNativeHandlerTags = useRef<Set<number>>(new Set<number>());
+
+  const shouldAttachGestureToChildView = (handlerTag: number): boolean => {
+    return (
+      RNGestureHandlerModule.getGestureHandlerNode(
+        handlerTag
+      ).shouldAttachGestureToChildView() && !!viewRef.current?.firstChild
```
j-piasecki:
```suggestion
      ).shouldAttachGestureToChildView()
```

I don't think we should be checking whether `firstChild` exists - later this method is used to determine whether to attach a gesture to the detector or its child. Even if the handler returns `shouldAttachGestureToChildView`, the gesture would be attached to the detector in case the child isn't there. I think we should assume the child is there, and throw if not.

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -13,13 +13,24 @@ export interface GestureHandlerDetectorProps extends PropsRef {
...
+    );
+  };
```
j-piasecki: And if this function no longer needs access to `viewRef`, it can be moved out of the component.

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -33,19 +44,33 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
...
+      } else {
+        RNGestureHandlerModule.attachGestureHandler(
+          tag,
+          viewRef.current,
+          ActionType.NATIVE_DETECTOR,
+          propsRef
+        );
+      }
     });
     attachedHandlerTags.current = currentHandlerTags;
```
j-piasecki, line 66: Should `attachedHandlerTags` overlap with `attachedNativeHandlerTags`?

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
j-piasecki: Ah, yeah - `Element` is the way to go. I forgot (again) that there exist elements other than divs 😅.

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: Ok, fixed in [01a7d654e](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/01a7d654e670c7fd91a643a20d493af109f754e8)

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: Yes, this was intended. It is a redundacy to decrease code complexity. `attachedNativeHandlerTags` is only used to detach handlers if the child changes (for them to be reattached to a new chilld).

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: I don't understand,  I think we still need `viewRef` in `attachHandlers`?

j-piasecki: I meant that `shouldAttachGestureToChildView` is currently defined inside the component, and thus, recreated on every render. Now, it only depends on `handlerTag`, which it receives as an argument so it can be moved outside of the component.

m-bert: Actually, why don't we simply do 

```jsx
const shouldAttachToChild = RNGestureHandlerModule.getGestureHandlerNode(
    handlerTag
  ).shouldAttachGestureToChildView();
```

and then use it in condition? Isn't wrapping it in another function and moving it to `utils` a bit too much?

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -33,19 +38,38 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
...
+        );
+        attachedNativeHandlerTags.current.add(tag);
+      } else {
+        RNGestureHandlerModule.attachGestureHandler(
+          tag,
+          viewRef.current,
+          ActionType.NATIVE_DETECTOR,
+          propsRef
+        );
+      }
```
m-bert: What about something like this? 

```jsx
const target = shouldAttachGestureToChildView(tag) ? viewRef.current.firstChild : viewRef.current;

if (!target) {
  throw new Error(tagMessage('Detector expected to have a child element'));
}

RNGestureHandlerModule.attachGestureHandler(
  tag,
  target,
  ActionType.NATIVE_DETECTOR,
  propsRef
);

```

packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
```diff
@@ -170,6 +170,10 @@ export default class NativeViewGestureHandler extends GestureHandler {
     return !this.disallowInterruption;
   }
 
+  public shouldAttachGestureToChildView(): boolean {
```
m-bert:
```suggestion
  public override shouldAttachGestureToChildView(): boolean {
```

i know we don't have it yet, we were a bit behind with typescript versions. I'm planning to add it, but in this case we can do it right away 😅

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: It is, my bad, fixed in [e26a0f5d](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/e26a0f5d1a3d78d19d908dc3aa1966058f8d9beb)

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
akwasniewski: Yeah, it used to look like [this](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/5a133217e9bd6e86634c6c2c8a5b30993cd290d1) but after I added `attachedNativeHandlerTags.current.add(tag);` it's not possible without additional calls to `shouldAttachGestueToChildView(tag)` (or memoising its value), so I decided it's not worth it. Do you think those redundancies are worth it to improve readability?

packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
akwasniewski: Hah okay, added in [94a3aca6f](https://github.com/software-mansion/react-native-gesture-handler/pull/3653/commits/94a3aca6faa2d4d02f3a54ff6d2493c645b617ae)

packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
m-bert: Oh right, I've missed `attachedNativeHandlerTags.current.add(tag);`. In that case I think we can leave it.

m-bert (approved):
LGTM

PR #3658, opened 2025-08-05
Bind `SharedValues` in handler config

m-bert:
## Description

This PR introduces `SharedValue` bindings to config. This will allow users to specify values in config as`SharedValues`, e.g:

```tsx
const taps = useSharedValue(2);

const gesture = useGesture('TapGestureHandler', {
  onEnd: () => {
    console.log('Tap detected. Required number of taps:', taps.value);
    taps.set(taps.value + 1);
  },
  numberOfTaps: taps,
});
```

## Test plan

<details>

```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,

  const taps = useSharedValue(2);

  const gesture = useGesture('TapGestureHandler', {
    onEnd: () => {
      console.log('Tap detected. Required number of taps:', taps.value);
      taps.set(taps.value + 1);
    },
    numberOfTaps: taps,
  });

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>


packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -175,13 +175,19 @@ - (void)attachGestureHandler:(double)handlerTag newView:(double)viewTag actionTy
...
   [self addOperationBlock:^(RNGestureHandlerManager *manager) {
-    [manager updateGestureHandler:[NSNumber numberWithDouble:handlerTag] config:config];
+    [manager setGestureHandlerConfig:[NSNumber numberWithDouble:handlerTag] config:config];
   }];
 }
 
+- (void)updateGestureHandlerConfig:(double)handlerTag newConfig:(NSDictionary *)config
+{
+  RNGestureHandlerManager *manager = [RNGestureHandlerModule handlerManagerForModuleId:_moduleId];
+  [manager updateGestureHandlerConfig:[NSNumber numberWithDouble:handlerTag] config:config];
```
m-bert: I wasn't sure if this should be done like this, or like in `setGestureHandlerConfig`. However, if we use `addOperationBlock` we have to call `flushOperations`, so I think we can leave it as it is (cc @j-piasecki)

m-bert:
Also let me know if **_web_** should be updated in this PR or not (cc @j-piasecki) 😅

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -90,19 +90,36 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
...
+
+    updateGestureHandlerHelper<GestureHandler>(handlerTag, config, false)
+  }
+
+  @ReactMethod
+  override fun updateGestureHandlerConfig(handlerTagDouble: Double, config: ReadableMap) {
     val handlerTag = handlerTagDouble.toInt()
 
-    updateGestureHandlerHelper<GestureHandler>(handlerTag, config)
+    updateGestureHandlerHelper<GestureHandler>(handlerTag, config, true)
```
j-piasecki: I think `updateGestureHandlerHelper` can be inlined in both places. Now that handlers are no longer generic classes, the helper doesn't add much, and the result should be easier to read.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -94,18 +160,21 @@ export function useGesture(
...
   useEffect(() => {
+    maybeExtractSharedValues(config, tag);
+
     return () => {
+      maybeDetachSharedValues();
       RNGestureHandlerModule.dropGestureHandler(tag);
       RNGestureHandlerModule.flushOperations();
     };
-  }, [type, tag]);
+  }, [type, tag, config]);
```
j-piasecki: This will cause the gesture to be dropped when the config object is changed. Shared value binding should be handled by a separate effect.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
   );
 }
 
+const sharedValues = new Map<
+  string,
+  { id: number; sharedValue: SharedValue }
+>();
```
j-piasecki: I don't think there's a need to store shared values in a map.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+
+let nextSharedValueListenerID = 0;
```
j-piasecki: Reanimated also numbers from 0, we need some kind of offset - pick your favorite, but not obvious (rational or rounded irrational) number. We shouldn't need that counter anyway.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+
+  for (const [key, maybeSharedValue] of Object.entries(config)) {
+    if (!Reanimated.isSharedValue(maybeSharedValue)) {
+      continue;
+    }
+
+    sharedValues.set(key, {
+      id: nextSharedValueListenerID,
+      sharedValue: maybeSharedValue,
+    });
```
```suggestion
```

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+function maybeExtractSharedValues(config: any, tag: number) {
+  if (Reanimated === undefined) {
+    return;
+  }
+
+  const attachListener = (
+    sharedValue: SharedValue,
+    id: number,
+    configKey: string
+  ) => {
```
```suggestion
  const listenerId = tag + MAGIC_OFFSET;

  const attachListener = (
    sharedValue: SharedValue,
    configKey: string
  ) => {
```

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+    sharedValues.set(key, {
+      id: nextSharedValueListenerID,
+      sharedValue: maybeSharedValue,
+    });
+
+    config[key] = maybeSharedValue.value;
+
+    Reanimated.runOnUI(attachListener)(
+      maybeSharedValue,
+      nextSharedValueListenerID,
```
```suggestion
```

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+      key
+    );
+
+    nextSharedValueListenerID++;
```
```suggestion
```

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+    return;
+  }
+
+  for (const [key, { id, sharedValue }] of sharedValues.entries()) {
+    Reanimated.runOnUI(() => {
+      sharedValue.removeListener(id);
+    })();
+
+    sharedValues.delete(key);
+  }
```
```suggestion
  for (const [key, maybeSharedValue] of Object.entries(config)) {
    if (!Reanimated.isSharedValue(maybeSharedValue)) {
      continue;
    }

    Reanimated.runOnUI(() => {
      maybeSharedValue.removeListener(handlerTag + MAGIC_OFFSET);
    })();
  }
```

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+  string,
+  { id: number; sharedValue: SharedValue }
+>();
+
+let nextSharedValueListenerID = 0;
+
+// This is used to obtain HostFunction that can be executed on the UI thread
+const { updateGestureHandlerConfig } = RNGestureHandlerModule;
+
+function maybeExtractSharedValues(config: any, tag: number) {
```
```suggestion
function bindSharedValues(config: any, tag: number) {
```

Imo, we don't need `maybe` prefix here. If there's no Reanimated, there will be no shared values - every shared value would be bound. Same, if there's Reanimated but no shared values in the config. And if there's Reanimated and shared values, they will be bound correctly.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,69 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+      maybeSharedValue,
+      nextSharedValueListenerID,
+      key
+    );
+
+    nextSharedValueListenerID++;
+  }
+}
+
+function maybeDetachSharedValues() {
```
```suggestion
function unbindSharedValues() {
```

Same

m-bert: Done in [fd6eb79](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/fd6eb79354defaff8a91f5ec0f19bd3008ec0ff2)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: Sure, done in [fd6eb79](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/fd6eb79354defaff8a91f5ec0f19bd3008ec0ff2)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
m-bert: I agree, done in [50ffbcd](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/50ffbcdc66e3bfc1b775eea3208fd31bd1faf3b0)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: Good catch, changed in [8d1ad9b](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/8d1ad9b4455eb957f9cb646f826526a12e0ac0e7)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: Done in [2af12c1](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/2af12c12e6b635c681d7d57a9e862c33771330fa)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: >Reanimated also numbers from 0

Reanimated starts from [9999](https://github.com/software-mansion/react-native-reanimated/blob/10f035676c15a37cd0d321f9fa9e37d3e9c04126/packages/react-native-reanimated/src/mappers.ts#L201) 😅 But I do agree that we may want to use offset. 

Also, do we need to use `tag + offset`? If each `shared value` has its own map of listeners, it should be fine to use just `offset` and then we could simply call it `listenerId`.

Changed in [9a98e19](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/9a98e197adde375e9f63d51f09fb3574aa304800), though this commit has `SharedValues` in array. It was removed later on in [2af12c1](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/2af12c12e6b635c681d7d57a9e862c33771330fa).

j-piasecki: > Also, do we need to use tag + offset? If each shared value has its own map of listeners, it should be fine to use just offset and then we could simply call it listenerId.

I think this breaks when the same shared value is used in multiple configs.

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
j-piasecki: I guess we can add it for consistency now and revisit the entire scheduling and flushing soon, now that the old arch support is gone.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -46,6 +49,54 @@ function shouldHandleTouchEvents(config: Record<string, unknown>) {
...
+      updateGestureHandlerConfig(tag, { [configKey]: value });
+    });
+  };
+
+  for (const [key, maybeSharedValue] of Object.entries(config)) {
+    if (!Reanimated.isSharedValue(maybeSharedValue)) {
+      continue;
+    }
+
+    config[key] = maybeSharedValue.value;
```
j-piasecki: Doesn't this break the unbinding? This looks like it overwrites the shared value in the config with its value. If so, isn't `unbindSharedValues` doing effectively nothing?

m-bert: Yes, it does, that's why I had used `Map` before. I've changed it to keep `SharedValue` in config and create copy before we pass it to the native side. Done in [ee63683](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/ee6368303ca4eeb4d9ae28c24e03856b33dd490a)

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
m-bert: Okay, I agree. Done in [73bd21b](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/73bd21b1aadb92e1d8c5d878c3a52333a20d637d), then moved `flashOperations` into JS in [ed5e9d8](https://github.com/software-mansion/react-native-gesture-handler/pull/3658/commits/ed5e9d8c5e5feae893b9d4f27d40d8d260ee07dd).

PR #3602, opened 2025-07-05
fix: turn off build IDs for reproducibility

Rexogamer:
## Description

see https://gitlab.com/IzzyOnDroid/repo/-/wikis/Reproducible-Builds/RB-Hints-for-Developers#no-funny-build-time-generated-ids for context.

## Test plan

I've successfully used these changes for a little while on my own app.

PR #3659, opened 2025-08-06
[Android] Move linker flags to cmake

m-bert:
## Description

This PR moves build id flags from `build.gradle` to `CMakeLists.txt`

## Test plan

Checked that basic-example builds


m-bert:
Closed in favor of reverting #3602, as we want to stay consistent with `react-native`.

PR #3664, opened 2025-08-07
Implement gesture relations

m-bert:
## Test plan

Not yet 🤷 

<details>

```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  useSimultaneous,
  useGesture,

  const event = Animated.event(
    [{ nativeEvent: { handlerData: { translationX: value } } }],
    {
      useNativeDriver: true,
    }
  );

  const pan1 = useGesture('PanGestureHandler', {
    onUpdate: () => console.log('Pan 1'),
  });

  const pan2 = useGesture('PanGestureHandler', {
    onUpdate: () => console.log('Pan 2'),
  });

  const composedGesture = useSimultaneous(pan2, pan1);

  console.log(composedGesture);
  console.log(pan1.config, pan2.config);

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
        <NativeDetector gesture={composedGesture}>
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/index.ts
+++ b/packages/react-native-gesture-handler/src/index.ts
@@ -166,5 +166,6 @@ export type { NativeDetectorProps } from './v3/NativeDetector';
 export { NativeDetector } from './v3/NativeDetector';
 
 export * from './v3/hooks/useGesture';
+export * from './v3/hooks/relations/useSimultaneous';
 
 initialize();
--- a/packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { NativeGesture } from './hooks/useGesture';
+import { NativeGesture } from './types';
 import { Reanimated } from '../handlers/gestures/reanimatedWrapper';
 
 import { Animated, StyleSheet } from 'react-native';
@@ -49,7 +49,7 @@ export function NativeDetector({ gesture, children }: NativeDetectorProps) {
         gesture.gestureEvents.onGestureHandlerTouchEvent
       }
       moduleId={globalThis._RNGH_MODULE_ID}
-      handlerTags={[gesture.tag]}
+      handlerTags={gesture.tag}
       style={styles.detector}>
       {children}
     </NativeDetectorComponent>
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
@@ -7,32 +7,7 @@ import {
   SharedValue,
 } from '../../handlers/gestures/reanimatedWrapper';
 import { tagMessage } from '../../utils';
-import { AnimatedEvent } from '../types';
-
-type GestureType =
-  | 'TapGestureHandler'
-  | 'LongPressGestureHandler'
-  | 'PanGestureHandler'
-  | 'PinchGestureHandler'
-  | 'RotationGestureHandler'
-  | 'FlingGestureHandler'
-  | 'ForceTouchGestureHandler'
-  | 'ManualGestureHandler'
-  | 'NativeViewGestureHandler';
-
-type GestureEvents = {
-  onGestureHandlerStateChange: (event: any) => void;
-  onGestureHandlerEvent: undefined | ((event: any) => void);
-  onGestureHandlerTouchEvent: (event: any) => void;
-  onGestureHandlerAnimatedEvent: undefined | AnimatedEvent;
-};
-
-export interface NativeGesture {
-  tag: number;
-  name: GestureType;
-  config: Record<string, unknown>;
-  gestureEvents: GestureEvents;
-}
+import { GestureType, NativeGesture } from '../types';
 
 function hasWorkletEventHandlers(config: Record<string, unknown>) {
   return Object.values(config).some(
@@ -181,7 +156,7 @@ export function useGesture(
   }, [config, tag]);
 
   return {
-    tag: tag,
+    tag: [tag],
     name: type,
     config,
     gestureEvents: {
--- a/packages/react-native-gesture-handler/src/v3/types.ts
+++ b/packages/react-native-gesture-handler/src/v3/types.ts
@@ -48,3 +48,28 @@ export type CallbackHandlers = Omit<
 export type AnimatedEvent = ((...args: any[]) => void) & {
   _argMapping?: unknown;
 };
+
+export type GestureType =
+  | 'TapGestureHandler'
+  | 'LongPressGestureHandler'
+  | 'PanGestureHandler'
+  | 'PinchGestureHandler'
+  | 'RotationGestureHandler'
+  | 'FlingGestureHandler'
+  | 'ForceTouchGestureHandler'
+  | 'ManualGestureHandler'
+  | 'NativeViewGestureHandler';
+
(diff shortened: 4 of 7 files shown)


packages/react-native-gesture-handler/src/v3/hooks/relations/useExclusive.ts
```diff
@@ -0,0 +1,16 @@
...
+import { useComposedGesture } from './useComposedGesture';
+
+export function useExclusive(...gestures: NativeGesture[]) {
+  const composedGesture = useComposedGesture(...gestures);
+
+  const tags = gestures.flatMap((gesture) => gesture.tag);
+
+  for (let i = 0; i < gestures.length; i++) {
+    gestures[i].config.waitFor = tags.slice(0, i);
+  }
```
akwasniewski: Storing and checking every dependency seems like a waste of memory and time. If you use a bidirectional linked list you can check dependencies in $O(1)$ time and $O(n)$ memory. Currently we have $O(n)$ time and $\Theta\left(\frac{n^2-n}{2}\right)$ memory. State could then only activate if it's first in the list - when its `prevReference` is null. When a state is failed you simply remove it from the linked list. In theory there is no cap on number of gestures, thus I think it might be worth it to optimise.

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
```diff
@@ -34,20 +35,130 @@ export function NativeDetector({ gesture, children }: NativeDetectorProps) {
...
+    node.gestures.forEach((child) => {
+      // If child is composed gesture, we have to correctly fill `waitFor` and `simultaneousHandlers`.
+      if (isComposedGesture(child)) {
+        // We have to update `simultaneousHandlers` before traversing the child.
+
+        // If we go from a non-simultaneous gesture to a simultaneous gesture,
+        // we add the tags of the simultaneous gesture to the `simultaneousHandlers`.
+        // This way when we traverse the child, we already have the tags of the simultaneous gestures
+        if (
+          node.name !== 'SimultaneousGesture' &&
```
akwasniewski: Why determining the type is not done with an Enum? Comparing strings seems wasteful

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
```diff
@@ -34,20 +35,130 @@ export function NativeDetector({ gesture, children }: NativeDetectorProps) {
...
+  // arrays for each gesture. It traverses the tree recursively using DFS.
+  // `waitFor` and `simultaneousHandlers` are global data structures that will be populated into each gesture.
+  // For `waitFor` we need array as order of the gestures matters.
+  // For `simultaneousHandlers` we use Set as the order doesn't matter.
+  // The tree consists of ComposedGestures and NativeGestures. NativeGestures are always leaf nodes.
+  const dfs = (
+    node: NativeGesture | ComposedGesture,
+    waitFor: number[] = [],
+    simultaneousHandlers: Set<number> = new Set()
+  ) => {
```
akwasniewski: We might want to extract this function logic to a seperate file as it will be reused on web.

packages/react-native-gesture-handler/src/index.ts
```diff
@@ -166,5 +166,8 @@ export type { NativeDetectorProps } from './v3/NativeDetector';
 export { NativeDetector } from './v3/NativeDetector';
 
 export * from './v3/hooks/useGesture';
+export * from './v3/hooks/relations/useSimultaneous';
+export * from './v3/hooks/relations/useExclusive';
+export * from './v3/hooks/relations/useRace';
```
akwasniewski, line 171: If the `useSimultanous`/`useExclusive`/`useRace` only differ by assigning different name, is it necessary to have a separate file for them? We could export it right from `useComposed`.

packages/react-native-gesture-handler/src/v3/NativeDetector.tsx
m-bert: It will be, I remind you that it is still a draft PR 😉

packages/react-native-gesture-handler/src/index.ts
m-bert: I've not finished this implementation yet so I'm not sure if this will be the only difference between them 😉

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -48,3 +49,64 @@ export type CallbackHandlers = Omit<
...
+  Fling: 'FlingGestureHandler',
+  Manual: 'ManualGestureHandler',
+  Native: 'NativeGestureHandler',
+} as const;
+
+export const ComposedGestureType = {
+  Simultaneous: 'SimultaneousGesture',
+  Exclusive: 'ExclusiveGesture',
+  Race: 'RaceGesture',
+} as const;
```
j-piasecki: These could also use a different name, `Type` suffix is a bit misleading.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -97,8 +80,39 @@ function unbindSharedValues(config: any, tag: number) {
...
+      config.simultaneousWithExternalGesture
+    ),
+    waitFor: extractHandlerTags(config.requireExternalGestureToFail),
+    blocksHandlers: extractHandlerTags(config.blocksExternalGesture),
+  };
+}
+
 export function useGesture(
-  type: GestureType,
+  type: ValueOf<typeof HandlerType>,
```
j-piasecki: Is there a scenario where this is `ComposedGestureType`?

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+// `waitFor` and `simultaneousHandlers` are global data structures that will be populated into each gesture.
+// For `waitFor` we need array as order of the gestures matters.
+// For `simultaneousHandlers` we use Set as the order doesn't matter.
+
+import RNGestureHandlerModule from '../../RNGestureHandlerModule';
+import { isComposedGesture } from '../hooks/utils';
+import { ComposedGesture, ComposedGestureType, NativeGesture } from '../types';
+
+// The tree consists of ComposedGestures and NativeGestures. NativeGestures are always leaf nodes.
+export const dfs = (
```
j-piasecki, line 12: This could also use a more descriptive name.

m-bert: Yes, I know, I left it for now as there are some other things to do 😅

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
m-bert: I don't think so, good catch.

m-bert:
>[!CAUTION]
> - Suppressed by #3693

j-piasecki:
> Suppressed by https://github.com/software-mansion/react-native-gesture-handler/pull/3693

I think the word you're looking for is `superseded`

Issue #3643, opened 2025-07-26
ReanimatedSwipeable's Left Swipeable Card is not working after Expo SDK53 update


farhadjaman:
### Description

After upgrading from Expo SDK 52 to Expo SDK 53, the left-action button inside my ReanimatedSwipeable (from react-native-gesture-handler) no longer receives press events on iOS—unless the New Architecture (Fabric) is enabled. With Fabric disabled, tapping the left action button does not invoke its onPress handler at all, regardless of the button’s width. I have disabled the new architecture because of some UI related issues

https://github.com/user-attachments/assets/f36be0b9-a225-4b93-82f6-157879fe88b7

### Steps to reproduce

1. Create a ReanimatedSwipeable with a left action button:
2. Disable the New Architecture (Fabric) in iOS/Podfile or via Expo config plugin.
3. Make sure you are using the new EXPO SDK53
4. Swipe right to reveal the red “Delete” button.
5. Tap anywhere on the red “Delete” button.

### A link to a [Gist](https://gist.github.com/), an [Expo Snack](https://snack.expo.io/) or a link to a repository based on [this template](https://github.com/react-native-community/reproducer-react-native) that reproduces the bug.

https://gist.github.com/farhadjaman/46e6a0d10bb2d181b54104ad6524ec98

### Gesture Handler version

2.24.0

### React Native version

0.79.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

Using Expo Prebuild or an Expo development build

### Architecture

Old Architecture (Paper)

### Build type

Debug mode

### Device

Real device

### Acknowledgements

Yes

RalissonMattias:
Here too! In my case it breaks in both directions and throws an error. Obs: I'm not using Expo

React Native CLI: 0.79.5
Reanimated: 4.0.1
Gesture Handler: 2.27.2
New Arch: ON

<img width="415" height="841" alt="Image" src="https://github.com/user-attachments/assets/f2a0ef5e-19c6-42be-b6da-ad973e815844" />

deepktp:
@RalissonMattias 

i had this issue fixed with an simple patch by changing import path from `.` to `./index`

issue was on `react-native-gesture-handler: "2.28.0"`  with `react-native: "81.4"`

got to file `node_modules/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx`
```diff
[...]

Issue #3665, opened 2025-08-10
ReanimatedError: Cannot read property 'LEFT' of undefined from SwipeDirection enum after upgrade


christianbach:
### Description

When upgrading react-native-gesture-handler from 2.25.0 to 2.28.0 on React Native 0.80, our app crashes with the following error:

```ReanimatedError: Cannot read property 'LEFT' of undefined, js engine: reanimated. Cannot read LEFT of undefined```

The error points to the TypeScript enum:
```typescript
export enum SwipeDirection {
  LEFT = 'left',
  RIGHT = 'right',
}
```

This enum usage is standard and works fine elsewhere in our codebase. After upgrading, it fails specifically in this context, referencing the SwipeDirection enum. Enums work as expected elsewhere.

The only thing that I can think of is that we use `unstable_enablePackageExports` and `unstable_conditionNames ` in our `metro.config`

```javascript
resolver: {
    assetExts: assetExts.filter(ext => ext !== 'svg'),
    sourceExts: [...sourceExts, 'svg', 'mjs', 'cjs'],
    // disable watchman for now, its just too slow and unstable
    useWatchman: false,
    // This needs to be here for react-strict-dom to work
    unstable_enablePackageExports: true,
    unstable_conditionNames: ['react-native', 'require', 'default'],
  },
```

### Steps to reproduce

1. Upgrade react-native-gesture-handler from 2.25.0 to 2.28.0
2. Use `react-native-gesture-handler/ReanimatedSwipeable` 

### A link to a [Gist](https://gist.github.com/), an [Expo Snack](https://snack.expo.io/) or a link to a repository based on [this template](https://github.com/react-native-community/reproducer-react-native) that reproduces the bug.

none

### Gesture Handler version

2.28.0 (upgraded from 2.25.0)

### React Native version

0.80

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Old Architecture (Paper)

### Build type

Debug mode

### Device

iOS simulator

### Acknowledgements

Yes

christianbach:
Can't be reproduced with a minimal bare react-native project: https://github.com/christianbach/react-native-gesture-handler-swipable-enum

gboudreau:
I am having the same issue. On both iOS and Android. Works fine in 2.26.0, broken in 2.27.2.

danielcrk:
Same issue here as well. RNGH `2.28.0`.

[...]

PR #3666, opened 2025-08-11
fix: ensure SwipeDirection is imported as a value in ReanimatedSwipeable

christianbach:
**Summary**
Fixes `SwipeDirection` being `undefined` at runtime in `ReanimatedSwipeable` when the package
is consumed from TypeScript source (e.g. in some Metro configs).

**Details**
`SwipeDirection` was imported from the local barrel (`./index.ts`), which re-exports it
as a type-only export. This removes the value in the compiled JS, causing it to be `undefined`
at runtime.

This PR imports `SwipeDirection` directly from `ReanimatedSwipeableProps`, which exports it
as a value.

See closed issue for details: [#3665](https://github.com/software-mansion/react-native-gesture-handler/issues/3665)

**Impact**
No API changes. Only affects internal import, ensures enum value is present at runtime.


christianbach:
Perhaps @latekvo could take a look?

christianbach:
Forced push a new commit that fixes the lint error

latekvo (commented):
Hey @christianbach, before I merge this I'd like to make sure: have you tested this fix on your project, and did it resolve your issues?

Your explanation for why this fix works does not seem to be true.
You're writing that `./index.ts [...] re-exports it as a type-only export `, but that's not the case, `SwipeDirection` is exported as a value within the `index.ts`.

Regardless, if the fix you're proposing does fix this issue, I have nothing against merging it. Especially since there was another person reporting a similar problem.

Perhaps one last thing, is that if this issue occurs in `ReanimatedSwipeable`, then it'd be worth checking other components for a similar issue. Would you mind checking if you're encountering similar issues with the `Pressable` or the `ReanimatedDrawerLayout`? Both export numerous `Enums`. 

(c.c. @j-piasecki)

christianbach:
> Your explanation for why this fix works does not seem to be true. You're writing that `./index.ts [...] re-exports it as a type-only export `, but that's not the case, `SwipeDirection` is exported as a value within the `index.ts`.

You're absolutely right, my bad! 

I did some more digging and this is really strange, if I change the import from:
```javascript
import { SwipeableProps, SwipeableMethods, SwipeDirection } from '.';
```
to:
```javascript
import { SwipeableProps, SwipeableMethods, SwipeDirection, } from './index';
```

Everything starts working as expected. It seems to me that metro is somehow striping `SwipeDirection` as a type or omitting the `index.ts` file entirely...

latekvo:
Hey, would you mind sharing your `metro.config.js` and `tsconfig.json` files? These could contain some useful insights.

christianbach:
Nothing out of the ordinary, we don't use the default tsconfig from `react-native` but it's similar.

`metro.config.js`

```javascript
const path = require('path');
const fs = require('fs');
const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');
const {withSentryConfig} = require('@sentry/react-native/metro');
const {
  wrapWithReanimatedMetroConfig,
} = require('react-native-reanimated/metro-config');
const defaultConfig = getDefaultConfig(__dirname);
const {assetExts, sourceExts} = defaultConfig.resolver;

/**
 * Metro configuration
 * https://facebook.github.io/metro/docs/configuration
 *
 * @type {import('@react-native/metro-config').MetroConfig}
 */

// Exclude devDependencies from watchFolders, include everything else
const excludesPackages = ['eslint-config', 'tsconfig', 'prettier-config'];
const watchFolders = fs
  .readdirSync(path.resolve(__dirname, '../../packages'))
  .filter(
    folder =>
      !excludesPackages.some(
        exclude => folder.includes(exclude) || folder.startsWith('.'),
      ),
  )
  .filter(Boolean)
  .map(folder => path.resolve(__dirname, `../../packages/${folder}`));

const config = {
  watchFolders: [
    path.resolve(__dirname, '../../node_modules'),
    ...watchFolders,
  ],

  transformer: {
    babelTransformerPath: require.resolve(
      'react-native-svg-transformer/react-native',
    ),
  },

  resolver: {
    assetExts: assetExts.filter(ext => ext !== 'svg'),
    sourceExts: [...sourceExts, 'svg', 'mjs', 'cjs'],
    // disable watchman for now, its just too slow and unstable
    useWatchman: false,
    // This needs to be here for react-strict-dom to work
    unstable_enablePackageExports: true,
    unstable_conditionNames: ['react-native', 'require', 'default'],
  },
};

const mergedConfig = mergeConfig(defaultConfig, config);

module.exports = withSentryConfig(wrapWithReanimatedMetroConfig(mergedConfig));
```

`tsconfig.json`

```json
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "display": "React Native",
  "compilerOptions": {
    "target": "esnext",
    "module": "esnext",
    "types": ["react-native", "jest"],
    "lib": [
      "es2019",
      "es2020.bigint",
      "es2020.date",
      "es2020.number",
      "es2020.promise",
      "es2020.string",
      "es2020.symbol.wellknown",
      "es2021.promise",
      "es2021.string",
      "es2021.weakref",
      "es2022.array",
      "es2022.object",
      "es2022.string"
    ],
    "allowJs": true,
    "jsx": "react-native",
    "noEmit": true,
    "isolatedModules": true,
    "strict": true,
    "incremental": true,
    "moduleResolution": "bundler",
    "customConditions": ["react-native"],
    "moduleSuffixes": [".native", ""],
    "allowImportingTsExtensions": true,
    "allowArbitraryExtensions": true,
    "resolveJsonModule": true,
    "resolvePackageJsonImports": false,
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": false,
    "strictFunctionTypes": true
  },
  "exclude": ["**/node_modules", "**/Pods", "**/ios", "**/android"]
}
```

latekvo:
Hey @christianbach, apologies for getting back to you so late.
Could you please go ahead and commit the `'.' -> './index'` fix? You can also remove the other import changes you've made so far, i don't think they're necessary once that fix is applied.
There's been other reports of that patch working, so we'll go ahead and merge it.

latekvo:
This also solves this Issue: #3720, neat

Here's a consistant repro that works on iOS: [link](https://github.com/PrivioSupport/rn-reanimated-gesture-repro)

christianbach:
> This also solves this Issue: #3720, neat
> 
> Here's a consistant repro that works on iOS: [link](https://github.com/user-attachments/assets/f6139d7a-3b81-4469-9334-5c369016b8b6)

Sure, I will make time for this tomorrow.

christianbach:
@latekvo removed the type import and imported everything at once.

Vashiru:
From my testing this doesn't seem to resolve #3643. I tried it with and without temporary solution from https://github.com/software-mansion/react-native-gesture-handler/issues/3643#issuecomment-3343123344, but to no avail.

naxey:
I second this, patching the import with `./index` or `./ReanimatedSwipeableProps` doesn't seem to resolve #3720

latekvo:
@Vashiru & @naxey, I'm aware. This PR resolves #3720, which is a different bug, and for now I'm still working on resolving your issue (#3643).

latekvo (approved):
Alright, merging. Thank you again for contributing.

AndreMaz:
@latekvo any ETA for a release containing this fix?

PR #3667, opened 2025-08-11
Add `override` keyword to `TypeScript` part of codebase

m-bert:
## Description

This PR adds `override` keyword to part of our codebase that is written in `TypeScript`. I think it will improve readability and some cases will be easier to understand.

It also enables `noImplicitOverride` flag in ts compiler, which means that `override` keyword is mandatory when overriding a method. 

## Test plan

Run example app


akwasniewski (approved):
I love it, looks good.

PR #3669, opened 2025-08-11
Fix gesture detach on component drop

akwasniewski:
## Description

Gestures were  not detached properly on Android on component drop.

### Android component drop
Add a logging message in attach/detach in `android/.../react/RNGestureHandlerRegistry.kt`
```ts

import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,

    onBegin: (e: any) => {
      console.log('onGestureHandlerEvent', e.nativeEvent.handlerData);
    },
    onUpdate: (e: any) => {
      const { translationX, translationY } = e.nativeEvent.handlerData;
      const distance = Math.sqrt(translationX ** 2 + translationY ** 2);
      if (distance > 100) {
        setVisible(false);
      }
    },
  });

  return (
    <GestureHandlerRootView
      style={{ flex: 1, backgroundColor: 'white', paddingTop: 100 }}>
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={{
              width: 150,
              height: 150,
              backgroundColor: 'blue',
              opacity: 0.5,
              borderWidth: 10,
              borderColor: 'red',
              marginTop: 20,
              marginLeft: 40,
            }}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```


BoomchainLabs (approved):

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -124,6 +124,15 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
     eventDispatcher?.dispatchEvent(event)
   }
 
+  fun onViewDrop() {
+    val registry = RNGestureHandlerModule.registries[moduleId]
+      ?: throw Exception("Tried to access a non-existent registry")
+    for (tag in attachedHandlers) {
```
m-bert:
```suggestion
      ?: throw Exception("Tried to access a non-existent registry")
      
    for (tag in attachedHandlers) {
```

m-bert (commented):
Shouldn't iOS have the same logic?

packages/react-native-gesture-handler/android/kls_database.db
m-bert, line 1: Is this file needed?

akwasniewski:
> Shouldn't iOS have the same logic?

The issue doesn't occur on iOS. Gestures are properly detached in `willMoveToWindow`.

PR #3672, opened 2025-08-12
Fix native gesture reattach

akwasniewski:
## Description

In `NativeDetector` native gesture failed to reattach when child changed. 

## Test plan

Add a logging message in attach/detach in `android/.../react/RNGestureHandlerRegistry.kt` and `apple/RNGestureHandlerRegistry.m`
`
```ts
import { StyleSheet, Text, View, ScrollView, Button } from 'react-native';
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,

  const items = Array.from({ length: 5 }, (_, index) => `Item ${index + 1}`);
  const [enabled, setEnabled] = React.useState(true);
    onBegin: (e) => {
      console.log('onBegin');
    },
    onStart: (e) => {
      console.log('onStart');
    }
  });

  const SV1 = () => (
    <ScrollView style={styles.scrollView1}>
      {items.map((item, index) => (
        </View>
      ))}
    </ScrollView>
  );

  const SV2 = () => (
    <ScrollView style={styles.scrollView2}>
      {items.map((item, index) => (
        </View>
      ))}
    </ScrollView>
  );

  return (
    <GestureHandlerRootView
      style={{ flex: 1, backgroundColor: 'white', paddingTop: 100 }}>
      <Button
        title="Swap the child"
        onPress={() => {
          setEnabled(!enabled);
        }}
      />
        {enabled ? <SV1 /> : <SV2 />}
      </NativeDetector >
    </GestureHandlerRootView >
  );
}

  scrollView1: {
    backgroundColor: 'pink',
    marginHorizontal: 20,
  },
  scrollView2: {
    backgroundColor: 'lightblue',
    marginHorizontal: 20,
  },

  item: {
    flexDirection: 'row',
    justifyContent: 'center',
    alignItems: 'center',
    padding: 20,
    margin: 2,
    backgroundColor: 'white',
    borderRadius: 10,
  },
  text: {
    fontSize: 20,
    color: 'black',
  },
});
```


packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -152,7 +154,8 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
 
     if (handlerChange.second == RNGestureHandlerMutationAttach) {
       if ([self shouldAttachGestureToSubview:handlerTag]) {
-        [_nativeHandlersToAttach addObject:handlerTag];
+        [_attachedNativeHandlers addObject:handlerTag];
```
m-bert: Why don't we move this into `tryAttachHandlerToChildView`?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -94,11 +94,13 @@ - (BOOL)shouldAttachGestureToSubview:(NSNumber *)handlerTag
   return [[[handlerManager registry] handlerWithTag:handlerTag] wantsToAttachDirectlyToView];
 }
 
-- (void)addSubview:(UIView *)view
+- (void)didAddSubview:(UIView *)view
 {
-  [super addSubview:view];
+  for (const id handlerTag : _attachedNativeHandlers) {
+    [self tryAttachHandlerToChildView:handlerTag];
+  }
```
m-bert: Why was this loop moved outside of `tryAttachHandlerToChildView`? 

Also we could find better name for this method, like `tryAttachNativeHandlersToChildView`

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -102,24 +103,20 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
     val registry = RNGestureHandlerModule.registries[moduleId]
       ?: throw Exception("Tried to access a non-existent registry")
 
-    for (tag in nativeHandlersToAttach) {
+    for (tag in attachedNativeHandlers) {
```
m-bert: Maybe a cosmetic note, but I don't like that we iterate through `attachedNativeHandlers` and call `attachHandlerToView` - it looks like we try to attach it again, which is not the case.

m-bert (commented):
I can see that android got a bit "out of sync" with iOS, can we make it so that one is analogous to another?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
akwasniewski: This is intentional, as `tryAttachHandlerToChildView` is also used in `didAddSubview`. `attachedNativeHandlers` is intentionally not cleaned on reattach, thus there is no need to call `addObject` again.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
akwasniewski: Done for [ios](https://github.com/software-mansion/react-native-gesture-handler/pull/3672/commits/0b4363fa16d1ea253d1997540c08fe1d634066b5) and [android](https://github.com/software-mansion/react-native-gesture-handler/pull/3672/commits/3bfb86c0b90a004387112057a7ed8b7336f7bc61).

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
akwasniewski: I changed the variable name to `nativeHandlers`

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -163,40 +163,38 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
...
   [super updateProps:propsBase oldProps:oldPropsBase];
   // Override to force hittesting to work outside bounds
   self.clipsToBounds = NO;
 }
 
-- (void)tryAttachHandlerToChildView
+- (void)tryAttachNativeHandlersToChildView
 {
+  if (!self.subviews[0])
+    return;
```
m-bert: Correct me if I'm wrong, but this should never happen with `NativeHandler` and if it does, I believe we would like to throw an error instead of silent fail.

akwasniewski: That was my attempt to synchronise android and iOS as on Android we had [this](https://github.com/software-mansion/react-native-gesture-handler/blob/3bfb86c0b90a004387112057a7ed8b7336f7bc61/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt#L97C1-L99C6), on Android there is a comment that explains it. I'll move it to the right place 
>  It might happen that `attachHandlers` will be called before children are added into view hierarchy

akwasniewski: Ok, you are right it can really only happen in the `attachHandlers`, thus it's better to have a check only there. Fixed in [4916ba5](https://github.com/software-mansion/react-native-gesture-handler/pull/3672/commits/4916ba5882d7d07c4cdb2a338600192e4b3a5e4e).

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -75,51 +75,46 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
       val tag = entry.key
 
       if (entry.value == GestureHandlerMutation.Attach) {
-        // It might happen that `attachHandlers` will be called before children are added into view hierarchy. In that case we cannot
-        // attach `NativeViewGestureHandlers` here and we have to do it in `addView` method.
```
m-bert, line 79: I don't really see the point of moving this comment and removing part of its information. It was written in this place so that when you see adding to set instead of attaching handler you know why it happens.

akwasniewski: I thought, that the set would have been used regardless as we need it for reattach, so it wasn't the place for this comment. But ok, I don't have a strong opinion thus I restored it and synced ios, [c4f18a7d](https://github.com/software-mansion/react-native-gesture-handler/pull/3672/commits/c4f18a7df73205b670724616e602871c6c28010d)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -152,7 +152,9 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
 
     if (handlerChange.second == RNGestureHandlerMutationAttach) {
       if ([self shouldAttachGestureToSubview:handlerTag]) {
-        [_nativeHandlersToAttach addObject:handlerTag];
+        // It might happen that `attachHandlers` will be called before children are added into view hierarchy. In that
+        // case we cannot attach `NativeViewGestureHandlers` here and we have to do it in `addView` method.
```
```suggestion
        // case we cannot attach `NativeViewGestureHandlers` here and we have to do it in `didAddSubview` method.
```

PR #3673, opened 2025-08-12
[Web] Bind `SharedValues` in handler config

akwasniewski:
## Description

Following #3658, this PR introduces web version of `SharedValue` bindings to config. This will allow users to specify values in config as SharedValues, e.g

```ts
const taps = useSharedValue(2);

const gesture = useGesture('TapGestureHandler', {
  onEnd: () => {
    console.log('Tap detected. Required number of taps:', taps.value);
    taps.set(taps.value + 1);
  },
  numberOfTaps: taps,
});
```
## Test plan

<details>

```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,

  const taps = useSharedValue(2);

  const gesture = useGesture('TapGestureHandler', {
    onEnd: () => {
      console.log('Tap detected. Required number of taps:', taps.value);
      taps.set(taps.value + 1);
    },
    numberOfTaps: taps,
  });

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
@@ -614,6 +614,11 @@ export default abstract class GestureHandler implements IGestureHandler {
   // Handling config
   //
 
+  public setGestureConfig(config: Config) {
+    this.resetConfig();
+    this.updateGestureConfig(config);
+  }
+
   public updateGestureConfig({
     enabled = true,
     dispatchesAnimatedEvents = false,
--- a/packages/react-native-gesture-handler/src/RNGestureHandlerModule.web.ts
+++ b/packages/react-native-gesture-handler/src/RNGestureHandlerModule.web.ts
@@ -74,15 +74,15 @@ export default {
     NodeManager.detachGestureHandler(handlerTag);
   },
   setGestureHandlerConfig(handlerTag: number, newConfig: Config) {
-    NodeManager.getHandler(handlerTag).updateGestureConfig(newConfig);
+    NodeManager.getHandler(handlerTag).setGestureConfig(newConfig);
 
     InteractionManager.instance.configureInteractions(
       NodeManager.getHandler(handlerTag),
       newConfig
     );
   },
-  updateGestureHandlerConfig(_handlerTag: number, _newConfig: Config) {
-    // TODO: To be implemented
+  updateGestureHandlerConfig(handlerTag: number, newConfig: Config) {
+    NodeManager.getHandler(handlerTag).updateGestureConfig(newConfig);
   },
   getGestureHandlerNode(handlerTag: number) {
     return NodeManager.getHandler(handlerTag);
--- a/packages/react-native-gesture-handler/src/web/handlers/IGestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/IGestureHandler.ts
@@ -45,6 +45,7 @@ export default interface IGestureHandler {
 
   sendEvent: (newState: State, oldState: State) => void;
 
+  setGestureConfig: (config: Config) => void;
   updateGestureConfig: (config: Config) => void;
 
   isButton?: () => boolean;


m-bert (changes requested):
Given that in `updateGestureConfig` we have:


```jsx
this._config = {
  enabled,
  dispatchesAnimatedEvents,
  ...props,
};
```

`updateGestureConfig` acts as `setGestureConfig`. The point of separating this method in module was to have `set` method which sets the config to new one and `update` method which updates only changed fields. 

Right now doing this:

```jsx
console.log(this.config);
this._config = {
  enabled,
  dispatchesAnimatedEvents,
  ...props,
};
console.log(this.config);
```

yields this:

Image

akwasniewski:
> The point of separating this method in module was to have set method which sets the config to new one and update method which updates only changed fields.

Okay, I had previously not understood this was the point, as your PR didn't add this, it simply renamed the function. Even before your PR checking if the fields exist was already built into `setConfig`. I added it in [41f615ef](https://github.com/software-mansion/react-native-gesture-handler/pull/3673/commits/41f615ef9f3b1594b85d5ae13dd09e8793142888).

m-bert:
> Okay, I had previously not understood this was the point, as your PR didn't add this, it simply renamed the function.

Not really. It added `setGestureHandlerConfig` into module spec and updated behavior of `updateGestureHandlerConfig` such that it really updates the config, not resets it 😉

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -614,29 +614,46 @@ export default abstract class GestureHandler implements IGestureHandler {
...
-      ...props,
-    };
+  public setGestureConfig(config: Config) {
+    if (config.enabled === undefined) {
+      config.enabled = true;
+    }
+    if (config.dispatchesAnimatedEvents === undefined) {
+      config.dispatchesAnimatedEvents = false;
+    }
+    this.resetConfig();
```
m-bert, line 692: I can see that `shouldCancelWhenOutside` is set to default value on Android, while here it is not. Could we check which fields should be reset in this method, as now it is only overriden in child classes?

akwasniewski:
> Not really. It added setGestureHandlerConfig into module spec and updated behavior of updateGestureHandlerConfig such that it really updates the config, not resets it 😉

Yeah, I could have worded that better. I meant that iOS / android implementations required fewer changes within the `GestureHandler`, I did not notice that web required more.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -614,29 +614,46 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+      if (this.config[key] !== config[key]) {
+        this.config[key] = config[key];
+      }
+    }
 
-    if (this.config.shouldCancelWhenOutside !== undefined) {
+    if (
+      config.shouldCancelWhenOutside !== undefined &&
+      this.config.shouldCancelWhenOutside !== undefined
+    ) {
```
m-bert:
```suggestion
 if (config.shouldCancelWhenOutside !== undefined) {
```

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -614,29 +614,46 @@ export default abstract class GestureHandler implements IGestureHandler {
...
       this.shouldCancelWhenOutside = this.config.shouldCancelWhenOutside;
     }
 
-    this.forAnimated = dispatchesAnimatedEvents;
-    this.validateHitSlops();
+    if (
+      config.dispatchesAnimatedEvent !== undefined &&
+      this.config.dispatchesAnimatedEvents !== undefined
+    ) {
```
m-bert:
```suggestion
    if (config.dispatchesAnimatedEvent !== undefined) {
```

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
m-bert: Also this check is quite weird - if `config.shouldCancelWhenOutside` is defined, then it will be assigned into `this.config.shouldCancelWhenOutside` in the loop above. And if `config.shouldCancelWhenOutside` is `undefined`, then nothing really happens.

Note, I assumed that the for loop above doesn't check for key differences.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -614,29 +614,46 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+      this.delegate.onEnabledChange(this.enabled);
+      this.enabled = config.enabled;
     }
 
-    this.enabled = enabled;
+    for (const key of Object.keys(config)) {
+      if (this.config[key] !== config[key]) {
+        this.config[key] = config[key];
+      }
+    }
```
m-bert:
```suggestion
    for (const key of Object.keys(config)) {
      this.config[key] = config[key];
    }
```

I don't think we need this check, I'd just assign new values.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
m-bert: Same as above

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -614,29 +614,27 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+      this.delegate.onEnabledChange(this.enabled);
     }
 
-    this.enabled = enabled;
+    for (const key of Object.keys(config)) {
+      this.config[key] = config[key];
+    }
 
-    if (this.config.shouldCancelWhenOutside !== undefined) {
-      this.shouldCancelWhenOutside = this.config.shouldCancelWhenOutside;
```
m-bert, line 699: Why was it removed?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Ok, my bad did not notice it was not implemented. Fixed in [4fd97f8a1b8](https://github.com/software-mansion/react-native-gesture-handler/pull/3673/commits/4fd97f8a1b804833bc4941c989b0daee08a704cc). I included `dispatchesAnimatedEvent` there, I'm not sure what the correct default value for `mouseButton`, but from what I see android default 0 is used as undefined so I put undefined. Also, different fields are reset on Android and different on iOS. It seems weird, I'll have to look into whether or not reset iOS fields as well. For now I'm mimicking android

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Fixed, got carried away [4fd97f8a1b8](https://github.com/software-mansion/react-native-gesture-handler/pull/3673/commits/4fd97f8a1b804833bc4941c989b0daee08a704cc)

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Checking undefined was just to make eslint happy, fixed in [4fd97f8a1b8](https://github.com/software-mansion/react-native-gesture-handler/pull/3673/commits/4fd97f8a1b804833bc4941c989b0daee08a704cc)

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: I agree, fixed [4fd97f8a1b8](https://github.com/software-mansion/react-native-gesture-handler/pull/3673/commits/4fd97f8a1b804833bc4941c989b0daee08a704cc)

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Mistake when fixing, ammend was incoming but because you commented I added a new commit [891df81b3](https://github.com/software-mansion/react-native-gesture-handler/pull/3673/commits/891df81b3f0abf939e922d37098e0dcdaa3cff0d)

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -614,29 +614,31 @@ export default abstract class GestureHandler implements IGestureHandler {
...
-      this.delegate.onEnabledChange(enabled);
+  public updateGestureConfig(config: Config): void {
+    if (config.enabled !== undefined && this.enabled !== config.enabled) {
+      this.delegate.onEnabledChange(this.enabled);
     }
 
-    this.enabled = enabled;
+    for (const key of Object.keys(config)) {
+      this.config[key] = config[key];
+    }
```
j-piasecki: Hmm, this may be a good moment to explicitly set individual properties here, instead of copying everything.

Then, the issue of what should be reset in the default implementation would be solved (those that are set here).

akwasniewski: Currently Child Handlers (Tap, Pan ...) don't set those config values themselves so I settled on this approach. But, I agree it might be a good time to fix it and set the config manually everywhere.

akwasniewski: Ok, in last few commits I tediously starting setting them manually. The loop has to stay for now as we use config in its entirety in certain places so we need it to have it updated with all the fields. But, looking at the native implementations, we don't have a config field there. 
Do we want to get rid of the config field in `GestureHandler` entirely? Currently the fields are doubled, they are both in config and as a class field. This seems both ugly and ineffective. @m-bert, @j-piasecki ? 
I feel if we decide to do so, that should be a separate PR.

m-bert: > Do we want to get rid of the config field in GestureHandler entirely? Currently the fields are doubled, they are both in config and as a class field. This seems both ugly and ineffective.

I don't mind getting rid of this field 😅 

> I feel if we decide to do so, that should be a separate PR.

It could, but if you look at this PR it doesn't really add `SharedValue` support as it is already done in the former PR. This one only changes the way how config is handled (by moving to `set`/`update` methods in module), so I don't think we need another PR. I'd just rename this one and update description to clarify what it really does.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -678,29 +688,58 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+
+    if (config.hitSlop !== undefined) {
+      this.hitSlop = config.hitSlop;
+      this.hitSlop = config.hitSlop;
+      this.validateHitSlops();
+    }
+
+    if (config.dispatchesAnimatedEvents !== undefined) {
+      this.forAnimated = config.dispatchesAnimatedEvents;
```
j-piasecki:
```suggestion
      this.dispatchesAnimatedEvents = config.dispatchesAnimatedEvents;
```

Can we align those names?

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
```diff
@@ -62,113 +49,136 @@ export default class PanGestureHandler extends GestureHandler {
...
+      this.activeOffsetXStart !== undefined &&
+      this.activeOffsetXEnd === undefined
+    ) {
+      this.activeOffsetXEnd = Number.MAX_SAFE_INTEGER;
+    } else if (
+      this.activeOffsetXStart === undefined &&
+      this.activeOffsetXEnd !== undefined
+    ) {
+      this.activeOffsetXStart = Number.MIN_SAFE_INTEGER;
     }
```
j-piasecki: Is the logic with `Number.MAX_SAFE_INTEGER` needed? Wouldn't that be handled by `resetConfig`?

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
```diff
@@ -62,113 +49,136 @@ export default class PanGestureHandler extends GestureHandler {
...
+      this.failOffsetXStart !== undefined &&
+      this.failOffsetXEnd === undefined
+    ) {
+      this.failOffsetXEnd = Number.MAX_SAFE_INTEGER;
+    } else if (
+      this.failOffsetXStart === undefined &&
+      this.failOffsetXEnd !== undefined
+    ) {
+      this.failOffsetXStart = Number.MIN_SAFE_INTEGER;
     }
```
j-piasecki: Same

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
```diff
@@ -62,113 +49,136 @@ export default class PanGestureHandler extends GestureHandler {
...
+      this.activeOffsetYStart !== undefined &&
+      this.activeOffsetYEnd === undefined
+    ) {
+      this.activeOffsetYEnd = Number.MAX_SAFE_INTEGER;
+    } else if (
+      this.activeOffsetYStart === undefined &&
+      this.activeOffsetYEnd !== undefined
+    ) {
+      this.activeOffsetYStart = Number.MIN_SAFE_INTEGER;
+    }
```
j-piasecki: Same

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
```diff
@@ -62,113 +49,136 @@ export default class PanGestureHandler extends GestureHandler {
...
+      this.failOffsetYStart !== undefined &&
+      this.failOffsetYEnd === undefined
+    ) {
+      this.failOffsetYEnd = Number.MAX_SAFE_INTEGER;
+    } else if (
+      this.failOffsetYStart === undefined &&
+      this.failOffsetYEnd !== undefined
+    ) {
+      this.failOffsetYStart = Number.MIN_SAFE_INTEGER;
     }
```
j-piasecki: Same

j-piasecki (commented):
Took a quick look, and the PR generally looks good. Can you please merge the `next` branch so that only actually modified files are shown in the diff?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Done in 2da4640d16f89c8218f8d2e076de3e935fbb5fb0

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
akwasniewski: Indeed, it is handled by `resetConfig`. Thanks, removed in f490152047a8b4c3f29cc8b80afd1739f58e71e4.

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
akwasniewski: Removed in https://github.com/software-mansion/react-native-gesture-handler/commit/f490152047a8b4c3f29cc8b80afd1739f58e71e4.

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
akwasniewski: Removed in https://github.com/software-mansion/react-native-gesture-handler/commit/f490152047a8b4c3f29cc8b80afd1739f58e71e4.

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
akwasniewski: Removed in https://github.com/software-mansion/react-native-gesture-handler/commit/f490152047a8b4c3f29cc8b80afd1739f58e71e4.

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
```diff
@@ -62,113 +49,88 @@ export default class PanGestureHandler extends GestureHandler {
...
+        config.enableTrackpadTwoFingerGesture;
     }
 
-    if (this.config.enableTrackpadTwoFingerGesture !== undefined) {
-      this.enableTrackpadTwoFingerGesture =
-        this.config.enableTrackpadTwoFingerGesture;
+    if (config.minDist !== undefined) {
+      this.minDist = config.minDist;
+      this.minDistSq = config.minDist * config.minDist;
+    } else if (this.minDist === undefined && this.hasCustomActivationCriteria) {
```
m-bert, line 132:
```suggestion
    } else if (this.hasCustomActivationCriteria) {
```

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -156,7 +153,7 @@ export class GestureHandlerWebDelegate
   }
 
   private setUserSelect(isHandlerEnabled: boolean) {
```
m-bert: It is probably out of scope of this PR, but if we have access to `this.gestureHandler` we could do `this.gestureHandler.enabled` instead of passing it as argument.

akwasniewski: I think If the delegate is already being refactored we might as well add it.

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
akwasniewski: We actually do need this check. Without it on next config change which doesn't change `minDist` and `hasCustomActivationCriteria` is true it would reset the value, even if we specified it in a previous config update.

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
akwasniewski: Added in 847f3edf758d35be927b5f7164434a32dc07804c

packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
m-bert: Okay, I've missed that one uses `this` and the other `config`

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -937,6 +976,34 @@ export default abstract class GestureHandler implements IGestureHandler {
     this._shouldResetProgress = value;
   }
 
+  public get enableContextMenu() {
+    return this._enableContextMenu;
+  }
+  protected set enableContextMenu(value) {
```
m-bert, line 976: Should they be `protected`? If they're not used outside of this class we could assign directly to backing field and remove setters. 

However this approach is also ok, so feel free to choose the one you prefer.

m-bert (approved):
LGTM

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Sure, I removed setters that are not used outside of the class, there is no reason to bloat the code e87ea1114591818e4c5f37d40191855e6dc31021

PR #3674, opened 2025-08-14
Lazy addition of callbacks to `useHandler`

m-bert:
## Description

Currently we call `useHandler` with all callbacks, for example:

```js
const handlers: CallbackHandlers = {
  onBegin: config.onBegin,
  onStart: config.onStart,
  onEnd: config.onEnd,
  onFinalize: config.onFinalize,
};
```

this means that if user does not pass them into config, those values will be `undefined`. This works on `android` and `iOS`, but it seems to break on `web` as Reanimated throws error that `useHandler` received non-worklet callback. To avoid potential issues, we only assign callbacks that were passed by user.

>[!WARNING]
> This solves only one issue - passing `undefined` callbacks. However, passing callback that is not marked as `worklet` will result in the same error. This problem will be handled by autoworkletization.

## Test plan

Tested with `console.logs` in event hooks.


j-piasecki:
Please remove the logs though

PR #3678, opened 2025-08-20
Fix `Pressable`'s `onPress` being called on cancellation

### LINKED ISSUE #3609: [Pressable] onPress is called if user swipes while not lifting finger on iOS
### Description

I observe this behaviour only on iOS and I'm not sure if it's correct, looks more like a bug. My patch for now is to check for `success` inside [onFinalize](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx#L304).

### Steps to reproduce

1. Press and hold the Pressable
2. Move finger
3. Observe onPress immediately called, even though I haven't lifted my finger

https://github.com/user-attachments/assets/3ef158d9-ecbe-4814-833b-589ee52cdbfa

https://gist.github.com/gigobyte/5d1d0bf9e5f188f6590210a21fc6eb3a

### Gesture Handler version

2.27.1

### React Native version

0.79.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

New Architecture (Fabric)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

latekvo:
## Description

This PR fixes the `Pressable`'s `onPress` callback being called despite the press being cancelled.

Credit for finding and fixing this issue goes to @gigobyte.

Fixes #3609

## Test plan

- Run the following example: https://gist.github.com/gigobyte/5d1d0bf9e5f188f6590210a21fc6eb3a
- Swipe the pink `Pressable` to the right.
- Before this PR, the `onPress` callback would be called without the finger being lifted.
- Since this PR, the `onPress` is no longer called when the finger isn't lifted.


AfrazHussain:
Hey @latekvo , 

Thanks for the PR! 

Any plans to get this merged? This seems to be a major issue on iOS as mentioned on the comment linked above. I've also applied these changes via a patch file, and all seems to be good for now. 

My issue: For me, the issue is when a Pressable component is either in a FlatList or a FlashList. The Pressable component just instantly triggers as soon as I touch the screen to scroll (both on simulator and on a physical device; new arch; iOS-only). It works just fine on Android. Happening since 2.27.0, not before that.

latekvo:
@AfrazHussain we'll include this fix in the next release. Apologies for the delay.

PR #3639, opened 2025-07-25
Remove paper support

latekvo:
## Description

This PR removes all paper support.

## Test plan

TODO

PR #3681, opened 2025-08-21
Remove `sync-architectures` script from commit hook

m-bert:
## Description

#3639 removed support for React Native old architecture. While it removed script for architectures sync, it is still called on commit hook, thus trying to add commit with updated spec fails.

## Test plan

Run `git commit`


coado (approved):
LGTM!

PR #3682, opened 2025-08-21
Separate `Reanimated` from `JS`

m-bert:
## Description

Currently we have 2 ways of handling events - `Animated` and `JS`/`Reanimated`. Handling both `JS` and `Reanimated` in the same places results in more complex codebase. It also introduces problems with composing gestures. We decided to split those implementations.

## Test plan

<details>
<summary>Tested on the following code:</summary>

```tsx
import * as React from 'react';
import { Animated, Button, useAnimatedValue } from 'react-native';
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,
} from 'react-native-gesture-handler';

export default function App() {
  const [visible, setVisible] = React.useState(true);

  const value = useAnimatedValue(0);

  const event = Animated.event(
    [{ nativeEvent: { handlerData: { translationX: value } } }],
    {
      useNativeDriver: true,
    }
  );

  const gesture = useGesture('PanGestureHandler', {
    onBegin: (e: unknown) => {
      'worklet';
      console.log('onBegin', e);
    },
    onStart: (e: unknown) => {
      'worklet';
      console.log('onStart', e);
    },
    // onUpdate: event,
    onUpdate: (e: unknown) => {
      'worklet';
      console.log('onUpdate', e);
    },
    onEnd: (e: unknown) => {
      'worklet';
      console.log('onEnd', e);
    },
    onFinalize: (e: unknown) => {
      'worklet';
      console.log('onFinalize', e);
    },
    onTouchesDown: (e: unknown) => {
      'worklet';
      console.log('onTouchesDown', e);
    },
    onTouchesMove: (e: unknown) => {
      'worklet';
      console.log('onTouchesMoved', e);
    },
    onTouchesUp: (e: unknown) => {
      'worklet';
      console.log('onTouchesUp', e);
    },
    onTouchesCancelled: (e: unknown) => {
      'worklet';
      console.log('onTouchesCancelled', e);
    },
  });

  return (
    <GestureHandlerRootView
      style={{ flex: 1, backgroundColor: 'white', paddingTop: 8 }}>
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
        <NativeDetector gesture={gesture}>
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
[...]

PR #3683, opened 2025-08-21
[Web] Separate Reanimated from JS

akwasniewski:
## Description

Web portion of functionality from PR #3682 
> Currently we have 2 ways of handling events - Animated and JS/Reanimated. Handling both JS and Reanimated in the same places results in more complex codebase. It also introduces problems with composing gestures. We decided to spit those implementations.

## Test plan
```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,

  const av = React.useRef(new Animated.Value(0)).current

  const event = Animated.event(
    [{ handlerData: { translationX: av } }],
    {
      useNativeDriver: false,
    }
  );

    onBegin: (e: unknown) => {
      'worklet';
      console.log('onBegin', e);
    },
    onStart: (e: unknown) => {
      'worklet';
      console.log('onStart', e);
    },
    // onUpdate: event,
    onUpdate: (e: unknown) => {
      'worklet';
      console.log('onUpdate', e);
    },
    onEnd: (e: unknown) => {
      'worklet';
      console.log('onEnd', e);
    },
    onFinalize: (e: unknown) => {
      'worklet';
      console.log('onFinalize', e);
    },
    onTouchesDown: (e: unknown) => {
      'worklet';
      console.log('onTouchesDown', e);
    },
    onTouchesMove: (e: unknown) => {
      'worklet';
      console.log('onTouchesMoved', e);
    },
    onTouchesUp: (e: unknown) => {
      'worklet';
      console.log('onTouchesUp', e);
    },
    onTouchesCancelled: (e: unknown) => {
      'worklet';
      console.log('onTouchesCancelled', e);
    },
  });

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
              { transform: [{ translateX: av }] },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```


packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -383,21 +385,29 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+            onGestureHandlerReanimatedTouchEvent,
+            touchEvent
+          );
+          return;
+        } else if (onGestureHandlerTouchEvent) {
+          invokeNullableMethod(onGestureHandlerTouchEvent, touchEvent);
+          return;
+        }
       }
+      invokeNullableMethod(onGestureHandlerEvent, touchEvent);
```
m-bert, line 406: Shouldn't we invoke `onGestureHandlerReanimatedEvent` if `forReanimated` is `true`?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -425,14 +437,23 @@ export default abstract class GestureHandler implements IGestureHandler {
 
     if (this.lastSentState !== newState) {
       this.lastSentState = newState;
-      invokeNullableMethod(onGestureHandlerStateChange, resultEvent);
+      if (this.forReanimated && onGestureHandlerReanimatedStateChange) {
```
m-bert: Wouldn't checking only `this.forReanimated` be enough? It is assigned based on `shouldUseReanimated` and reanimated hooks won't exist only if `disableReanimated` is `true`, which means that `shouldUseReanimated` is `false`. Then you could do:

```js
invokeNullableMethod(
  this.forReanimated ? onGestureHandlerReanimatedStateChange : onGestureHandlerStateChange,
  resultEvent
);
```

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -425,14 +437,23 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+        invokeNullableMethod(onGestureHandlerStateChange, resultEvent);
+      }
     }
     if (this.state === State.ACTIVE) {
       if (this.actionType !== ActionType.NATIVE_DETECTOR) {
         (resultEvent.nativeEvent as GestureHandlerNativeEvent).oldState =
           undefined;
       }
-      if (onGestureHandlerAnimatedEvent && this.dispatchesAnimatedEvents) {
+    } else if (onGestureHandlerReanimatedEvent && this.forReanimated) {
```
m-bert: Same here, I don't think we have to check for both conditions.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -425,14 +437,23 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+      invokeNullableMethod(onGestureHandlerReanimatedEvent, resultEvent);
+      if (onGestureHandlerAnimatedEvent && this.forAnimated) {
```
m-bert: Same in this case, if `this.forAnimated` is `true`, `onGestureHandlerAnimatedEvent` will exist.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
m-bert: Also, why did this move to `else` block? I think we want to send those events in `ACTIVE` state

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -425,14 +437,23 @@ export default abstract class GestureHandler implements IGestureHandler {
...
         invokeNullableMethod(onGestureHandlerAnimatedEvent, resultEvent);
       }
       invokeNullableMethod(onGestureHandlerEvent, resultEvent);
```
m-bert: This will result in sending both `Reanimated` and `JS` events at the same time and I'm not sure if we want that. Especially that now `onGestureHandlerEvent` is invoked under `forReanimated` condition

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: It logically would be enough, though `onGestureHandlerReanimatedStateChange` is nullable, therefore eslint will not allow it. I would have to  override it with the ! operator

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Ok, done in 8a2eeccad5425a977f6678f9d7b6da47f3f9609a

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: >Also, why did this move to else block

I lost the battle with brackets. Fixed in 2713e82b40c528ba3efc49841152ebfab90ac933

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Right, fixed in 2713e82b40c528ba3efc49841152ebfab90ac933

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Fixed in b3cdd6a2fa1f2f7ff16d83f571416b366a033044

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Ok, it may be better, done  in b3cdd6a2fa1f2f7ff16d83f571416b366a033044

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -383,21 +385,36 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+        if (this.forReanimated) {
+          invokeNullableMethod(
+            onGestureHandlerReanimatedTouchEvent!,
+            touchEvent
+          );
+          return;
+        } else if (onGestureHandlerTouchEvent) {
+          invokeNullableMethod(onGestureHandlerTouchEvent, touchEvent);
+          return;
+        }
```
m-bert: Why not 

```jsx
invokeNullableMethod(
  this.forReanimated ? onGestureHandlerReanimatedTouchEvent! : onGestureHandlerTouchEvent,
  touchEvent
);
```

?

akwasniewski: I wanted to preserve preexisting logic as I was not sure whethere `onGestureHandlerTouchEvent` is always defined, but I think it has to be so done in d55a970c4155c0fe27de0eb212130ca6b3a2fc34

m-bert: Yes, it will be defined, it is not called conditionally, unlike reanimated 😅

m-bert: Actually why was it marked as optional? To keep compatibility with v2?

akwasniewski: >Actually why was it marked as optional? To keep compatibility with v2?

Indeed, to keep compatibility with v2, we use it [here](https://github.com/software-mansion/react-native-gesture-handler/blob/15478f6eefb0d5f3b81ae00dc4869b906c0b842a/packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts#L170) without `onGestureHandlerTouchEvent`.

m-bert: TBH I'm not yet sure what do we want to do with these types. I'd prefer some redundancy to keep v3 isolated, however I can see that it might be easier to use existing types and adapt them to v3 later, if necessary. One possibility is to remove `?` from definition and use `Omit` in `useWebEventHandlers` (if that's the only occurrence). That would be a change in v2, but we change those types anyway. We can also leave that as it is.

Let me know what you think, we could also use @latekvo and @j-piasecki opinion - for me both options are ok.

akwasniewski: While implementing [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/3651) I tried to separate the types, but settled on reusing it, as it would require a lot of redundancy to achieve complete separation. 

> if that's the only occurrence

As far as I see it is the only occurence, I don't have a strong opinion whether to use `omit` or leave as is, but I think it's not worth it to separate completely due to redundancy.

m-bert: > but I think it's not worth it to separate completely due to redundancy.

We wanted to do so in order to separate v3 from old APIs. If we decide to get rid of the others, migration would be easier. I don't say that we should do this in this PR, I'm just saying that this was intended (unless something has changed).

cc @j-piasecki

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -383,21 +385,32 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+          touchEvent
+        );
       }
+
+      invokeNullableMethod(
+        this.forReanimated
+          ? onGestureHandlerReanimatedEvent!
+          : onGestureHandlerEvent,
+        touchEvent
+      );
```
m-bert: Shouldn't this be in `else` branch?

Also, could we do 

```js
if (!touchEvent) {
  return;
}
```

to avoid unnecessary nesting?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
j-piasecki: If this is only type issue, and this branch won't be hit in the API v2, then can't we update the type to non-optional and pass a no-op from v2?

akwasniewski: Marked as non-optional and passing no-op in https://github.com/software-mansion/react-native-gesture-handler/pull/3683/commits/aa48bb210e04c5bb40403bbecd5f8c4f8a22a814

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -175,5 +175,7 @@ export function useWebEventHandlers() {
     onGestureHandlerStateChange: (e: ResultEvent) => {
       onGestureHandlerEvent(e.nativeEvent as GestureHandlerNativeEvent);
     },
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    onGestureHandlerTouchEvent: () => {}, // no-op
```
m-bert: Let's do either this:

```suggestion
    onGestureHandlerTouchEvent: () => {
      // no-op
    },
```

or this:

```suggestion
    // eslint-disable-next-line @typescript-eslint/no-empty-function
    onGestureHandlerTouchEvent: () => {}
```

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: > Shouldn't this be in else branch? 

It should, yet again I did not notice and brackets won.

Fixed this and improved nesting in 9094afe4989048a9f0ea526603dd59584e823330

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
akwasniewski: Applied first suggestion in 240d48e0f19cc1d61464da13187661010f1fce63

akwasniewski:
>I can see that in your example, if you use both Animated.Event as onUpdate and touch events, you get the following error:
> > Uncaught Invariant Violation: Bad event of type undefined for key handlerData

Thank you! Great catch, It took me a while as I couldn't replicate the error -> it only occurs on chrome. I did not notice that I needed to have attached a condition on where to send the cancel event. I think the error might have been there ever since [this](https://github.com/software-mansion/react-native-gesture-handler/pull/3651) PR. It is fixed as of c9c856fb734123d7bb59aa4571efa65d00961161.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -425,17 +442,27 @@ export default abstract class GestureHandler implements IGestureHandler {
...
-      invokeNullableMethod(onGestureHandlerEvent, resultEvent);
+
+      invokeNullableMethod(
+        this.forReanimated
+          ? onGestureHandlerReanimatedEvent!
+          : this.forAnimated
+            ? onGestureHandlerAnimatedEvent!
+            : onGestureHandlerEvent,
+        resultEvent
+      );
```
m-bert: Cosmetic thing, we could extract this to a separate variable `const method = ...` and then use `invokeNullableMethod(method, resultEvent)`. For me it looks cleaner, but I'll leave that up to you.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -383,21 +385,34 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+          : onGestureHandlerTouchEvent,
+        touchEvent
+      );
+    } else {
+      invokeNullableMethod(
+        this.forReanimated
+          ? onGestureHandlerReanimatedEvent!
+          : onGestureHandlerEvent,
+        touchEvent
+      );
```
m-bert: Correct me if I'm wrong, but when we do not use V3, it means that `forReanimated` is `false`. In that case, why do we have

```js
invokeNullableMethod(
        this.forReanimated
          ? onGestureHandlerReanimatedEvent!
          : onGestureHandlerEvent,
        touchEvent
      );
```

instead of simply

```js
invokeNullableMethod(onGestureHandlerEvent touchEvent);
```

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -652,9 +679,28 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+        cancelEvent
+      );
+    } else {
+      invokeNullableMethod(
+        this.forReanimated
+          ? onGestureHandlerReanimatedEvent!
+          : onGestureHandlerEvent,
+        cancelEvent
+      );
+    }
```
m-bert: Same as above

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -383,21 +385,34 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+      return;
+    }
+
+    if (this.actionType === ActionType.NATIVE_DETECTOR) {
+      invokeNullableMethod(
+        this.forReanimated
+          ? onGestureHandlerReanimatedTouchEvent!
+          : onGestureHandlerTouchEvent,
+        touchEvent
+      );
```
j-piasecki: Can we assert before invocation that if `forReanimated` is true, then `onGestureHandlerReanimatedTouchEvent` is defined?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -425,17 +442,27 @@ export default abstract class GestureHandler implements IGestureHandler {
 
     if (this.lastSentState !== newState) {
       this.lastSentState = newState;
-      invokeNullableMethod(onGestureHandlerStateChange, resultEvent);
+      invokeNullableMethod(
+        this.forReanimated
+          ? onGestureHandlerReanimatedStateChange!
```
j-piasecki: Same

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -652,9 +679,28 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+      onGestureHandlerReanimatedTouchEvent,
+      onGestureHandlerTouchEvent,
+      onGestureHandlerReanimatedEvent,
+    }: PropsRef = this.propsRef!.current;
 
-    invokeNullableMethod(onGestureHandlerEvent, cancelEvent);
+    if (this.actionType === ActionType.NATIVE_DETECTOR) {
+      invokeNullableMethod(
+        this.forReanimated
+          ? onGestureHandlerReanimatedTouchEvent!
```
j-piasecki: Same

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Okay, done in https://github.com/software-mansion/react-native-gesture-handler/pull/3683/commits/5d315ca3dc37fc86c0659a48728c7d6b851d2aae

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Okay, done in https://github.com/software-mansion/react-native-gesture-handler/pull/3683/commits/5d315ca3dc37fc86c0659a48728c7d6b851d2aae

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Do we want an explicit check? I think it will always be defined, and using ! should be fine.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: I think it looks cleaner right now, moreover we avoid having additional variables.

m-bert (approved):
Looks good!

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
j-piasecki: Yeah, it's always defined *for now*. If we change it in the future, we likely won't remember this assertion.

akwasniewski: Okay, I understand it may cause problems in the future. What do you think about asserting it in `updateConfig` rather than here before each invocation? Eslint will still require ! here, but we will ensure that we will get an understandable error if something goes wrong.

j-piasecki: I've looked at the code, and this may be the wrong way to approach this - I don't think we should throw (or assert) on whether the callback is defined or not. Users may choose not to handle specific events, in which case we shouldn't try emitting them.

It looks like `invokeNullableMethod` already handles the case where it receives a falsy argument, so I think we should just update its definition to make it explicit and remove asserts.

akwasniewski: Ok, thank you great idea! Done in f8d91bb1289e429e0b349a80d0045339c2aef1dc

PR #3684, opened 2025-08-22
[Web] many children check

akwasniewski:
## Description

Web implementation of native detector did not check whether there is exactly one check. This PR resolves this issue and throws an error if necessary.

## Test plan

Code below should throw an error
```ts
import { Button } from 'react-native';
import {
  GestureHandlerRootView,
  NativeDetector,
  useGesture,
import Animated, { useAnimatedStyle, useSharedValue } from 'react-native-reanimated';

  const visible = useSharedValue(true);
    onBegin: (e: any) => {
      'worklet'
      console.log('onGestureHandlerEvent', e.handlerData);
    },
    onUpdate: (e: any) => {
      'worklet'
      const { translationX, translationY } = e.handlerData;
      console.log(e.handlerData)
      const distance = Math.sqrt(translationX ** 2 + translationY ** 2);

      if (distance > 100) {
        visible.value = false;
      }
    },
  });
    opacity: visible.value ? 1 : 0,
    transform: [{ scale: visible.value ? 1 : 0.5 }],
    width: 150,
    height: 150,
    backgroundColor: 'blue',
    borderWidth: 10,
    borderColor: 'green',
    marginTop: 20,
    marginLeft: 40,
  }));

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          visible.value = !visible.value
        }}
      />
        <Animated.View
          style={animatedStyle}
        />
        <Animated.View
          style={animatedStyle}
        />
      </NativeDetector>
    </GestureHandlerRootView >
  );
}
```


packages/react-native-gesture-handler/src/v3/HostGestureDetector.web.tsx
```diff
@@ -42,14 +42,9 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
           tag
         ).shouldAttachGestureToChildView()
       ) {
-        if (!viewRef.current?.firstChild) {
-          throw new Error(
-            tagMessage('Detector expected to have a child element')
-          );
-        }
```
akwasniewski, line 49: I deleted the check here as it is checked in the calling function, so it is safe

PR #3690, opened 2025-08-25
Bump Reanimated in `expo-example`

m-bert:
## Description

Currently expo-example uses Reanimated 3.19 which seems to break. Given that `next` targets only new architecture, we can bump version of Reanimated, which seems to resolve the problem.

>[!IMPORTANT]
> We should also check if example works on `main`.

## Test plan

Tested that `expo-example` builds on both platforms.


j-piasecki (approved):

latekvo (approved):

MatiPl01 (approved):
FYI: you can bump reanimated to `4.1.0` and worklets to `0.5.0`

PR #3693, opened 2025-08-26
Gesture relations

m-bert:
> [!IMPORTANT]
> Supersede #3664.
>
> I've decided to create a separate PR since it was easier to start working on it directly than waiting for #3682 to be merged.

## Description

This PR introduces hooks to set relations between handlers. 

## API

New API replaces the old one as follows:

- `Gesture.Race(g1, g2)` $\rightarrow$ `useRace(g1, g2)`
- `Gesture.Exclusive(g1, g2)` $\rightarrow$ `useExclusive(g1, g2)`
- `Gesture.Simultaneous(g1, g2)` $\rightarrow$ `useSimultaneous(g1, g2)`

### Handling external relations
In order to properly handle gesture relations, we need to pass 3 arrays to the native side:

- `waitFor` - responsible for handling `Exclusive` and `requireExternalGestureToFail` relations
- `simultaneousHandlers` - responsible for `Simultaneous` and `simultaneousWithExternalGesture` relations
- `blocksHandlers` - responsible for `blocksExternalGesture` relation

At first, these arrays are filled with external relations in `useGesture` hook. Then we use `DFS` algorithm to add remaining relations, added with relation hooks. Since `Race` doesn't really change anything when it comes to gesture interactions, we can ignore it in our algorithm. 

### DFS overview

We use `DFS` because gesture relations form tree structure. 

<details>
<summary>The algorithm works as follows:</summary>

- Initialize two arrays: `waitFor` and `simultaneousHandlers`. If root node is `SimultaneousGesture`, we also add its handler tags into `simultaneousHandlers` array. This ensures that the algorithm works even if we have only `Simultaneous` as the root node (e.g. `useSimultaneous(g1, g1)`)
- Traverse the gesture tree:
  - If we are not in the `ComposedGesture`, it means that we reached leaf node. In that case we populate `waitFor` and `simultanoursHandlers` arrays into `node` arrays and then update relations on the native side
  - If we are in the `ComposedGesture`, then for each child:
    - If the child is not `ComposedGesture`:
      - We call `traverseGestureRelations` to reach stop condition and configure relations on the native side
      - If current node is `Exclusive`, then we add child `tag` to `waitFor` array
    - If the child is `ComposedGesture`:
      - On the way down:
        - Going from `non-simultaneous` gesture to `simultaneous` gesture we add all **child** tags into global `simulatneousHandlers` array
        - If we go from `simultaneous` to `non-simultaneous` gesture, we remove **child** tags instead of adding. 
      - We store length of `waitFor` to reset it later. 
      - We call `traverseGestureRelations`
      - On the way back:
        - if we go from `simultaneous` (child) to `non-simultaneous` (node) gesture, we remove **node** tags from `simultaneousHandlers`
        - Going from `non-simultaneous` (child) to `simultaneous` (node) gesture we add **node** tags instead of removing
        - Returning to `Exclusive` gesture means that we want to add all children tags into `waitFor`
        - If we return from `Exclusive` child to `non-exclusive` node, we want to reset `waitFor` to previous state, using `length` variable.

</details>

### Example 

Below you can see example of the algorithm. 

<details>
<summary>We use the following notation:</summary>

- Handlers and composition:
  - `E` - `Exclusive`
  - `S` - `Simultaneous`
  - `P` - `Pan`
  - `T` - `Tap`
- Relation arrays:
  - `SH` - `simultaneousHandlers`
  - `WF` - `waitFor`
- Operators:
  - `+=` -  adding tags
  - `-=` -  removing tags

</details>

_**Note:**_ vertex label in relation arrays expands to all tags in the composed gesture.

```mermaid
graph TB
    E1["E₁"] --> |SH += S₁| S1
    E1["E₁"] --> |SH += S₂| S2

    S1["S₁"] --> |SH -= S₁ <br/> WF += S₁| E1
    S1 --> |SH -= E₂| E2
    S1 --> P3
    E2["E₂"] --> |SH += E₂ <br/> WF -= E₂| S1
    P3["P₃ <br/> SH: {T₁, T₂, P₃}<br/>WF: #91;#93;"] --> S1

    E2 --> T1
    E2 --> T2
    T1["T₁ <br/> SH: {P₃}<br/>WF: #91;#93;"] --> |WF += T₁| E2
    T2["T₂ <br/> SH: {P₃}<br/>WF: #91;T₁#93;"] --> |WF += T₂| E2

    S2["S₂"] -->|SH -= S₂| E1
    S2 --> P4
    S2 --> P5
    P4["P₄ <br/> SH: {P₄, P₅}<br/>WF: #91;T₁, T₂, P₃#93;"] --> S2
    P5["P₅ <br/> SH: {P₄, P₅}<br/>WF: #91;T₁, T₂, P₃#93;"] --> S2

    style DFS fill-opacity:0,stroke-opacity:0,stroke-width:0px
```

## Limitations

Currently the following setup doesn't work on `android`:
```js
const composedGesture = useExclusive(tap1, useRace(pan1, pan2));
```

I've managed to find out what is the difference between this and using only `useRace`.

>[!WARNING]
> This problem seems to be present also on `main`, so I think it will be better to solve it in the follow-up PR.

For now, external relation props do not support composed gestures. Let me know if this should be done in this PR, or in a follow-up.

### Same detector interactions

Verified that the following relations work:

<details>
<summary>Android</summary>

- [x] `Simultaneous`
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated 
  - [x] JS + Animated
- [x] `Exclusive` 
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated
  - [x] JS + Animated
- [x] `Race` 
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated
  - [x] JS + Animated
- [x] Simple composition
  - [x] `Exclusive` + `Simultaneous` 

</details>

<details>
<summary>iOS</summary>

- [x] `Simultaneous`
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated 
  - [x] JS + Animated
- [x] `Exclusive` 
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated
  - [x] JS + Animated
- [x] `Race` 
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated
  - [x] JS + Animated
- [x] Simple composition
  - [x] `Exclusive` + `Simultaneous` 

</details>

<details>
<summary>Base code used for testing:</summary>

```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  useSimultaneous,
  useGesture,
  useExclusive,
  useRace,

  const event = Animated.event(
    [{ nativeEvent: { handlerData: { translationX: value } } }],
    {
      useNativeDriver: true,
    }
  );

  const tap1 = useGesture('TapGestureHandler', {
    onEnd: () => {
      // 'worklet';
      console.log('Tap 1');
    },
    numberOfTaps: 1,
    disableReanimated: true,
  });

  const tap2 = useGesture('TapGestureHandler', {
    onEnd: () => {
      // 'worklet';
      console.log('Tap 2');
    },
    numberOfTaps: 2,
    disableReanimated: true,
  });

  const pan1 = useGesture('PanGestureHandler', {
    // onUpdate: event,
    onUpdate: (e) => {
      // 'worklet';
      console.log('Pan 1');
    },
    disableReanimated: true,
  });

  const pan2 = useGesture('PanGestureHandler', {
    onUpdate: (e) => {
      // 'worklet';
      console.log('Pan 2');
    },
    disableReanimated: true,
  });

  const composedGesture = useSimultaneous(pan1, pan2);

  // const composedGesture = useExclusive(tap2, tap1);
  // const composedGesture = useExclusive(pan2, pan1); // For Animtaed.Event
  // const composedGesture = useExclusive(pan1, pan2); // For Animtaed.Event

  // const composedGesture = useRace(pan1, pan2);
  // const composedGesture = useRace(pan2, pan1);

  // const composedGesture = useExclusive(tap1, useSimultaneous(pan1, pan2));

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
        <NativeDetector gesture={composedGesture}>
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>

### Cross detector interactions

Verified that the following relations work:

<details>
<summary>Android</summary>

- [x] `simultaneousWithExternalGesture`
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated 
  - [x] JS + Animated
- [x] `requireExternalGestureToFail` 
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated
  - [x] JS + Animated
- [x] `blocksExternalGesture` 
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated
  - [x] JS + Animated

</details>

<details>
<summary>iOS</summary>

- [x] `simultaneousWithExternalGesture`
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated 
  - [x] JS + Animated
- [x] `requireExternalGestureToFail` 
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated
  - [x] JS + Animated
- [x] `blocksExternalGesture` 
  - [x] Only JS
  - [x] Only Reanimated
  - [x] JS + Reanimated
  - [x] JS + Animated

</details>

<details>
<summary>Base code used for testing:</summary>

```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  useSimultaneous,
  useGesture,
  useExclusive,
  useRace,

  const event = Animated.event(
    [{ nativeEvent: { handlerData: { translationX: value } } }],
    {
      useNativeDriver: true,
    }
  );

  const tap1 = useGesture('TapGestureHandler', {
    onEnd: () => {
      // 'worklet';
      console.log('Tap 1');
    },
    numberOfTaps: 1,
    disableReanimated: true,
  });

  const tap2 = useGesture('TapGestureHandler', {
    onEnd: () => {
      // 'worklet';
      console.log('Tap 2');
    },
    numberOfTaps: 2,
    disableReanimated: true,
    blocksExternalGesture: tap1,
  });

  // const tap1 = useGesture('TapGestureHandler', {
  //   onEnd: () => {
  //     'worklet';
  //     console.log('Tap 1');
  //   },
  //   numberOfTaps: 1,
  //   // disableReanimated: true,
  //   requireExternalGestureToFail: tap2,
  // });

  const pan1 = useGesture('PanGestureHandler', {
    // onUpdate: event,
    onUpdate: (e) => {
      'worklet';
      console.log('Pan 1');
    },
    // disableReanimated: true,
  });

  const pan2 = useGesture('PanGestureHandler', {
    onUpdate: (e) => {
      'worklet';
      console.log('Pan 2');
    },
    simultaneousWithExternalGesture: pan1,
    // requireExternalGestureToFail: pan1,
    // blocksExternalGesture: pan1,
    // disableReanimated: true,
  });

  // const composedGesture = useSimultaneous(pan1, pan2);

  // const composedGesture = useExclusive(tap2, tap1);
  // const composedGesture = useExclusive(pan2, pan1); // For Animated.Event
  // const composedGesture = useExclusive(pan1, pan2); // For Animated.Event

  // const composedGesture = useRace(pan1, pan2);
  // const composedGesture = useRace(pan2, pan1);

  // const composedGesture = useExclusive(tap1, useSimultaneous(pan1, pan2));

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
        <NativeDetector gesture={pan1}>
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
                display: 'flex',
                alignItems: 'center',
              },
            ]}>
              <Animated.View
              />
            </NativeDetector>
          </Animated.View>
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
@@ -6,7 +6,9 @@ export function isComposedGesture(
   return 'tags' in gesture;
 }
 
-function extractHandlerTags(otherHandler: Gesture | Gesture[]): number[] {
+function extractHandlerTags(
+  otherHandler: Gesture | Gesture[] | undefined
+): number[] {
   if (!otherHandler) {
     return [];
   }
@@ -27,7 +29,7 @@ function extractHandlerTags(otherHandler: Gesture | Gesture[]): number[] {
 }
 
 function makeSimultaneousWithExternalGestureSymmetric(
-  otherHandler: Gesture | Gesture[],
+  otherHandler: Gesture | Gesture[] | undefined,
   handlerTag: number
 ) {
   if (!otherHandler) {


packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -48,3 +49,80 @@ export type CallbackHandlers = Omit<
...
+  Simultaneous: 'SimultaneousGesture',
+  Exclusive: 'ExclusiveGesture',
+  Race: 'RaceGesture',
+} as const;
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type ComposedGestureType = ValueOf<typeof ComposedGestureType>;
+
+// TODO: Find better name
+export const HandlerType = {
```
m-bert: I'm not sure about this name so if you have any idea how to name it let me know 😅

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -48,3 +49,80 @@ export type CallbackHandlers = Omit<
...
+  Pan: 'PanGestureHandler',
+  Pinch: 'PinchGestureHandler',
+  Rotation: 'RotationGestureHandler',
+  Fling: 'FlingGestureHandler',
+  Manual: 'ManualGestureHandler',
+  Native: 'NativeGestureHandler',
+} as const;
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type SingleGestureType = ValueOf<typeof SingleGestureType>;
```
j-piasecki: Wouldn't making it an enum solve the need to redeclare?

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -48,3 +49,80 @@ export type CallbackHandlers = Omit<
...
+
+export type GestureRelations = {
+  simultaneousHandlers: number[];
+  waitFor: number[];
+  blocksHandlers: number[];
+};
+
+export type NativeGesture = {
+  tag: number;
+  type: HandlerType;
```
j-piasecki: Shouldn't this be `SingleGestureType`?

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -48,3 +49,80 @@ export type CallbackHandlers = Omit<
...
+
+export type ComposedGesture = {
+  tags: number[];
+  type: ComposedGestureType;
+  config: {
+    shouldUseReanimated: boolean;
+    dispatchesAnimatedEvents: boolean;
+  };
+  gestureEvents: GestureEvents;
+  gestures: (NativeGesture | ComposedGesture)[];
```
j-piasecki: We may want to have that (`NativeGesture | ComposedGesture`) as a named type

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -103,9 +103,6 @@ - (void)resetConfig
   self.enabled = YES;
   self.manualActivation = NO;
   _shouldCancelWhenOutside = NO;
-  _handlersToWaitFor = nil;
-  _simultaneousHandlers = nil;
-  _handlersThatShouldWait = nil;
```
j-piasecki, line 108: 1. Why don't we want to reset it?
2. Why only on iOS?

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
```diff
@@ -34,29 +37,51 @@ export function NativeDetector({ gesture, children }: NativeDetectorProps) {
...
+        gesture.type === ComposedGestureType.Simultaneous ? gesture.tags : []
+      )
+    );
+  } else {
+    RNGestureHandlerModule.configureRelations(gesture.tag, {
+      waitFor: gesture.gestureRelations.waitFor,
+      simultaneousHandlers: gesture.gestureRelations.simultaneousHandlers,
+      blocksHandlers: gesture.gestureRelations.blocksHandlers,
+    });
+  }
```
j-piasecki: I feel like this should be its own function

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+// `waitFor` and `simultaneousHandlers` are global data structures that will be populated into each gesture.
+// For `waitFor` we need array as order of the gestures matters.
+// For `simultaneousHandlers` we use Set as the order doesn't matter.
+
+import RNGestureHandlerModule from '../../RNGestureHandlerModule';
+import { isComposedGesture } from '../hooks/utils/relationUtils';
+import { ComposedGesture, ComposedGestureType, NativeGesture } from '../types';
+
+// The tree consists of ComposedGestures and NativeGestures. NativeGestures are always leaf nodes.
+export const traverseGestureRelations = (
```
j-piasecki: Uhm, can we have tests for that 🥺 (follow-up is fine, this is quite large already)?

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
```diff
@@ -0,0 +1,110 @@
...
+      onGestureHandlerAnimatedEvent =
+        gesture.gestureEvents.onGestureHandlerAnimatedEvent;
+
+      break;
+    }
+  }
+
+  return {
+    tags,
+    type: ComposedGestureType.Race,
```
j-piasecki: IMO, having this as the first argument would be cleaner

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
```diff
@@ -0,0 +1,110 @@
...
+  );
+
+  let onGestureHandlerAnimatedEvent;
+
+  for (const gesture of gestures) {
+    if (gesture.gestureEvents.onGestureHandlerAnimatedEvent) {
+      onGestureHandlerAnimatedEvent =
+        gesture.gestureEvents.onGestureHandlerAnimatedEvent;
+
+      break;
```
j-piasecki: We may want to go over the entire thing (in dev) to make sure that only one is passed and warn if more are set.

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
```diff
@@ -0,0 +1,110 @@
...
+    gestures.map(
+      (gesture) => gesture.gestureEvents.onReanimatedUpdateEvent || null
+    )
+  );
+
+  const onReanimatedTouchEvent = Reanimated?.useComposedEventHandler(
+    gestures.map(
+      (gesture) => gesture.gestureEvents.onReanimatedTouchEvent || null
+    )
+  );
```
j-piasecki, line 91: Is passing `nulls` to `useComposedEventHandler` safe?

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
```diff
@@ -0,0 +1,110 @@
...
+  UpdateEvent,
+  TouchEvent,
+  ComposedGesture,
+  ComposedGestureType,
+} from '../../types';
+import { tagMessage } from '../../../utils';
+import { Reanimated } from '../../../handlers/gestures/reanimatedWrapper';
+import { isComposedGesture } from '../utils/relationUtils';
+
+// TODO: Simplify repeated relations (Simultaneous with Simultaneous, Exclusive with Exclusive, etc.)
```
j-piasecki, line 13: What do you mean? This comment seems a bit out of place, but I may be simply not understanding it.

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+          isComposedGesture(gesture) ? gesture.tags : gesture.tag
+      );
+    } else {
+      otherTags = isComposedGesture(otherHandler)
+        ? otherHandler.tags
+        : [otherHandler.tag];
+    }
+
+    return otherTags;
+  };
```
j-piasecki: Why is that function not in global scope?

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+  gesture: NativeGesture | ComposedGesture
+): gesture is ComposedGesture {
+  return 'tags' in gesture;
+}
+
+export function prepareRelations(
+  config: any,
+  handlerTag: number
+): GestureRelations {
+  // TODO: Handle composed gestures passed into external relations
```
j-piasecki: Is this still valid?

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+  // waitFor and simultaneousHandlers. We also want to configure relations on the native side.
+  if (!isComposedGesture(node)) {
+    node.gestureRelations.simultaneousHandlers.push(...simultaneousHandlers);
+    node.gestureRelations.waitFor.push(...waitFor);
+
+    RNGestureHandlerModule.configureRelations(node.tag, {
+      waitFor: node.gestureRelations.waitFor,
+      simultaneousHandlers: node.gestureRelations.simultaneousHandlers,
+      blocksHandlers: node.gestureRelations.blocksHandlers,
+    });
```
j-piasecki, line 32: Can this be done outside of this function? If not, can this be renamed? `traverseGestureRelations` in any way doesn't suggest that it updates the gestures on the native side.

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+export const traverseGestureRelations = (
+  node: NativeGesture | ComposedGesture,
+  simultaneousHandlers: Set<number>,
+  waitFor: number[] = []
+) => {
+  // If we are in the leaf node, we want to fill gesture relations arrays with current
+  // waitFor and simultaneousHandlers. We also want to configure relations on the native side.
+  if (!isComposedGesture(node)) {
+    node.gestureRelations.simultaneousHandlers.push(...simultaneousHandlers);
+    node.gestureRelations.waitFor.push(...waitFor);
```
j-piasecki, line 26: Doesn't this cause the arrays to grow with duplicates on every render?

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
```diff
@@ -34,29 +37,51 @@ export function NativeDetector({ gesture, children }: NativeDetectorProps) {
     );
   }
 
+  if (isComposedGesture(gesture)) {
+    traverseGestureRelations(
```
j-piasecki: It might be a good idea to check if a gesture isn't passed multiple times (i.e. `useSimultaneous(tap1, useRace(tap1, tap2))`)

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+      }
+
+      // If we go back from an exclusive gesture to a non-exclusive gesture, we want to reset the `waitFor` array
+      // to the previous state, siblings of the exclusive gesture are not exclusive with it. Since we use `push` method to
+      // add tags to the `waitFor` array, we can override `length` property to reset it to the previous state.
+      if (
+        child.type === ComposedGestureType.Exclusive &&
+        node.type !== ComposedGestureType.Exclusive
+      ) {
+        waitFor.length = length;
```
j-piasecki, line 107: What is this sorcery?

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    return;
+  }
+
+  // If we are in the composed gesture, we want to traverse its children.
+  node.gestures.forEach((child) => {
+    // If child is composed gesture, we have to correctly fill `waitFor` and `simultaneousHandlers`.
+    if (isComposedGesture(child)) {
+      // We have to update `simultaneousHandlers` before traversing the child.
```
j-piasecki:
```suggestion
      // We have to update `simultaneousHandlers` before traversing the child (going top-down). Simultaneous is an all-to-all relation - it needs to be prepared when entering the node. Exclusive is a one-to-many relation - gesture depends on the preceding ones and not on itself - it should be prepared when leaving the node (bottom-up).
```

packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+  const len = str.length;
+  let h = 5381;
+
+  for (let i = 0; i < len; i++) {
+    h = (h * 33) ^ str.charCodeAt(i);
+  }
+  return h >>> 0;
+}
+
+const SHARED_VALUE_OFFSET = 1.618;
```
akwasniewski, line 20: Why is it a float?

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
```diff
@@ -34,29 +37,51 @@ export function NativeDetector({ gesture, children }: NativeDetectorProps) {
...
+      gesture,
+      new Set(
+        // If root is simultaneous, we want to add its tags to the set
+        gesture.type === ComposedGestureType.Simultaneous ? gesture.tags : []
```
akwasniewski: I feel like, because of this conditional, it would be cleaner to not call DFS directly, but wrap it and check the condition there. This component is already complicated enough

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+// The tree consists of ComposedGestures and NativeGestures. NativeGestures are always leaf nodes.
+export const traverseGestureRelations = (
+  node: NativeGesture | ComposedGesture,
+  simultaneousHandlers: Set<number>,
+  waitFor: number[] = []
+) => {
+  // If we are in the leaf node, we want to fill gesture relations arrays with current
+  // waitFor and simultaneousHandlers. We also want to configure relations on the native side.
+  if (!isComposedGesture(node)) {
+    node.gestureRelations.simultaneousHandlers.push(...simultaneousHandlers);
```
akwasniewski, line 25: Why do we push rather than assign to `waitFor` and `simulatanous`?

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
akwasniewski: Not necesarilly, I think we should add a warning, but allow the user to do it. For example one might do something like `useSimulatanous(useEclusive(tap1, tap2), useExcluseive(tap1, tap3))`. Then `tap2`, `tap3` can run simultanously, but `tap1` blocks both.

akwasniewski: Not necesarilly, I think we should add a warning, but allow the user to do it. For example one might do something like useSimulatanous(useEclusive(tap1, tap2), useExcluseive(tap1, tap3)). Then tap2, tap3 can run simultanously, but tap1 blocks both.

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: The problem here was that I wanted to have separate `enum` for single and composed gestures, but it is not trivial to merge 2 enums into one. 

On the other hand, I don't think we really need to give users access into those enums as we want to create separate hooks. In that case I think we can stick to enums in our internal implementation.

Changed in [619aed9](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/619aed94bdcc70d14affca67b5fbef5414f0f68f).

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: Probably, I'm open to suggestions how we can change those names to better represent what they stand for. I think this one doesn't need `Type` suffix, so `SingleGesture` should be enough. I'm not sure about `SingleGestureType` though, as it represents all possible single gestures.

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: Actually we do, 2 lines below 😄 I must have missed it for some reason 😅 

Changed in [b12824b](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/b12824bfa9e444af68219abccd99b84813b23492)

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
m-bert: > Why don't we want to reset it?

Doing so in `resetConfig` resulted in relations not working on `iOS` since it was reset after setting those relations. Also now they are not part of the config so it is not the best place to do it. Given that calling `updateRelations` sets, not updates this array:

```
  _handlersToWaitFor = [RCTConvert NSNumberArray:relations[@"waitFor"]];
```

I'm not sure if reset is necessary in that case. If so, let me know. 

> Why only on iOS?

On android it is handled in `InteractionsManager`, so it is a bit different flow and everything worked out of the box.

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
m-bert: Sure, done in [4ad7f5b](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/4ad7f5b5f4c4c41dbf70b87a5641a6be3bc5ad52)

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
m-bert: I think so 😅 Do you mean jest tests that will check result arrays after DFS?

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
m-bert: I agree. Done in [a6f1ecb](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/a6f1ecb98bad534deae58a56c3d100f5df03ee0f)

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
m-bert: Yes, it first filters the array removing all `nulls`. You can check that [here](https://github.com/software-mansion/react-native-reanimated/blob/dfbce898ab87d348fcb8b01d2ae6f5d664951cda/packages/react-native-reanimated/src/hook/useComposedEventHandler.ts#L53).

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
m-bert: It may be a bit out of place, that's true. What I meant by it is if you have something like

```js
useSimultaneous(useSimultaneous(p1, p2), p3))
```

then in reality you have

```js
useSimultaneous(p1, p2, p3)
```

This is just a TODO that I've written not to forget about it, but I have not been thinking about it yet (maybe it doesn't  make much sense to simplify it, or it won't be trivial, I don't know yet)

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
m-bert: Well, I've left it here but you're right, it will be better to have it in global scope. Changed in [761e22a](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/761e22abba890a5edf1588eeabc423b4e7b4f7ed)

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
m-bert: Yes it is

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
m-bert: I think it will be better to rename it, though I'm open for a discussion (cc @akwasniewski). 

If we stick to rename, what about `traverseAndUpdateRelations`?

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
m-bert: It resets `waitFor` to previous state without having to use functions like `splice` 😅

packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
m-bert: We don't want to use popular offsets in order to avoid conflicts with other libraries

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
m-bert: I've already moved it to `Utils` in [4ad7f5b](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/4ad7f5b5f4c4c41dbf70b87a5641a6be3bc5ad52), let me know if you'd like to change something else

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
m-bert: Because they can contain tags added by external relations. However I can see that this causes troubles during re-renders 🙈

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
m-bert: Added in [648f862](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/648f8629fe996d1507bbdfcb15dc6e2adb01baf8)

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
m-bert: Oh, you're right 🙈

m-bert: I've changed it a bit in [ce2de67](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/ce2de67a53388e265b5b4c99055345e699b38ef0), let me know what you think

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
m-bert: I've made some changes in [ce2de67](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/ce2de67a53388e265b5b4c99055345e699b38ef0), let me know what you think

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
m-bert: Though we've talked about it, I agree with @j-piasecki. I think adding warning here will be the safest option - if it works than we won't break something, but on the other hand we do warn about potentially unstable behavior.

j-piasecki: > For example one might do something like useSimulatanous(useEclusive(tap1, tap2), useExcluseive(tap1, tap3)). Then tap2, tap3 can run simultanously, but tap1 blocks both.

```
useExclusive(tap1, useSimultaneous(tap2, tap3))
```

I think that you should be able to simplify those relations so that each gesture appears once.

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
akwasniewski: Looks good, altough we repopulate the array with external gestures even when they were not changed. But i guess this will only happen durring rerenders, so it is not an issue, and fixing it would probably complicate the code a lot. I think it can stay as is.

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
akwasniewski: > I think that you should be able to simplify those relations so that each gesture appears once.

Yes of course, but for more complicated examples it might be harder to figure out, I think it's best to leave the users with some freedom. We also didn't prove that it is always possible to reduce it in such a way, there might exist a irreducible case.

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
akwasniewski: I think it is most logical to leave it here. Maybe we could simply adhere to the name of `configureRelations` and name it `traverseAndConfigureRelations`?

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
```diff
@@ -19,6 +19,14 @@ export function useComposedGesture(
     isComposedGesture(gesture) ? gesture.tags : gesture.tag
   );
 
+  if (containsDuplicates(tags)) {
+    console.warn(
+      tagMessage(
+        'Using the same gesture more than once in gesture composition can lead to unexpected behavior.'
```
akwasniewski: Warning is a good addition. Do simple cases with duplicates work? Something like 
`useSimultaneous(useSimultaneous(p1, p2), p3))` or `useSimulatenous(useExclusive(p1,p2),useExclusive(p1,p3))?

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
j-piasecki: Having it constrained to one instance of a specific gesture would prevent some weird scenarios like
```
useSimultaneous(useExclusive(tap1, tap2), tap2)
```
which doesn't make much sense. Generally, I don't think it's likely that more than 4-5 gestures will be needed at once on the same component, so in reality, the complexity shouldn't be very high.

And if there's a case which cannot be reduced, the relations can always be setup manually with `simultaneousWithExternalGesture`, `requireExternalGestureToFail` and `blocksExternalGesture` (names TBD).

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
```diff
@@ -51,3 +51,9 @@ export function prepareRelations(
     blocksHandlers: extractHandlerTags(config.blocksExternalGesture),
   };
 }
+
+export function containsDuplicates(tags: number[]) {
+  const tagSet = new Set(tags);
+
+  return tagSet.size !== tags.length;
```
akwasniewski:
```suggestion
  return new Set(tags).size !== tags.length;
```

m-bert: [2e99481](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/2e994810caae3edd45e0e369c5e9bce1edb97cbb)

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
m-bert: I've changed it to error in [adfada7](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/adfada714777e3c92cb6fc6fce693997391e8076). If one gesture appears as a leaf node more than once, we may have some unexpected behavior.

Consider @akwasniewski example:

```js
useSimultaneous(
  useExclusive(pan1, pan2),
  useExclusive(pan1, pan3)
)
```

This will result in the following structure:

```mermaid
graph TB
    S["S <br/> SH = {P₁, P₂, P₁, P₃}"] --> |SH -= E₁| E1
    S --> |SH -= E₂| E2

    E1["E₁ <br/> SH = {P₃}"] --> P1["P₁  <br/> SH = {P₂}"]
    E1 --> P2["P₂ <br/> SH = {P₃}"]
    
    E2["E₂ <br/> SH = {P₂}"] --> P1
    E2 --> P3["P₃  <br/> SH = {P₂}"]
```

When traversing `E₂`, `P₁` gets its `simultaneousHandlers` array overwritten to contain only `P₂`, while it should contain both, `P₂` and `P₃`.

Now, we can fix it by not resetting these arrays, but I think that if we leave possibility of gestures being multiple leaves at the same time it can lead to unexpected behaviors that will be hard to fix later on. Moreover, this situation is not possible on `main` so we do not break something that is working now.

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
m-bert: Let me know what you think about [7ab9b3b](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/7ab9b3bebae01ffb3696c70bfcfde6fd5912baef)

packages/react-native-gesture-handler/src/v3/types.ts
j-piasecki: `NativeGestureName`?

packages/react-native-gesture-handler/src/v3/types.ts
j-piasecki: Can you update the occurrences of `NativeGesture | ComposedGesture` to be `Gesture` instead?

packages/react-native-gesture-handler/src/v3/hooks/relations/useComposedGesture.ts
j-piasecki: I think the word is "nested"

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
m-bert: Okay, I've decided to include it in this PR as it doesn't add that much of a logic. We only had to handle `simultaneous` relation to make sure that the relation is symmetric. Done in [db2e974](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/db2e974fb2cc1029659b06de7064c0a1ab6a47e9).

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: I've done it in #3700, but I guess we can also include it here. Done in [1429ecd](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/1429ecdc323d94f8505a6a8dce233be40faae59b)

packages/react-native-gesture-handler/src/v3/NativeDetector/utils.ts
m-bert: Sure, done in [81677c7](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/81677c751b8c0761d5a6015884d88d25baa30e70)

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: What about changing `NativeGesture` to `SingleGesture` and than `SingleGestureType` to `SingleGestureName`, along with `ComposedGestureType` to `ComposedGestureName`. Or alternatively leaving `Type` suffix? 

TBH `NativeGesture` reminds me of `NativeViewGestureHandler` so I'm not a big fan of keeping it. But that's not a strong opinion, we can do this, especially that it is our internal implementation 

cc @akwasniewski @latekvo

j-piasecki: That sounds reasonable - `SingleGesture` and `SingleGestureName` imo.

m-bert: Okay, done in [7c001fe](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/7c001fe789367bd1e45c5e9d42f041497795e470)

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
```diff
@@ -26,31 +26,40 @@ function extractHandlerTags(otherHandler: Gesture | Gesture[]): number[] {
...
-): GestureRelations {
-  if (config.simultaneousWithExternalGesture) {
-    if (Array.isArray(config.simultaneousWithExternalGesture)) {
-      for (const gesture of config.simultaneousWithExternalGesture) {
-        const simultaneousHandlers =
-          gesture.gestureRelations.simultaneousHandlers;
+) {
+  if (!otherHandler) {
+    return;
+  }
```
j-piasecki, line 37: Types say this is not necessary

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
j-piasecki: So, the todo is no longer needed, right?

m-bert: Yep, I've already removed it in the follow-up commit since I've already pushed this one. Done in https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/6fe39a57346a3ddaeadb6b6aa91a36cd9da03ff9

packages/react-native-gesture-handler/src/v3/hooks/utils/relationUtils.ts
m-bert: Good catch, they should be marked as possibly `undefined`, as those props (`simultaneousWithExternalGesture`, ...) are not required in config. Changed in [c3d63cc](https://github.com/software-mansion/react-native-gesture-handler/pull/3693/commits/c3d63cc99d7b711c6f4acc0156478dd7ee2ec378).

PR #3700, opened 2025-09-05
Properly type gesture config

m-bert:
## Description

This PR adds `BaseGestureConfig` type, so that `config` is no longer typed as `any`.

> [!NOTE]
> For now it introduces generic gesture config. Configs for handlers along with typed hooks will be introduced in follow-ups.

## Test plan

`yarn ts-check`


packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -66,20 +55,18 @@ export enum ComposedGestureName {
   Race = 'RaceGesture',
 }
 
-export type GestureEvents = {
-  onGestureHandlerStateChange: (
-    event: StateChangeEvent<Record<string, unknown>>
-  ) => void;
+export type GestureEvents<THandlerData> = {
```
latekvo, line 58: I understand what `THandlerData` does, but I don't think i ever really saw this pattern in our codebase, and it doesn't really change how the types work. I'd say it's less clear what it does than a reguler `T`, just because of how conventional `T` is compared to `THandlerData`.

I would prefer to stick to `T` instead of `THandlerData`, but that's just my opinion.

akwasniewski: I like the idea of using `THandlerData`, it is clear at first glance that this is the type of the events `handlerData` field. Moreover since we have both `THandlerData` and `TConfig` so we have to use the full name for one of them anyway, I think it's better to use for both for consisency.

m-bert: To be honest I don't have strong opinion in this exact case. Since we have only one type argument here, we can go with `T`. WDYT @j-piasecki?

j-piasecki: Not a strong opinion, but I like `THandlerData`. From the definition it's obvious what part of the type is generic, and when there are more types, then it's clear which type is which.

If consistency is the goal, we can always update existing types to follow this pattern.


> Since we have only one type argument here, we can go with T

What about places where there are more than one arguments (`THandlerData, TConfig`)?

j-piasecki (approved):
LGTM, feel free to update the generic type names if you reach a different conclusion

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: > What about places where there are more than one arguments (`THandlerData`, `TConfig`)?

In those I'd leave it as it is.

m-bert:
> LGTM, feel free to update the generic type names if you reach a different conclusion

I think I'll keep `THandlerData` for now. What I also thought about is switching the order of those two, but I don't think it matters for now. 

In the meantime, I've also removed [this unnecessary cast](https://github.com/software-mansion/react-native-gesture-handler/pull/3700/commits/0a672ee1772e3b32450a014beb8391723ad16286).

PR #3701, opened 2025-09-09
[Tests] Tests for relations traversal algorithm 

m-bert:
## Description

This PR adds `jest` tests for our `DFS` algorithm that traverses gesture relations tree and configures relations on the native side.

## Test plan

Well... `yarn test RelationsTraverse`


m-bert:
I've converted it to draft since I forgot to include external relations in these test cases 😅

j-piasecki:
Can we also have a github action that would run jest?

m-bert:
> Can we also have a github action that would run jest?

Sure. Would running only these test be sufficient?

j-piasecki:
Why not all js tests?

Re: some old tests are failing, will be investigated in a follow-up.

PR #3706, opened 2025-09-15
Typed gesture hooks

m-bert:
## Description

This PR introduces typed hooks for new API.

## Test plan

<details>

```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  usePan,

  const gesture = usePan({
    onUpdate: (e) => {
      'worklet';
      console.log(e.handlerData);
    },
  });

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>

Also on other handlers.


packages/react-native-gesture-handler/src/v3/hooks/gestures/index.ts
```diff
@@ -0,0 +1,26 @@
...
+export type { RotationGestureConfig } from './useRotation';
+export { useRotation } from './useRotation';
+
+export type { HoverGestureConfig } from './useHover';
+export { useHover } from './useHover';
+
+export type { ManualGestureConfig } from './useManual';
+export { useManual } from './useManual';
+
+export type { NativeViewGestureConfig } from './useNative';
```
m-bert, line 22: Can change to `NativeGestureConfig` (or something else) if you prefer.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -0,0 +1,255 @@
...
+  PanHandlerData,
+  PanGestureInternalProps
+>;
+
+const PanPropsMapping = new Map<
+  keyof PanGestureProps,
+  keyof PanGestureInternalProps
+>([['minDistance', 'minDist']]);
+
+function validatePanConfig(config: PanGestureConfig) {
```
m-bert, line 133: This one is taken from [PanGestureHandler](https://github.com/software-mansion/react-native-gesture-handler/blob/18be3b30617161f5d11d21a3b7e711571cb1fc39/packages/react-native-gesture-handler/src/handlers/PanGestureHandler.ts#L161), if you prefer to import it instead of duplicating let me know.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  if (config.minDistance && (config.activeOffsetX || config.activeOffsetY)) {
+    throw new Error(
+      `It is not supported to use minDistance with activeOffsetX or activeOffsetY`
+    );
+  }
+}
+
+function transformPanProps(
```
m-bert, line 183: Same as above ([link](https://github.com/software-mansion/react-native-gesture-handler/blob/18be3b30617161f5d11d21a3b7e711571cb1fc39/packages/react-native-gesture-handler/src/handlers/PanGestureHandler.ts#L211)), though slightly modified.

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -1,21 +1,37 @@
...
+export type BaseHandlerData = {
+  numberOfPointers: number;
+  pointerType: PointerType;
+};
+
+export type HandlerData<T> = BaseHandlerData & T;
+
 export type GestureUpdateEvent<THandlerData> = EventPayload & {
-  handlerData: THandlerData;
+  handlerData: HandlerData<THandlerData>;
```
m-bert: If you have better idea for these names, let me know. Maybe 

```js
GestureUpdateEvent<T>
```
and then 

```js
handlerData: HandlerData<T>;
```

would be better?

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -95,3 +96,30 @@ export function shouldHandleTouchEvents<THandlerData, TConfig>(
     !!config.onTouchesCancelled
   );
 }
+
+export function cloneConfig<THandlerData, TConfig>(
+  config: ExcludeInternalConfigProps<BaseGestureConfig<THandlerData, TConfig>>
+): BaseGestureConfig<THandlerData, TConfig> {
+  return { ...config } as BaseGestureConfig<THandlerData, TConfig>;
```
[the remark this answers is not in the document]
m-bert: That's on purpose, we need this return type.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (Array.isArray(config.activeOffsetX)) {
+      config.activeOffsetXStart = config.activeOffsetX[0];
+      config.activeOffsetXEnd = config.activeOffsetX[1];
+    } else if (config.activeOffsetX < 0) {
+      config.activeOffsetXStart = config.activeOffsetX;
+    } else {
+      config.activeOffsetXEnd = config.activeOffsetX;
+    }
+
+    delete config.activeOffsetX;
```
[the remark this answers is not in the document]
m-bert: That's what should be happening.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (Array.isArray(config.activeOffsetY)) {
+      config.activeOffsetYStart = config.activeOffsetY[0];
+      config.activeOffsetYEnd = config.activeOffsetY[1];
+    } else if (config.activeOffsetY < 0) {
+      config.activeOffsetYStart = config.activeOffsetY;
+    } else {
+      config.activeOffsetYEnd = config.activeOffsetY;
+    }
+
+    delete config.activeOffsetY;
```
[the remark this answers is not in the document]
m-bert: Same as above.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (Array.isArray(config.failOffsetX)) {
+      config.failOffsetXStart = config.failOffsetX[0];
+      config.failOffsetXEnd = config.failOffsetX[1];
+    } else if (config.failOffsetX < 0) {
+      config.failOffsetXStart = config.failOffsetX;
+    } else {
+      config.failOffsetXEnd = config.failOffsetX;
+    }
+
+    delete config.failOffsetX;
```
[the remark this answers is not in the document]
m-bert: Same as above.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (Array.isArray(config.failOffsetY)) {
+      config.failOffsetYStart = config.failOffsetY[0];
+      config.failOffsetYEnd = config.failOffsetY[1];
+    } else if (config.failOffsetY < 0) {
+      config.failOffsetYStart = config.failOffsetY;
+    } else {
+      config.failOffsetYEnd = config.failOffsetY;
+    }
+
+    delete config.failOffsetY;
```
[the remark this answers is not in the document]
m-bert: Same as above.

packages/react-native-gesture-handler/src/v3/types.ts
akwasniewski: Yeah, I think we should only use <THandlerData> rather than <T> when it gives additional info. Here we already see that it corrresponds to handler data. Therefore I would name it with <T>.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
akwasniewski: If it's the same, I think we should import.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
akwasniewski: If it's modified I think it can stay duplicated, though please add a comment before we forget they are duplicated.

m-bert: What do you expect to be in this comment? Just a mention that it is duplicate? If so, I don't think it is necessary.

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: > Therefore I would name it with .

I assume you meant `T`?

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: The only problem would be if we decide to get rid of API v1. Then, we need to bring it back. 

WDYT @j-piasecki?

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
akwasniewski: Just a mention if it is a duplicate, so if we ever change one we know we have to change the other one. But it's not a strong opinion

packages/react-native-gesture-handler/src/v3/types.ts
akwasniewski: Yes

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: I don't know if we want to make any changes in v1 at this point, especially given that it is deprecated. I can add a comment if you insist, but I don't think we need that.

cc @j-piasecki

m-bert: Actually, it will be further modified to correctly handle `SharedValues`, so I don't think this comment is necessary.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: Actually we can't do that as we have to handle `SharedValues` later.

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
```diff
@@ -7,9 +7,9 @@ import { tagMessage } from '../../utils';
 import { configureRelations } from './utils';
 import { isComposedGesture } from '../hooks/utils/relationUtils';
 
-export interface NativeDetectorProps {
+export interface NativeDetectorProps<THandlerData, TConfig> {
```
j-piasecki, line 10: Does this need to be generic? I'm not sure we're getting anything from those not being `unknown`.

packages/react-native-gesture-handler/src/v3/hooks/gestures/useFling.ts
```diff
@@ -0,0 +1,51 @@
+import {
+  BaseGestureConfig,
+  ExcludeInternalConfigProps,
+  SingleGestureName,
+} from '../../types';
+import { useGesture } from '../useGesture';
+import { cloneConfig } from '../utils';
+
+type FlingGestureProps = {
```
j-piasecki: I'd prefer that those not be called `_Props`, since they aren't props.

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
j-piasecki, line 196: Is this the same config as the one passed by the user?

If so, this may be a bit dirty with memoization, i.e.:

```js
const config = useMemo(() => ({
  activeOffsetX: [0, 1]
}), []);

const pan = usePan(config);

// config is now { activeOffsetXStart: 0, activeOffsetXEnd: 1 }
```

I'm not sure if we want it to behave that way.

packages/react-native-gesture-handler/src/v3/NativeDetector/NativeDetector.tsx
m-bert: Unfortunately yes, if it isn't, then we get ts errors that we cannot assign `SingleGesture` with types into `SingleGesture<unknown, unknown>` (events lack required fields) 😞

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: No, `remapProps` [creates shallow copy of config](https://github.com/software-mansion/react-native-gesture-handler/blob/78a8f1a9abd28989d4386ba4b2e42439b5c3a182/packages/react-native-gesture-handler/src/v3/hooks/utils.ts#L112).

But I can see that it might be confusing that some hooks call `cloneConfig` while others don't. To avoid much changes, what about calling `cloneConfig` in all hooks, and then passing this copy into `remapProps`?

packages/react-native-gesture-handler/src/v3/hooks/gestures/useFling.ts
m-bert: Okay, I can see the confusion here. Changed to `*Properties` in [dc472e1](https://github.com/software-mansion/react-native-gesture-handler/pull/3706/commits/dc472e11930d46f0717413e12f5c240a7b56acf2).

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: I've decided to add this approach in [5d0ef4d](https://github.com/software-mansion/react-native-gesture-handler/pull/3706/commits/5d0ef4d5a5fcbd4906c586558e1aa884d1cb63da).

j-piasecki (approved):

PR #3708, opened 2025-09-15
Add `changeEventCalculator` to handlers

m-bert:
## Description

This PR adds `changeEventCalculator` functions into continuous handlers.

## Test plan

<details>

```tsx
import {
  GestureHandlerRootView,
  NativeDetector,
  usePan,

  const gesture = usePan({
    onUpdate: (e) => {
      'worklet';
      console.log(e.handlerData.changeX);
    },
  });

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/web/tools/PointerEventManager.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/PointerEventManager.ts
@@ -1,5 +1,4 @@
 import EventManager from './EventManager';
-import { MouseButton } from '../../handlers/gestureHandlerCommon';
 import { AdaptedEvent, EventTypes, Point } from '../interfaces';
 import {
   PointerTypeMapping,
@@ -13,17 +12,10 @@ const POINTER_CAPTURE_EXCLUDE_LIST = new Set<string>(['SELECT', 'INPUT']);
 
 export default class PointerEventManager extends EventManager<HTMLElement> {
   private trackedPointers = new Set<number>();
-  private readonly mouseButtonsMapper = new Map<number, MouseButton>();
   private lastPosition: Point;
 
   constructor(view: HTMLElement) {
     super(view);
-
-    this.mouseButtonsMapper.set(0, MouseButton.LEFT);
-    this.mouseButtonsMapper.set(1, MouseButton.MIDDLE);
-    this.mouseButtonsMapper.set(2, MouseButton.RIGHT);
-    this.mouseButtonsMapper.set(3, MouseButton.BUTTON_4);
-    this.mouseButtonsMapper.set(4, MouseButton.BUTTON_5);
 
     this.lastPosition = {
       x: -Infinity,
@@ -213,7 +205,7 @@ export default class PointerEventManager extends EventManager<HTMLElement> {
       eventType: eventType,
       pointerType:
         PointerTypeMapping.get(event.pointerType) ?? PointerType.OTHER,
-      button: this.mouseButtonsMapper.get(event.button),
+      button: event.buttons,
       time: event.timeStamp,
       stylusData: tryExtractStylusData(event),
     };


packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -114,6 +115,8 @@ type PanHandlerData = {
   velocityX: number;
   velocityY: number;
   stylusData: StylusData;
+  changeX?: number;
+  changeY?: number;
```
j-piasecki: I think the only case where this will be undefined is when being used with Animated.

If that's the case, I'd prefer those to be typed as non-nullable and throw an error when validating mapping.

(same with the other events)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -121,3 +130,24 @@ export function remapProps<TConfig extends object, TInternalConfig>(
...
+    'worklet';
+    const currentEventData = current.handlerData;
+    const previousEventData = previous ? previous.handlerData : null;
+
+    const changePayload = diffCalculator(currentEventData, previousEventData);
+
+    const resultEvent = { ...current };
+    resultEvent.handlerData = { ...currentEventData, ...changePayload };
+
+    return resultEvent;
```
j-piasecki: This should be safe to do in-place. Is there something preventing that? I don't think Reanimated is freezing events.

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -39,6 +39,11 @@ export type GestureHandlerEvent<THandlerData> =
   | StateChangeEvent<THandlerData>
   | TouchEvent;
 
+export type ExtractedGestureHandlerEvent<THandlerData> =
```
j-piasecki: What do you think about `UnboxedGestureHandlerEvent` or `UnpackedGestureHandlerEvent`?

packages/react-native-gesture-handler/src/v3/hooks/gestures/useHover.ts
```diff
@@ -37,10 +40,23 @@ type HoverGestureInternalConfig = BaseGestureConfig<
...
   ExcludeInternalConfigProps<HoverGestureInternalConfig>;
 
+function diffCalculator(
+  current: HandlerData<HoverHandlerData>,
+  previous: HandlerData<HoverHandlerData> | null
+) {
+  'worklet';
+  return {
+    changeX: previous ? current.x - previous.x : current.x,
+    changeY: previous ? current.y - previous.y : current.y,
```
j-piasecki: I know this matches the current behavior, but I think it doesn't make sense to return the current coordinates when the previous event isn't defined. Zeroes would be better, as there's no change to the previous event since there isn't one.

packages/react-native-gesture-handler/src/v3/hooks/callbacks/updateHandler.ts
```diff
@@ -36,7 +41,6 @@ export function getUpdateHandler<THandlerData>(
         : event
     );
```
j-piasecki, line 42: Cannot add a comment to the relevant event, but is this correct?
```js
      changeEventCalculator
        ? changeEventCalculator(event, context.lastUpdateEvent)
        : event
```

It looks like it will just run the change calculator instead of calling the event.

m-bert: I'm not sure what you mean. It is placed inside `runCallback`, so we either pass `event` as argument, or we call `changeEventCalculator` in order to get event with additional fields.

packages/react-native-gesture-handler/src/v3/hooks/gestures/useHover.ts
m-bert: I agree, done in [312c954](https://github.com/software-mansion/react-native-gesture-handler/pull/3708/commits/312c9541b1a2de86eeba1850174d2e57d7682f63)

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: I think that `Unpacked` looks much better! Done in [16b1cc2](https://github.com/software-mansion/react-native-gesture-handler/pull/3708/commits/16b1cc22d14694ab36384e7f09285c0bd861824c)

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: We already have [this error](https://github.com/software-mansion/react-native-gesture-handler/blob/e93030d87c012b0294892e27061ae07dcb289758/packages/react-native-gesture-handler/src/v3/hooks/utils.ts#L57). Removed `?` in [68a24ca](https://github.com/software-mansion/react-native-gesture-handler/pull/3708/commits/68a24caae4bc0f3b369d77d1c4e76347bdebaf0f)

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Let me know what you think about [a28e2c9](https://github.com/software-mansion/react-native-gesture-handler/pull/3708/commits/a28e2c991eec1e3670b19671c36480994eee47d7)

packages/react-native-gesture-handler/src/v3/hooks/callbacks/updateHandler.ts
j-piasecki: Nvm, I didn't look at the signature of `changeEventCalculator` 🤦.

packages/react-native-gesture-handler/src/v3/hooks/gestures/useHover.ts
j-piasecki: I think that everything but hover was fine, no? `translation`, `scale`, or `rotation` for the first event can be defined, and now we are effectively ignoring those.

I meant specifically hover, which was returning the coords of the event. If you started hovering at `(100, 100)`, those would be the values of change props, which is clearly wrong.

packages/react-native-gesture-handler/src/v3/hooks/callbacks/updateHandler.ts
m-bert: We could change this name if you want to, I've just followed old implementation 😅

packages/react-native-gesture-handler/src/v3/hooks/gestures/useHover.ts
m-bert: Well, you're right 😅 Reverted in [771e25c](https://github.com/software-mansion/react-native-gesture-handler/pull/3708/commits/771e25cb4822b2683a5d5ba2f8fa4aa6780b0ebe) and `Hover` changed in [f399d36](https://github.com/software-mansion/react-native-gesture-handler/pull/3708/commits/f399d361bbc562b81f056b75b0f5f7ed4d4615fa)

packages/react-native-gesture-handler/src/v3/hooks/callbacks/updateHandler.ts
j-piasecki: I think it's fine for now. Unless you have an idea for another approach?

PR #3709, opened 2025-09-16
Remove remaining `any` in gesture config type

m-bert:
## Description

Seems like I've missed one place where `config` was typed as `any` 🙊 

## Test plan

`yarn ts-check`


akwasniewski (approved):
LGTM

PR #3689, opened 2025-08-25
Logic Detector

akwasniewski:
## Description

This PR implements new component -> `LogicDetector`. It resolves the issue of attaching gestures to inner SVG components. `LogicDetector` communicates with a `NativeDetector` higher in the hierarchy, which will be responsible for attaching gestures.

**Note:** attaching `Native` gestures to `LogicDetector` will be a added in a follow up, as it is a niche feature - thus not a priority - and we don't want to block this PR.

**Note 2:** Reanimated handlers currently only work on web for reanimated: ^4.1

## Test plan
tested on the following code
```tsx
import React from 'react';
import { Text, View, StyleSheet } from 'react-native';
import { NativeDetector, LogicDetector, useGesture, SingleGestureName } from 'react-native-gesture-handler';

import Svg, { Circle, Rect } from 'react-native-svg';

export default function SvgExample() {
  const circleElementTap = useGesture(SingleGestureName.Tap, {
    onStart: () => {
      'worklet';
      console.log('RNGH: clicked circle')
    }
  });
  const rectElementTap = useGesture(SingleGestureName.Tap, {
    onStart: () => {
      'worklet';
      console.log('RNGH: clicked parallelogram')
    }
  });
  const containerTap = useGesture(SingleGestureName.Tap, {
    onStart: () => {
      'worklet';
      console.log('RNGH: clicked container')
    }
  });
  const vbContainerTap = useGesture(SingleGestureName.Tap, {
    onStart: () => {
      'worklet';
      console.log('RNGH: clicked viewbox container')
    }
  });
  const vbInnerContainerTap = useGesture(SingleGestureName.Tap, {
    onStart: () => {
      'worklet';
      console.log('RNGH: clicked inner viewbox container')
    }
  });
  const vbCircleTap = useGesture(SingleGestureName.Tap, {
    onStart: () => {
      'worklet';
      console.log('RNGH: clicked viewbox circle')
    }
  });

  return (
    <View>
      <View style={styles.container}>
        <Text style={styles.header}>
          Overlapping SVGs with gesture detectors
        </Text>
        <View style={{ backgroundColor: 'tomato' }}>
          <NativeDetector gesture={containerTap}>
            <Svg
              height="250"
              width="250"
              onPress={() => console.log('SVG: clicked container')}>
              <LogicDetector gesture={circleElementTap}>
                <Circle
                  cx="125"
                  cy="125"
                  r="125"
                  fill="green"
                  onPress={() => console.log('SVG: clicked circle')}
                />
[...]

PR #3710, opened 2025-09-17
Filter config

m-bert:
## Description

This PR adds logic that filters `config` object before passing it to native side. 

## Test plan

<details>

```jsx
import {
  GestureHandlerRootView,
  NativeDetector,
  useTap,

  const gesture = useTap({
    onEnd: (e) => {
      'worklet';
      console.log('tap', e.handlerData);
    },

    numberOfTaps: 2,
  });

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/utils.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/utils.ts
@@ -144,7 +144,7 @@ export const PropsWhiteLists = new Map<
   [SingleGestureName.LongPress, LongPressNativeProperties],
 ]);
 
-const EMPTY_WHILE_LIST = new Set<string>();
+const EMPTY_WHITE_LIST = new Set<string>();
 
 export function prepareConfig<THandlerData, TConfig extends object>(
   handlerType: SingleGestureName,
@@ -153,7 +153,7 @@ export function prepareConfig<THandlerData, TConfig extends object>(
   // @ts-ignore Seems like TypeScript can't infer the type here properly because of generic
   const filteredConfig: BaseGestureConfig<THandlerData, TConfig> = {};
   const handlerPropsWhiteList =
-    PropsWhiteLists.get(handlerType) ?? EMPTY_WHILE_LIST;
+    PropsWhiteLists.get(handlerType) ?? EMPTY_WHITE_LIST;
 
   for (const [key, value] of Object.entries(config)) {
     // @ts-ignore That's the point, we want to see if key exists in the whitelists


packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -72,27 +72,45 @@ export function checkMappingForChangeProperties(animatedEvent: AnimatedEvent) {
...
 
-  for (const key in copy) {
-    // @ts-ignore It is fine to use string as index
-    if (Reanimated?.isSharedValue(copy[key])) {
-      // @ts-ignore It is fine to use string as index
-      copy[key] = copy[key].value;
+  for (const [key, value] of Object.entries(config)) {
+    if (PropsToFilter.has(key)) {
+      continue;
     }
```
j-piasecki: I think the whitelist would be a better pick here. We could have known passable props (the configs), known unpassable props (what is currently in `PropsToFilter`), and we could warn about anything else.

m-bert: Do you mean something like

```js
if (whitelist.has(prop)) {
  ...
} else if (PropsToFilter.has(prop)) {
  continue;
} else {
  console.warn(...)
}
```

?

j-piasecki: Basically, yes. There should be arrays of all possible values for each handler somewhere already.

m-bert: The problem here is that we cannot simply take specific gesture properties. Let's say we pick `Pan` props to be in whitelist, then, for example, `mouseButton` is not direct `Pan` property, therefore it will be ignored. 

On the other hand I can see why this may be useful, so we could make list of common props, then merge them with handlers' properties 😅

m-bert: This one [4ff105a](https://github.com/software-mansion/react-native-gesture-handler/pull/3710/commits/4ff105adba66c964254e912c3521a6724856788c) seems to do the job. Let me know what you think.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -79,27 +88,91 @@ export function checkMappingForChangeProperties(animatedEvent: AnimatedEvent) {
...
+  HandlersPropsWhiteList
+>([
+  [SingleGestureName.Pan, withCommonProps(panGestureHandlerProps)],
+  [SingleGestureName.Tap, withCommonProps(tapGestureHandlerProps)],
+  [SingleGestureName.Native, withCommonProps(nativeViewGestureHandlerProps)],
+  [SingleGestureName.LongPress, withCommonProps(longPressGestureHandlerProps)],
+  [SingleGestureName.Fling, withCommonProps(flingGestureHandlerProps)],
+]);
+
+const EMPTY_WHILE_LIST = new Set<string>();
```
m-bert: Similar thing will be added in #3689, so we will have to find good place to keep it 😅

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Actually not really, I forgot that this is function which prepares config to be passed into native side. In that case some of the props are wrongly white-listed, as `Pan` doesn't accept for example `activeOffsetX`, but rather `activeOffsetXStart` and `activeOffsetXEnd` etc. 

In that case I think we will have to make those lists, since they don't exist in old types.

m-bert: > I forgot that this is function which prepares config to be passed into native side.

Okay, I've fixed that... though to get rid of circular dependencies I had to split types and hooks, so it grew quite large 🙈 

Let me know what you think about this whitelist approach, and if you have any thoughts/ideas how to make it look better than it is now. Also, I'm not sure about how it will work with #3731 🙈 

cc @akwasniewski

akwasniewski: As far as I see it won't conflict with #3731. But we'll have to test to make sure.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
j-piasecki: Shouldn't it be `EMPTY_WHITE_LIST`?

PR #3713, opened 2025-09-18
[Web] Gesture relations

akwasniewski:
## Description

This PR adds a function in web module to configure gesture relations via hooks from  #3693. 

## Test plan
```ts
import {
  GestureHandlerRootView,
  NativeDetector,
  useSimultaneous,
  useGesture,
  useExclusive,
  useRace,
  SingleGestureName,

  const av = React.useRef(new Animated.Value(0)).current
  const event = Animated.event(
    [{ nativeEvent: { handlerData: { translationX: av } } }],
    {
      useNativeDriver: true,
    }
  );

  const tap1 = useGesture(SingleGestureName.Tap, {
    onEnd: () => {
      // 'worklet';
      console.log('Tap 1');
    },
    numberOfTaps: 1,
    disableReanimated: true,
  });

  const tap2 = useGesture(SingleGestureName.Tap, {
    onEnd: () => {
      // 'worklet';
      console.log('Tap 2');
    },
    numberOfTaps: 2,
    disableReanimated: true,
    blocksExternalGesture: tap1,
  });

  // const tap1 = useGesture('TapGestureHandler', {
  //   onEnd: () => {
  //     'worklet';
  //     console.log('Tap 1');
  //   },
  //   numberOfTaps: 1,
  //   // disableReanimated: true,
  //   requireExternalGestureToFail: tap2,
  // });

  const pan1 = useGesture(SingleGestureName.Pan, {
    // onUpdate: event,
    onUpdate: (e) => {
      'worklet';
      console.log('Pan 1');
    },
    // disableReanimated: true,
  });

  const pan2 = useGesture(SingleGestureName.Pan, {
    onUpdate: (e) => {
      'worklet';
      console.log('Pan 2');
    },
    simultaneousWithExternalGesture: pan1,
    // requireExternalGestureToFail: pan1,
    // blocksExternalGesture: pan1,
    // disableReanimated: true,
  });

  // const composedGesture = useSimultaneous(pan1, pan2);

  // const composedGesture = useExclusive(tap2, tap1);
  // const composedGesture = useExclusive(pan2, pan1); // For Animated.Event
  // const composedGesture = useExclusive(pan1, pan2); // For Animated.Event

  // const composedGesture = useRace(pan1, pan2);
  // const composedGesture = useRace(pan2, pan1);

  // const composedGesture = useExclusive(tap1, useSimultaneous(pan1, pan2));

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
        <NativeDetector gesture={pan1}>
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
                display: 'flex',
                alignItems: 'center',
              },
              { transform: [{ translateX: av }] },
            ]}>
              <Animated.View
              />
            </NativeDetector>
          </Animated.View>
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```


packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -726,6 +726,15 @@ export default abstract class GestureHandler implements IGestureHandler {
...
 
+    if (
+      config.waitFor !== undefined &&
+      config.simultaneousHandlers !== undefined &&
+      config.blocksHandlers !== undefined
+    ) {
+      // Compatiblity with old api, it will never happen on new api
+      InteractionManager.instance.configureInteractions(this, config);
+    }
+
```
m-bert: Aren't those 3 always defined? If so, this could be read as `if (true)` which doesn't make much sense. 

> Compatibility with old api, it will never happen on new api

Are you sure? `updateGestureConfig` is called regardless of the API version, isn't it?

packages/react-native-gesture-handler/src/web/tools/InteractionManager.ts
```diff
@@ -11,12 +12,15 @@ export default class InteractionManager {
...
+  public configureInteractions(
+    handler: IGestureHandler,
+    config: GestureRelations | Config
+  ) {
     this.dropRelationsForHandlerWithTag(handler.handlerTag);
 
     if (config.waitFor) {
       const waitFor: number[] = [];
-      config.waitFor.forEach((otherHandler: Handler): void => {
+      config.waitFor.forEach((otherHandler): void => {
```
m-bert: I'm not sure if removing types is the correct way, is it inferred correctly?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: I checked, the condition is only satisfied on old API. so it's not if(true). We cannot simply check the action type as update may change only a subset of the fields, then iirc they will be undefined even on old api

packages/react-native-gesture-handler/src/web/tools/InteractionManager.ts
akwasniewski: On old api `otherHandler` has `Handler` type and on new it has `number` type, thus we cannot type it here.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Ok, I forgot update is a v3 feature. Changed it to simply checking the action type in 5ec58f8f7d6a19b93aa7d703e51d30bee93b72ca

packages/react-native-gesture-handler/src/web/tools/InteractionManager.ts
m-bert: So can't we do `Handler | number`? And then correctly handle it inside?

akwasniewski: Ok, the types were inferred correctly so for me it looked cleaner without explicit typing, but it's not a strong opinion, thus i applied your suggestion in https://github.com/software-mansion/react-native-gesture-handler/pull/3713/commits/ea968ba5e48291787f8df67d0f93070693e9874e. Both cases are already handled correctly. However, I looked into it and as far as I can see `Handler` type is back from v1? Do we want to continue to support v1? It might be the time to cut this case out.

m-bert: It is not from V1, but I guess it was just a helper type to ensure that `handlerTag` exists. If types are inferred correctly then I guess it is fine, but IIRC in many cases we specify the type and it allows us to see what it really is without hovering on the variable.

m-bert: And even if it refers to V1, we don't want to remove it yet as it is not certain that we will drop V1 in 3.

packages/react-native-gesture-handler/src/web/tools/InteractionManager.ts
```diff
@@ -31,20 +35,22 @@ export default class InteractionManager {
...
-          simultaneousHandlers.push(otherHandler);
-        } else {
-          simultaneousHandlers.push(otherHandler.handlerTag);
+      config.simultaneousHandlers.forEach(
+        (otherHandler: Handler | number): void => {
+          if (typeof otherHandler === 'number') {
+            simultaneousHandlers.push(otherHandler);
+          } else {
+            simultaneousHandlers.push(otherHandler.handlerTag);
+          }
```
m-bert:
```suggestion
(otherHandler: Handler | number): void => {
  simultaneousHandlers.push(typeof otherHandler === 'number' ? otherHandler : otherHandler.handlerTag);
}
```

Also I think the `void` can be removed in that case.

m-bert (approved):
Looks good 😅 You don't have to apply these suggestions if you don't like them 😅

packages/react-native-gesture-handler/src/web/tools/InteractionManager.ts
akwasniewski: I like them, it looks cleaner. https://github.com/software-mansion/react-native-gesture-handler/pull/3713/commits/98ce5c9be6c98294b217bd62c2d45cfa1dc3466a

PR #3714, opened 2025-09-18
Update hook usage in the NativeDetector

j-piasecki:
## Description

1. Merges two `useEffects` with the same dependency array into one in `useGesture`
2. Adds `useEffect` responsible for updating the `propsRef` in `HostGestureDetector.web.tsx`
3. Removes `useEffects` which were only calling `detachHandlers` and replaces them with a cleanup function in `HostGestureDetector.web.tsx`

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
+++ b/packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
@@ -62,8 +62,8 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
   };
 
   useEffect(() => {
-    detachHandlers(attachedNativeHandlerTags.current);
-  }, [children]);
+    propsRef.current = props;
+  }, [props]);
 
   useEffect(() => {
     if (React.Children.count(children) !== 1) {
@@ -73,13 +73,11 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
     }
 
     attachHandlers(new Set(handlerTags));
-  }, [handlerTags, children]);
 
-  useEffect(() => {
     return () => {
-      detachHandlers(attachedHandlerTags.current);
+      detachHandlers(attachedNativeHandlerTags.current);
     };
-  }, []);
+  }, [handlerTags, children]);
 
   return (
     <View style={{ display: 'contents' }} ref={viewRef as Ref<View>}>
--- a/packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
+++ b/packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
@@ -72,10 +72,11 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
       );
     }
 
-    attachHandlers(new Set(handlerTags));
+    const newHandlerTags = new Set(handlerTags);
+    attachHandlers(newHandlerTags);
 
     return () => {
-      detachHandlers(attachedNativeHandlerTags.current);
+      detachHandlers(newHandlerTags);
     };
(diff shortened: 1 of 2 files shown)


packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
```diff
@@ -62,8 +62,8 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
   };
 
   useEffect(() => {
-    detachHandlers(attachedNativeHandlerTags.current);
-  }, [children]);
+    propsRef.current = props;
+  }, [props]);
```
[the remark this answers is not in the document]
j-piasecki: That's the point

packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
```diff
@@ -73,13 +73,11 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
...
 
     attachHandlers(new Set(handlerTags));
-  }, [handlerTags, children]);
 
-  useEffect(() => {
     return () => {
-      detachHandlers(attachedHandlerTags.current);
+      detachHandlers(attachedNativeHandlerTags.current);
     };
-  }, []);
```
akwasniewski, line 82: Why the cleanup only detaches native gesture handlers? When the component unmounts other gestures won't be detached

packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
```diff
@@ -62,8 +62,8 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
   };
 
   useEffect(() => {
-    detachHandlers(attachedNativeHandlerTags.current);
-  }, [children]);
+    propsRef.current = props;
```
akwasniewski, line 65: Correct me if I am wrong, but without this useEffect, if child under native gesture changes (eg. we swap one scroll view with another) the gestures won't reattach to the new child.

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -94,21 +94,17 @@ export function useGesture<THandlerData, TConfig>(
   }, [type, tag]);
 
   useEffect(() => {
+    const preparedConfig = prepareConfig(config);
+    RNGestureHandlerModule.setGestureHandlerConfig(tag, preparedConfig);
+    RNGestureHandlerModule.flushOperations();
+
```
m-bert, line 100: I think it would look better if `bindSharedValues` would be called first, but I don't think that this order matters so we can leave that 😄

packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
j-piasecki: Nice catch - fixed in [1fb5444](https://github.com/software-mansion/react-native-gesture-handler/pull/3714/commits/1fb5444678e0f95cc85e0a52c784d9f243593910)

packages/react-native-gesture-handler/src/v3/NativeDetector/HostGestureDetector.web.tsx
j-piasecki: The other effect should be handling this case - it now runs attach/detach when either `handlerTags` or `children` change.

akwasniewski: Ok, I tested and it seems to work as intended.

PR #3715, opened 2025-09-18
Fix relations old api

akwasniewski:
## Description

After #3693 old api gesture relations did not work as they were never configured. This PR fixes this issue

## Test plan

Check if relations examples from [v2 docs](https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/gesture-composition) work

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -899,8 +899,7 @@ open class GestureHandler {
       if (config.hasKey(KEY_MOUSE_BUTTON)) {
         handler.mouseButton = config.getInt(KEY_MOUSE_BUTTON)
       }
-      if (config.hasKey(WAIT_FOR) && config.hasKey(SIMULTANEOUS_HANDLERS) && config.hasKey(BLOCKS_HANDLERS)) {
-        // Compatibility with old api, it will never happen on new api
+      if (handler.actionType != ACTION_TYPE_NATIVE_DETECTOR) {
         interactionManager.dropRelationsForHandlerWithTag(handler.tag)
         interactionManager.configureInteractions(handler, config)
       }
@@ -925,9 +924,6 @@ open class GestureHandler {
       private const val KEY_HIT_SLOP_HORIZONTAL = "horizontal"
       private const val KEY_HIT_SLOP_WIDTH = "width"
       private const val KEY_HIT_SLOP_HEIGHT = "height"
-      private const val WAIT_FOR = "waitFor"
-      private const val SIMULTANEOUS_HANDLERS = "simultaneousHandlers"
-      private const val BLOCKS_HANDLERS = "blocksHandlers"
 
       private fun handleHitSlopProperty(handler: GestureHandler, config: ReadableMap) {
         if (config.getType(KEY_HIT_SLOP) == ReadableType.Number) {
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -863,6 +863,7 @@ open class GestureHandler {
   abstract class Factory<T : GestureHandler> {
     abstract val type: Class<T>
     abstract val name: String
+
     protected abstract fun create(context: Context?): T
 
     fun create(context: Context?, handlerTag: Int): T = create(context).also { it.tag = handlerTag }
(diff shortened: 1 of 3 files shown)


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -898,6 +899,11 @@ open class GestureHandler {
       if (config.hasKey(KEY_MOUSE_BUTTON)) {
         handler.mouseButton = config.getInt(KEY_MOUSE_BUTTON)
       }
+      if (config.hasKey(WAIT_FOR) && config.hasKey(SIMULTANEOUS_HANDLERS) && config.hasKey(BLOCKS_HANDLERS)) {
```
m-bert: I don't like this condition, can't we check if `actionType` is `NativeDetector`?

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -182,6 +182,11 @@ - (void)updateConfig:(NSDictionary *)config
       RCTLogError(@"Cannot have all of top, bottom and height defined");
     }
   }
+
+  if (config[@"waitFor"] != nil and config[@"simultaneousHandlers"] != nil and config[@"blocksHandlers"] != nil) {
```
m-bert: Same here

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
akwasniewski: iirc when update happens and only a subset of the config has changed the unchanged fields will be undefined. If one is defined all must be, but it doesn't seem right to just check one.

akwasniewski: Ok, I forgot update is a v3 feature. Changed it to simply checking the action type in 362057d9916e265e151491138f078acff9c42041

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
akwasniewski: 362057d9916e265e151491138f078acff9c42041

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -863,7 +864,7 @@ open class GestureHandler {
   abstract class Factory<T : GestureHandler> {
     abstract val type: Class<T>
     abstract val name: String
-
+    private val interactionManager = RNGestureHandlerInteractionManager()
```
j-piasecki: This cannot be local to a single gesture, as it's responsible for deciding whether a gesture should fail or continue. It needs information about relations for all currently attached gestures.

akwasniewski: Oh, thanks, on web it is a singleton so I was certain that on android it is a singleton as well, I forgot to make sure.

akwasniewski: Fixed in d9f909863f51c55ccb413340bbec54af2a8522c4

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -107,6 +107,11 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
     val factory = RNGestureHandlerFactoryUtil.findFactoryForHandler<GestureHandler>(handler) ?: return
 
     factory.updateConfig(handler, config)
+
+    if (handler.actionType == GestureHandler.ACTION_TYPE_NATIVE_DETECTOR) {
+      interactionManager.dropRelationsForHandlerWithTag(handlerTag)
+      interactionManager.configureInteractions(handler, config)
+    }
```
m-bert: Shouldn't this be in `setGestureHandlerConfig`?

m-bert (changes requested):

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: As far as I remember, previous implementations rely only on setting the config, so yes, but that would need to be double-checked.

akwasniewski: Ok, I checked and you are right. Moved it toset a07db183c085b5a7d154c45c4ab0bb9d14c3a0db

m-bert (approved):

PR #3717, opened 2025-09-18
Add `SharedValue` type to config.

m-bert:
## Description

This PR updates types to accept `SharedValues` into config objects.

## Test plan

<details>

```tsx
import {
  GestureHandlerRootView,
  MouseButton,
  NativeDetector,
  usePan,

  const sv1 = useSharedValue(0);
  const sv2 = useSharedValue(MouseButton.LEFT);

  const gesture = usePan({
    onEnd: (e) => {
      'worklet';
      console.log('tap', e.handlerData);
    },

    activeOffsetX: [sv1, 120],
    mouseButton: sv2,
  });

  return (
    <GestureHandlerRootView
      <Button
        title="Toggle visibility"
        onPress={() => {
          setVisible(!visible);
        }}
      />

      {visible && (
          <Animated.View
            style={[
              {
                width: 150,
                height: 150,
                backgroundColor: 'blue',
                opacity: 0.5,
                borderWidth: 10,
                borderColor: 'green',
                marginTop: 20,
                marginLeft: 40,
              },
            ]}
          />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}
```

</details>


packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -133,50 +139,50 @@ const PanPropsMapping = new Map<
...
+  if (!Array.isArray(offsets)) {
+    return;
   }
 
-  if (
-    Array.isArray(config.failOffsetX) &&
-    (config.failOffsetX[0] > 0 || config.failOffsetX[1] < 0)
-  ) {
+  const first = maybeUnpackValue(offsets[0]);
+  const second = maybeUnpackValue(offsets[1]);
```
MatiPl01: Is `first` the min value and `second` the max allowed value here? If yes, can you use  more explicit names like `minValue` and `maxValue`?

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -133,50 +139,50 @@ const PanPropsMapping = new Map<
...
+
+  if (first > 0 || second < 0) {
     throw new Error(
-      `First element of failOffsetX should be negative, and the second one should be positive`
+      `First element of ${propName} should be negative, and the second one should be positive`
```
MatiPl01: If the above comment applies and they are min and max values, shouldn't we also check if the max is >= than min?

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -133,50 +139,50 @@ const PanPropsMapping = new Map<
...
+function validatePanConfig(config: PanGestureConfig) {
+  validateOffsetsArray(config.activeOffsetX, 'activeOffsetX');
+  validateOffsetsArray(config.activeOffsetY, 'activeOffsetY');
+  validateOffsetsArray(config.failOffsetX, 'failOffsetX');
+  validateOffsetsArray(config.failOffsetY, 'failOffsetY');
 
   if (
-    Array.isArray(config.failOffsetY) &&
-    (config.failOffsetY[0] > 0 || config.failOffsetY[1] < 0)
+    config.minDistance &&
```
MatiPl01: What if someone passes `0` as the `minDistance` value? Don't we want to throw error as well?

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
```diff
@@ -190,10 +196,14 @@ function transformPanProps(
...
-      config.activeOffsetXStart = config.activeOffsetX;
     } else {
-      config.activeOffsetXEnd = config.activeOffsetX;
+      const offsetValue = maybeUnpackValue(config.activeOffsetX);
+
+      if (offsetValue < 0) {
+        config.activeOffsetXStart = config.activeOffsetX;
+      } else {
+        config.activeOffsetXEnd = config.activeOffsetX;
+      }
```
MatiPl01: Since this and the following 3 if blocks (handling of `activeOffsetX`, `activeOffsetY`, `failOffsetX`, `failOffsetY` props) are similar, is it possible to extract this logic to a separate helper function?

I can see a pattern here that you only add the `Start` and the `End` suffix to the prop name. What about creating a function like this:

```ts
function transformOffsetProp(config: PanGestureConfig & PanGestureInternalConfig, propName: keyof OffsetProps) {
  const propValue = config[propName];
  
  if (propValue !== undefined) {
    if (Array.isArray(propValue)) {
      config[`${propName}Start`] = propValue[0];
      config[`${propName}End`] = propValue[1];
    } else {
      const offsetValue = maybeUnpackValue(propValue);

      if (offsetValue < 0) {
        config[`${propName}Start`] = propValue;
      } else {
        config[`${propName}End`] = propValue;
      }
    }

    delete config[propName];
  }
}

function transformPanProps(
  config: PanGestureConfig & PanGestureInternalConfig
) {
  transformOffsetProp(config, 'activeOffsetY');
  transformOffsetProp(config, 'failOffsetX');
  transformOffsetProp(config, 'failOffsetY');
  transformOffsetProp(config, 'activeOffsetX');
}
```

I extracted these 4 prop types to the `OffsetProps` type:

```ts

type OffsetProps = {
    /**
     * Range along X axis (in points) where fingers travels without activation of
     * handler. Moving outside of this range implies activation of handler. Range
     * can be given as an array or a single number. If range is set as an array,
     * first value must be lower or equal to 0, a the second one higher or equal
     * to 0. If only one number `p` is given a range of `(-inf, p)` will be used
     * if `p` is higher or equal to 0 and `(-p, inf)` otherwise.
     */
    activeOffsetY?: number | [number, number];

    /**
     * Range along X axis (in points) where fingers travels without activation of
     * handler. Moving outside of this range implies activation of handler. Range
     * can be given as an array or a single number. If range is set as an array,
     * first value must be lower or equal to 0, a the second one higher or equal
     * to 0. If only one number `p` is given a range of `(-inf, p)` will be used
     * if `p` is higher or equal to 0 and `(-p, inf)` otherwise.
     */
    activeOffsetX?: number | [number, number];

    /**
     * When the finger moves outside this range (in points) along Y axis and
     * handler hasn't yet activated it will fail recognizing the gesture. Range
     * can be given as an array or a single number. If range is set as an array,
     * first value must be lower or equal to 0, a the second one higher or equal
     * to 0. If only one number `p` is given a range of `(-inf, p)` will be used
     * if `p` is higher or equal to 0 and `(-p, inf)` otherwise.
     */
    failOffsetY?: number | [number, number];

    /**
     * When the finger moves outside this range (in points) along X axis and
     * handler hasn't yet activated it will fail recognizing the gesture. Range
     * can be given as an array or a single number. If range is set as an array,
     * first value must be lower or equal to 0, a the second one higher or equal
     * to 0. If only one number `p` is given a range of `(-inf, p)` will be used
     * if `p` is higher or equal to 0 and `(-p, inf)` otherwise.
     */
    failOffsetX?: number | [number, number];
  }
```

MatiPl01: Maybe add these changes in a separate PR as this one is meant to add only type changes.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -53,6 +54,12 @@ export function isAnimatedEvent<THandlerData>(
   return !!callback && '_argMapping' in callback;
 }
 
+export function maybeUnpackValue<T>(v: TOrSharedValue<T>) {
+  'worklet';
+
```
MatiPl01, line 59:
```suggestion
```

nit: Not sure if this line break was intended here

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -199,3 +205,52 @@ export type ExcludeInternalConfigProps<T> = Omit<
...
+  set(value: Value | ((value: Value) => Value)): void;
+  addListener: (listenerID: number, listener: (value: Value) => void) => void;
+  removeListener: (listenerID: number) => void;
+  modify: (
+    modifier?: <T extends Value>(value: T) => T,
+    forceUpdate?: boolean
+  ) => void;
+}
+
+type ExcludeUndefined<T> = T extends undefined ? never : T;
```
MatiPl01: Chat suggested sth like that `type ExcludeUndefined<T> = Exclude<T, undefined>;`. Maybe it can be simplified in this way?

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -199,3 +205,52 @@ export type ExcludeInternalConfigProps<T> = Omit<
...
+        : // Default case: apply the MaybeWithSharedValue logic recursively or as a direct SharedValue wrap.
+          MaybeWithSharedValue<T[K], P>;
+};
+
+export type TOrSharedValue<T> = T | SharedValue<ExcludeUndefined<T>>;
+
+// Utility type that decides whether to recurse for objects or apply SharedValue directly.
+type MaybeWithSharedValue<T, P> = T extends object
+  ? WithSharedValue<T, P>
+  : Simplify<TOrSharedValue<T>>;
```
MatiPl01: Maybe we would like to make the `MaybeWithSharedValue` the main type? It could be used e.g. for the `offsets` param in the `validateOffsetsArray` function then as follows:

```ts
function validateOffsetsArray(
  offsets: MaybeWithSharedValue<number | [number, number] | undefined>,
  propName: string
) { ... }
```

MatiPl01: I think that the current `WithSharedValue` can be renamed to sth like `WithSharedValueObject` or `WithSharedValueRecur` (indicating that it recursively traverses the object) and changed to the non-exported local type.

The current `MaybeWithSharedValue` type can be renamed to `WithSharedValue` and updated like this:

```ts
type MaybeWithSharedValue<T, P = never> = T extends object
  ? WithSharedValueRecur<T, P>
  : Simplify<TOrSharedValue<T>>;
```

<details>
<summary>Final types</summary>

```ts
type WithSharedValueRecur<T extends object, P> = {
  [K in keyof T]: ExcludeUndefined<T[K]> extends P
    ? Simplify<TOrSharedValue<T[K]>>
    : // Special case for boolean as passing `boolean` as P doesn't look ok.
      boolean extends T[K]
      ? boolean | SharedValue<boolean>
      : // Special handling for tuples [number, number].
        T[K] extends [number, number]
        ? [WithSharedValue<number, P>, WithSharedValue<number, P>]
        : // Default case: apply the WithSharedValue logic recursively or as a direct SharedValue wrap.
          WithSharedValue<T[K], P>;
};


export type WithSharedValue<T, P = never> = T extends object
  ? WithSharedValueRecur<T, P>
  : Simplify<TOrSharedValue<T>>;
```
</details>

MatiPl01 (commented):
Left a few suggestions, some of which I find might be worth implementing. Overall, looks good except some improvements that can be made.

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: In that case we can just use `Exclude<T, undefined>` directly. Done in [1a2a038](https://github.com/software-mansion/react-native-gesture-handler/pull/3717/commits/1a2a038180f85362a1e4a16180cb1b59b3efc4fc).

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: Given that all those "oneliners" have this break yes, it was 😅

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: Well, I thought about it but I'm not a fan of using template as object index 

```js
config[`${propName}Start`]
```

On the other hand, I don't like this logic being repeated again and again, so it might be worth it. WDYT @j-piasecki?

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: You're right, done in [bbb81b3](https://github.com/software-mansion/react-native-gesture-handler/pull/3717/commits/bbb81b39765f270471c15f12c52bf7c7f38d4c6a).

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: You're right, looks better when changed. Done in [3df1572](https://github.com/software-mansion/react-native-gesture-handler/pull/3717/commits/3df15720a4bc9b8fa6be27cf2d995935fedaae1f).

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: Assume other names, `a` for `first` and `b` for `second`.

Our [requirements](https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/pan-gesture#activeoffsetxvalue-number--number) are: 

```math
a \leq 0 \wedge b \geq 0
```

Otherwise we want to throw an error:

```math
\sim(a \leq 0 \wedge b \geq 0)
```

Now, this is equivalent to 


```math
a \gt 0 \vee b \lt 0
```

So the only tricky case is if both are `0`, but I'm not sure if that breaks anything 😅

MatiPl01: Thanks for the detailed explanation

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: Let me know what you think about [47fa845](https://github.com/software-mansion/react-native-gesture-handler/pull/3717/commits/47fa8450acdf82392503baac606012649a2ed3fb).

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
```diff
@@ -9,6 +9,7 @@ import {
   GestureHandlerEvent,
   GestureStateChangeEvent,
   GestureUpdateEvent,
+  TOrSharedValue,
```
```suggestion
  SharedValueOrT,
```

Imo it's less confusing this way since you're not wondering what it `TOr` 😆.

packages/react-native-gesture-handler/src/v3/hooks/gestures/useTap.ts
```diff
@@ -55,17 +56,17 @@ type TapGestureProperties = {
    * activated, it will fail to recognize the gesture.
    */
   maxDistance?: number;
-};
+}>;
 
-type TapGestureInternalProperties = {
+type TapGestureInternalProperties = WithSharedValue<{
```
j-piasecki, line 61: Does the internal type also need the SV wrapper? The values sent to the native side should be primitive, no?

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
j-piasecki: @m-bert templates should be fine as long as you type it properly.

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -181,7 +179,15 @@ export type BaseGestureConfig<THandlerData, TConfig> = ExternalRelations &
     mouseButton?: MouseButton;
     enableContextMenu?: boolean;
     touchAction?: TouchAction;
-  };
+  },
+  HitSlop | UserSelect | ActiveCursor | MouseButton | TouchAction
+>;
+
+export type BaseGestureConfig<THandlerData, TConfig> = ExternalRelations &
```
j-piasecki, line 186: Is there a benefit to applying `WithSharedValue` to each individual type as opposed to applying it here?

I know callbacks and relations shouldn't be SVs, but other parts could be handled here, no?

packages/react-native-gesture-handler/src/v3/hooks/gestures/useTap.ts
m-bert: At this point config still contains `SharedValues`. Internal configs are used for remapping props, like `maxDistance` to `maxDist` etc. Actual values are extracted later, in `useGesture` hook.

packages/react-native-gesture-handler/src/v3/hooks/utils.ts
m-bert: I didn't like it either 😞 Thanks for suggestion 😃 Changed in [94dfe3d](https://github.com/software-mansion/react-native-gesture-handler/pull/3717/commits/94dfe3dacc6485f44b6958951404d708fe6a8756)

packages/react-native-gesture-handler/src/v3/hooks/gestures/usePan.ts
m-bert: Okay, makes sense. Done in [d5c7189](https://github.com/software-mansion/react-native-gesture-handler/pull/3717/commits/d5c718961597c292b45fe68e68120bfa9f61fae1)

packages/react-native-gesture-handler/src/v3/types.ts
m-bert: So if I understand correctly, instead of doing 

```ts
type CommonGestureConfig = WithSharedValue<...>
```

you'd like to do 

```ts
export type BaseGestureConfig<THandlerData, TConfig> =
  ... &
  WithSharedValue<CommonGestureConfig, ...>;
```

Right? 

I don't think that it benefits us, as the only thing in `BaseGestureConfig` that should be wrapped with `WtihSharedValue` is `CommonGestureConfig`. Also I think it is cleaner that `CommonGestureConfig` clearly states that it may contain `SharedValues`

Or do you mean that we should also do `WithSharedValue<TConfig, ...>` here, and do not change for example `Pan` config?

j-piasecki: I meant the first one

PR #3718, opened 2025-09-19
Update `metro.config.js` in example apps

m-bert:
## Description

In `metro` 0.83.2 `exclusionList` was changed to `default` export (see [this commit](https://github.com/facebook/metro/commit/8610b2c9e5bce0b686d7ffc8e5f3b59868e245c5#diff-a6e36af38cd6fde5360c601ce8f2f02e4e6b350b7fea9b0f049ab63f3b8f5709)). This means that we have to update our configs, or otherwise we won't be able to start server. 

> [!NOTE]
> `macos-example` uses older versions of `react-native` so we don't have to update its config.

## Test plan

1. `yarn clean`
2. `yarn`
3. `yarn start` in both examples

### DIFF
--- a/apps/basic-example/metro.config.js
+++ b/apps/basic-example/metro.config.js
@@ -2,7 +2,7 @@ const { getDefaultConfig } = require('@react-native/metro-config');
 const { mergeConfig } = require('metro-config');
 
 const path = require('path');
-const exclusionList = require('metro-config/private/defaults/exclusionList');
+const exclusionList = require('metro-config/private/defaults/exclusionList').default;
 const escape = require('escape-string-regexp');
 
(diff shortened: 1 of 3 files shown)


j-piasecki (commented):
Where do the yarn.lock changes come from?

m-bert:
> Where do the yarn.lock changes come from?

`metro` changed from 0.83.1 to 0.83.2, hence `yarn.lock` changed. Without generating new lock `yarn` will install old versions.

I can go through all updated packages in `yarn.lock` if you want to.

j-piasecki:
I was confused because there was no `package.json` change

j-piasecki (approved):

PR #3726, opened 2025-10-01
[Web] Fix handling of `enabled` prop

m-bert:
## Description

Right now, if one changes `enabled` property of handlers to `false`, they can never activate again. In the following example:

```ts
const [enabled, setEnabled] = useState(false);
const pan = Gesture.Pan().enabled(enabled)
```

gesture will never become active, even if `enabled` changes to `true`. Same happens when `enabled` is set to `true` and then changed to `false` and `true` again. 

This happens because we set `enabled` to `true` by default in `setGestureConfig`, thus the following check:

```ts
if (config.enabled !== undefined && this.enabled !== config.enabled) { ... }
```
 passes only when `config.enabled` is `false`.

I've set `enabled` to `null` by default, and then if `config.enabled` is defined, it is assigned given value. Else default will be `true`. To fix the issue, I've removed reset of `enabled` in `resetConfig` method.

Alternative approach would be to follow Android implementation and always react to changes in `enabled` even in cases like `false` $\rightarrow$ `false` and `true` $\rightarrow$ `true`.

## Test plan

<details>

```tsx
import React from 'react';
import { Button, StyleSheet, View } from 'react-native';
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,
  NativeDetector,
  usePan,
import Animated, {
  useAnimatedStyle,
  useSharedValue,

  const [showDetector, setShowDetector] = React.useState(true);

  const [enablePan1, setEnablePan1] = React.useState(true);
  const enablePan2 = useSharedValue(true);

  const pan1 = Gesture.Pan().enabled(enablePan1);

  const pan2 = usePan({
    enabled: enablePan2,
  });

  const as = useAnimatedStyle(() => {
    return {
      backgroundColor: enablePan2.value ? 'lightgreen' : 'crimson',
    };
  });

  return (
      <Button
        title="Toggle old API enabled!"
        onPress={() => {
          setEnablePan1((prev) => !prev);
        }}
      />

      <GestureDetector gesture={pan1}>
        <View
          style={[
            styles.box,
            { backgroundColor: enablePan1 ? 'lightgreen' : 'crimson' },
          ]}
        />
      </GestureDetector>

      <Button
        title="Toggle new API enabled!"
        onPress={() => {
          enablePan2.value = !enablePan2.value;
        }}
      />

      <Button
        title="Toggle NativeDetetor"
        onPress={() => {
          setShowDetector((prev) => !prev);
        }}
      />

      {showDetector && (
          <Animated.View style={[styles.box, as]} />
        </NativeDetector>
      )}
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },

  box: {
    width: 200,
    height: 200,
    borderRadius: 20,
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
@@ -75,6 +75,8 @@ export class GestureHandlerWebDelegate
     this.removeContextMenuListeners();
     this._view = null;
     this.eventManagers = [];
+
+    this.isInitialized = false;
   }
 
   isPointerInBounds({ x, y }: { x: number; y: number }): boolean {
@@ -237,6 +239,8 @@ export class GestureHandlerWebDelegate
     this.eventManagers.forEach((manager) => {
       manager.unregisterListeners();
     });
+
+    this.isInitialized = false;
   }
 
   private ensureView(view: any): asserts view is HTMLElement {


packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -707,16 +707,28 @@ export default abstract class GestureHandler implements IGestureHandler {
...
     this.resetConfig();
     this.updateGestureConfig(config);
   }
 
   public updateGestureConfig(config: Config): void {
-    if (config.enabled !== undefined && this.enabled !== config.enabled) {
-      this._enabled = config.enabled;
-      this.delegate.onEnabledChange();
-    }
+    this.updateEnabled(config.enabled);
```
akwasniewski, line 736: Correct me if I am wrong, but wouldn't this work? I don't think we need to allow `enabled` to be null
```suggestion
if (config.enabled !== undefined) {
this.enabled = config.enabled;
}
if(this.enabled !== config.enabled) {
this.delagate.onEnabledChange();
}
```

m-bert: Let's break it down into two cases (note, I assume that default value for `this.enabled` is `true`):

1. `config.enabled === undefined`

In this case second `if` will always execute, regardless of `enabled` change. 

2. `config.enabled !== undefined`

Second `if` will never execute, as those values will be the same.

Or have I missed something?

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -48,17 +46,15 @@ export class GestureHandlerWebDelegate
       touchAction: this.view.style.touchAction,
     };
 
-    this.setUserSelect();
-    this.setTouchAction();
-    this.setContextMenu();
```
j-piasecki, line 54: Was that unnecessary?

packages/react-native-gesture-handler/src/web/tools/EventManager.ts
```diff
@@ -94,6 +94,15 @@ export default abstract class EventManager<T> {
     this.pointersInBounds.splice(index, 1);
   }
 
+  public enable(value: boolean | null) {
```
j-piasecki: `enable`/`disable` or `setEnabled`

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -707,16 +707,28 @@ export default abstract class GestureHandler implements IGestureHandler {
   // Handling config
   //
 
+  // Helper function to correctly set enabled property
+  private updateEnabled(enabled: boolean | null) {
+    if (enabled === undefined) {
```
j-piasecki, line 717: This will never happen if `enabled` is of type `boolean | null`

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -707,16 +707,28 @@ export default abstract class GestureHandler implements IGestureHandler {
...
+      this._enabled = true;
+    } else if (this._enabled !== enabled) {
+      this._enabled = enabled;
+    }
+
+    this.delegate.onEnabledChange();
+  }
+
   public setGestureConfig(config: Config) {
     this.resetConfig();
```
j-piasecki, line 731: If I understand correctly, the issue is that we reset the config here, but if the gesture was disabled before, we don't reattach the event listeners. Shouldn't we also try to attach listeners when resetting the config? At first glance, it looks like it could simplify this a lot (especially if it were to be done in a setter).

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
m-bert: With changes in this PR, `onEnabledChange` is called so this is no longer necessary. This is because `setConfig` resets `enabled` into `null` and then in `updateEnabled` we call this method on delegate. It didn't work that way before though, so calls in `init` were necessary.

packages/react-native-gesture-handler/src/web/tools/EventManager.ts
m-bert: Sure, changed in [24b57ac](https://github.com/software-mansion/react-native-gesture-handler/pull/3726/commits/24b57ac5b6ec4146891bf3eacc6db495e069f715).

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
m-bert: Right, actually this type should be `boolean | undefined` as if user doesn't provide this prop in config it will be undefined. Changed in [ad10a30](https://github.com/software-mansion/react-native-gesture-handler/pull/3726/commits/ad10a30dceb36836896243bacacb76baa7efa4c3).

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
m-bert: > Shouldn't we also try to attach listeners when resetting the config?

But in that case if we have `config` that has `enabled` set to `false` we will first attach and then immediately detach listeners. Is that something that we want?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
akwasniewski: Ok, you are right

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
j-piasecki: Can we reverse that and start with detached listeners which are then attached when config has `enabled !== false`?

m-bert: So instead of attaching listeners in `resetConfig`, we should detach them, and then attach if `enabled !== false`?

m-bert: > If I understand correctly, the issue is that we reset the config here, but if the gesture was disabled before, we don't reattach the event listeners.

Also, I don't know whether we will have any other logic that depends on `enabled` change, but I think that we should say that what the issue really is is that in current setup it is impossible to detect that `enabled` has changed to `true`*. This may (or may not) cause other troubles in the future.

\* If we go through `resetConfig` first, with `SharedValue` it should be fine.

j-piasecki: Do we need to start with listeners attached? I think with the changes this PR adds, we could remove that part of the logic and keep it centralized in `updateGestureConfig`:

- start with listeners detached
- pass config with `enabled: false`
- listeners stay detached
- pass config with `enabled: true`
- listeners attach
- pass config with `enabled: undefined`
- listeners stay attached
- pass config with `enabled: false`
- listeners detach

m-bert: If I understand correctly, you want to react only to `config.enabled`, without comparing it to previous value is that true? If so, what in situation where two subsequent `updateGestureConfig` will have `enabled: true`? I guess we could check if listeners are already attached and do nothing in that case, but I think it solves only part of the problem. 

Right now, `onEnabledChange` in `GestureHandlerWebDelegate` does some other things, i.e.:

```ts
    this.setUserSelect();
    this.setTouchAction();
    this.setContextMenu();
```

So even if we solve problem with listeners, we still have issues where those functions won't be called because we can't detect when `enabled` value changes to `true`.

j-piasecki: > If I understand correctly, you want to react only to config.enabled, without comparing it to previous value is that true?

No - IIRC, currently, we start every gesture with `enabled: true` and listeners attached. We could start every gesture with `enabled: false` and listeners detached. Then compare `config.enabled ?? true` with the previous value and attach/detach if needed.

m-bert: But wouldn't it make it the same problem, but with `false` instead of `true`? If we make `false` default value, and then in `updateGestureConfig` we receive `config.enabled = false`, we won't trigger `onEnabledChanged`, even if user changed it from `true`.

j-piasecki: I'm not sure I see it. Could you do a step-by-step on what needs to happen to get that?

m-bert: > We could start every gesture with enabled: false and listeners detached.

I assume this means that when we reset handler, we set `this.enabled` into `false`. Let's say that we have this example:

```js
const [enabled, setEnabled] = useState(true);
const pan = Gesture.Pan().enabled(enabled)
```

The flow here is:

1. Reset `this.enabled` to `false` (via calling `setGestureConfig`)
2. Compare `config.enabled` with `this.enabled`

At first this works fine, we compare `config.enabled` which is `true` with `this.enabled` which is `false`. But now, when we change `enabled` to `false`, we have the same comparison, but both values are `false`, so `onEnabledChange` will not be called as we did not detect change from `true` to `false` - we lost that information in `resetConfig`.

j-piasecki: Resolved offline - there's still an issue, no matter the starting value. Ternary logic is likely the simplest solution here.

m-bert:
I've changed it a bit and removed reset of `enabled` in `resetConfig` as discussed offline. I've also updated an example on which I've tested these changed. Let me know what you think (cc @j-piasecki)

PR #3727, opened 2025-10-01
[Web] Disable `onPointerMove` on hover on gestures other than `Hover`

m-bert:
## Description

I've noticed that `onPointerMove` callback is triggered on each gesture, even if correct button is not pressed. Though events are not send, it still means that handlers do unnecessary work under the hood. This PR changes `PointerEventManager` so that it doesn't trigger `onPointerMove` callback without active pointer on gestures other than `Hover`

## Test plan

Tested on example app with `console.log` added into `onPointerMove` in `PanGestureHandler`

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/utils.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/utils.ts
@@ -14,6 +14,7 @@ import {
   SingleGestureName,
   InternalConfigProps,
   HandlersPropsWhiteList,
+  GestureCallbacks,
 } from '../types';
 import { GestureTouchEvent } from '../../handlers/gestureHandlerCommon';
 import { tagMessage } from '../../utils';
@@ -110,8 +111,9 @@ const allowedNativeProps = new Set<
   'changeEventCalculator',
 ]);
 
-const PropsToFilter = new Set<BaseGestureConfig<unknown, unknown>>([
-  // Callbacks
+export const HandlerCallbacks = new Set<
+  keyof Required<GestureCallbacks<unknown>>
+>([
   'onBegin',
   'onStart',
   'onUpdate',
@@ -121,6 +123,10 @@ const PropsToFilter = new Set<BaseGestureConfig<unknown, unknown>>([
   'onTouchesMove',
   'onTouchesUp',
   'onTouchesCancelled',
+]);
+
+const PropsToFilter = new Set<BaseGestureConfig<unknown, unknown>>([
+  ...HandlerCallbacks,
 
   // Config props
   'changeEventCalculator',
--- a/packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
@@ -1,6 +1,7 @@
 import RNGestureHandlerModule from '../../../RNGestureHandlerModule';
 import { Reanimated } from '../../../handlers/gestures/reanimatedWrapper';
 import { BaseGestureConfig, SharedValue } from '../../types';
+import { HandlerCallbacks } from '../utils';
 
 // Variant of djb2 hash function.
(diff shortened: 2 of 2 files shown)


packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -12,6 +12,7 @@ import { MouseButton } from '../../handlers/gestureHandlerCommon';
 import KeyboardEventManager from './KeyboardEventManager';
 import WheelEventManager from './WheelEventManager';
 import { tagMessage } from '../../utils';
+import HoverGestureHandler from '../handlers/HoverGestureHandler';
```
j-piasecki: Can we handle that in another way (like a method or property)?

m-bert: Sure, we can after #3729 is merged 😅

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -52,7 +53,11 @@ export class GestureHandlerWebDelegate
     this.setTouchAction();
     this.setContextMenu();
 
-    this.eventManagers.push(new PointerEventManager(this.view));
+    const shouldSendHoverEvents = handler instanceof HoverGestureHandler;
+
+    this.eventManagers.push(
+      new PointerEventManager(this.view, shouldSendHoverEvents)
+    );
```
akwasniewski: Why the unnecessary const? I think if you want to clarify what it does it is better to use a comment
```suggestion
    this.eventManagers.push(
      new PointerEventManager(this.view, handler instanceof HoverGestureHandler)
    );
```

akwasniewski (changes requested):

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
m-bert: I don't agree with that. With this named variable you can immediately see what it does. It is much better then looking at confusing code and then reading comment (cc @j-piasecki)

akwasniewski:
Correct me if I am wrong, but shouldn't `onPointerMove` be also triggered for `ManualGestures`?

m-bert:
> Correct me if I am wrong, but shouldn't `onPointerMove` be also triggered for ManualGestures?

tl;dr 
I don't think it is necessary to consider this.


This PR doesn't disable `onPointerMove` callbacks completely, but only if there's no active pointer on the given view. If we follow android implementation, you can see that the only method that `Manual` overrides is [onHandle ](https://github.com/software-mansion/react-native-gesture-handler/blob/3a28871b4a5a5aa3bc3ed2f193f418a5b574beac/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/ManualGestureHandler.kt#L8). Now, `onHandle` is called inside `handle` method in parent class. If you look into implementation, you'll see that [first it checks for active pointers](https://github.com/software-mansion/react-native-gesture-handler/blob/3a28871b4a5a5aa3bc3ed2f193f418a5b574beac/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt#L347). So it will behave in the same way 😅

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
j-piasecki: It's merged 🎉

m-bert: Right, I've changed it in [ba1ac1b](https://github.com/software-mansion/react-native-gesture-handler/pull/3727/commits/ba1ac1b63a3a2f2f7045cab4118596dd0f61e116) 😅

akwasniewski (approved):

PR #3729, opened 2025-10-02
[Web] Add `name` property to handlers

m-bert:
## Description

This PR adds `name` property to handlers on web, so that we can avoid checks like in `InteractionsManager`:

```js
   const isNativeHandler =
      otherHandler.constructor.name === 'NativeViewGestureHandler';
```

## Test plan

Tested on **Swipeable** example on mobile version in chrome.

### DIFF
--- a/packages/react-native-gesture-handler/src/web/interfaces.ts
+++ b/packages/react-native-gesture-handler/src/web/interfaces.ts
@@ -172,16 +172,3 @@ export type GestureHandlerRef = {
 export type SVGRef = {
   elementRef: { current: SVGElement };
 };
-
-export enum GestureHandlerName {
-  Tap = 'TapGestureHandler',
-  Pan = 'PanGestureHandler',
-  LongPress = 'LongPressGestureHandler',
-  Pinch = 'PinchGestureHandler',
-  Rotation = 'RotationGestureHandler',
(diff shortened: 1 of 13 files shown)


packages/react-native-gesture-handler/src/web/interfaces.ts
```diff
@@ -172,3 +172,16 @@ export type GestureHandlerRef = {
...
+  Pan = 'PanGestureHandler',
+  LongPress = 'LongPressGestureHandler',
+  Pinch = 'PinchGestureHandler',
+  Rotation = 'RotationGestureHandler',
+  Fling = 'FlingGestureHandler',
+  NativeView = 'NativeViewGestureHandler',
+  ForceTouch = 'ForceTouchGestureHandler',
+  Manual = 'ManualGestureHandler',
+  Hover = 'HoverGestureHandler',
+}
```
j-piasecki: Don't we already have that in types.ts?

m-bert: My bad, I had old API in mind and forgot that we target `next`. Used `SingleGestureName` in [cc87901](https://github.com/software-mansion/react-native-gesture-handler/pull/3729/commits/cc87901907c31580c08dfb530028ccacf7da9ca8) instead.

PR #3731, opened 2025-10-03
Add missing types

akwasniewski:
## Description

While writing examples I noticed that some types would be useful. This includes types for each gesture and their events. This PR adds those types.

**Important**: The v3 gesture types take over the names of old v2 types, the v2 types are getting a `Legacy` prefix. This change will have to be added to the migration guide. Added this as a [task](https://github.com/orgs/software-mansion/projects/31/views/1?pane=issue&itemId=132253302) to roadmap.

## Test plan

`yarn lint-js`


m-bert (commented):
Looks ok. Do we need `SingleGestureType` exported?

packages/react-native-gesture-handler/src/index.ts
```diff
@@ -168,7 +168,11 @@ export { NativeDetector } from './v3/NativeDetector/NativeDetector';
 export * from './v3/hooks/useGesture';
 export * from './v3/hooks/relations';
 
-export { SingleGestureName } from './v3/types';
+export {
+  SingleGestureName,
+  SingleGestureType,
+  ComposedGesture as ComposedGestureType,
```
j-piasecki: Hmm, what do you think about going in the other direction and exporting `SingleGesture`, `LongPressGesture`, ...?

That will conflict with the API v2, so we may want to rename those and keep that in mind for the migration guide.

packages/react-native-gesture-handler/src/index.ts
```diff
@@ -168,7 +168,11 @@ export { NativeDetector } from './v3/NativeDetector/NativeDetector';
 export * from './v3/hooks/useGesture';
 export * from './v3/hooks/relations';
 
-export { SingleGestureName } from './v3/types';
+export {
+  SingleGestureName,
```
j-piasecki: Why is that needed?

packages/react-native-gesture-handler/src/v3/types.ts
```diff
@@ -106,6 +106,8 @@ export type SingleGesture<THandlerData, TConfig> = {
   gestureRelations: GestureRelations;
 };
 
+export type SingleGestureType = SingleGesture<unknown, unknown>;
```
j-piasecki: What do you think about a union: `TapGestureType | PanGestureType | ...`

It would have to be defined elsewhere, though.

akwasniewski:
> Looks ok. Do we need `SingleGestureType` exported?

I figured there is a need for a more general type than handlers corresponding to each gesture, so the user doesn't have to do `PanGesture | TapGesture | ...`. `Gesture` and `GestureType` were taken so I figured I can stick with `SingleGesture` name

packages/react-native-gesture-handler/src/index.ts
akwasniewski: Yeah, that would be prefarable, and was my first choice, but I didn't want to conflict with v2 api. If you think it is Ok I will gladly change it.

packages/react-native-gesture-handler/src/index.ts
akwasniewski: `SingleGestureName`? I'm preety sure it is no longer needed after we added dedicated gesture hooks, am I correct cc @m-bert? I could remove it in this PR

m-bert: Yes, I think it is no longer required. It was with `useGesture` but now it is only internal (but please make sure that it's true before removing it 😅)

packages/react-native-gesture-handler/src/v3/types.ts
akwasniewski: Great idea, done in https://github.com/software-mansion/react-native-gesture-handler/pull/3731/commits/43bd77d4d1c02683ee846a8b90f84dfff591a599

packages/react-native-gesture-handler/src/index.ts
j-piasecki: Yeah, I think we can rename the old types to something else to free up the "makes sense" names for the future. We would have to include that in the migration guide.

packages/react-native-gesture-handler/src/v3/hooks/gestures/index.ts
```diff
@@ -1,26 +1,67 @@
+import { FlingGestureEvent, FlingGesture } from './useFling';
+import { HoverGestureEvent, HoverGesture } from './useHover';
+import { LongPressGestureEvent, LongPressGesture } from './useLongPress';
+import { ManualGestureEvent, ManualGesture } from './useManual';
+import { NativeGestureEvent, NativeGesture } from './useNative';
+import { PanGestureEvent, PanGesture } from './usePan';
+import { PinchGestureEvent, PinchGesture } from './usePinch';
+import { RotationGestureEvent, RotationGesture } from './useRotation';
+import { TapGestureEvent, TapGesture } from './useTap';
```
m-bert: If this imports only types, please use `import type` instead of only `import`

packages/react-native-gesture-handler/src/index.ts
m-bert: I'll just point out that we should write it somewhere in case we forget. Not sure if doing `TODO` comment` is the best, but but we could include it in the roadmap.

packages/react-native-gesture-handler/src/v3/hooks/gestures/index.ts
akwasniewski: Ok, done in 704d604722fe511cadf2d5d958629069600a2ae6

packages/react-native-gesture-handler/src/v3/hooks/gestures/useFling.ts
```diff
@@ -52,3 +55,12 @@ export function useFling(config: FlingGestureConfig) {
 
   return useGesture(SingleGestureName.Fling, flingConfig);
 }
+
+export type FlingGestureEvent =
+  | GestureStateChangeEvent<FlingHandlerData>
+  | GestureUpdateEvent<FlingHandlerData>;
```
m-bert: Do we actually want to combine these? At the end of the day, those events are accessed in different callbacks

akwasniewski: You are right, it is better to separate it. Done in 5113634c2dd173e7242d90f67ce8b133533252b3

packages/react-native-gesture-handler/src/index.ts
```diff
@@ -169,7 +169,8 @@ export { LogicDetector } from './v3/LogicDetector';
 export * from './v3/hooks/useGesture';
 export * from './v3/hooks/relations';
 
-export { SingleGestureName } from './v3/types';
+export type { ComposedGesture } from './v3/types';
+export type { GestureTouchEvent as SingleGestureTouchEvent } from './handlers/gestureHandlerCommon';
```
akwasniewski, line 161: I added this reexport to keep naming consistency, but I'm not sure whether this is the best approach. Let me know what you think

packages/react-native-gesture-handler/src/index.ts
```diff
@@ -54,20 +54,20 @@ export { default as createNativeWrapper } from './handlers/createNativeWrapper';
...
+export type { HoverGestureType } from './handlers/gestures/hoverGesture';
 export type {
-  ComposedGestureType as ComposedGesture,
-  RaceGestureType as RaceGesture,
-  SimultaneousGestureType as SimultaneousGesture,
-  ExclusiveGestureType as ExclusiveGesture,
+  ComposedGestureType,
+  RaceGestureType,
+  SimultaneousGestureType,
+  ExclusiveGestureType,
```
j-piasecki: Not sure about that, `TapGestureType` may still be confused with `TapGesture`. What about `OldTapGesture` / `LegacyTapGesture`?

akwasniewski: Ok, changed it to `LegacyTapGesture`, though I'm not sure it is the best name.

packages/react-native-gesture-handler/src/index.ts
akwasniewski: Removed it in c7a25efa33292dd2a0ecfffab7dda8d91772482f

packages/react-native-gesture-handler/src/index.ts
m-bert: `Legacy*Gesture` sounds fine 😅

packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFling.ts
```diff
@@ -32,3 +35,12 @@ export function useFling(config: FlingGestureConfig) {
 
   return useGesture(SingleGestureName.Fling, flingConfig);
 }
+
+export type FlingGestureStateChangeEvent =
+  GestureStateChangeEvent<FlingHandlerData>;
+export type FlingGestureUpdateEvent = GestureUpdateEvent<FlingHandlerData>;
```
m-bert, line 41: Maybe we could move it (along with `FlingHandlerData` to `FlingProperties` file (and probably rename it so it would make sense). 

But this one is also fine, if we decide to change it we can do it later 😅

m-bert (approved):

PR #3732, opened 2025-10-03
One detector to rule them all

akwasniewski:
## Description

To simplify migration the `GestureDetector` will check the type of the gesture it receives and render the detector accordingly it:
* Checks if the gesture it received is V2 or V2:
* If it's V2, it renders the old `GestureDetector` component
* If it's V3, it checks for the `InterceptingGestureDetector` context:
* If context is there, it's inside boundary and renders `LogicDetector`
* If the context is not there, it renders `NativeDetector`

### Why separate `NativeDetector` and `InterceptingGestureDetector`?

We did some [performance tests](https://github.com/software-mansion/react-native-gesture-handler/pull/3689#issuecomment-3350411993) on the `NativeDetector` after adding changes that handle attaching `LogicDetector`, which revealed that the new logic adds quite a bit of overhead, but only on the JS side. New logic on the native side does not seem to have a significant effect. 
We concluded that the best solution is to create a separate component that will have all functionalities of `NativeDetector` and also allow `LogicDetector` attachment, while NativeDetector will be reverted to how it had been before implementing `LogicDetector`.

## Test plan

Rendering `LogicDetector`:
<details>

```tsx
import React from 'react';
import { Text, View, StyleSheet } from 'react-native';
import { GestureDetectorBoundary, GestureDetector, useTap, GestureHandlerRootView } from 'react-native-gesture-handler';

import Svg, { Circle, Rect } from 'react-native-svg';

export default function SvgExample() {
  const circleElementTap = useTap({
    onStart: () => {
      'worklet';
      console.log('RNGH: clicked circle')
    },
  });
  const rectElementTap = useTap({
    onStart: () => {
      'worklet';
      console.log('RNGH: clicked parallelogram')
    },
  });

  return (
    <GestureHandlerRootView>
        <Text style={styles.header}>
          Overlapping SVGs with gesture detectors
        </Text>
        <View style={{ backgroundColor: 'tomato' }}>
          <GestureDetectorBoundary>
            <Svg
              height="250"
              width="250"
              onPress={() => console.log('SVG: clicked container')}>
              <GestureDetector gesture={circleElementTap}>
                <Circle
                  cx="125"
                  cy="125"
                  r="125"
                  fill="green"
                  onPress={() => console.log('SVG: clicked circle')}
                />
              </GestureDetector>
              <GestureDetector gesture={rectElementTap}>
                <Rect
                  skewX="45"
                  width="125"
                  height="250"
                  fill="yellow"
                  onPress={() => console.log('SVG: clicked parallelogram')}
                />
              </GestureDetector>
            </Svg>
          </GestureDetectorBoundary>
        </View>
        <Text>
          Tapping each color should read to a different console.log output
        </Text>
      </View>
    </GestureHandlerRootView>
  );
}

  container: {
    alignItems: 'center',
    justifyContent: 'center',
    marginBottom: 48,
  },
  header: {
    fontSize: 18,
    fontWeight: 'bold',
    margin: 10,
  },
});
```

</details

Rendering `NativeDetector`:
<details>

```tsx
import React from 'react';
import {
  GestureDetector,
  useTap,
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withTiming,
  Easing,
  interpolateColor,

export default function SimpleTap() {
  const colorValue = useSharedValue(0);

  const tapGesture = useTap({
    onStart: () => {
      'worklet';
      colorValue.value = withTiming(colorValue.value === 0 ? 1 : 0, {
        duration: 400,
        easing: Easing.inOut(Easing.ease),
      });
    },
  });

      colorValue.value,
      [0, 1],
      ['#b58df1', '#ff7f50'] // purple → coral
    );

    return {
      backgroundColor,
    };
  });

  return (
    <View style={styles.centerView}>
      </GestureDetector>
    </View>
  );
}

  centerView: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
  box: {
    height: 120,
    width: 120,
    backgroundColor: '#b58df1',
    marginBottom: 30,
    borderRadius: 12,
  },
});
```

</details

Rendering old v2 `GestureDetector`:
<details>

```tsx
import React from 'react';
import {
  Gesture,
  GestureDetector,
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  withTiming,
  Easing,
  interpolateColor,

export default function SimpleTap() {
  const colorValue = useSharedValue(0);

  const tapGesture = Gesture.Tap()
    .onStart(() => {
      'worklet';
      colorValue.value = withTiming(colorValue.value === 0 ? 1 : 0, {
        duration: 400,
        easing: Easing.inOut(Easing.ease),
      });
    });

      colorValue.value,
      [0, 1],
      ['#b58df1', '#ff7f50'] // purple → coral
    );

    return {
      backgroundColor,
    };
  });

  return (
    <View style={styles.centerView}>
      </GestureDetector>
    </View>
  );
}

  centerView: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
  box: {
    height: 120,
    width: 120,
    backgroundColor: '#b58df1',
    marginBottom: 30,
    borderRadius: 12,
  },
});

```

</details

### DIFF
--- a/apps/basic-example/src/Text.tsx
+++ b/apps/basic-example/src/Text.tsx
@@ -3,8 +3,7 @@ import { StyleSheet, Text, View } from 'react-native';
 import {
   Gesture,
   GestureDetector,
-  LogicDetector,
-  NativeDetector,
+  InterceptingGestureDetector,
   useTap,
 } from 'react-native-gesture-handler';
 
@@ -37,24 +36,24 @@ function NativeDetectorExample() {
       <Text style={styles.header}>
         Native Detector example - this one should work
       </Text>
-      <NativeDetector gesture={tapAll}>
+      <InterceptingGestureDetector gesture={tapAll}>
         <Text style={{ fontSize: 18, textAlign: 'center' }}>
           Some text example running with RNGH
-          <LogicDetector gesture={tapFirstPart}>
+          <GestureDetector gesture={tapFirstPart}>
             <Text style={{ fontSize: 24, color: COLORS.NAVY }}>
               {' '}
               try tapping on this part
             </Text>
-          </LogicDetector>
-          <LogicDetector gesture={tapSecondPart}>
+          </GestureDetector>
+          <GestureDetector gesture={tapSecondPart}>
             <Text style={{ fontSize: 28, color: COLORS.KINDA_BLUE }}>
               {' '}
               or on this part
             </Text>
-          </LogicDetector>
+          </GestureDetector>
           this part is not special :(
         </Text>
-      </NativeDetector>
+      </InterceptingGestureDetector>
     </View>
   );
 }


akwasniewski:
Currently the new Detector is named `DelegateDetector` as LogicDetector delegates their gestures to it. However, I'm not sure if it is a good name, I couldn't think of a better one. I'd appreciate any input on the naming.

packages/react-native-gesture-handler/src/index.ts
```diff
@@ -162,10 +162,12 @@ export type {
 } from './components/DrawerLayout';
 export { default as DrawerLayout } from './components/DrawerLayout';
 
-export type { NativeDetectorProps } from './v3/NativeDetector/NativeDetector';
-export { NativeDetector } from './v3/NativeDetector/NativeDetector';
+export type { NativeDetectorProps } from './v3/Detectors/common';
+export { NativeDetector } from './v3/Detectors/NativeDetector';
+
+export { LogicDetector } from './v3/Detectors/LogicDetector/LogicDetector';
+export { DelegateDetector } from './v3/Detectors/LogicDetector/DelegateDetector';
```
m-bert: Maybe let's also create `index.ts` in `v3/detectors` (and yes, I'd use lowercase for that directory 😅) which will export these?

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/useDetectorContext.ts
```diff
@@ -14,7 +14,9 @@ export const DetectorContext = createContext<DetectorContextType | null>(null);
 export function useDetectorContext() {
   const ctx = useContext(DetectorContext);
   if (!ctx) {
-    throw new Error('Logic detector must be a descendant of a Native Detector');
+    throw new Error(
+      'Logic detector must be a descendant of a delegate detector'
```
m-bert:
```suggestion
      tagMessage('Logic detector must be a descendant of a delegate detector')
```

packages/react-native-gesture-handler/src/index.ts
akwasniewski: Oki-dokie, I renamed directory in 88d4290f9de24e8107cdd00efa680b9f8f389268 and created `detectors/index` in b7e9fcdd10838724661a611358cf622e00d8ebfb

j-piasecki:
What do you think about changing the naming (and logic) this way:
- rename `DelegateDetector` to something along the lines of `GestureDetectorBoundary`
- update `GestureDetector` so it:
  1. Checks if the gesture it received is V2 or V2:
  2. If it's V2, it renders whatever it's rendering currently
  3. If it's V3, it checks for the `GestureDetectorBoundary` context:
    - If context is there, it's inside boundary and renders `LogicalDetector`
    - If the context is not there, it renders `NativeDetector`

This way we have a `GestureDetector` component which works with both APIs (which was a plan since the beginning), and the only change needed to migrate is to update `Gesture.*` to `use*`, and add `GestureDetectorBoundary` around SVGs, etc.

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/useDetectorContext.ts
akwasniewski: Added `tagMessage` and moved the check into `LogicDetector` in https://github.com/software-mansion/react-native-gesture-handler/pull/3732/commits/90cf5d5f0cb302102cdbfdbe59efc11d7a5ffb80

akwasniewski:
> What do you think about changing the naming (and logic) this way:
> 
>     * rename `DelegateDetector` to something along the lines of `GestureDetectorBoundary`
> 
>     * update `GestureDetector` so it:
>       
>       1. Checks if the gesture it received is V2 or V2:
>       2. If it's V2, it renders whatever it's rendering currently
>       3. If it's V3, it checks for the `GestureDetectorBoundary` context:
>       
>       
>       * If context is there, it's inside boundary and renders `LogicalDetector`
>       * If the context is not there, it renders `NativeDetector`
> 
> 
> This way we have a `GestureDetector` component which works with both APIs (which was a plan since the beginning), and the only change needed to migrate is to update `Gesture.*` to `use*`, and add `GestureDetectorBoundary` around SVGs, etc.

It seems that it is the right moment for that, so I implemented this functionality. Now we have one tag - `GestureDetector`, that renders either `NativeDetector`, `LogicDetector` or the old `GestureDetector` (v2) based on the gesture type and context.

akwasniewski:
I'm not sure about the name of the `GestureDetectorBoundary`, but couldn't figure out a better one so I settled on this one. Any ideas cc @m-bert? If that name stays I don't think it should accept any gestures as props, but ideally I would like to leave this functionality, while changing the name.

j-piasecki:
> I'm not sure about the name of the GestureDetectorBoundary, but couldn't figure out a better one so I settled on this one. Any ideas cc @m-bert? If that name stays I don't think it should accept any gestures as props, but ideally I would like to leave this functionality, while changing the name.

Maybe `InterceptingGestureDetector` when keeping the option to attach gestures directly to it?

packages/react-native-gesture-handler/src/index.ts
```diff
@@ -150,10 +150,8 @@ export type {
 } from './components/Pressable';
 export { default as Pressable } from './components/Pressable';
 
-export type { NativeDetectorProps } from './v3/NativeDetector/NativeDetector';
-export { NativeDetector } from './v3/NativeDetector/NativeDetector';
+export * from './v3/detectors';
```
j-piasecki: I'd avoid "export everything" syntax - we may expose some internal stuff accidentally.

In this case, I don't think we should be exporting `NativeDetector`, `NativeDetectorProps`, and `LogicDetector`. The end users should rely on `GestureDetector`, which will auto-select the correct one.

Exported props type should also be an union of all the relevant types.

packages/react-native-gesture-handler/src/v3/detectors/GestureDetector.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { BaseGesture } from '../../handlers/gestures/gesture';
+import { NativeDetector } from './NativeDetector';
+import { ComposedGesture } from '../../handlers/gestures/gestureComposition';
+import {
+  GestureDetectorProps as LegacyGestureDetectorProps,
+  GestureDetector as LegacyGestureDetector,
+} from '../../handlers/gestures/GestureDetector';
+import { useDetectorContext } from './LogicDetector/useDetectorContext';
+import { LogicDetector } from './LogicDetector/LogicDetector';
+export function GestureDetector<THandlerData, TConfig>(
```
m-bert, line 13:
```suggestion
import { LogicDetector } from './LogicDetector/LogicDetector';

export function GestureDetector<THandlerData, TConfig>(
```

Please 🙏

packages/react-native-gesture-handler/src/v3/detectors/GestureDetector.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return props.gesture instanceof ComposedGesture ||
+    props.gesture instanceof BaseGesture ? (
+    <LegacyGestureDetector {...(props as LegacyGestureDetectorProps)} />
+  ) : context ? (
+    <LogicDetector {...(props as NativeDetectorProps<THandlerData, TConfig>)} />
+  ) : (
+    <NativeDetector
+      {...(props as NativeDetectorProps<THandlerData, TConfig>)}
+    />
+  );
```
j-piasecki:
```suggestion
  if (props.gesture instanceof ComposedGesture ||
    props.gesture instanceof BaseGesture) {
    return <LegacyGestureDetector {...(props as LegacyGestureDetectorProps)} />;
  }

  const context = use(DetectorContext);
  return context ? (
    <LogicDetector {...(props as NativeDetectorProps<THandlerData, TConfig>)} />
  ) : (
    <NativeDetector
      {...(props as NativeDetectorProps<THandlerData, TConfig>)}
    />
  );
```

If this works, then there's no need to read context when rendering the legacy detector (not sure if syntax is correct, but you get the idea).

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/LogicDetector.tsx
```diff
@@ -1,19 +1,27 @@
...
+import { tagMessage } from '../../../utils';
 
 export function LogicDetector<THandlerData, TConfig>(
   props: NativeDetectorProps<THandlerData, TConfig>
 ) {
-  const { register, unregister } = useDetectorContext();
+  const context = useDetectorContext();
+  if (!context) {
+    throw new Error(
+      tagMessage('Logic detector must be a descendant of a delegate detector')
```
j-piasecki: The error needs to be updated once we settle on the final name.

m-bert: Shouldn't we update this error message?

m-bert:
Also, do we have to use `DetectorContext.Provider`? It should be enough to just use `DetectorContext`.

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/LogicDetector.tsx
akwasniewski: Done in 103f7782ae5bb4d026c2cb8e598534f7316812bd

akwasniewski: Done in 103f7782ae5bb4d026c2cb8e598534f7316812bd

packages/react-native-gesture-handler/src/v3/detectors/GestureDetector.tsx
akwasniewski: Ok, i tried it before but did not realise ou can do `use(DetectorContext)`, so I got conditional hook usage. Thanks, done in eb6e762d505efdb5f143fc5fdd3a01af4c628de1

packages/react-native-gesture-handler/src/v3/detectors/GestureDetector.tsx
akwasniewski: 2c7df129e4183abdd7dfa1e54f0042cc15dfbea9

packages/react-native-gesture-handler/src/index.ts
akwasniewski: Sure, all done in 0de96b3b4d5114147a060aca47dcd8e03d470371

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/LogicDetector.tsx
```diff
@@ -60,12 +70,19 @@ export function LogicDetector<THandlerData, TConfig>(
       Object.assign(logicProps, { viewRef });
     }
 
-    register(logicProps, logicMethods as RefObject<GestureEvents<unknown>>);
+    register(
+      logicProps,
+      logicMethods as RefObject<GestureEvents<unknown>>,
+      props.gesture.config.shouldUseReanimatedDetector,
+      props.gesture.config.dispatchesAnimatedEvents
```
j-piasecki, line 69: Can this be folded into `logicProps`?

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/LogicDetector.tsx
```diff
@@ -1,19 +1,29 @@
...
 
 export function LogicDetector<THandlerData, TConfig>(
   props: NativeDetectorProps<THandlerData, TConfig>
 ) {
-  const { register, unregister } = useDetectorContext();
+  const context = useDetectorContext();
+  if (!context) {
+    throw new Error(
+      tagMessage(
+        'Logic detector must be a descendant of an intercepting decector'
```
j-piasecki:
```suggestion
        'Logic detector must be a descendant of an InterceptingGestureDetector'
```

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/InterceptingGestureDetector.tsx
```diff
@@ -112,19 +103,25 @@ export function NativeDetector<THandlerData, TConfig>({
...
   );
 
+  ensureNativeDetectorComponent(NativeDetectorComponent);
+
+  if (gesture) {
+    configureRelations(gesture);
+  }
+
   return (
     <DetectorContext.Provider value={{ register, unregister }}>
```
m-bert: As mentioned [here](https://github.com/software-mansion/react-native-gesture-handler/pull/3732#issuecomment-3450457093), isn't `DetectorContext` enough?

akwasniewski: Yes, it seems to be. Thanks, f7542cbd20359bdfc2960ee0ce1c2fae89ea7162

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/LogicDetector.tsx
akwasniewski: It could, but I intended `logicProps` to contain only the types passed down to native. We don't have to pass those value down, thus I think it should stay seperate.

packages/react-native-gesture-handler/src/v3/detectors/LogicDetector/LogicDetector.tsx
akwasniewski: 0dbbb2c0162b3000fceca84bd111b40a683cc421

akwasniewski:
After merging with main, some issues with gestureless `InterceptingGestureDetector` occured, it crashed on web and ios when switching between detector types. After a lot of playing around I've made some fixes, but I'm not sure whether they are the best way to fix the problem. Let me know. [IOS](https://github.com/software-mansion/react-native-gesture-handler/pull/3732/commits/a6e6b588d8c7bcfd89414f8945bea43f6fd5e420), [web](https://github.com/software-mansion/react-native-gesture-handler/pull/3732/commits/8bcb96f20e607c21dd5ab96b8814264d131aec97)

m-bert:
> After a lot of playing around I've made some fixes, but I'm not sure whether they are the best way to fix the problem. Let me know. [IOS](https://github.com/software-mansion/react-native-gesture-handler/pull/3732/commits/a6e6b588d8c7bcfd89414f8945bea43f6fd5e420), [web](https://github.com/software-mansion/react-native-gesture-handler/pull/3732/commits/8bcb96f20e607c21dd5ab96b8814264d131aec97)

How does [this commit](https://github.com/software-mansion/react-native-gesture-handler/pull/3732/commits/a6e6b588d8c7bcfd89414f8945bea43f6fd5e420) fix `iOS`?

j-piasecki:
What do you mean by "switching between detector types"?

akwasniewski:
>What do you mean by "switching between detector types"?

When gestureless `InterceptingGestureDetector` gets animated/reanimated gesture attached (and `register` is called) it switches its underlying component from `HostGestureDetector` to Animated/Reanimated wrappers over it, this caused some issued on web as for a moment the `viewRefs` of the children were undefined.

packages/react-native-gesture-handler/src/v3/detectors/HostGestureDetector.web.tsx
```diff
@@ -110,6 +110,11 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
     );
 
     props.logicChildren?.forEach((child) => {
+      if (child.viewRef.current == null) {
+        // We must check whether viewRef is  not null as otherwise we get an error when intercepting gesture detector
+        // switches its component based on whether animated/reanimated events should run.
+        return;
+      }
```
j-piasecki, line 117: If the viewRef isn't set when this effect is run, when are the gestures attached?

akwasniewski: From what I tested it is comes down to the fact, that when switching between the underlying components, the new detector will get the old array passed as `LogicChildren` with refs to children that don't exist anymore (where children of the other component). This will then be cleaned up quickly, but we want to make sure we don't get an error before that.

j-piasecki: Would it be possible to prevent that (passing old refs) from happening at all?

akwasniewski: I've tried using useEffect to clean up `logicMethods` and `logicChildren` when `shouldUseReanimated` or `dispatchesAnimatedEvents` changed, but it doesn't seem to work, this was the best solution that worked

j-piasecki (approved):

akwasniewski:
> What do you think about renaming LogicDetector to VirtualDetector/VirtualGestureDetector?

Yes, I think it should be done, however let's leave it for a follow up PR for clarity.

PR #3740, opened 2025-10-07
[iOS] Fix gestures not ending properly on `iOS` 26

### LINKED ISSUE #3733: iOS 26: Pan gesture events don't work properly
### Description

When I use the Pan Gesture handler on latest iOS 26, the events are not triggered same as they do on iOS 18.x, Particularly when it's just a touch without drag.

on iOS 26 if we:
- touch and then remove the gesture only onBegin is called.
- touch, drag then remove the gesture all events are triggered properly,

Please see the attached video, you can see the difference, sample code and the package version used:
https://github.com/user-attachments/assets/583a200b-f0dc-44de-a0c5-f46dd04b0aec

### Steps to reproduce

Just try the sample code in the attached video on latest iOS 26

[Expo Snack](https://snack.expo.dev/@ashu_dubey/gesture-handler-ios26-issue)

### Gesture Handler version

2.28.0

### React Native version

0.81.4

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 16 Pro Max (iOS 18.4) and iPhone 17 Pro Max (iOS 26.0)

### Acknowledgements

Yes

m-bert:
## Description

There's a change in gesture recognizers introduced in `iOS` 26. Now when `reset` method is called, recognizers go back to `UIGestureRecognizerStatePossible` state. This breaks our current behavior, because this state is mapped into `RNGestureHandlerStateBegan`, so if for example `Pan` fails, it tries to send event with `Began` state. 

Unfortunately, changing recognizer state is not possible outside of `touches*` methods, therefore we had to move `triggerAction` into those callbacks.

Let me know if you see a different approach into this problem. 

Fixes #3733 

> [!WARNING]
> `triggerAction` call was already present in `Tap` right before `reset` ([see here](https://github.com/software-mansion/react-native-gesture-handler/blob/21c4943d5769d3fff60f7bf0550c5810f6011e13/packages/react-native-gesture-handler/apple/Handlers/RNTapHandler.m#L120)). Looks like it was called twice for some reason (but I believe that [check for _lastState](https://github.com/software-mansion/react-native-gesture-handler/blob/21c4943d5769d3fff60f7bf0550c5810f6011e13/packages/react-native-gesture-handler/apple/RNGestureHandler.mm#L311) prevented any problems with this redundancy). For now I have not included second call. If you think it is required, let me know.

## Test plan

Tested on the code provided below, on the following platforms: 

- [x] iOS 26.0 (iPhone 17 Pro)
- [x] iOS 18.5 (iPhone 16e)
- [x] OSX (macOS 15.6.1)  

<details>
<summary>Test code:</summary>

```tsx

import {
  GestureHandlerRootView,
  Gesture,
  GestureDetector,
  GestureType,

function TestBox({
  gestureType,
  bgColor,
}: {
  gestureType: GestureType;
  bgColor: string;
}) {
  const handlerName = gestureType.handlerName;

  const gesture = gestureType
    .onEnd(() => {
      console.log(`[${handlerName}] onEnd`);
    })
    .onFinalize(() => {
      console.log(`[${handlerName}] onFinalize`);
    })
    .runOnJS(true);

  return (
    <View style={styles.center}>
      <Text>{handlerName}</Text>
        <View style={[styles.box, { backgroundColor: bgColor }]} />
      </GestureDetector>
    </View>
  );
}

  return (
    <GestureHandlerRootView style={[{ flex: 1, padding: 50 }, styles.center]}>
      <TestBox gestureType={Gesture.Pan()} bgColor="#b58df1" />
      <TestBox gestureType={Gesture.LongPress()} bgColor="#f1a85d" />
      <TestBox gestureType={Gesture.Fling()} bgColor="#5df1a8" />
      <TestBox gestureType={Gesture.Tap()} bgColor="#5d8ef1" />
    </GestureHandlerRootView>
  );
}

  center: {
    display: 'flex',
    alignItems: 'center',
  },
  box: {
    height: 100,
    width: 100,
    backgroundColor: '#b58df1',
    borderRadius: 20,
    marginBottom: 30,
  },
});
```

</details>


packages/react-native-gesture-handler/apple/RNGHUIKit.h
```diff
@@ -12,6 +12,8 @@ typedef UIScrollView RNGHUIScrollView;
 #define RNGHGestureRecognizerStateBegan UIGestureRecognizerStateBegan;
 #define RNGHGestureRecognizerStateEnded UIGestureRecognizerStateEnded;
 
+#define iOS_VERSION [[[UIDevice currentDevice] systemVersion] floatValue]
```
m-bert: Now, since `iOS_VERSION` macro returns `float`, we _may_ want to do comparisons up to a constant $\epsilon$, or round this value. I don't think it is necessary at this point, but let me know what you think.

akwasniewski: I dont't really see a case where this would be an issue. I know floats behave weirdly, but we will always be checking it via a \leq or \geq comparison with another float and and the mapping from real values to floats is monotonic - it keeps the ordering.

packages/react-native-gesture-handler/apple/Handlers/RNLongPressHandler.m
```diff
@@ -102,12 +102,20 @@ - (void)touchesEnded:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
...
+    [self triggerAction];
+  }
 }
 
 - (void)touchesCancelled:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
 {
   [super touchesCancelled:touches withEvent:event];
   [_gestureHandler.pointerTracker touchesCancelled:touches withEvent:event];
+
+  if (iOS_VERSION >= 26.0 && self.state == UIGestureRecognizerStateFailed) {
```
akwasniewski: If this appropach also works on the old iOS i think the best approach would be to have `triggerAction` only called in touches*, and never in reset. iOS 26 will become the default soon, so we might as well treat it as such.

m-bert: The only thing I'm worried about is that whether `reset` is called manually somewhere else, but I'll check that.

m-bert: Okay, there were some direct calls, I've added `triggerAction` before them 😅

akwasniewski:
I'll just mention that we might want to release this fix fast, in a release, as currently pan is broken on all iOS 26 devices.

mysport12:
Any chance this can get merged and pushed in a new release?

sabicd:
@akwasniewski hey there - is there a timeline for a new release?

m-bert:
> Any chance this can get merged and pushed in a new release?

I can't give you exact ETA, but it should be the beginning of the next week.

j-piasecki (approved):

PR #3745, opened 2025-10-14
Change logic of `worklet` detection in config

m-bert:
## Description

This PR fixes issue where using `Animated.Event` without `worklet` callbacks was throwing error. This was because current implementation of `hasWorkletEventHandlers` was checking all functions present in `config`, including `changeEventCalculator` (which is in fact `worklet`). 

> [!NOTE]
> It previously contained #3748

## Test plan

Tested on the current `basic-example` app


m-bert:
If you prefer to split these changes into separate PRs, let me know.

m-bert:
> Looks good, though I think it makes sense to split it into two PRs, as fixing Animated.Event is probably a temporary fix, so when we reference it in the future it's going to be cleaner.

Sure, done!

> Should we add a task with a proper fix to the roadmap?

We could. The question is whether we will forget about it or not 😄 If there's already a task connected with `LogicDetector` (I'm not sure if it is marked as done or not), we could add it to description.

akwasniewski:
> If there's already a task connected with LogicDetector (I'm not sure if it is marked as done or not), we could add it to description.

I don't think we ever created one. I guess we might add it to the one with fixing `LogicDetector` handling of NativeViewGestures

j-piasecki (approved):

PR #3747, opened 2025-10-14
Fix swipeable on iOS

akwasniewski:
## Description

On iOS the swipeable component did not register events from child button components. This PR fixes this issue.

## Test plan

Open `v3examples/swipeable ` example in the example app and check whether you get an alert after tapping one of the rows.

### DIFF
--- a/.github/workflows/npm-gesture-handler-publish.yml
+++ b/.github/workflows/npm-gesture-handler-publish.yml
@@ -1,0 +1,55 @@
+name: Build package and publish on npm
+env:
+  YARN_ENABLE_HARDENED_MODE: 0
+on:
+  workflow_dispatch:
+
+jobs:
+  npm-build:
+    if: github.repository == 'software-mansion/react-native-gesture-handler'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write # for OIDC
+    env:
+      PACKAGE_NAME: PLACEHOLDER # Will be reassigned later on.
+    steps:
+      - name: Check out
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: 'yarn'
+          registry-url: https://registry.npmjs.org/
+
+      # Ensure npm 11.5.1 or later is installed for OIDC
+      - name: Update npm
+        run: npm install -g npm@latest
+
+      - name: Install node dependencies
+        run: yarn install --immutable
+
+      - name: Build package
+        id: build
+        working-directory: packages/react-native-gesture-handler
+        run: npm pack
+
+      - name: Add package name to env
+        working-directory: packages/react-native-gesture-handler
+        run: echo "PACKAGE_NAME=$(ls -l | egrep -o "react-native-gesture-handler-(.*)(=?\.tgz)")" >> $GITHUB_ENV
+
+      - name: Assert PACKAGE_NAME
+        if: ${{ env.PACKAGE_NAME == 'PLACEHOLDER' }}
+        run: exit 1 # If we end up here, it means that package was not generated.
+
+      - name: Upload npm package to GitHub
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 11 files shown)


akwasniewski:
> Could you explain why this fixes issue in example?
> 
> If I understand correctly, this `Tap` gesture is used to close swipeable when it is opened, so `enabled` can be `false` if swipeable is `closed`. If this is the only purpose of this gesture, then I agree that it is enough.
> 
> However, isn't there another underlying issue? That for some reason `Tap` blocks `Native` from `RectButton` on `iOS`, while it doesn't on `android`?
> 
> cc @j-piasecki

The swipeable tap gesture is only used to close the swipeable, and since it has no effect when it is closed we can simply disable it so it doesn't interefere with other gestures, currently it interferes with the tap gesture on iOS. Adding `enabled` prop was inspired by what had been done in old[ non-reanimated `Swipeable`](https://github.com/software-mansion/react-native-gesture-handler/blob/757265d7d776ca0e0ac3db847561290fe89f1ff2/packages/react-native-gesture-handler/src/components/Swipeable.tsx#L561C1-L563C37). Without the enabled flag the old swipeable also doesn't work on iOS.

  I've done some carefull testing, and it seems that the tap from `RectButton` is triggered only after the native tap fails, which is especially strange since RectButton is a child of the Swipeable, thus its gestures should get priority. What is even more bizzare, without this fix, on android both tap seem to be simultanous - they don't interfere with each other, which is also wrong as I think the `RectButton` one should cancel the native one. Only the web wersion works as I think is inteded. 

I settled on this fix, as if it was good enough for old Swipeable it probably is for the new one.

m-bert (approved):
> Only the web wersion works as I think is inteded.

I've not looked into it, is it that you get an `alert` from tapping, but `Swipeable` doesn't close?

> I settled on this fix, as if it was good enough for old Swipeable it probably is for the new one.

Sure 👍

packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
```diff
@@ -472,6 +472,7 @@ const Swipeable = (props: SwipeableProps) => {
   const tapGesture = useMemo(() => {
     const tap = Gesture.Tap()
       .shouldCancelWhenOutside(true)
+      .enabled(rowState.value !== 0)
```
j-piasecki: We can't read Shared values like that. I mean, we can, but it won't rerender the component when the value changes

j-piasecki (changes requested):

packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
akwasniewski: Oh, you are right. Should be fixed in [52fb5bc](https://github.com/software-mansion/react-native-gesture-handler/pull/3747/commits/52fb5bc8c6c041741af41505dcf6f356968f3f4a).

akwasniewski:
> I've not looked into it, is it that you get an alert from tapping, but Swipeable doesn't close?

No, no it's now that I meant, everything works as intended. However without this, when swipeable was closed we had two taps registered, one from `Swipeable` and one from `RectButton`. When swipeable on closed its tap gesture did not do anything, however it behaved weirdly.
* on android both taps were registered, as if they were marked as simultanous -> this in effect produced the intended behaviour, but probably shouldn't happen.
* on ios only the swipeable one was registered, resulting in a bug.
* on web only the RectButton one is registered -> this I think is what should happen, the gesture from the child component should cancel the gesture from parent component.

This illustrates that native gestures behave quite weirdly when combined with v1 gestures. (which are within `RectButton`). As long as it doesn't cause any other issues I don't think it's a big deal, it will probably go away if we decide to deprecate v1 after v3 is out.

m-bert:
> (...) if we decide to deprecate v1 after v3 is out.

It is already deprecated 😅

Issue #2920, opened 2024-05-20
GestureDetector is not responsive 


rezkiy37:
### Description

The app uses **GestureDetector** to handle a tap on the send button. **GestureDetector** works properly when the user opens a chat screen. However, if the user opens nested chats (_threads_), returns to the 1st one, and taps on the send button, **GestureDetector** does not react to this action. The expected behavior is to have the button responsive. 

**Note:** the bug is replicated for a case when there are 2 nested screens e.g. Main chat -> Subchat -> Subchat.

There is an issue with the **Expensify** app - https://github.com/Expensify/App/issues/41528.

<details><summary>Video</summary>
<p>

[...]

PR #3752, opened 2025-10-16
Fix SSR error with `useLayoutEffect`

akwasniewski:
## Description
Using `useLayoutEffect` in combination with SSR results in a warning. This PR adds [isomorphicLayoutEffect](https://github.com/reduxjs/react-redux/blob/7e2fdd4ee2021e4282e12ba9fc722f09124e30cd/src/utils/useIsomorphicLayoutEffect.ts#L36), which acts as a `useEffect` when running on server and as `useLayoutEffect` everywhere else. This fixes https://github.com/software-mansion/react-native-gesture-handler/issues/3375.

Note: This may cause #2920 to happen when running SSR, however as we couldn't replicate the issue on main, probably due to same changes to view recycling within react native, I think it would even be ok to simply revert https://github.com/software-mansion/react-native-gesture-handler/pull/2925, and simply only `useEffect`, so it shouldn't be an issue.

I decided to use `isomorphicLayoutEffect` instead of reverting to ensure correctness.

## Test plan

See that there is no warning mentioned in #3375

### DIFF
--- a/.github/workflows/android-build-paper.yml
+++ b/.github/workflows/android-build-paper.yml
@@ -31,10 +31,10 @@ jobs:
           distribution: oracle
           java-version: 17
 
-      - name: Use Node.js 18
-        uses: actions/setup-node@v4
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
         with:
-          node-version: 18
+          node-version: 24
           cache: yarn
 
       - name: Install node dependencies
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -33,10 +33,10 @@ jobs:
           distribution: oracle
           java-version: 17
 
-      - name: Use Node.js 18
-        uses: actions/setup-node@v4
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
         with:
-          node-version: 18
+          node-version: 24
           cache: yarn
 
       - name: Install node dependencies
--- a/.github/workflows/check-archs-consistency.yml
+++ b/.github/workflows/check-archs-consistency.yml
@@ -27,10 +27,10 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
 
-      - name: Use Node.js 18
-        uses: actions/setup-node@v4
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
         with:
-          node-version: 18
+          node-version: 24
           cache: yarn
 
       - name: Install node dependencies
--- a/.github/workflows/docs-check.yml
+++ b/.github/workflows/docs-check.yml
@@ -21,10 +21,10 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
 
-      - name: Use Node.js 18
-        uses: actions/setup-node@v4
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
         with:
-          node-version: 18
+          node-version: 24
           cache: yarn
 
       - name: Install node dependencies
--- a/.github/workflows/ios-build-paper.yml
+++ b/.github/workflows/ios-build-paper.yml
@@ -31,10 +31,10 @@ jobs:
         with:
           xcode-version: '16.1'
 
-      - name: Use Node.js 18
-        uses: actions/setup-node@v4
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
         with:
-          node-version: 18
+          node-version: 24
           cache: yarn
 
       - name: Install node dependencies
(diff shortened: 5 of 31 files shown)


packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -179,3 +179,44 @@ export function useWebEventHandlers() {
...
+// sourced from:
+// https://github.com/reduxjs/react-redux/blob/7e2fdd4ee2021e4282e12ba9fc722f09124e30cd/src/utils/useIsomorphicLayoutEffect.ts#L36
+// React currently throws a warning when using useLayoutEffect on the server.
+// To get around it, we can conditionally useEffect on the server (no-op) and
+// useLayoutEffect in the browser. We need useLayoutEffect to ensure the store
+// subscription callback always has the selector from the latest render commit
+// available, otherwise a store update may happen between render and the effect,
+// which may cause missed updates; we also must ensure the store subscription
+// is created synchronously, otherwise a store update may occur before the
+// subscription is created and an inconsistent state may be observed
```
j-piasecki: This comment is redux-specific and doesn't apply to RNGH.

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -179,3 +179,44 @@ export function useWebEventHandlers() {
...
+
+// Matches logic in React's `shared/ExecutionEnvironment` file
+const canUseDOM = () =>
+  !!(
+    typeof window !== 'undefined' &&
+    typeof window.document !== 'undefined' &&
+    typeof window.document.createElement !== 'undefined'
+  );
+
+const isDOM = /* @__PURE__ */ canUseDOM();
```
j-piasecki:
```suggestion
const isDOM = canUseDOM();
```

I think we can get rid of `PURE` annotations - we have static checks that should ensure there are no unused variables.

m-bert: Actually, do we need that as function? Can't we simply do

```tsx
const isDOM = !!(
    typeof window !== 'undefined' &&
    typeof window.document !== 'undefined' &&
    typeof window.document.createElement !== 'undefined'
  );
```

I don't think we have other use cases for this function, and if we will have, we can extract it later.

Same for other functions in this PR.

m-bert (commented):
We have a file called [ghQueueMicrotask.ts](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/ghQueueMicrotask.ts), maybe we could do the same for `isomorphicLayoutEffect`? (cc @j-piasecki)

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
akwasniewski: Hah okay, fixed in 2423909b2c4bb2b490bd6dd5ed216d777ce7018f

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
akwasniewski: Sure, both fixed in 2423909b2c4bb2b490bd6dd5ed216d777ce7018f

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -179,3 +179,23 @@ export function useWebEventHandlers() {
...
+);
+
+// Under React Native, we know that we always want to use useLayoutEffect
+const isReactNative =
+  typeof navigator !== 'undefined' && navigator.product === 'ReactNative';
+
+const getUseIsomorphicLayoutEffect = () =>
+  isDOM || isReactNative ? React.useLayoutEffect : React.useEffect;
+
+export const useIsomorphicLayoutEffect = getUseIsomorphicLayoutEffect();
```
j-piasecki: Can we go deeper?

```suggestion
export const useIsomorphicLayoutEffect = isDOM || isReactNative ? React.useLayoutEffect : React.useEffect;
```

akwasniewski: dde38b09cfaca511c75870a5c9cd757c98548149. Sure, I was certain I've done it, turns out I must have forgotten to commit it.

PR #3753, opened 2025-10-17
Bump `node` on CI

m-bert:
## Description

`node` versions in our workflows were quite outdated, this PR bumps them from `18` to `24`

## Test plan

Check CIs


j-piasecki (approved):
Hmm, they are quite **updated** as of this PR, before they were quite **outdated**.

PR #3756, opened 2025-10-20
[iOS] Fix gestures on iOS 26

m-bert:
## Description

Some changes in #3740 required follow up.

#### Tap

Turns out that we can leave `triggerAction` in `reset` (thank you apple for transparent docs that clearly describe what changed between 18.5 and 26 ❤️).

It also seems that double `onFinalize` issue is fixed on 26.

#### Fling

Removing `[self triggerAction]` from `reset` broke things on old iOS. Unfortunately now swiping in different direction does not result in calling `onFinalize` - this should be handled by `triggerAction` in `reset`, but now by default recognizers in `reset` have `Possible` state (once again,  thank you apple 😄). To fix that we will probably have to rewrite fling to custom logic.

#### LongPress

`LongPress` was missing on `triggerAction` call that sends `fail`.

#### Pan 

In `Pan`, I've moved `triggerAction` calls from `touches*` methods to `interactions*` methods.

## Test plan

<details>
<summary>Tested on basic-example and the following code:</summary>

```tsx

import {
  GestureHandlerRootView,
  Gesture,
  GestureDetector,
  GestureType,

function TestBox({
  gestureType,
  bgColor,
}: {
  gestureType: GestureType;
  bgColor: string;
}) {
  const handlerName = gestureType.handlerName;

  const gesture = gestureType
    .onBegin(() => {
      console.log(`[${handlerName}] onBegin`);
    })
    .onEnd(() => {
      console.log(`[${handlerName}] onEnd`);
    })
    .onFinalize(() => {
      console.log(`[${handlerName}] onFinalize`);
    })
    .runOnJS(true);

  return (
    <View style={styles.center}>
      <Text>{handlerName}</Text>
        <View style={[styles.box, { backgroundColor: bgColor }]} />
      </GestureDetector>
    </View>
  );
}

  return (
    <GestureHandlerRootView style={[{ flex: 1, padding: 50 }, styles.center]}>
      <TestBox gestureType={Gesture.Pan()} bgColor="#b58df1" />
      <TestBox gestureType={Gesture.LongPress()} bgColor="#f1a85d" />
      <TestBox gestureType={Gesture.Fling()} bgColor="#5df1a8" />
      <TestBox gestureType={Gesture.Tap()} bgColor="#5d8ef1" />
    </GestureHandlerRootView>
  );
}

  center: {
    display: 'flex',
    alignItems: 'center',
  },
  box: {
    height: 100,
    width: 100,
    backgroundColor: '#b58df1',
    borderRadius: 20,
    marginBottom: 30,
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
+++ b/packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
@@ -75,7 +75,7 @@
 {
   // TODO: On iOS 26 swiping in "wrong" direction doesn't send `onFinalize` callback. This is because now in `reset`
   // default state is `UIGestureRecognizerStatePossible`
-  [self triggerAction]; // Keeping it will not brake old iOS because we check if we do not send the same state twice.
+  [self triggerAction]; // Keeping it will not break old iOS because we check if we do not send the same state twice.
   [_gestureHandler.pointerTracker reset];
   _hasBegan = NO;
   [super reset];


packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
```diff
@@ -73,6 +73,9 @@ - (void)triggerAction
 
 - (void)reset
 {
+  // TODO: On iOS 26 swiping in "wrong" direction doesn't send `onFinalize` callback. This is because now in `reset`
+  // default state is `UIGestureRecognizerStatePossible`
+  [self triggerAction]; // Keeping it will not brake old iOS because we check if we do not send the same state twice.
```
j-piasecki:
```suggestion
  [self triggerAction]; // Keeping it will not break old iOS because we check if we do not send the same state twice.
```

akwasniewski (approved):

voxlz:
I'm still rarely get a onBegin without a onFinalize, even on this version. (Pan handler) any way to prevent this?

PR #3757, opened 2025-10-20
fixed reanimated handlers

akwasniewski:
## Description

After recent changes `hasWorkletEventHandlers` function was broken, it alwasy returned false, as `HandlerCallbacks` is a set so we have to check with `.has` not `is` operator.

Without this `LogicDetector` will throw weird errors as  `gesture.config.shouldUseReanimatedDetector` is always false.

_Note: I still have to check whether this change did not break anything else
## Test plan

Check whether `NativeDetector` with a reanimated `LogicDetector` underneath loads on iOS.


m-bert:
Already fixed in #3758 😅

PR #3760, opened 2025-10-21
Add workflow for releases

m-bert:
## Description

This PR adds workflow that will publish new versions of **Gesture Handler**  to `npm`. This is similar to [react-native-screens CI](https://github.com/software-mansion/react-native-screens/blob/15f83cf00b2d46c17c8e0f4c157c76f10bb57d53/.github/workflows/npm-screens-publish-nightly.yml), but this one will publish stable releases. 

> [!NOTE]
> We will dive deeper and check how to make this CI publish both, nightly and stable releases. For now we need this for stable release, so we want to make sure it works without any problems.

## Test plan

Downloaded artifact from CI.

### DIFF
--- a/.github/workflows/npm-gesture-handler-publish.yml
+++ b/.github/workflows/npm-gesture-handler-publish.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: Assert PACKAGE_NAME
         if: ${{ env.PACKAGE_NAME == 'PLACEHOLDER' }}
-        run: exit 1 # This should never happen.
+        run: exit 1 # If we end up here, it means that package was not generated.
 
       - name: Upload npm package to GitHub
         uses: actions/upload-artifact@v4


.github/workflows/npm-gesture-handler-publish.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  npm-build:
+    if: github.repository == 'software-mansion/react-native-gesture-handler'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write # for OIDC
+    env:
+      PACKAGE_NAME: PLACEHOLDER # Will be reassigned later on.
```
j-piasecki, line 15: Why do we need that?

m-bert: Comment changed in [585a77b](https://github.com/software-mansion/react-native-gesture-handler/pull/3760/commits/585a77b61d8f210a9c88cec7e73f80418cba7b89).

PR #3777, opened 2025-10-27
[iOS] Fix `Pressable` `onPressOut` callback

### LINKED ISSUE #3769: [Pressable] onPressOut no longer called if user swipes while not lifting finger on iOS
### Description

There was a bug in 2.28.0 where onPress was still triggering even after a scroll should have cancelled it: #3609 
It was fixed in 2.29.0 by: #3678

After updating, I am observing that onPressOut is no longer being triggered when onPress is cancelled by scrolling. This only happens on iOS.

### Steps to reproduce

1. Press and hold the Pressable (onPressIn is triggered)
2. Move finger to trigger scroll, and then lift finger
3. Observe that "on press out" alert never appears (neither onPress nor onPressOut are triggered)

The gist is the same setup as issue #3609, just with onPressOut instead of onPress.

https://gist.github.com/terreng/4845edbaac9a568a2aabcf57ed6eefc9

### Gesture Handler version

2.29.0

### React Native version

0.81.4

### Platforms

iOS

### JavaScript runtime

None

### Workflow

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

Real device

### Device model

iPhone 11 Pro, iOS 26.0

### Acknowledgements

Yes

m-bert:
## Description

Currently `onPressOut` is not called when `Pressable` gestures are cancelled. `iOS` and `android` are different when it comes to order of callbacks. In that particular case, when gesture is cancelled, on `android` we first receive `onTouchesCancelled`. However, on `iOS` `onFinalize` is called first. Inside `handleFinalize` function `isCurrentlyPressed` is reset to false, which means that `onPressOut` will not be called when we enter `onTouchesCancelled` callback. The thing is, in that place we don't have information required for event, so it is not possible to call `handlePressOut` from this place.

In this PR, I disabled calling `handleFinalize` on `iOS` in `onFinalize` callback, so that it calls `handlePressOut` when inside `onTouchesCancelled` callback. It also aligns with `android` implementation, as now the flow is the same.

Fixes #3769

## Test plan

Tested on:

- `Pressable` examples on `expo-example` app.

<details>
<summary> On the following code: </summary>

```tsx
import {
  GestureHandlerRootView,
  Pressable,
  ScrollView,

export default function Test() {
  return (
    <GestureHandlerRootView>
      <ScrollView style={{ flexGrow: 1, backgroundColor: 'white' }}>
        <View style={{ paddingTop: 200, height: 2000 }}>
          <Pressable
            style={{ width: 30, height: 30, backgroundColor: 'pink' }}
            onPress={() => console.log('press')}
            onPressOut={() => console.log('press out')}
          />
        </View>
      </ScrollView>
    </GestureHandlerRootView>
  );
}

```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/index.tsx
+++ b/packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/index.tsx
@@ -13,11 +13,12 @@ import { useAnimatedGesture } from './useAnimatedGesture';
 import { attachHandlers } from './attachHandlers';
 import { needsToReattach } from './needsToReattach';
 import { dropHandlers } from './dropHandlers';
-import { useIsomorphicLayoutEffect, useWebEventHandlers } from './utils';
+import { useWebEventHandlers } from './utils';
 import { Wrap, AnimatedWrap } from './Wrap';
 import { useDetectorUpdater } from './useDetectorUpdater';
 import { useViewRefHandler } from './useViewRefHandler';
 import { useMountReactions } from './useMountReactions';
+import { useIsomorphicLayoutEffect } from '../../../useIsomorphicLayoutEffect';
 
 function propagateDetectorConfig(
   props: GestureDetectorProps,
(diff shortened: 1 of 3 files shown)


RohovDmytro:
Can it be the reason why onPress stopped working recently on web?

j-piasecki:
@RohovDmytro This PR hasn't been included in any release yet.

PR #3778, opened 2025-10-27
Add support for nested Text to LogicDetector

j-piasecki:
## Description

Updates the Gesture Handler backend on the native platforms to support view hierarchies that get folded into a single host component.

On Android, it relies on [ReactCompoundView](https://github.com/facebook/react-native/blob/796d182d8989ff826eaa3a57458bdfc79750f820/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactCompoundView.kt). When encountering a view that implements this interface, it's also hit-tested using its `reactTagForTouch` method, and additional handlers are used for this particular touch.

On iOS, it relies on [`touchEventEmitterAtPoint`](https://github.com/facebook/react-native/blob/796d182d8989ff826eaa3a57458bdfc79750f820/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L1514-L1517) to achieve a similar effect to the one on Android. The main difference here is that iOS gesture recognizers need to be attached to a concrete view in order to be recognized. I made an assumption that any gesture with `LogicDetector` action type, which is being attached to a nonexistent view, falls into this category. It will get attached to the detector component, and a new prop - `virtualHandlerTag` - will be set to the tag of the "virtual" view. Then, inside `gestureRecognizerShouldBegin`, `reactTagForTouch` is used to determine whether the recognizer should be allowed to start processing events or not.

## Test plan

Added a new screen to the basic example app

### DIFF
--- a/apps/basic-example/src/Text.tsx
+++ b/apps/basic-example/src/Text.tsx
@@ -34,7 +34,9 @@ function NativeDetectorExample() {
 
   return (
     <View style={styles.subcontainer}>
-      <Text>Native Detector example</Text>
+      <Text style={styles.header}>
+        Native Detector example - this one should work
+      </Text>
       <NativeDetector gesture={tapAll}>
         <Text style={{ fontSize: 18, textAlign: 'center' }}>
           Some text example running with RNGH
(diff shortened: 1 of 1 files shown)


apps/basic-example/src/Text.tsx
```diff
@@ -0,0 +1,121 @@
...
+            </Text>
+          </GestureDetector>
+          this part is not special :(
+        </Text>
+      </GestureDetector>
+    </View>
+  );
+}
+
+export default function RuntimeDecorationExample() {
```
m-bert: Does this name fit here? 😅

j-piasecki:
> Should it work with legacy detector, or is this example here only to show difference between those 2? 🤔

It shouldn't work with the legacy detector. I updated the text to make it clearer.

m-bert (approved):

PR #3783, opened 2025-10-28
Remove remaining paper references

m-bert:
## Description

This PR removes:

1. `handleSetJSResponder` and `handleClearJSResponder` functions from our module
2. `isFabric` function
3. some of unnecessary suppresions

## Test plan

Tested on `expo-example` on:

1. Bouncing Box
4. Gesturized Pressable
5. Nested Buttons


packages/react-native-gesture-handler/src/utils.ts
```diff
@@ -45,13 +45,6 @@ export function tagMessage(msg: string) {
...
-// Helper method to check whether Fabric is enabled, however global.nativeFabricUIManager
-// may not be initialized before the first render
-export function isFabric(): boolean {
-  // @ts-expect-error nativeFabricUIManager is not yet included in the RN types
-  return !!global?.nativeFabricUIManager;
-}
-
 export function isReact19() {
   return React.version.startsWith('19.');
 }
```
j-piasecki, line 50: I think all versions of React Native that we support require React 19 by now, no? This could be removed as well if that's the case (in a follow-up).

m-bert: Sure, here you go: #3785

PR #3784, opened 2025-10-28
Add automated release scripts

j-piasecki:
## Description

Implements automated release scripts for the repository.

A new nightly release should be triggered on every commit pushed to the `main` branch. Version resolution happens automatically - the `latest` tag is pulled from npm, 1 is added to the minor, and the patch is set to 0. The full version is `$major.$minor$.$patch-$year$month$day-$sha`.

To release a stable version, cut a branch with the name matching format of `$major.$minor-stable` and trigger the `publish-stable-release` on that branch. By default, it does a dry run, skipping publishing the package and committing changes. Use that to make sure everything is correct before an actual release. It also uploads the package that would've been published as a GitHub artifact, which could be useful for testing. Version resolution happens automatically:
- the major and minor are inferred from the branch name
- next patch name is figured out from npm, in the dumbest possible way - iteratively check the patch versions until a not-released one is found
- if the minor is exactly +1 compared to the latest and the patch is 0 or the minor matches the latest and the patch is +1, the release is tagged as latest

In the future, we may want to run static checks automatically and app builds before releasing, but that would require refactoring our CI to rely more on reusable workflows.

## Test plan

Tested on a fork.
Release commitly run: https://github.com/j-piasecki/react-native-gesture-handler/actions/runs/18902148546/job/53951686147 (failed on publish due to missing provenance)

Release stable run for existing version: https://github.com/j-piasecki/react-native-gesture-handler/actions/runs/18902287389/job/53952137361 (this one succeeded because I commented out `npm publish` and replaced it with `echo`)

And here is the created release commit: https://github.com/j-piasecki/react-native-gesture-handler/commit/211296a9805e68d5e5ae24850c9f03b0a8cfd265. You can notice that the tag was also created successfully.

Release stable run for a new version: https://github.com/j-piasecki/react-native-gesture-handler/actions/runs/18903154587/job/53954932559 (this one was a dry run).

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,4 +1,4 @@
-name: Publish stable release to npm
+name: Publish release to npm
 on:
   # For commitlies
   push:


scripts/release/parse-version.js
```diff
@@ -0,0 +1,10 @@
+const VERSION_REGEX = /^(\d+)\.(\d+)\.(\d+)$/;
+
+function parseVersion(version) {
+  const [, major, minor, patch] = version.match(VERSION_REGEX);
+  return [Number(major), Number(minor), Number(patch)];
+}
+
+module.exports = {
+  parseVersion,
+};
```
m-bert:
```suggestion
};

```

scripts/release/parse-version.js
```diff
@@ -0,0 +1,10 @@
+const VERSION_REGEX = /^(\d+)\.(\d+)\.(\d+)$/;
+
+function parseVersion(version) {
+  const [, major, minor, patch] = version.match(VERSION_REGEX);
+  return [Number(major), Number(minor), Number(patch)];
+}
```
m-bert: Also we could do 

```suggestion
function parseVersion(version) {
  return version.match(VERSION_REGEX).slice(1).map(Number)
}
```

scripts/release/set-package-version.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+function getNextStableVersion() {
+  const currentBranch = execSync('git rev-parse --abbrev-ref HEAD').toString().trim();
+
+  try {
+    const [, major, minor] = currentBranch.match(BRANCH_REGEX);
+
+    // TODO: We'll worry about 3.x.x later :)
```
m-bert: 😄

.github/actions/publish-npm-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+    default: "true"
+  dry-run:
+    description: Whether to perform a dry run of the publish.
+    default: "true"
+
+runs:
+  using: composite
+  steps:
+    - name: Set up environment
+      shell: bash
```
m-bert, line 15: Do we need to specify shell in each step?

.github/actions/publish-npm-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+      run: |
+        echo "YARN_ENABLE_HARDENED_MODE=0" >> $GITHUB_ENV
+        echo "PACKAGE_NAME=PLACEHOLDER" >> $GITHUB_ENV
```
m-bert, line 18: Is there a specific reason why we don't keep them in yaml `env`?

.github/actions/publish-npm-package/action.yml
```diff
@@ -0,0 +1,134 @@
...
+      run: |
+        git add packages/react-native-gesture-handler/package.json
+        git commit -m "Release v${{ steps.set-package-version.outputs.version }}"
+        
+        if [ "${{ inputs.dry-run }}" = 'false' ]; then
+          echo "Pushing release commit to origin..."
+          git push
+        else
+          echo "Dry run mode: skipping git push"
+        fi
```
m-bert, line 121: Right now our release commits also update `Podfile.locks`. I don't think it is a problem, I just wanted to mention that 😅

scripts/release/version-utils.js
```diff
@@ -0,0 +1,23 @@
...
+
+function parseVersion(version) {
+  return version.match(VERSION_REGEX).slice(1).map(Number)
+}
+
+function getStableBranchVersion() {
+  const currentBranch = execSync('git rev-parse --abbrev-ref HEAD').toString().trim();
+  try {
+    const [, major, minor] = currentBranch.match(BRANCH_REGEX);
+    return [Number(major), Number(minor)];
```
m-bert, line 15: If we use `map` above, why don't we do the same here?



```suggestion
    return currentBranch.match(BRANCH_REGEX).slice(1).map(Number)
```

m-bert: If you want to keep the former for readability it is fine, just stick to the one convention 😅

.github/actions/publish-npm-package/action.yml
j-piasecki: yes

.github/actions/publish-npm-package/action.yml
j-piasecki: Reusable actions cannot have top-level `env` section

.github/actions/publish-npm-package/action.yml
j-piasecki: I'm not sure if it matters that much, those are only in the test apps so they don't end up in the package anyway.

We may even consider not versioning on the `main` branch and set it to `0.0.0` or something like that.

scripts/release/version-utils.js
j-piasecki: Now that I think about it, I think it's easier to understand what's happening when it's explicit, changed it back in [Restore explicit version parsing](https://github.com/software-mansion/react-native-gesture-handler/pull/3784/commits/e7a7272061373572e7e1dffc7e39d4ae5d571128)

m-bert (approved):
Should be fine, remember to change workflow in `npm` settings 😅

PR #3795, opened 2025-10-31
[General] Fix `DiagonalDirections` type

j-piasecki:
## Description

It seems like TypeScript isn't treating an expression where every node is `const` as `const`:

```ts
const a = 1; // a: 1
const b = 2; // b: 2
const c = a + b; // c: number
```

This caused `DiagonalDirections` type to be a number instead of a union of consts.

## Test plan

Check the inferred type of `DiagonalDirections`

### DIFF
--- a/packages/react-native-gesture-handler/src/Directions.ts
+++ b/packages/react-native-gesture-handler/src/Directions.ts
@@ -12,10 +12,10 @@ const DOWN_LEFT = 10; // DOWN | LEFT
 
 // Public interface
 export const Directions = {
-  RIGHT: RIGHT,
-  LEFT: LEFT,
-  UP: UP,
-  DOWN: DOWN,
+  RIGHT,
+  LEFT,
+  UP,
+  DOWN,
 } as const;
 
 // Internal interface


packages/react-native-gesture-handler/src/Directions.ts
```diff
@@ -13,10 +20,10 @@ export const Directions = {
...
 // Internal interface
 export const DiagonalDirections = {
-  UP_RIGHT: UP | RIGHT,
-  DOWN_RIGHT: DOWN | RIGHT,
-  UP_LEFT: UP | LEFT,
-  DOWN_LEFT: DOWN | LEFT,
+  UP_RIGHT: UP_RIGHT,
+  DOWN_RIGHT: DOWN_RIGHT,
+  UP_LEFT: UP_LEFT,
+  DOWN_LEFT: DOWN_LEFT,
```
m-bert: Isn't this enough?

```tsx

export const DiagonalDirections = {
  UP_RIGHT,
  DOWN_RIGHT,
  UP_LEFT,
  DOWN_LEFT,
} as const;

```

m-bert (approved):

PR #3798, opened 2025-11-03
Throw error when detectors are not placed inside `GestureHandlerRootView`

m-bert:
## Description

For now, `NativeDetector` doesn't throw error when it is used outside of `GestureHandlerRootView`. This PR moves check from `LegacyDetector` to `GestureDetector`.

## Test plan

<details>

```tsx
import React from 'react';
import { usePan, GestureDetector } from 'react-native-gesture-handler';

  const pan = usePan({});

  return (
        <View style={{ width: 300, height: 300, backgroundColor: 'green' }} />
      </GestureDetector>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
@@ -29,7 +29,7 @@ export function NativeDetector<THandlerData, TConfig>({
         gesture.detectorCallbacks.onGestureHandlerStateChange
       }
       // @ts-ignore This is a type mismatch between RNGH types and RN Codegen types
-      onGestureHandlerEvent={gesture.gestureEvents.onGestureHandlerEvent}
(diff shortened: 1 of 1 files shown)


j-piasecki:
Will we be exporting `LegacyGestureDetector` as a standalone component? If so, the check should stay in both.

m-bert:
> Will we be exporting LegacyGestureDetector as a standalone component? If so, the check should stay in both.

Well, if exporting everything as `GestureDetector` works fine I don't think we have to export it as a standalone component. WDYT @akwasniewski?

akwasniewski:
> > Will we be exporting LegacyGestureDetector as a standalone component? If so, the check should stay in both.
> 
> Well, if exporting everything as `GestureDetector` works fine I don't think we have to export it as a standalone component. WDYT @akwasniewski?

The idea of putting all detectors in the new `GestureDetector` was not to export them seperately. We only want to export this multi-use component, I think it's less effort for the user.

PR #3800, opened 2025-11-04
Change components implementation

m-bert:
## Description

This PR changes components implementations to use new hooks API.

> [!WARNING]
> It requires changes in module on iOS - call `createGestureHandler` immediately, not adding operation block. 

## Test plan

Example apps

### DIFF
--- a/packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerDetectorShadowNode.cpp
+++ b/packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerDetectorShadowNode.cpp
@@ -38,12 +38,12 @@ void RNGestureHandlerDetectorShadowNode::layout(LayoutContext layoutContext) {
   // TODO: consider allowing more than one child and doing bounding box
   react_native_assert(getChildren().size() == 1);
 
-  auto child = std::static_pointer_cast<const YogaLayoutableShadowNode>(
-      getChildren()[0]);
-
   if (!this->yogaNode_.getHasNewLayout()) {
     return;
   }
+
+  auto child = std::static_pointer_cast<const YogaLayoutableShadowNode>(
+      getChildren()[0]);
 
   child->ensureUnsealed();
   auto mutableChild = std::const_pointer_cast<YogaLayoutableShadowNode>(child);


packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+          enabled: props.enabled,
+          hitSlop: props.hitSlop,
+          // testID: props.testID,
+        } as P,
+      }
+    );
+
+    const native = useNative(gestureHandlerProps);
+
+    const componentRef = useRef<React.ComponentType<P>>(null);
```
j-piasecki: Is this set anywhere?

m-bert: `componentRef`? It is passed into `Component` few lines below:

```tsx
return (
    <DetectorComponent gesture={native}>
      <Component {...childProps} ref={componentRef} />
    </DetectorComponent>
  );
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -47,7 +48,13 @@ open class GestureHandler {
...
         this.hostDetectorView
       } else if (this is NativeViewGestureHandler) {
-        this.view?.parent
+        val parent = this.view?.parent
+
+        if (parent is ReactSwipeRefreshLayout) {
+          parent.parent
+        } else {
+          parent
+        }
```
j-piasecki: Can we extend `hostDetectorView` to always be set when using v3 API instead of this? I think we already do something like that on iOS.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -149,12 +150,18 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
-  private fun tryAttachNativeHandlersToChildView(childId: Int) {
+  private fun tryAttachNativeHandlersToChildView(child: View) {
     val registry = RNGestureHandlerModule.registries[moduleId]
       ?: throw Exception("Tried to access a non-existent registry")
 
+    val id = if (child is ReactSwipeRefreshLayout) {
+      child.getChildAt(0).id
+    } else {
+      child.id
+    }
```
j-piasecki, line 165: 1. Why do we need that? It looks like it makes attaching to `ReactSwipeRefreshLayout` impossible
2. Can this be done in the `NativeViewGestureHandlerHook`?

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
```diff
@@ -0,0 +1,87 @@
...
+          // @ts-ignore FIXME(TS)
+          res.childProps[key] = props[key];
+        }
+        return res;
+      },
+      {
+        gestureHandlerProps: { ...config }, // Watch out not to modify config
+        childProps: {
+          enabled: props.enabled,
+          hitSlop: props.hitSlop,
```
j-piasecki, line 44: The previous implementation was also passing `testID`, so why aren't we doing this here?

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -83,6 +83,12 @@ export class GestureHandlerWebDelegate
     this.isInitialized = false;
   }
 
+  configure(): void {
```
j-piasecki, line 86: I don't really like the name. Maybe `updateListeners`?

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -0,0 +1,195 @@
...
+            )
+          : undefined
+      }
+    />
+  );
+};
+// Backward type compatibility with https://github.com/software-mansion/react-native-gesture-handler/blob/db78d3ca7d48e8ba57482d3fe9b0a15aa79d9932/react-native-gesture-handler.d.ts#L440-L457
+// include methods of wrapped components by creating an intersection type with the RN component instead of duplicating them.
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type ScrollView = typeof GHScrollView & RNScrollView;
```
j-piasecki: Is this trickery still needed (and also for the rest of the components)?

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -0,0 +1,195 @@
...
+);
+
+export type ImperativeScrollViewRef =
+  ComponentWrapperRef<RNScrollViewProps> | null;
+
+export const ScrollView = (
+  props: RNScrollViewProps &
+    NativeWrapperProperties & {
+      ref?: React.RefObject<ImperativeScrollViewRef>;
+      // This prop existst because using `ref` in `renderScrollComponent` doesn't work (it is overwritten by RN internals).
```
j-piasecki:
```suggestion
      // This prop exists because using `ref` in `renderScrollComponent` doesn't work (it is overwritten by RN internals).
```

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -0,0 +1,195 @@
...
+export type RefreshControl = typeof RefreshControl & RNRefreshControl;
+
+const GHScrollView = createNativeWrapper<PropsWithChildren<RNScrollViewProps>>(
+  RNScrollView,
+  {
+    disallowInterruption: true,
+    shouldCancelWhenOutside: false,
+  },
+  DetectorType.Intercepting
+);
```
j-piasecki, line 46: How does this work when, correct me if I'm wrong, Android and iOS render them in different positions, e.g.:

```
<RefreshControl>
  <ScrollView />
</RefreshControl>
```

on one and 

```
<ScrollView>
  <RefreshControl />
</ScrollView>
```

on the other. I'd imagine that would also impact where the detector is rendered.

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
m-bert: `testID` is handled in #3838. I can bring it here, but I've decided to leave it for another PR.

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
m-bert: I don't like it either, but I don't think `updateListeners` fits here, as 2 functions inside are responsible for style, not for listening to events 😞

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
m-bert: >  It looks like it makes attaching to `ReactSwipeRefreshLayout` impossible

This should be handled by `attachVirtualChildren`, right?

> Can this be done in the `NativeViewGestureHandlerHook`?

We are inside NativeDetector, so  I don't exactly know what do you mean. Something like `handler.shouldAttachToChild` that would return `hook.shouldAttachToChild`?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
m-bert: Let me know if [110d333](https://github.com/software-mansion/react-native-gesture-handler/pull/3800/commits/110d333dd9dcaafd2aa5df5d987f2f44545d026d) is fine 😅

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/events/RNGestureHandlerEvent.kt
```diff
@@ -21,7 +21,7 @@ class RNGestureHandlerEvent private constructor() : Event<RNGestureHandlerEvent>
     eventHandlerType: EventHandlerType,
   ) {
     val view = if (GestureHandler.usesNativeOrVirtualDetector(handler.actionType)) {
-      handler.viewForEvents
+      handler.hostDetectorView!!
     } else {
       handler.view!!
     }
```
j-piasecki: What do you think about making this the new `.viewForEvents`?

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
m-bert: Yes, but in slightly different form - [5f7ec9c](https://github.com/software-mansion/react-native-gesture-handler/pull/3800/commits/5f7ec9cb22201b294f5c96ba32c3586c45f6327a).

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
m-bert: I've also thought about it for a while, but it doesn't. In case of `ScrollView` we have the following structure, thanks to `createNativeWrapper`:

```tsx
<InterceptingGestureDetector>
  <RNScrollView>
```

Now, it doesn't matter which one will be rendered first `RefreshControl` or `ScrollView` - they both fall under this Detector. While **React Native** does change this order, it does it [inside ScrollView](https://github.com/facebook/react-native/blob/7c603ee261c3eb2d9bf160505427ba06607d3465/packages/react-native/Libraries/Components/ScrollView/ScrollView.js#L1830), so effectively we have:

```tsx
{/* Android */}
<InterceptingGestureDetector>
  <RefreshControl>
    <ScrollView>

{/* iOS */}
<InterceptingGestureDetector>
  <ScrollView>
    <RefreshControl>
```

`InterceptingGestureDetector` stays at the same level.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/events/RNGestureHandlerEvent.kt
m-bert: I like the idea! [aabc5b4](https://github.com/software-mansion/react-native-gesture-handler/pull/3800/commits/aabc5b4e7071582e1158e3c1256f684fe1de0b71).

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -0,0 +1,215 @@
...
+  RNRefreshControl,
+  {
+    disallowInterruption: true,
+    shouldCancelWhenOutside: false,
+  },
+  DetectorType.Virtual
+);
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export type RefreshControl = RNRefreshControlProps;
```
j-piasecki: Why is this one mapped to the `Props` type?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: If there's no way to get rid of it, can we get a comment explaining the "why" behind this workaround?

packages/react-native-gesture-handler/src/components/GestureComponents.tsx
```diff
@@ -42,11 +47,15 @@ const GHScrollView = createNativeWrapper<PropsWithChildren<RNScrollViewProps>>(
     shouldCancelWhenOutside: false,
   }
 );
-export const ScrollView = React.forwardRef<
+
+/**
+ * @deprecated use `ScrollView` instead
+ */
+export const LegacyScrollView = React.forwardRef<
```
akwasniewski, line 54: Why do we keep legacy implementations?

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    const native = useNativeGesture(gestureHandlerProps);
+
+    const componentRef = useRef<React.ComponentType<P>>(null);
+    const gestureRef = useRef<NativeGesture>(native);
+
+    useImperativeHandle(props.ref, () => ({
+      componentRef: componentRef.current,
+      gestureRef: gestureRef.current,
+    }));
```
j-piasecki: I'm not sure if this is at all useful:
1. It's a breaking change compared to the previous api and the components stop being a drop-in replacement - `scrollRef.current.scrollTo` becomes `scrollRef.current.componentRef.scrollTo`.
2. Is `gestureRef` actually usable? There's no way to react to changes happening to it in an easy way. It will be `null` after creating the ref, and that's what will be used for relations during the first render. To actually read the ref, something would need to cause a render after it's set. If it happens to change (like in the case of child remount), there's no signal that the parent (gesture) should react to it.

packages/react-native-gesture-handler/src/components/GestureComponents.tsx
j-piasecki: We'd like to have a way for people to upgrade to 3.0 and keep their apps working as they did before. Since the new components aren't compatible API-wise with the old ones, it makes sense to keep both for the transition period.

m-bert: I'd add that we do this because in that case it is easier - we want to leave old `createNativeWrapper` either way and those components don't have to be re-written, unlike those based on `Reanimated`.

I believe that's the root of this question - why `Reanimated` components don't have legacy versions.

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
m-bert: Ad 1. I agree that this may be a problem.

Ad 2. Isn't it necessary if we want to set relation between `ScrollView` and `RefreshControl`? For now `RefreshControl` blocks `ScrollView`, so we need that ref. On the other hand, we can't pass `RefreshControl` to `ScrollView` to wait for. I mean we could, but then we would have to call `createNativeWrapper` inside `ScrollView` component. 

Am I missing something?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
m-bert: For now I can't see a clean way of doing it in hook. Added a comment in [7ad7ad7](https://github.com/software-mansion/react-native-gesture-handler/pull/3800/commits/7ad7ad70a9da17a29980bf2963edd259aaa7df55).

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
m-bert: My bad, I've missed that. Changed in [06428d4](https://github.com/software-mansion/react-native-gesture-handler/pull/3800/commits/06428d4b97591acfa116a9a50a491f40922152e2).

packages/react-native-gesture-handler/src/components/GestureComponents.tsx
j-piasecki: > why Reanimated components don't have legacy versions

Because they were built on Animated and due to some breaking changes on the new arch, they no longer worked. We'd need to rewrite them to keep them.

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
j-piasecki: Hmm, would it be possible to split this? Maybe it would be possible to keep `ref` as a view ref and add a callback like `onGestureUpdate` which would run every time the gesture changes? This way, it's reactive, so you don't have to do any tricks to trigger updates.

Internally, we only care about the handler tag, which doesn't change so I don't think we care about any updates, just the first time it's set.

m-bert: Let me know what you think about [a66158c](https://github.com/software-mansion/react-native-gesture-handler/pull/3800/commits/a66158c6ef96304730f6ce1387220cc169991d74)

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
```diff
@@ -48,14 +51,7 @@ export default function createNativeWrapper<P>(
...
     const native = useNativeGesture(gestureHandlerProps);
-
-    const componentRef = useRef<React.ComponentType<P>>(null);
-    const gestureRef = useRef<NativeGesture>(native);
-
-    useImperativeHandle(props.ref, () => ({
-      componentRef: componentRef.current,
-      gestureRef: gestureRef.current,
-    }));
+    updateGesture_CAN_CAUSE_INFINITE_RERENDER?.(native);
```
j-piasecki: Shouldn't this happen inside of an effect?

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
```diff
@@ -8,21 +8,24 @@ import { NativeGesture } from './hooks/gestures/native/useNativeGesture';
...
   Component: React.ComponentType<P>,
   config: Readonly<NativeWrapperProperties> = {},
   detectorType: DetectorType = DetectorType.Native
 ) {
   const ComponentWrapper = (
-    props: P & NativeWrapperProperties & { ref?: React.RefObject<unknown> }
+    props: P &
+      NativeWrapperProperties & {
+        ref?: React.RefObject<unknown>;
+        updateGesture_CAN_CAUSE_INFINITE_RERENDER?: (
```
j-piasecki: I feel like this should be named along the lines of `onUpdateGesture...`, `onGestureUpdate...` or something like that. It's essentially a weird kind of event.

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
m-bert: You're right, it should - added in [f35b66a](https://github.com/software-mansion/react-native-gesture-handler/pull/3800/commits/f35b66a3ab0e288d962d337682de85f8c1c0ec3f).

packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
m-bert: Renamed in [f35b66a](https://github.com/software-mansion/react-native-gesture-handler/pull/3800/commits/f35b66a3ab0e288d962d337682de85f8c1c0ec3f).

PR #3802, opened 2025-11-04
[iOS] Correctly attach `Native` gesture

m-bert:
## Description

`Native` gesture was incorrectly attached when underlying view was `UIControl`. This PR aims to fix this. It also changes `hostDetectorTag` to `hostDetectorView`, which simplifies `viewForEvents` logic. 

## Test plan

<details>

```tsx
import { StyleSheet } from 'react-native';
import {
  GestureDetector,
  GestureHandlerRootView,
  PureNativeButton,
  useNative,

const Button = ({ nativeGestureProperties }: any) => {
  const native = useNative(nativeGestureProperties);

  return (
    <GestureDetector gesture={native}>
      <PureNativeButton
        style={{ width: 100, height: 20, backgroundColor: 'green' }}
      />
    </GestureDetector>
  );
};

  return (
      <Button
        nativeGestureProperties={{
          onEnd: () => {
            console.log('end');
          },
          disableReanimated: true,
        }}
      />
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.m
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.m
@@ -35,13 +35,6 @@
 - (void)attachHandlerWithTag:(NSNumber *)handlerTag
                       toView:(RNGHUIView *)view
               withActionType:(RNGestureHandlerActionType)actionType
-{
-  [self attachHandlerWithTag:handlerTag toView:view withActionType:actionType withHostDetector:nil];
-}
-
-- (void)attachHandlerWithTag:(NSNumber *)handlerTag
-                      toView:(RNGHUIView *)view
-              withActionType:(RNGestureHandlerActionType)actionType
             withHostDetector:(nullable RNGHUIView *)hostDetector
 {
   RNGestureHandler *handler = _handlers[handlerTag];
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
@@ -184,11 +184,7 @@ constexpr int NEW_ARCH_NUMBER_OF_ATTACH_RETRIES = 25;
   view.reactTag = viewTag; // necessary for RNReanimated eventHash (e.g. "42onGestureHandlerEvent"), also will be
                            // returned as event.target
 
-  if (hostDetector != nil) {
-    [_registry attachHandlerWithTag:handlerTag toView:view withActionType:actionType withHostDetector:hostDetector];
-  } else {
-    [_registry attachHandlerWithTag:handlerTag toView:view withActionType:actionType];
-  }
+  [_registry attachHandlerWithTag:handlerTag toView:view withActionType:actionType withHostDetector:hostDetector];
 
   // register view if not already there
   [self registerViewWithGestureRecognizerAttachedIfNeeded:view];
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.h
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.h
@@ -14,9 +14,6 @@
 - (void)registerGestureHandler:(nonnull RNGestureHandler *)gestureHandler;
 - (void)attachHandlerWithTag:(nonnull NSNumber *)handlerTag
                       toView:(nonnull RNGHUIView *)view
-              withActionType:(RNGestureHandlerActionType)actionType;
-- (void)attachHandlerWithTag:(nonnull NSNumber *)handlerTag
-                      toView:(nonnull RNGHUIView *)view
               withActionType:(RNGestureHandlerActionType)actionType
             withHostDetector:(nullable RNGHUIView *)hostDetector;
 - (void)detachHandlerWithTag:(nonnull NSNumber *)handlerTag;


packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -389,10 +389,14 @@ - (void)sendEventsInState:(RNGestureHandlerState)state
...
-      ? self.recognizer.view.superview
-      : self.recognizer.view;
+  if ([self isKindOfClass:[RNNativeViewGestureHandler class]] &&
+      _actionType == RNGestureHandlerActionTypeNativeDetector) {
+    RNGHUIView *view = self.recognizer.view.superview;
+
+    return [view isKindOfClass:[RCTViewComponentView class]] ? view.superview : view;
+  }
+
+  return self.recognizer.view;
```
akwasniewski:
```suggestion
return ([self isKindOfClass:[RNNativeViewGestureHandler class]] &&
        _actionType == RNGestureHandlerActionTypeNativeDetector)
  ? ([self.recognizer.view.superview isKindOfClass:[RCTViewComponentView class]]
      ? self.recognizer.view.superview.superview
      : self.recognizer.view.superview)
  : self.recognizer.view;
```
?

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -148,6 +148,11 @@ - (void)bindToView:(UIView *)view
     [control addTarget:self action:@selector(handleDragExit:forEvent:) forControlEvents:UIControlEventTouchDragExit];
     [control addTarget:self action:@selector(handleDragEnter:forEvent:) forControlEvents:UIControlEventTouchDragEnter];
     [control addTarget:self action:@selector(handleTouchCancel:forEvent:) forControlEvents:UIControlEventTouchCancel];
+
+    if (self.actionType == RNGestureHandlerActionTypeNativeDetector) {
+      [view addGestureRecognizer:_recognizer];
+    }
```
j-piasecki: Why do we need the dummy recognizer on ui controls? It should do nothing unless attached to a scroll view.

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -389,10 +391,22 @@ - (void)sendEventsInState:(RNGestureHandlerState)state
 
 - (RNGHUIView *)findViewForEvents
 {
-  return
-      [self isKindOfClass:[RNNativeViewGestureHandler class]] && _actionType == RNGestureHandlerActionTypeNativeDetector
-      ? self.recognizer.view.superview
-      : self.recognizer.view;
+  RNGHUIView *view = self.recognizer.view;
+
+  if ([self isKindOfClass:[RNNativeViewGestureHandler class]] &&
```
j-piasecki: Not related, but shouldn't [this](https://github.com/software-mansion/react-native-gesture-handler/blob/76caf4444323e7b71c3ac4fdec6fc333f2abf3f1/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm#L299-L300) also be a part of this method?

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -389,10 +391,22 @@ - (void)sendEventsInState:(RNGestureHandlerState)state
...
+      _actionType == RNGestureHandlerActionTypeNativeDetector) {
+    if ([view.superview conformsToProtocol:@protocol(RCTRNGestureHandlerDetectorViewProtocol)]) {
+      return view.superview;
+    }
+
+    if ([view.superview isKindOfClass:[RCTViewComponentView class]]) {
+      return view.superview.superview;
+    }
+
+    return view.superview;
```
j-piasecki: This gets really convoluted. Can this be moved to the `NativeViewHandler` itself as an override?

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
m-bert: In `GestureHandler` base class we have `findViewForEvents` method that uses `self.recognizer.view`. If we do not set it here, buttons won't work as the view for events is `nil`.

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
m-bert: I'm not exactly sure what you mean. Is it to also handle the case of `VirtualDetector`?

j-piasecki: It's basically "finding the view for events" so the behavior fits this method, and it goes through this path anyway.

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
j-piasecki: What about virtual detector then?

m-bert:
As discussed offline, I've changed this approach. Let me know what you think (cc @j-piasecki, @akwasniewski)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -5,6 +5,8 @@
 #import <React/RCTConversions.h>
 #import <React/RCTFabricComponentsPlugins.h>
 
+#import "RNGestureHandlerButtonComponentView.h"
```
j-piasecki: Why is this needed?

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.m
```diff
@@ -46,6 +46,7 @@ - (void)attachHandlerWithTag:(NSNumber *)handlerTag
 - (void)detachHandlerWithTag:(NSNumber *)handlerTag
 {
   RNGestureHandler *handler = _handlers[handlerTag];
+  handler.hostDetectorView = nil;
```
j-piasecki: Can we keep that inside `unbindFromView`?

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -389,9 +388,9 @@ - (void)sendEventsInState:(RNGestureHandlerState)state
 
 - (RNGHUIView *)findViewForEvents
 {
-  return
-      [self isKindOfClass:[RNNativeViewGestureHandler class]] && _actionType == RNGestureHandlerActionTypeNativeDetector
-      ? self.recognizer.view.superview
+  return _actionType == RNGestureHandlerActionTypeNativeDetector ||
+          _actionType == RNGestureHandlerActionTypeVirtualDetector
```
j-piasecki: Didn't we have a helper for this? Or was this Android?

m-bert: As far as I remember it was added only on android. I'll check that and add it if not present.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
m-bert: Leftovers, removed in [e54c63b](https://github.com/software-mansion/react-native-gesture-handler/pull/3802/commits/e54c63b35b1a0843172f6fc31214b3300f1ed851).

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
m-bert: Added in [029b77a](https://github.com/software-mansion/react-native-gesture-handler/pull/3802/commits/029b77aff2f1e2aad18830f31294399aeab1ed81)

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.m
m-bert: Sure, changed in [3094904](https://github.com/software-mansion/react-native-gesture-handler/pull/3802/commits/3094904362972b967c0835880af380343a50ac5c)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -180,19 +180,28 @@ - (void)attachHandlers:(const std::vector<int> &)handlerTags
...
-          [handlerManager.registry attachHandlerWithTag:@(tag) toView:self withActionType:actionType];
+          [handlerManager.registry attachHandlerWithTag:@(tag)
+                                                 toView:self
+                                         withActionType:actionType
+                                       withHostDetector:self];
         }
         [attachedHandlers addObject:@(tag)];
       }
-      [[handlerManager registry] handlerWithTag:@(tag)].hostDetectorTag = @(self.tag);
+      [[handlerManager registry] handlerWithTag:@(tag)].hostDetectorView = self;
```
j-piasecki: This shouldn't be necessary with `withHostDetector`.

packages/react-native-gesture-handler/apple/RNGestureHandlerManager.h
```diff
@@ -24,6 +24,11 @@
                toViewWithTag:(nonnull NSNumber *)viewTag
               withActionType:(RNGestureHandlerActionType)actionType;
 
+- (void)attachGestureHandler:(nonnull NSNumber *)handlerTag
+               toViewWithTag:(nonnull NSNumber *)viewTag
+              withActionType:(RNGestureHandlerActionType)actionType
+            withHostDetector:(nullable RNGHUIView *)hostDetector;
```
j-piasecki, line 30: Is this actually used anywhere?

m-bert: Isn't it here, in `Detector`?

```obj-c
[handlerManager attachGestureHandler:@(tag)
                                   toViewWithTag:@(viewTag)
                                  withActionType:actionType
                                withHostDetector:self];
```

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.m
```diff
@@ -43,6 +43,17 @@ - (void)attachHandlerWithTag:(NSNumber *)handlerTag
   [handler bindToView:view];
 }
 
+- (void)attachHandlerWithTag:(NSNumber *)handlerTag
```
akwasniewski, line 35: In the manager we did it the other way around. The old function (without hostDetector) only calls the new one with `hostDetector=nil`, it's not a strong opinion, but it'd be better to make the logic consistent.

packages/react-native-gesture-handler/apple/RNGestureHandlerManager.h
j-piasecki: Am dumb

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: Remember to remove this assignment before merging

j-piasecki (approved):

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
j-piasecki: IMO, those spells are always fun to write and mostly not fun to read 😞. Since it will be written once, and read (presumably) more often, I'd keep the more verbose one.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
m-bert: I already did, it is not pushed yet since I wanted to test something but I had to wait for iOS 26.1 simulator download to finish 😞

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
m-bert: I agree with @j-piasecki. While I like ternary operator, this one seems too complicated.

Also, isn't it to specific commit? I don't think we have that logic and github also says that this comment is `outdated`
image

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.m
m-bert: In manager it was easier to make older implementation call new one than the other way around. If you prefer to stay consistent I can change it, but then I believe that the one in `registry` should be changed, not ther other way around (cc @j-piasecki).

m-bert: As far as I can see, the other implementation in registry is not used anywhere else but in **manager**, so I've decided to get rid of it in [a0ed055](https://github.com/software-mansion/react-native-gesture-handler/pull/3802/commits/a0ed05553d68770556ff45236ce4ffd8c8184ea3). Let me know what you think.

PR #3803, opened 2025-11-05
Swipeable and drawer, new api

akwasniewski:
## Description

This PR changes reanimated swipeable and reanimated drawer components implementations to use new hooks API.

## Test plan

Check whether 'new swipable' and 'reanimated drawer layout' work in the example app

### DIFF
--- a/packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
+++ b/packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
@@ -457,9 +457,9 @@ const Swipeable = (props: SwipeableProps) => {
   const tapGesture = useTap({
     shouldCancelWhenOutside: true,
     enabled: shouldEnableTap,
-    simultaneousWithExternalGesture: simultaneousWithExternalGesture,
-    requireExternalGestureToFail: requireExternalGestureToFail,
-    blocksExternalGesture: blocksExternalGesture,
+    simultaneousWith: simultaneousWithExternalGesture,
+    requireToFail: requireExternalGestureToFail,
+    block: blocksExternalGesture,
     onStart: () => {
       'worklet';
       if (rowState.value !== 0) {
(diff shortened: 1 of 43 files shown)


akwasniewski:
> Looks like the main changes are due to migration from api v2 to v3. Can we also explore whether we can reduce number of renders using shared value binding?

I've made small adjsutment for swipeable 0f9daeb691b9974b23bdbd13603f835b53941e94. I played around a lot with trying to utilise this in drawer, however I could not manage it to work when making `drawerOpened` a shared value, we need it to be a state, and all other gesture props depend on it, so I don't think we can utilise it there.

packages/react-native-gesture-handler/src/v3/detectors/common.ts
```diff
@@ -29,7 +29,5 @@ export const ReanimatedNativeDetector =
 export const nativeDetectorStyles = StyleSheet.create({
   detector: {
     display: 'contents',
-    // TODO: remove, debug info only
-    backgroundColor: 'red',
```
j-piasecki, line 33: End of an era 😢

akwasniewski: Yeah, hah. Without it the swipeable is all red, so I decided it is time.

packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
```diff
@@ -478,88 +454,58 @@ const Swipeable = (props: SwipeableProps) => {
...
+  const tapGesture = useTap({
+    shouldCancelWhenOutside: true,
+    enabled: shouldEnableTap,
+    onStart: () => {
+      'worklet';
+      if (rowState.value !== 0) {
+        close();
+      }
+    },
+  });
```
m-bert, line 469: Why don't we pass relation props to `Tap`?

m-bert (changes requested):

packages/react-native-gesture-handler/src/components/ReanimatedSwipeable/ReanimatedSwipeable.tsx
akwasniewski: My bad, good catch 6978df75239df98e6758d94342028416b1532019

PR #3804, opened 2025-11-06
[Web] Fixed `Accessing element.ref was removed in React 19` error

rmarquois:
## Description

Since version 19, React doesn't allow direct access to ref from an element. 
This error occurs when using React Native Gesture Handler on web.

## Test plan

Tested my fix in my web project and error in console disappears.


fbartho:
@j-piasecki What would it take to get a release of this to be published?

I just went to integrate react-native-draggable-flatlist on web, and the only functionality that isn't working is the dragging, haha!

(I haven't yet tried if I can install this package directly from GitHub-branch-references yet, if that's possible, I'd love a hint if that's a faster process).

j-piasecki:
@fbartho We do have commitly releases, the one including this PR would be `2.30.0-20251106-b53b8ef67`.

PR #3811, opened 2025-11-12
[Types] Remove `onUpdate` from discrete gestures

j-piasecki:
## Description

Discrete gestures shouldn't be sending `update` events, so the `onUpdate` method shouldn't be accepted by the type definitions.

It also moves the type definitions for events and gestures above the hook definition. I'm pretty sure it was also correct before, but I think it's cleaner when things are declared before they are used. LMK what you think.

## Test plan

```
  const tap = useTap({
    onUpdate: () => {
      'worklet';
    },
  });

  const fling = useFling({
    onUpdate: () => {
      'worklet';
    },
  });

  const longpress = useLongPress({
    onUpdate: () => {
      'worklet';
    },
  });

  const native = useNative({
    onUpdate: () => {
      'worklet';
    },
  });
```

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFling.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFling.ts
@@ -1,5 +1,5 @@
 import {
-  BaseGestureConfig,
+  BaseDiscreteGestureConfig,
   ExcludeInternalConfigProps,
   SingleGesture,
   SingleGestureName,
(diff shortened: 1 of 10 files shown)


packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFling.ts
```diff
@@ -20,27 +20,28 @@ type FlingHandlerData = {
...
   GestureStateChangeEvent<FlingHandlerData>;
+
 export type FlingGestureUpdateEvent = GestureUpdateEvent<FlingHandlerData>;
 
 export type FlingGesture = SingleGesture<
   FlingHandlerData,
   FlingGestureProperties
 >;
+
+export function useFling(config: FlingGestureConfig): FlingGesture {
```
m-bert, line 41: Is specifying return type necessary?

j-piasecki: Without it, the type is inferred from `useGesture`, which doesn't know what kind of gesture it creates, so all are typed the same as continuous gestures. Another benefit is that gestures are now using named types.

PR #3813, opened 2025-11-12
Simplify `VirtualDetector`/`InterceptingGestureDetector` and reduce the number of renders

j-piasecki:
## Description

The communication logic between `VirtualDetector`/`InterceptingGestureDetector` was complex, since it was passing callbacks through ref and everything else through state via context.

Since we're relying on Reanimated's `useEvent` and `useComposedEventHandler`, we need to trigger a rerender anyway to update the handler. This means there's no point in trying to optimize passing callbacks via refs, as we need to end with a render, and synchronizing renders and state updates requires additional logic.

This PR simplifies the communication layer to pass everything through the state, which should greatly simplify logic.

It also:
- changes behavior of `shouldUseReanimated`, `dispatchesAnimatedEvents` flags - now it checks every virtual gesture, where previously the last registered one was the deciding factor
- explicitly disables auto-memoization for `VirtualDetector` so that children changes can be detected - this should be changed to use MutationObserver once it's rolled out in RN
- adds manual memoization to `InterceptingGestureDetector`

## Test plan

Tested on the following snippet:
```
import { Button, StyleSheet, Text, View } from 'react-native';
import {
  GestureDetector,
  InterceptingGestureDetector,
  useTap,

import { COLORS } from './colors';

function TextWithTap() {
  const tap = useTap({
    onStart: () => {
      'worklet';
      console.log('Tapped on text in its own component!');
    },
  });

  return (
      <Text style={{ fontSize: 24, color: COLORS.KINDA_GREEN }}>
        This text is rendered in a separate component.
      </Text>
    </GestureDetector>
  );
}

function NativeDetectorExample() {
  const [entireVisible, setEntireVisible] = React.useState(true);
  const [firstVisible, setFirstVisible] = React.useState(true);
  const [secondVisible, setSecondVisible] = React.useState(true);
  const [thirdVisible, setThirdVisible] = React.useState(true);
  const [secondKey, setSecondKey] = React.useState(0);
  const [firstHasCallback, setFirstHasCallback] = React.useState(true);

  const tapAll = useTap({
    onStart: () => {
      'worklet';
      console.log('Tapped on text!');
    },
  });

  const tapFirstPart = useTap({
    onStart: firstHasCallback
      ? () => {
          'worklet';
          console.log('Tapped on first part!');
        }
      : () => {
          'worklet';
          console.log('First part tapped, but no callback set.');
        },
  });

  const tapSecondPart = useTap({
    onStart: () => {
      'worklet';
      console.log('Tapped on second part!');
    },
  });

  return (
    <View style={styles.subcontainer}>
      <Button
        title={(firstVisible ? 'Hide' : 'Show') + ' entire text'}
        onPress={() => setEntireVisible((v) => !v)}
      />
      <Button
        title={(firstVisible ? 'Hide' : 'Show') + ' first part'}
        onPress={() => setFirstVisible((v) => !v)}
      />
      <Button
        title={(secondVisible ? 'Hide' : 'Show') + ' second part'}
        onPress={() => setSecondVisible((v) => !v)}
      />
      <Button
        title={(thirdVisible ? 'Hide' : 'Show') + ' third part'}
        onPress={() => setThirdVisible((v) => !v)}
      />

      <Button
        title="Re-mount second part"
        onPress={() => setSecondKey((k) => k + 1)}
      />
      <Button
        title={
          (firstHasCallback ? 'Disable' : 'Enable') + ' callback on first text'
        }
        onPress={() => setFirstHasCallback((v) => !v)}
      />

      {entireVisible && (
        <InterceptingGestureDetector gesture={tapAll}>
          <Text style={{ fontSize: 18, textAlign: 'center' }}>
            Some text example running with RNGH
            {firstVisible && (
              <GestureDetector gesture={tapFirstPart}>
                <Text style={{ fontSize: 24, color: COLORS.NAVY }}>
                  {' '}
                  try tapping on this part
                </Text>
              </GestureDetector>
            )}
            {secondVisible && (
              <GestureDetector gesture={tapSecondPart}>
                <Text
                  key={secondKey}
                  style={{ fontSize: 28, color: COLORS.KINDA_BLUE }}>
                  {' '}
                  or on this part
                </Text>
              </GestureDetector>
            )}
            {thirdVisible && (
              <>
                {' '}
                <TextWithTap />
              </>
            )}{' '}
            this part is not special :(
          </Text>
        </InterceptingGestureDetector>
      )}
    </View>
  );
}

export default function NativeTextExample() {
  return (
      <NativeDetectorExample />
    </View>
  );
}

  container: {
    flex: 1,
  },
  subcontainer: {
    flex: 1,
    gap: 8,
    alignItems: 'center',
    justifyContent: 'center',
  },
  header: {
    fontSize: 18,
    textAlign: 'center',
    paddingHorizontal: 24,
  },
});
```

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
@@ -21,7 +21,7 @@ import {
 } from '../common';
 import { tagMessage } from '../../../utils';
 
-interface SafeVirtualChildren {
+interface VirtualChildrenForNative {
   viewTag: number;
   handlerTags: number[];
   viewRef: unknown;
@@ -31,10 +31,12 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
   gesture,
   children,
 }: InterceptingGestureDetectorProps<THandlerData, TConfig>) {
-  const [virtualChildren, setVirtualChildren] = useState<VirtualChild[]>([]);
-  const safeVirtualChildren: SafeVirtualChildren[] = useMemo(
+  const [virtualChildren, setVirtualChildren] = useState<Set<VirtualChild>>(
+    () => new Set()
+  );
+  const virtualChildrenForNativeComponent: VirtualChildrenForNative[] = useMemo(
     () =>
-      virtualChildren.map((child) => ({
+      Array.from(virtualChildren).map((child) => ({
         viewTag: child.viewTag,
         handlerTags: child.handlerTags,
         viewRef: child.viewRef,
@@ -61,19 +63,18 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
 
   const register = useCallback((child: VirtualChild) => {
     setVirtualChildren((prev) => {
-      const index = prev.findIndex((c) => c.viewTag === child.viewTag);
-      if (index !== -1) {
-        const updated = [...prev];
-        updated[index] = child;
-        return updated;
-      }
-
-      return [...prev, child];
+      const newSet = new Set(prev);
+      newSet.add(child);
+      return newSet;
     });
   }, []);
 
-  const unregister = useCallback((childTag: number) => {
-    setVirtualChildren((prev) => prev.filter((c) => c.viewTag !== childTag));
+  const unregister = useCallback((child: VirtualChild) => {
+    setVirtualChildren((prev) => {
+      const newSet = new Set(prev);
+      newSet.delete(child);
+      return newSet;
+    });
   }, []);
 
   const contextValue: InterceptingDetectorContextValue = useMemo(
@@ -239,7 +240,7 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
         }
         handlerTags={handlerTags}
         style={nativeDetectorStyles.detector}
-        virtualChildren={safeVirtualChildren}
+        virtualChildren={virtualChildrenForNativeComponent}
         moduleId={globalThis._RNGH_MODULE_ID}>
         {children}
       </NativeDetectorComponent>
--- a/packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/VirtualDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/VirtualDetector.tsx
@@ -81,7 +81,7 @@ export function VirtualDetector<THandlerData, TConfig>(
     register(virtualChild);
 
     return () => {
-      unregister(viewTag);
+      unregister(virtualChild);
     };
   }, [viewTag, props.gesture, register, unregister, setMode]);
 
--- a/packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/useInterceptingDetectorContext.ts
+++ b/packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/useInterceptingDetectorContext.ts
@@ -11,7 +11,7 @@ export type InterceptingDetectorContextValue = {
   mode: InterceptingDetectorMode;
   setMode: (mode: InterceptingDetectorMode) => void;
   register: (child: VirtualChild) => void;
-  unregister: (child: number) => void;
+  unregister: (child: VirtualChild) => void;
 };
 
 export const InterceptingDetectorContext =


packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -21,17 +21,24 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
...
-
-  const virtualMethods = useRef<
-    Map<number, RefObject<DetectorCallbacks<unknown>>>
-  >(new Map());
-
-  const [shouldUseReanimated, setShouldUseReanimated] = useState(
-    gesture ? gesture.config.shouldUseReanimatedDetector : false
+  const [virtualChildren, setVirtualChildren] = useState<VirtualChild[]>([]);
+
+  const shouldUseReanimated = useMemo(
```
m-bert: I'd probably call it `shouldUseReanimatedDetector`, but we can leave that.

m-bert (commented):
LGTM, though I'll leave approval for @akwasniewski as he is the author of `VirtualDetector` 😅

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -142,42 +155,48 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
...
         // @ts-ignore This is a type mismatch between RNGH types and RN Codegen types
-        onGestureHandlerEvent={handleGestureEvent('onGestureHandlerEvent')}
+        onGestureHandlerEvent={useMemo(
+          () => createGestureEventHandler('onGestureHandlerEvent'),
+          [createGestureEventHandler]
+        )}
         // @ts-ignore This is a type mismatch between RNGH types and RN Codegen types
         onGestureHandlerAnimatedEvent={
           gesture?.detectorCallbacks.onGestureHandlerAnimatedEvent
         }
```
akwasniewski, line 223: Does the new approach with useMemo( () => createGestureEventHandler allow us to allow support for AnimatedEvents?

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -21,61 +21,57 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
...
+      ),
+    [virtualChildren, gesture]
   );
 
   const NativeDetectorComponent = dispatchesAnimatedEvents
     ? AnimatedNativeDetector
-    : shouldUseReanimated
+    : shouldUseReanimatedDetector
       ? ReanimatedNativeDetector
       : HostGestureDetector;
```
akwasniewski, line 62: Shouldn't we throw an error when both `shouldUseReanimatedDetector` and `dispatchesAnimatedEvents` are true?

akwasniewski (commented):
Looks good, you managed to solve most of my workarounds, congrats.

j-piasecki:
Found an infinite render loop 😞

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
j-piasecki: Only for non-native-driver ones, but I assume this is also working currently. The ones using the native driver must be attached directly to the intercepting detector. I added an explicit error message for this.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
j-piasecki: I think that in the new approach, the error that is thrown when the mode changes between Animated and Reanimated handles this. Let me know if you see any issues with that.

j-piasecki:
The issue was switching between the Animated and non-animated intercepting detectors, which caused the tree to unmount, switch back to the other detector, and remount, triggering the whole process again.

Should be gone with the new approach using `InterceptingDetectorMode`.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/VirtualDetector.tsx
```diff
@@ -65,26 +60,30 @@ export function VirtualDetector<THandlerData, TConfig>(
       ? props.gesture.tags
       : [props.gesture.tag];
 
-    const virtualProps = {
+    if (props.gesture.config.dispatchesAnimatedEvents) {
+      throw new Error(
+        tagMessage(
+          'VirtualGestureDetector cannot handle Animated events with native driver when used inside InterceptingGestureDetector. Use Reanimated or Animated events without native driver instead.'
```
akwasniewski, line 66:
```suggestion
          'GestureDetector cannot handle Animated events with native driver when used inside InterceptingGestureDetector. Use Reanimated or Animated events without native driver instead.'
```
VirtualDetector is exposed as GestureDetector, thus it might be less confusing to write the exposed name

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -17,65 +21,98 @@ import {
...
-    []
-  );
+  const register = useCallback((child: VirtualChild) => {
+    setVirtualChildren((prev) => {
+      const index = prev.findIndex((c) => c.viewTag === child.viewTag);
+      if (index !== -1) {
+        const updated = [...prev];
+        updated[index] = child;
+        return updated;
+      }
```
akwasniewski: `VirtualChildren` is an array only because it was passed to the native side. Now, if we have `safeVirtualChildren` it would be better to make `VirtualChildren` a set or a map.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -17,65 +21,98 @@ import {
...
-  const [virtualChildren, setVirtualChildren] = useState<VirtualChildren[]>([]);
-
-  const virtualMethods = useRef<
-    Map<number, RefObject<DetectorCallbacks<unknown>>>
-  >(new Map());
-
-  const [shouldUseReanimated, setShouldUseReanimated] = useState(
-    gesture ? gesture.config.shouldUseReanimatedDetector : false
+  const [virtualChildren, setVirtualChildren] = useState<VirtualChild[]>([]);
+  const safeVirtualChildren: SafeVirtualChildren[] = useMemo(
```
akwasniewski: It's not a strong opinion, but I'm not really a fan of the name `safeVirtualChildren`, but I'm not sure what would be better. Maybe `filteredVirtualChildren`?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -211,7 +211,7 @@ - (void)attachHandlers:(const std::vector<int> &)handlerTags
   }
 
   // This covers the case where `NativeViewGestureHandlers` are attached after child views were created.
-  if (self.subviews[0]) {
+  if (self.subviews.count != 0) {
```
m-bert, line 214: Do we need this change?

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -64,14 +76,43 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
...
+      setMode: (newMode: InterceptingDetectorMode) => {
+        if (
+          (newMode === InterceptingDetectorMode.REANIMATED &&
+            mode === InterceptingDetectorMode.ANIMATED) ||
+          (newMode === InterceptingDetectorMode.ANIMATED &&
+            mode === InterceptingDetectorMode.REANIMATED)
+        ) {
+          throw new Error(
+            tagMessage(
+              'InterceptingGestureDetector can only handle either Reanimated or Animated events at the same time.'
```
m-bert:
```suggestion
              'InterceptingGestureDetector can only handle either Reanimated or Animated events.'
```

Correct me if I'm wrong, but this prevents from switching between those two, not handling them simultaneously. If so, I think this message fits better without "at the same time".

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/useInterceptingDetectorContext.ts
```diff
@@ -0,0 +1,26 @@
...
+export const InterceptingDetectorContext =
+  createContext<InterceptingDetectorContextValue | null>(null);
+
+export function useInterceptingDetectorContext() {
+  const ctx = use(InterceptingDetectorContext);
+  if (!ctx) {
+    return null;
+  }
+  return ctx;
+}
```
m-bert, line 22:
```suggestion
export const InterceptingDetectorContext =
  createContext<InterceptingDetectorContextValue | null>(null);

export function useInterceptingDetectorContext() {
  return use(InterceptingDetectorContext);
```

Given that `InterceptingDetectorContext` is either `null` or `InterceptingDetectorContextValue`, isn't this enough?

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/VirtualDetector.tsx
m-bert: I'd suggest to leave that for now as we may want to split those components.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/useInterceptingDetectorContext.ts
j-piasecki: Done in [ddec85e](https://github.com/software-mansion/react-native-gesture-handler/pull/3813/commits/ddec85e7d53fb79747d470333e59d5dbf5aa0649)

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
j-piasecki: [3021f35](https://github.com/software-mansion/react-native-gesture-handler/pull/3813/commits/3021f35988851438f6188f8b344b810ea64c8256)

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: Yes, accessing `self.subviews[0]` crashes when there are no subviews.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
j-piasecki: Changed it to set in [Store virtual children in a set, rename variable](https://github.com/software-mansion/react-native-gesture-handler/pull/3813/commits/842521873101c33283f3034a95fd67fd2e293089), lmk if something's missing.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
j-piasecki: Agree, not the perfect name, but I'm also not a fan of `filtered...` 😅. In https://github.com/software-mansion/react-native-gesture-handler/pull/3813/commits/842521873101c33283f3034a95fd67fd2e293089 I changed it to `virtualChildrenForNativeComponent`, which is also not perfect, but at least explains what it does. What do you think?

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/VirtualDetector.tsx
j-piasecki: @m-bert found some cases where the current API makes it impossible to create the `NativeDetector` in legitimate cases. We'd need to make a `StopIntercepting` component to disable `InterceptingGestureDetector`, at which point we have 3 components again.

At this point, it may make the most sense to reduce the ✨magic✨ we're doing and make all three detectors a public API.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
akwasniewski: A bit wordy hah, but it's clear why it exists. Thanks

PR #3815, opened 2025-11-13
Apply more granular memoization

j-piasecki:
## Description

Hooks and utils for creating event handlers and gesture relations were always relying on the entire config, while they require only a subset of its fields. This PR changes those to only work with what they actually need and updates the dependency arrays.

## Test plan

Verify that created handlers are stable between renders, assuming the original callback is memoized.

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
@@ -73,19 +73,13 @@ export function useGesture<THandlerData, TConfig>(
     () =>
       prepareRelations(
         {
-          simultaneousWithExternalGesture:
-            config.simultaneousWithExternalGesture,
-          requireExternalGestureToFail: config.requireExternalGestureToFail,
-          blocksExternalGesture: config.blocksExternalGesture,
+          simultaneousWith: config.simultaneousWith,
+          requireToFail: config.requireToFail,
+          block: config.block,
         },
         tag
       ),
-    [
-      tag,
-      config.simultaneousWithExternalGesture,
-      config.requireExternalGestureToFail,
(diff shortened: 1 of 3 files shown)


packages/react-native-gesture-handler/src/v3/hooks/callbacks/js/useGestureStateChangeEvent.ts
```diff
@@ -1,14 +1,25 @@
...
+import { ensureStateChangeHandlers } from '../../utils';
 import { getStateChangeHandler } from '../stateChangeHandler';
 
 export function useGestureStateChangeEvent<THandlerData, TConfig>(
   handlerTag: number,
   config: BaseGestureConfig<THandlerData, TConfig>
 ) {
   return useMemo(() => {
-    const handlers = extractStateChangeHandlers(config);
+    const handlers = ensureStateChangeHandlers({
```
m-bert: I'm not sure if `ensure` is the correct prefix here. As far as I remember @akwasniewski  used it to mark required variables and we decided to throws errors if those are `undefined`. In that case it is perfectly fine that some of them are `undefined`.

akwasniewski: Yeah, i used it in this way, thus I don't think the rename is fitting. It might be better to leave the old `extract`, or use something like `prepare`

j-piasecki: Changed to `prepare` in [9e63d24](https://github.com/software-mansion/react-native-gesture-handler/pull/3815/commits/9e63d24c6aeb6cca2e81b60c0c708b70528433d4)

PR #3816, opened 2025-11-13
Memoize cloned configs in gesture hooks

j-piasecki:
## Description

Currently config is cloned and modified on every render in gesture hooks. This PR changes that, so that it's memoized assuming the input is also stable.

## Test plan

Verify that if the input config is memoized, the returned gesture reference is stable as long as the config doesn't change (will require other PRs to be merged first).
- https://github.com/software-mansion/react-native-gesture-handler/pull/3815
- https://github.com/software-mansion/react-native-gesture-handler/pull/3812

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/utils/configUtils.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/utils/configUtils.ts
@@ -91,13 +91,11 @@ export function useClonedAndRemappedConfig<
   THandlerData,
   TConfig extends object,
   TInternalConfig extends Record<string, unknown>,
-  TTransformedConfig extends Record<string, unknown> = TInternalConfig,
 >(
   config: ExcludeInternalConfigProps<BaseGestureConfig<THandlerData, TConfig>>,
   propsMapping: Map<string, string> = new Map(),
-  propsTransformer: (config: TInternalConfig) => TTransformedConfig = (cfg) =>
-    cfg as unknown as TTransformedConfig
-): BaseGestureConfig<THandlerData, TTransformedConfig> {
+  propsTransformer: (config: TInternalConfig) => TInternalConfig = (cfg) => cfg
+): BaseGestureConfig<THandlerData, TInternalConfig> {
   return useMemo(() => {
     const clonedConfig = cloneConfig<THandlerData, TConfig>(config);
 


packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFling.ts
```diff
@@ -39,9 +39,11 @@ export type FlingGesture = SingleGesture<
 >;
 
 export function useFling(config: FlingGestureConfig): FlingGesture {
-  const flingConfig = cloneConfig<FlingHandlerData, FlingGestureProperties>(
-    config
-  );
+  const flingConfig = useClonedAndRemappedConfig<
+    FlingHandlerData,
+    FlingGestureProperties,
+    FlingGestureProperties
```
m-bert, line 45: If last generic parameter is optional, do we need to pass it? I'd choose either making it required, or do not pass it when not necessary.

packages/react-native-gesture-handler/src/v3/hooks/gestures/pinch/usePinch.ts
```diff
@@ -49,10 +52,23 @@ function diffCalculator(
...
+
 export function usePinch(config: PinchGestureConfig): PinchGesture {
-  const pinchConfig = cloneConfig<PinchHandlerData, PinchGestureProperties>(
-    config
-  );
+  const pinchConfig = useClonedAndRemappedConfig<
+    PinchHandlerData,
+    PinchGestureProperties,
+    // no internal props, pass record as PinchGestureProperties maps everything to never
+    Record<string, unknown>
```
m-bert, line 70: Just in case, types for `pinch` are correct from users' perspective, right? What I mean is, you can only pass `pinch` properties into config when using `usePinch`?

packages/react-native-gesture-handler/src/v3/hooks/gestures/rotation/useRotation.ts
```diff
@@ -52,14 +55,23 @@ function diffCalculator(
...
+  const rotationConfig = useClonedAndRemappedConfig<
     RotationHandlerData,
-    RotationGestureProperties
-  >(config);
-
-  rotationConfig.changeEventCalculator =
-    getChangeEventCalculator(diffCalculator);
+    RotationGestureProperties,
+    // no internal props, pass record as RotationGestureProperties maps everything to never
+    Record<string, unknown>
```
m-bert, line 73: Same as for `pinch`.

packages/react-native-gesture-handler/src/v3/hooks/gestures/pinch/usePinch.ts
j-piasecki: image

packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFling.ts
j-piasecki: It's not passed 😄. There are 4 arguments accepted, but 3 passed.

packages/react-native-gesture-handler/src/v3/hooks/gestures/rotation/useRotation.ts
j-piasecki: image

packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFling.ts
m-bert: Oh, my bad 😅 4 is _a lot_ 😄 

Actually, does any of these use the 4th argument? I can't see it, but I may be blind 😄 And if not, is it necessary?

j-piasecki: You're right, it wasn't used in the final implementation. Removed in [01b6be2](https://github.com/software-mansion/react-native-gesture-handler/pull/3816/commits/01b6be2cc519d60eb9d85e14c7f2b597ecc5f2c6)

PR #3817, opened 2025-11-13
Rename relation props in config

j-piasecki:
## Description

Renames the configuration fields used for gesture relations:
- `simultaneousWithExternalGesture` -> `simultaneousWith`
- `requireExternalGestureToFail` -> `requireToFail`
- `blocksExternalGesture` -> `block`

I'm not entirely sold on `block`, I'd love to hear opinions.

## Test plan

Static checks


m-bert (approved):
> I'm not entirely sold on `block`, I'd love to hear opinions.

For me `blocks` would sound more natural, but on the other hand we have `requireToFail` not `requires`, so I'm not sure.

j-piasecki:
> For me blocks would sound more natural, but on the other hand we have requireToFail not requires, so I'm not sure.

Yeah, I went with `block` to be consistent.

PR #3818, opened 2025-11-13
Rename `onTouchesCancelled` to `onTouchesCancel`

j-piasecki:
## Description

Renames `onTouchesCancelled` to `onTouchesCancel` to unify the naming convention.

Also changes `success` to `didSucceed`, but this change should be transparent, besides code hints.

## Test plan

Static checks

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/types/ConfigTypes.ts
+++ b/packages/react-native-gesture-handler/src/v3/types/ConfigTypes.ts
@@ -19,11 +19,11 @@ export type GestureCallbacks<THandlerData> = {
   onStart?: (event: GestureStateChangeEvent<THandlerData>) => void;
   onEnd?: (
     event: GestureStateChangeEvent<THandlerData>,
-    success: boolean
+    didSucceed: boolean
   ) => void;
   onFinalize?: (
     event: GestureStateChangeEvent<THandlerData>,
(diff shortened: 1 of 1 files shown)


packages/react-native-gesture-handler/src/v3/types/ConfigTypes.ts
```diff
@@ -19,17 +19,17 @@ export type GestureCallbacks<THandlerData> = {
   onStart?: (event: GestureStateChangeEvent<THandlerData>) => void;
   onEnd?: (
     event: GestureStateChangeEvent<THandlerData>,
-    success: boolean
+    didSucceed: boolean
```
m-bert, line 22: I'll bring this topic once again, what about moving `didSucceed` into `event`? This can be both good and bad idea at the same time, so let me know what you think 😅

j-piasecki: It already is inside the event, kind of. `didSucceed` is `true` when the new state is `END`, and false when the new state is `CANCELLED` or `FAILED`.

We can come back to the idea together with changing the state info. Do we want to hide the state info from the public API entirely?

I'll go with merging this for now, since this rename doesn't change the discussion in any way 😅.

PR #3819, opened 2025-11-13
[iOS] Remove dropped native handlers

m-bert:
## What is happening

While working on #3800 I've noticed frequent crashes on `iOS`. I'll explain this using the demo from the test code, basically two screens, second screen has a `RectButton`.

1. Navigate to second screen
2. Go back
3. Navigate to the second screen again
4. App crashes

Sometime it crashes immediately on startup.

## Why is it happening?

`Native` handlers are handled differently compared to others. The main problem was that we tried to attach handler that was previously dropped, but `NativeDetector` has no information that this happened. The flow here looks as follows:

1. We navigate to screen with button
2. `addSubview` is called - at this point it won't attach anything as `_nativeHandlers` is empty
3. `updateProps` is called - handlers are attached
4. We navigate back to first screen
5. Handler is dropped
6. We navigate again to second screen
7. `addSubview` is called - it tries to attach handler that was dropped, therefore app crashes

## Solution

To fix this issue, we clear `_nativeHandlers` in `prepareForRecycle` method.

## Test plan

<details>
<summary>I've tested it on my branch with components re-written to new API, using the following code in EmptyExample:</summary>

```tsx
import React from 'react';

  return (
      <RectButton
        style={styles.button}
        onPress={() => console.log('Hello World!')}
      />
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
  button: {
    width: 100,
    height: 30,
    borderRadius: 10,
    backgroundColor: 'pink',
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import HostGestureDetector from './HostGestureDetector';
 import { configureRelations, ensureNativeDetectorComponent } from './utils';
 import { isComposedGesture } from '../hooks/utils/relationUtils';
@@ -21,6 +21,10 @@ export function NativeDetector<THandlerData, TConfig>({
 
   ensureNativeDetectorComponent(NativeDetectorComponent);
   configureRelations(gesture);
+
+  const handlerTags = useMemo(() => {
+    return isComposedGesture(gesture) ? gesture.tags : [gesture.tag];
+  }, [gesture]);
 
   return (
     <NativeDetectorComponent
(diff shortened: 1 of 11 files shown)


j-piasecki:
Hmm, I see that we already disabled recycling only on macOS. Disabling that would be ok, but a better, and correct, thing to do would be to reset all local fields inside `prepareForRecycle`. It can stay disabled on macOS, I'm sure there was a reason why our components do that but don't recall it.

m-bert:
> Disabling that would be ok, but a better, and correct, thing to do would be to reset all local fields inside prepareForRecycle. 

For now I've cleared only `_nativeHandlers`, as other sets seem to be correctly cleared at this point. WDYT?

j-piasecki:
> For now I've cleared only _nativeHandlers, as other sets seem to be correctly cleared at this point. WDYT?

Can everything be reset? I shouldn't impact performance in a meaningful way and will likely prevent some nasty bugs in the future, where that assumption may no longer hold.

PR #3824, opened 2025-11-19
[General] Allow explicit opt-in for the Animated codepaths when not using native driver

j-piasecki:
## Description

Currently, we're enabling the Animated codepath only when the handler passed to `onUpdate` is an `Animated.Event` with `useNativeDriver`. When `useNativeDriver` is not enabled, `Animated.event` returns a function, and we have no way of differentiating between it and a function passed by the user.

This is not ideal since the animated codepath receives events in a different format (wrapped with `nativeEvent`). This meant that the code would stop working when switching between using the native driver and not. The same would happen when trying to use a native driver in code supposed to also run on web, where it's not available - the gesture would work on mobile, but crash on web.

To avoid this, this PR adds an option to explicitly opt-in to the animated codepath by setting `useAnimated` in the gesture config. This will automatically disable Reanimated and throw an error when user if the user tries to reenable it.

## Test plan

Tested on the `Native Detector` example in the basic app, as well as on the expo app on the same example for web.

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
@@ -30,14 +30,6 @@ export function useGesture<THandlerData, TConfig>(
   // This has to be done ASAP as other hooks depend `shouldUseReanimatedDetector`.
   prepareConfig(config);
 
-  if (config.dispatchesAnimatedEvents && config.shouldUseReanimatedDetector) {
-    throw new Error(
-      tagMessage(
-        `${type}: You cannot use Animated.Event together with callbacks running on the UI thread. Either remove Animated.Event from onUpdate, or set runOnJS property to true on the gesture.`
-      )
-    );
-  }
-
   // TODO: Call only necessary hooks depending on which callbacks are defined (?)
   const {
     onGestureHandlerStateChange,


packages/react-native-gesture-handler/src/v3/types/ConfigTypes.ts
```diff
@@ -25,7 +25,9 @@ export type GestureCallbacks<THandlerData> = {
     event: GestureStateChangeEvent<THandlerData>,
     didSucceed: boolean
   ) => void;
-  onUpdate?: (event: GestureUpdateEvent<THandlerData>) => void | AnimatedEvent;
+  onUpdate?:
+    | ((event: GestureUpdateEvent<THandlerData>) => void)
+    | AnimatedEvent;
```
j-piasecki, line 30: Instead of a function or an `AnimatedEvent`, it was a function that was returning void or `AnimatedEvent`.

This was irritating, as it was the reason types in `InterceptingGestureDetector` were correct.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -126,13 +126,15 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
...
+        if (typeof gesture?.detectorCallbacks[key] === 'function') {
+          // @ts-expect-error passing event to a union of functions where only one is typed as such
           gesture.detectorCallbacks[key](e);
         }
 
         virtualChildren.forEach((child) => {
           const method = child.methods[key];
-          if (method) {
+          if (typeof method === 'function') {
+            // @ts-expect-error passing event to a union of functions where only one is typed as such
```
j-piasecki, line 137: Those types.

packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -126,13 +126,15 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
   const createGestureEventHandler = useCallback(
     (key: keyof DetectorCallbacks<THandlerData>) => {
       return (e: GestureHandlerEvent<THandlerData>) => {
-        if (gesture?.detectorCallbacks[key]) {
+        if (typeof gesture?.detectorCallbacks[key] === 'function') {
```
j-piasecki, line 129: This is also covering `undefined` check, `typeof undefined === 'undefined'`

packages/react-native-gesture-handler/src/v3/types/EventTypes.ts
```diff
@@ -48,10 +48,10 @@ export type TouchEvent =
...
-// This is almost how Animated.event is typed in React Native. We add _argMapping in order to:
-// 1. Distinguish it from a regular function,
-// 2. Have access to the _argMapping property to check for usage of `change*` callbacks.
-export type AnimatedEvent = ((...args: any[]) => void) & {
+// This is not how Animated.event is typed in React Native. We add _argMapping in order to
+// have access to the _argMapping property to check for usage of `change*` callbacks.
+// It's also not typed as a function, which is breaking Gesture Handler type definitions.
+export type AnimatedEvent = {
   _argMapping: (Animated.Mapping | null)[];
 };
```
j-piasecki, line 56: After fixing the type above, this started breaking type inference and every `onUpdate` was being typed as `(...any) => void`

j-piasecki (commented):
Let me know if you prefer the `AnimatedEvent` type fixes to be in a separate PR

packages/react-native-gesture-handler/src/v3/hooks/useGestureCallbacks.ts
```diff
@@ -1,11 +1,38 @@
...
+import {
+  checkMappingForChangeProperties,
+  isNativeAnimatedEvent,
+} from './utils';
 import { useReanimatedStateChangeEvent } from './callbacks/reanimated/useReanimatedStateChangeEvent';
 import { useReanimatedUpdateEvent } from './callbacks/reanimated/useReanimatedUpdateEvent';
 import { useReanimatedTouchEvent } from './callbacks/reanimated/useReanimatedTouchEvent';
+import { tagMessage } from '../../utils';
+
+function guardJSAnimatedEvent(handler: (...args: unknown[]) => void) {
```
m-bert, line 14: I'm not a fun of this name, but I cannot think of any better.

PR #3825, opened 2025-11-19
[General] Make `VirtualDetector` part of the public API

j-piasecki:
## Description

As @m-bert pointed out while working on V3 components, the current API makes certain patterns impossible to accomplish (e.g., `ScrollView` with `RefreshControl` where `InterceptingGestureDetector` needs to be used) without forcing the entire subtree to attach gestures to a single component.

There are two ways to handle that:
- add a mechanism for disabling the existing `InterceptingGestureDetector` for a given subtree
- stop automatically deciding between `NativeDetector` and `VirtualDetector` and require users to explicitly use the correct one

The first approach is confusing and not a perfect solution to a problem of our own making, while we can simply go with the second one, which gets rid of the problem.

Note: I exported `VirtualDetector` as `VirtualGestureDetector` to stay consistent.

## Test plan

Updated the example

### DIFF
--- a/apps/basic-example/src/Text.tsx
+++ b/apps/basic-example/src/Text.tsx
@@ -4,6 +4,7 @@ import {
   Gesture,
   GestureDetector,
   InterceptingGestureDetector,
+  VirtualGestureDetector,
   useTap,
 } from 'react-native-gesture-handler';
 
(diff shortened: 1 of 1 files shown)


packages/react-native-gesture-handler/src/v3/detectors/GestureDetector.tsx
```diff
@@ -31,12 +29,7 @@ export function GestureDetector<THandlerData, TConfig>(
     return <LegacyGestureDetector {...(props as LegacyGestureDetectorProps)} />;
   }
 
-  const context = use(InterceptingDetectorContext);
-  return context ? (
-    <VirtualDetector
-      {...(props as NativeDetectorProps<THandlerData, TConfig>)}
-    />
-  ) : (
+  return (
```
m-bert, line 32: I guess we keep that in order to simplify migration?

j-piasecki: Switching between V2 and V3? Yes, it explicitly depends on what the user passes to the detector, so I feel like this one is fine.

PR #3827, opened 2025-11-20
[General] Reset gesture context when it finishes

j-piasecki:
## Description

I noticed that the context used to store the last event used for calculating the change values was never reset. This resulted in the `change*` values in the first event being the difference between the starting event and the last event from the previous gesture, effectively resetting the sum to the starting value.

This PR updates the context to be shared between the `update` and `stateChange` handlers, and resets the context when the gesture finishes.

## Test plan

|Before|After|
|-|-|
|<video src='https://github.com/user-attachments/assets/6a30c5fa-8f16-4d06-88cc-87c676c9d3a0' />|<video src='https://github.com/user-attachments/assets/c11ffaae-13ae-48f9-b6ee-fc2bf7cc1431' />|

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/callbacks/reanimated/useReanimatedUpdateEvent.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/callbacks/reanimated/useReanimatedUpdateEvent.ts
@@ -11,7 +11,7 @@ export function useReanimatedUpdateEvent<THandlerData>(
   reanimatedHandler: ReanimatedHandler<THandlerData> | undefined,
   changeEventCalculator: ChangeCalculatorType<THandlerData> | undefined
 ) {
-  // We don't want to call hooks conditionally, therefore `useHandler` and `useEvent` will be always called.
+  // We don't want to call hooks conditionally, `useEvent` will be always called.
   // The only difference is whether we will send events to Reanimated or not.
   // The problem here is that if someone passes `Animated.event` as `onUpdate` prop,
   // it won't be workletized and therefore `useHandler` will throw. In that case we override it to empty `worklet`.


packages/react-native-gesture-handler/src/v3/hooks/callbacks/reanimated/useReanimatedUpdateEvent.ts
```diff
@@ -1,19 +1,16 @@
...
+  changeEventCalculator: ChangeCalculatorType<THandlerData> | undefined
 ) {
-  const { handlers, changeEventCalculator } = prepareUpdateHandlers(
-    {
-      onUpdate: config.onUpdate,
-    },
-    config.changeEventCalculator
-  );
-
   // We don't want to call hooks conditionally, therefore `useHandler` and `useEvent` will be always called.
```
m-bert:
```suggestion
  // We don't want to call hooks conditionally, `useEvent` will be always called.
```

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
```diff
@@ -33,10 +38,9 @@ let Reanimated:
           options?: unknown
         ): ComponentClass<P>;
       };
-      useHandler: <THandlerData>(handlers: GestureCallbacks<THandlerData>) => {
-        doDependenciesDiffer: boolean;
-        context: ReanimatedContext<THandlerData>;
-      };
+      useHandler: <THandlerData>(
+        handlers: GestureCallbacks<THandlerData>
+      ) => ReanimatedHandler<THandlerData>;
```
m-bert, line 43: Why do we remove `doDependenciesDiffer` from this type, if it is still used?

packages/react-native-gesture-handler/src/v3/hooks/useGestureCallbacks.ts
```diff
@@ -38,22 +45,48 @@ export function useGestureCallbacks<THandlerData, TConfig>(
...
 
   let onReanimatedStateChange;
   let onReanimatedUpdateEvent;
   let onReanimatedTouchEvent;
 
   if (!config.disableReanimated) {
+    const handlers = {
+      ...prepareStateChangeHandlers(config),
+      ...prepareUpdateHandlers(config).handlers,
+    };
```
m-bert, line 74: So the idea is to put all handlers into this object, but those are executed based on `event` (like `onGestureHandlerEvent`), right?

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
j-piasecki: What do you mean? It's just moving the return type from being defined inline to being named and exported.

packages/react-native-gesture-handler/src/v3/hooks/useGestureCallbacks.ts
j-piasecki: Yes, specific event handlers should run relevant methods on `handlers` object.

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
m-bert: Oh you're right, my bad 😅

PR #3830, opened 2025-11-21
[General] Schedule flushes instead of flushing immediately

j-piasecki:
## Description

Currently, when creating a gesture or configuring relations, `flushOperations` is called immediately. When rendering multiple detectors in the same batch, this adds unnecessary overhead as it can be done a single time at the end of the batch. This PR changes that.

I also noticed that `createGestureHandler` is scheduled on the UI anyway, so I figured there may be no point in keeping it synchronous.

## Test plan

<details>
<summary>Used this stress test to measure impact:</summary>

```js
import { GestureDetector, usePan } from 'react-native-gesture-handler';
import Animated, {
  useAnimatedStyle,
  useSharedValue,

const DATA = new Array(500).fill(null).map((_, i) => `Item ${i + 1}`);

function Item() {

    return {
    };
  });

  const pan = usePan({
    onUpdate: (event) => {
      'worklet';
      translateX.value += event.handlerData.changeX;
    },
  });

  return (
    <View style={{ height: 80, padding: 16, backgroundColor: 'gray' }}>
        {/* <View collapsable={false} style={{opacity: 0.5}}> */}
        <Animated.View
          style={[
            { backgroundColor: 'red', height: '100%', aspectRatio: 1 },
            style,
          ]}
        />
        {/* </View> */}
      </GestureDetector>
    </View>
  );
}

function Benchmark() {
  return (
    <ScrollView
      style={{ flex: 1 }}
      {DATA.map((_, index) => (
        <Item key={index} />
      ))}
    </ScrollView>
  );
}

const TIMES = 30;

  const [visible, setVisible] = React.useState(false);

  useEffect(() => {
      setTimeout(() => {
        setVisible(true);
      }, 24);
    }

    if (times.length === TIMES) {
      sortedTimes.shift();
      sortedTimes.pop();
      const avgTime =

    }
  }, [visible]);

  return (
      {visible && (
        <Profiler
          id="v3"
            times.push(actualDuration);
            setTimeout(() => {
              setVisible(false);
            }, 24);
          }}>
          <Benchmark />
        </Profiler>
      )}
    </View>
  );
}

  container: {
    flex: 1,
  },
});
```

</details>

I saw the average time fall from ~610ms to ~595ms.

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
@@ -82,7 +82,7 @@ export function useGesture<THandlerData, TConfig>(
   ) {
     currentGestureRef.current = { type, tag };
     RNGestureHandlerModule.createGestureHandler(type, tag, {});
-    // It's possible that this can cause errors about handler not being created when attemting to mound NativeDetector
+    // It's possible that this can cause errors about handler not being created when attempting to mount NativeDetector
     scheduleFlushOperations();
   }
 


packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -81,20 +82,21 @@ export function useGesture<THandlerData, TConfig>(
   ) {
     currentGestureRef.current = { type, tag };
     RNGestureHandlerModule.createGestureHandler(type, tag, {});
-    RNGestureHandlerModule.flushOperations();
+    // It's possible that this can cause errors about handler not being created when attemting to mound NativeDetector
```
m-bert:
```suggestion
    // It's possible that this can cause errors about handler not being created when attempting to mount NativeDetector
```

m-bert: Also, isn't it connected with the error that I've shown you offline? The one when app would sometimes crash on startup because handler has not yet been created?

j-piasecki: I don't know - it cannot be the cause, since it wasn't there when you found it, but I also was able to reproduce that only by modifying RNGH source while the app was running.

PR #3832, opened 2025-11-21
ReanimatedDrawerLayout: fix buttons not receiving onPress

peterpme:
## Description

When using ReanimatedDrawerLayout, none of the react-native-gesture-handler buttons respond to taps because of the `overlayDismissGesture`. The gesture is activated even when the Drawer is closed. 

By passing in `drawerOpened`, a state variable already enabled into `overlayDismissGesture` the problem is fixed and the drawer continues working as-is.

## Test plan

In my own app, I added `drawerOpened` into `overlayDismissGesture` and was able to successfully fix the problem.


j-piasecki (approved):
Thanks for fixing it ❤️!

PR #3833, opened 2025-11-24
[Web] Fix press out

akwasniewski:
## Description

After https://github.com/software-mansion/react-native-gesture-handler/pull/3678/ pressables `onPressOut` did not work on web. This PR fixes the issue. That PR added a check whether gesture has succeded or not. Because on web, native gestures fail on mouse move, the finalise was handled by LongPress.finalize, but the LongPress has minDuration set to INT_32 max so it never ends succesfully. This creates new tap gesture that which handles finalise on web.

## Test plan

Tested on the following example
```tsx
import { View } from 'react-native'
import { GestureHandlerRootView, Pressable, ScrollView } from 'react-native-gesture-handler'

export default function Test() {
  return (
    <GestureHandlerRootView >
      <ScrollView style={{ flexGrow: 1, backgroundColor: 'white' }}>
        <View style={{ paddingTop: 200, height: 2000 }}>
          <Pressable
            style={{ width: 30, height: 30, backgroundColor: 'pink' }}
            onPress={() => console.log("pressed")}
          />
        </View>
      </ScrollView>
    </GestureHandlerRootView >
  )
}
```

### DIFF
--- a/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
+++ b/packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
@@ -307,8 +307,8 @@ const Pressable = (props: PressableProps) => {
         })
         .onFinalize((_event, success) => {
           if (Platform.OS !== 'web') {
-            // On Web we use Tap().onFinalize() instead of Native().onFinalize(),
-            // as Native cancels on mouse move, and Tap does not.
+            // On Web we use LongPress().onFinalize() instead of Native().onFinalize(),
+            // as Native cancels on mouse move, and LongPress does not.
             if (success) {
               stateMachine.handleEvent(StateMachineEvent.FINALIZE);
             } else {


packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -323,9 +313,30 @@ const Pressable = (props: PressableProps) => {
     [stateMachine, handlePressOut, handleFinalize]
   );
 
+  const webFinalizeGesture = useMemo(
+    () =>
+      Gesture.Tap()
+        .maxDuration(INT32_MAX)
+        .maxDistance(INT32_MAX)
+        .onFinalize((_event, success) => {
+          if (Platform.OS === 'web') {
```
j-piasecki: Can this be a part of the gesture config instead (`.enabled(Platform.OS === 'web')`)?

m-bert: I'm not exactly sure how it will behave, but as the description says

> Because on web, native gestures fail on mouse move, the finalise was handled by LongPress.finalize, but the LongPress has minDuration set to INT_32 max so it never ends succesfully.

In that case (and based on `LongPress` usage in this component), wouldn't it be enough to do

```tsx
LongPress.minDuration(Platform.OS === 'web' ? 0 : INT_32)
```

?

Note that I have not tested that, just what came up to me 😅

j-piasecki: Hmm, sounds like it could work.

akwasniewski: I tried this approach, but did not settle on it as then the gesture did not cancel if the touch went outside. However after some investigation, I noticed that this is because in https://github.com/software-mansion/react-native-gesture-handler/pull/2981/files#diff-bf60ea818ab04625c9e68d6d04c3ca9444af86ff48a2aec7ed680687b2145eb4 we set `shouldCancelWhenOutside` to true for all gestures (it was later changed to only happen on web). I tested it, and I think this change was not necessary to fix the issue, thus after reverting we can use the LongPress with minDuration(0) edef18bb5dcc9260df68d2674e0be5254640c10b

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -331,7 +331,6 @@ const Pressable = (props: PressableProps) => {
     gesture.enabled(isPressableEnabled);
     gesture.runOnJS(true);
     gesture.hitSlop(appliedHitSlop);
-    gesture.shouldCancelWhenOutside(Platform.OS !== 'web');
```
m-bert, line 334: As far as I can see, `LongPress` does not override default value and keeps it as `false`. If so, we now change this value on native platforms. Doesn't it break anything? 

Android does set it to `true` in `LongPress`, but I have not found this on `iOS`.

akwasniewski: I've tested and LongPress seems to work as if the default was `shouldCancelWhenOutside = true` on all platforms.

m-bert: > LongPress seems to work as if the default was shouldCancelWhenOutside = true.

But it is not, right? What's important is that `Pressable` behavior doesn't break with these changes. I guess it is?

akwasniewski: Ok i see it now. For some reason the value is not overriden to true on the native side, but in the [gesture itself](https://github.com/software-mansion/react-native-gesture-handler/blob/9f0a5215adc274e02ea98029180021584cda9cb8/packages/react-native-gesture-handler/src/handlers/gestures/longPressGesture.ts#L12C1-L13C4). This means that the changes will not change any pan behaviour on the native side. Correct me if I am wrong, but `shouldCancelWhenOutside` should not affect native gestures. Possibly only hover may be affected, but I did not observe any changes. I think it is redundant, but if you think it is necessary, to ensure correctness I can add 
`shouldCancelWhenOutside=true` for all gestures, on all platforms.

m-bert: > the changes will not change any pan behaviour

I guess you mean `LongPress` right?

> Correct me if I am wrong, but shouldCancelWhenOutside should not affect native gestures.

If the default value is `true` then yes, it shouldn't. Though I think that settings this in different places is quite misleading (e.g. [here](https://github.com/software-mansion/react-native-gesture-handler/blob/9f0a5215adc274e02ea98029180021584cda9cb8/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt#L27))

We may want to keep that in one place (can be in JS as it was intended), but let's keep that outside of this PR.

gigobyte:
Hey, can you publish this change? :)

j-piasecki:
Hey @gigobyte! Recently, we've added commitly releases, so you can start using the changes from specific PRs right after they are merged. [This](https://www.npmjs.com/package/react-native-gesture-handler/v/2.30.0-20251127-78c9ed7d6) is the version published after this PR.

If you're asking about a stable release, then we'll most likely publish a new version early next week.

PR #3834, opened 2025-11-24
Rename gesture composition hooks

j-piasecki:
## Description

- Renames composition hooks to include `Gestures` suffix.
- Renames `useRace` to `useMultipleGestures`
  - `useRaceGesture`/`useRacingGestures` suggests that it sets up some kind of relation between gestures. It doesn't - this is the default behavior. I hope that `useMultipleGestures` sounds like the answer to "I just want to attach pan and pinch to the same view", or "I want to set the relations manually because they depend on other components".
- Changes the directory they are in from `relations` to `composition`.
- Removes `useGesture` from the public API

## Test plan

Static checks

### DIFF
--- a/packages/react-native-gesture-handler/src/__tests__/RelationsTraversal.test.tsx
+++ b/packages/react-native-gesture-handler/src/__tests__/RelationsTraversal.test.tsx
@@ -1,8 +1,8 @@
 import { tagMessage } from '../utils';
 import {
-  useExclusiveGesture,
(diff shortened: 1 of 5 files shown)


j-piasecki:
> What do you think about useSimultaneousGestures, etc.? So plural instead of singular.

Sounds good if you ignore `useRaceGesture`. I've renamed it to `useMultipleGestures` and elaborated in the description.

PR #3836, opened 2025-11-24
[General] Reduce the number of calls to `useEvent`

j-piasecki:
## Description

During testing, I've noticed that calling `useEvent` from Reanimated is relatively heavy. Since we're doing it three times, I thought it might be worth exploring how much impact it has on our hooks. This PR reduces the number of `useEvent` calls to one per gesture, which reduces the execution time of `usePan` by ~10%. I'd assume the results for other gestures will be the same, since they all depend on internal `useGesture`.

## Test plan

<details>
<summary>Tested on this stress-test</summary>

```jsx
import { GestureDetector, usePan } from 'react-native-gesture-handler';
// import { PerfMonitor } from 'react-native-gesture-handler/src/v3/PerfMonitor';
import Animated, {
  useAnimatedStyle,
  useSharedValue,

const DATA = new Array(1).fill(null).map((_, i) => `Item ${i + 1}`);

function Item() {

    return {
    };
  });

  const pan = usePan({
    // disableReanimated: true,
    onUpdate: (event) => {
      'worklet';
      console.log('pan onUpdate', event.handlerData.changeX);
    },
    onStart: () => {
      'worklet';
      console.log('pan onStart');
    },
    onEnd: () => {
      'worklet';
      console.log('pan onEnd');
    },
    onBegin: () => {
      'worklet';
      console.log('pan onBegin');
    },
    onFinalize: () => {
      'worklet';
      console.log('pan onFinalize');
    },
    onTouchesDown: () => {
      'worklet';
      console.log('pan onTouchesDown');
    },
  });

  return null;
}

function Benchmark() {
  return (
    <ScrollView
      style={{ flex: 1 }}
      {DATA.map((_, index) => (
        <Item key={index} />
      ))}
    </ScrollView>
  );
}

const TIMES = 35;

  useEffect(() => {
      setTimeout(() => {
        setVisible(true);
      }, 24);
    }

    if (times.length === TIMES) {
      sortedTimes.shift();
      sortedTimes.shift();
      sortedTimes.pop();
      sortedTimes.pop();
      const avgTime =

      // console.log(JSON.stringify(PerfMonitor.getMeasures(), null, 2));
      // PerfMonitor.clear();
    }
  }, [visible]);

  return (
      {visible && (
        <Profiler
          id="v3"
            times.push(actualDuration);
            setTimeout(() => {
              setVisible(false);
            }, 24);
          }}>
          <Benchmark />
        </Profiler>
      )}
    </View>
  );
}

  container: {
    flex: 1,
  },
});

```
</details>

Before: ~280ms, after: ~253ms

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
@@ -1,6 +1,5 @@
 import { useEffect, useMemo, useRef } from 'react';
 import { getNextHandlerTag } from '../../handlers/getNextHandlerTag';
-import RNGestureHandlerModule from '../../RNGestureHandlerModule';
 import { useGestureCallbacks } from './useGestureCallbacks';
 import {
   prepareConfig,
@@ -12,7 +11,7 @@ import {
 import { tagMessage } from '../../utils';
 import { BaseGestureConfig, SingleGesture, SingleGestureName } from '../types';
 import { Platform } from 'react-native';
-import { scheduleFlushOperations } from '../../handlers/utils';
+import { NativeProxy } from '../NativeProxy';
 
 export function useGesture<THandlerData, TConfig>(
   type: SingleGestureName,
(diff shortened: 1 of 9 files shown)


packages/react-native-gesture-handler/src/v3/hooks/callbacks/useReanimatedEventHandler.ts
```diff
@@ -0,0 +1,69 @@
...
+    handlers,
+    reanimatedHandler?.context,
+    changeEventCalculator
+  );
+
+  const touchCallback = getTouchEventHandler(handlerTag, handlers);
+
+  const callback = (event: UnpackedGestureHandlerEvent<THandlerData>) => {
+    'worklet';
+    // console.log('Reanimated update event received:', event);
```
m-bert:
```suggestion
```

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -114,10 +106,24 @@ export function useGesture<THandlerData, TConfig>(
...
         onGestureHandlerTouchEvent,
-        onReanimatedStateChange,
-        onReanimatedUpdateEvent,
-        onReanimatedTouchEvent,
         onGestureHandlerAnimatedEvent,
+        // On web, we're triggering Reanimated callbacks ourselves, based on the type.
+        // To handle this properly, we need to provide all three callbacks, so we set
+        // all three to the Reanimated event handler.
+        // On native, Reanimated handles routing internally based on the event names
+        // passed to the useEvent hook. We only nee to pass it once, so that Reanimated
```
m-bert:
```suggestion
        // passed to the useEvent hook. We only need to pass it once, so that Reanimated
```

akwasniewski (approved):

j-piasecki:
> I can see that this PR contains changes https://github.com/software-mansion/react-native-gesture-handler/pull/3827 and https://github.com/software-mansion/react-native-gesture-handler/pull/3835, should we wait for them with merge?

Hmm, I forgot about it a few hours after reading it :sweat_smile:. I'll close https://github.com/software-mansion/react-native-gesture-handler/pull/3827 then.

PR #3837, opened 2025-11-25
[General] Listen for Reanimated events directly, without `createAnimatedComponent` when possible

j-piasecki:
## Description

Shouldn't need https://github.com/software-mansion/react-native-reanimated/pull/8655 to work properly

Instead of using `createAnimatedComponent`, which does a lot more stuff other than setting up events, we can set up events ourselves using Reanimated's `NativeEventsManager` directly.

The implementation in the PR should be independent of the Reanimated version being used, as long as it's a relatively modern one.

## Test plan

<details>
<summary>Tested on this stress-test</summary>

```jsx
import { GestureDetector, usePanGesture } from 'react-native-gesture-handler';
// import { PerfMonitor } from 'react-native-gesture-handler/src/v3/PerfMonitor';
import Animated, {
  useAnimatedStyle,
  useSharedValue,

const DATA = new Array(500).fill(null).map((_, i) => `Item ${i + 1}`);

function Item() {

    return {
    };
  });

  const pan = usePanGesture({
    // disableReanimated: true,
    onUpdate: (event) => {
      'worklet';
      console.log('pan onUpdate', event.handlerData.changeX);
    },
    onStart: () => {
      'worklet';
      console.log('pan onStart', _WORKLET);
    },
    onEnd: () => {
      'worklet';
      console.log('pan onEnd');
    },
    onBegin: () => {
      'worklet';
      console.log('pan onBegin');
    },
    onFinalize: () => {
      'worklet';
      console.log('pan onFinalize');
    },
    onTouchesDown: () => {
      'worklet';
      console.log('pan onTouchesDown');
    },
  });

  return (
    <View style={{ height: 80, padding: 16, backgroundColor: 'gray' }}>
        {/* <View collapsable={false} style={{opacity: 0.5}}> */}
        <Animated.View
          style={[
            { backgroundColor: 'red', height: '100%', aspectRatio: 1 },
            style,
          ]}
        />
        {/* </View> */}
      </GestureDetector>
    </View>
  );
}

function Benchmark() {
  return (
    <ScrollView
      style={{ flex: 1 }}
      {DATA.map((_, index) => (
        <Item key={index} />
      ))}
    </ScrollView>
  );
}

const TIMES = 35;

  const [visible, setVisible] = React.useState(false);

  useEffect(() => {
      setTimeout(() => {
        setVisible(true);
      }, 24);
    }

    if (times.length === TIMES) {
      sortedTimes.shift();
      sortedTimes.shift();
      sortedTimes.pop();
      sortedTimes.pop();
      const avgTime =

      // console.log(JSON.stringify(PerfMonitor.getMeasures(), null, 2));
      // PerfMonitor.clear();
    }
  }, [visible]);

  return (
      {visible && (
        <Profiler
          id="v3"
            times.push(actualDuration);
            setTimeout(() => {
              setVisible(false);
            }, 24);
          }}>
          <Benchmark />
        </Profiler>
      )}
    </View>
  );
}

  container: {
    flex: 1,
  },
});

```
</details>

Before: 614ms, after: 576ms

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/detectors/ReanimatedNativeDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/detectors/ReanimatedNativeDetector.tsx
@@ -11,7 +11,10 @@ import { findNodeHandle } from 'react-native';
 let NativeEventsManagerImpl = Reanimated?.NativeEventsManager;
 
 if (!NativeEventsManagerImpl) {
-  // Fallback to empty object when Reanimated or NativeEventsManager is not available
+  // When Reanimated.NativeEventsManager is undefined, it may be that an older
+  // Reanimated version is used which doesn't export NativeEventsManager, or
+  // Reanimated is not installed at all. For the older versions, try to import
+  // NativeEventsManager by a subpath. Otherwise, it will stay undefined.
   try {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     NativeEventsManagerImpl =


packages/react-native-gesture-handler/src/v3/detectors/HostGestureDetector.tsx
```diff
@@ -1,3 +1,4 @@
 import RNGestureHandlerDetectorNativeComponent from '../../specs/RNGestureHandlerDetectorNativeComponent';
+export { NativeProps as RNGestureHandlerDetectorNativeComponentProps } from '../../specs/RNGestureHandlerDetectorNativeComponent';
```
m-bert:
```suggestion
export type { NativeProps as RNGestureHandlerDetectorNativeComponentProps } from '../../specs/RNGestureHandlerDetectorNativeComponent';
```

It's only a type, right?

packages/react-native-gesture-handler/src/v3/detectors/ReanimatedNativeDetector.tsx
```diff
@@ -0,0 +1,94 @@
...
+} from '../../handlers/gestures/reanimatedWrapper';
+import HostGestureDetector, {
+  type RNGestureHandlerDetectorNativeComponentProps,
+} from './HostGestureDetector';
+import { findNodeHandle } from 'react-native';
+
+let NativeEventsManagerImpl = Reanimated?.NativeEventsManager;
+
+if (!NativeEventsManagerImpl) {
+  // Fallback to empty object when Reanimated or NativeEventsManager is not available
```
m-bert: Maybe I've missed something, but is there a fallback to empty object?

packages/react-native-gesture-handler/src/v3/detectors/ReanimatedNativeDetector.tsx
```diff
@@ -0,0 +1,94 @@
...
+    viewRef.__nativeTag = nativeTag;
+    // @ts-expect-error NativeEventsManager should be defined here, if it isn't, we should
+    // go the fallback way and use Reanimated's createAnimatedComponent
+    eventManager.current = new NativeEventsManagerImpl({
+      props: reaProps,
+      _componentRef: viewRef,
+      _componentViewTag: nativeTag,
+      getComponentViewTag: () => {
+        return nativeTag;
+      },
```
m-bert:
```suggestion
      getComponentViewTag: () => nativeTag,
```

packages/react-native-gesture-handler/src/v3/detectors/ReanimatedNativeDetector.tsx
j-piasecki: I'm not even sure who hallucinated this, fixed in [df7eaac](https://github.com/software-mansion/react-native-gesture-handler/pull/3837/commits/df7eaac45bf4de25b968a6a16dbf87bf958bd6fa).

PR #3838, opened 2025-11-25
Handle `testID` prop

m-bert:
## Description

This PR adds handling of `testID` property.

## Test plan

`yarn test api_v3`

### DIFF
--- a/packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
+++ b/packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
@@ -17,9 +17,8 @@ export function registerGesture<THandlerData, TConfig>(
   handlerTag: number,
   gesture: SingleGesture<THandlerData, TConfig>
 ) {
-  hookGestures.set(handlerTag, gesture);
-
   if (isTestEnv() && gesture.config.testID) {
+    hookGestures.set(handlerTag, gesture);
     testIDs.set(gesture.config.testID, handlerTag);
   }
 }
@@ -29,9 +28,8 @@ export function unregisterGesture(handlerTag: number) {
 
   if (gesture && isTestEnv() && gesture.config.testID) {
     testIDs.delete(gesture.config.testID);
+    hookGestures.delete(handlerTag);
   }
-
-  hookGestures.delete(handlerTag);
 }
 
 export function registerHandler(
--- a/packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
+++ b/packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
@@ -60,7 +60,7 @@ import {
 } from '../handlers/TapGestureHandler';
 import { State } from '../State';
 import { hasProperty, withPrevAndCurrent } from '../utils';
-import { SingleGesture } from '../v3/types';
+import type { SingleGesture } from '../v3/types';
 import { maybeUnpackValue } from '../v3/hooks/utils';
 
 // Load fireEvent conditionally, so RNGH may be used in setups without testing-library


packages/react-native-gesture-handler/src/v3/createNativeWrapper.tsx
```diff
@@ -19,13 +19,22 @@ export default function createNativeWrapper<P>(
...
+      }
   ) => {
     // Filter out props that should be passed to gesture handler wrapper
     const { gestureHandlerProps, childProps } = Object.keys(props).reduce(
       (res, key) => {
+        if (key === 'testID') {
+          res.gestureHandlerProps[key] = props[key];
+          // @ts-ignore FIXME(TS)
+          res.childProps[key] = props[key];
+        }
```
m-bert: The idea here is to use the same `testID` for both, component and gesture. This way, it is possible to use `getByGestureTestId`, which returns gesture,  and locate component by the same ID (though I have not tested that with other testing tools).

If you can see that this fails, let me know.

packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
```diff
@@ -421,6 +429,33 @@ function getHandlerData(
...
+        };
+
+        if (eventName === 'onGestureHandlerStateChange') {
+          componentOrGesture.detectorCallbacks.onGestureHandlerStateChange({
+            oldState: oldState as State,
+            ...event,
+          });
+        } else if (eventName === 'onGestureHandlerEvent') {
+          componentOrGesture.detectorCallbacks.onGestureHandlerEvent?.(event);
+        }
```
j-piasecki, line 454: Don't we need to force `disableReanimated` to be sure those contain all handlers?

packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
```diff
@@ -1,12 +1,39 @@
...
+const hookGestures = new Map<number, SingleGesture<any, any>>();
 const gestures = new Map<number, GestureType>();
 const oldHandlers = new Map<number, GestureHandlerCallbacks>();
 const testIDs = new Map<string, number>();
 
+export function registerGesture<THandlerData, TConfig>(
+  handlerTag: number,
+  gesture: SingleGesture<THandlerData, TConfig>
+) {
+  hookGestures.set(handlerTag, gesture);
```
j-piasecki: Do we need to store it outside of Jest?

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -135,4 +120,28 @@ export function useGesture<THandlerData, TConfig>(
...
+    };
+  }, [type, tag]);
+
+  useEffect(() => {
+    const preparedConfig = prepareConfigForNativeSide(type, config);
+    RNGestureHandlerModule.setGestureHandlerConfig(tag, preparedConfig);
+    scheduleFlushOperations();
+
+    bindSharedValues(config, tag);
+
```
```suggestion
```

packages/react-native-gesture-handler/src/__tests__/api_v3.test.tsx
```diff
@@ -0,0 +1,60 @@
+import { usePanGesture } from '../v3/hooks/gestures';
```
j-piasecki, line 1: Will this run on our GH Actions?

m-bert: For now I have not added this to any workflow. If you want to I can do this, but then I believe we should write more sophisticated tests.

packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
m-bert: If we were to keep **key-based** relations concept this might be useful. But I didn't think about it that much so I'm not sure yet. 

So for now I believe the answer would be no.

packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
m-bert: Initially JS callbacks were always present and I don't think that this has changed. `disableReanimated` should only change Reanimated callbacks.

packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
j-piasecki: I don't think this is coming soon, and would most likely need to be implemented on the native side.

packages/react-native-gesture-handler/src/__tests__/api_v3.test.tsx
j-piasecki: I think we can set it up and work on the tests in the follow-ups.

m-bert: Ok, for now I've combined it with relations tests in [57db777](https://github.com/software-mansion/react-native-gesture-handler/pull/3838/commits/57db777b97146fa58513efdd142d56fe5e1f67a3). Let me know if you prefer to keep them separated.

packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
m-bert: Okay, I've moved this inside `isTestEnv` condition in [6b7fff0](https://github.com/software-mansion/react-native-gesture-handler/pull/3838/commits/6b7fff0363fc3796ffa6e2c6436de8e262c33e28).

packages/react-native-gesture-handler/src/__tests__/api_v3.test.tsx
j-piasecki: Why not just run `yarn test`?

m-bert: `yarn test` will run all of them, don't we want only tests for V3?

j-piasecki (approved):

PR #3840, opened 2025-11-26
[General] Unwrap `handlerData` when passing event to user-defined callbacks

j-piasecki:
## Description

Due to how codegen is structured, we rely on a dynamic `handlerData` field in events so we don't have to define an event handler for each event type we may be sending. Currently, this is reflected in the API - when reading data from events, it needs to be read from `handlerData`.

This PR adds flattened types and a flatten step to get rid of that step, so that data can be read directly from the event.

This doesn't impact animated, so the event mapping there needs to include both `nativeEvent` and `handlerData`.

## Test plan

`console.log` the events

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGestureCallbacks.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGestureCallbacks.ts
@@ -23,7 +23,6 @@ import { useMemo } from 'react';
 
 function guardJSAnimatedEvent(handler: (...args: unknown[]) => void) {
   return (...args: unknown[]) => {
-    console.log('Warning: Animated.event called in JS mode.');
     try {
       handler(...args);
     } catch (e) {


packages/react-native-gesture-handler/src/v3/hooks/useGestureCallbacks.ts
```diff
@@ -19,6 +23,7 @@ import { useMemo } from 'react';
 
 function guardJSAnimatedEvent(handler: (...args: unknown[]) => void) {
   return (...args: unknown[]) => {
+    console.log('Warning: Animated.event called in JS mode.');
```
m-bert: Is it necessary? If so, why don't we change it to `warn` or `error`?

j-piasecki: It's not, removed in [80d9035](https://github.com/software-mansion/react-native-gesture-handler/pull/3840/commits/80d90354df4b53e0c305c6e80613a152d879e7de)

PR #3841, opened 2025-11-26
[iOS] Fix Animated events with no native driver

j-piasecki:
## Description

I noticed that on iOS, Animated Events with no native driver weren't working. This was caused by the fact that Animated events were only sent using the old pipeline, so that NativeAnimatedModule would intercept them. However, when not using the native driver, the event handler is just a JS function, which was effectively ignored.

## Test plan

Checked the `NativeDetector` screen with the `useNativeDriver` set to `false`.

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
@@ -464,22 +464,21 @@ constexpr int NEW_ARCH_NUMBER_OF_ATTACH_RETRIES = 25;
                          forView:(RNGHUIView *)detectorView
 {
   if ([event isKindOfClass:[RNGestureHandlerEvent class]]) {
+    RNGestureHandlerEvent *gestureEvent = (RNGestureHandlerEvent *)event;
+
     switch (eventHandlerType) {
       case RNGestureHandlerEventHandlerTypeAnimated: {
         [self sendEventForNativeAnimatedEvent:event];
-        RNGestureHandlerEvent *gestureEvent = (RNGestureHandlerEvent *)event;
         auto nativeEvent = [gestureEvent getNativeEvent];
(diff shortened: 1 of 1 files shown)


packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
```diff
@@ -465,9 +465,13 @@ - (void)sendNativeOrVirtualEvent:(RNGestureHandlerStateChange *)event
 {
   if ([event isKindOfClass:[RNGestureHandlerEvent class]]) {
     switch (eventHandlerType) {
-      case RNGestureHandlerEventHandlerTypeAnimated:
+      case RNGestureHandlerEventHandlerTypeAnimated: {
         [self sendEventForNativeAnimatedEvent:event];
+        RNGestureHandlerEvent *gestureEvent = (RNGestureHandlerEvent *)event;
+        auto nativeEvent = [gestureEvent getNativeEvent];
```
m-bert: What about inline cast?


```suggestion
        auto nativeEvent = [(RNGestureHandlerEvent *)event getNativeEvent];
```

m-bert: Also, we have to send both simultaneously

```obj-c
[self sendEventForNativeAnimatedEvent:event];
```

and 

```obj-c
[(RNGestureHandlerDetector *)detectorView dispatchAnimatedGestureEvent:nativeEvent];
```

right?

j-piasecki: The cast was repeated in every branch, so I moved it outside of the switch expression. It's now just after the if, which checks if the cast is safe, so it should be easier to read.

j-piasecki: > Also, we have to send both simultaneously

We have to send one for the native driver and the other for the JS driver. Hopefully, the C++ Animated will be able to read the events sent through the codegened path, and we will be able to remove one of those in the future.

PR #3844, opened 2025-11-27
[General] Hide internal fields from gesture events

j-piasecki:
## Description

Hides internal fields (`state`, `oldState`, and `handlerTag`) from the Gesture Handler events exposed externally. Those shouldn't be needed when using the v3 api and clutter the events.

This also means that update events don't differ from the update events in shape, so they became redundant.

## Test plan

Verify that `state`, `oldState`, and `handlerTag` are hidden from events.

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFlingGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFlingGesture.ts
@@ -4,7 +4,7 @@ import {
   SingleGesture,
   SingleGestureName,
   WithSharedValue,
-  GestureUpdateEvent,
+  GestureEvent,
 } from '../../../types';
 import { useGesture } from '../../useGesture';
 import { useClonedAndRemappedConfig } from '../../utils';
@@ -27,7 +27,7 @@ type FlingGestureInternalConfig = BaseDiscreteGestureConfig<
 export type FlingGestureConfig =
   ExcludeInternalConfigProps<FlingGestureInternalConfig>;
 
-export type FlingGestureEvent = GestureUpdateEvent<FlingHandlerData>;
+export type FlingGestureEvent = GestureEvent<FlingHandlerData>;
 
 export type FlingGesture = SingleGesture<
   FlingHandlerData,
--- a/packages/react-native-gesture-handler/src/v3/types/EventTypes.ts
+++ b/packages/react-native-gesture-handler/src/v3/types/EventTypes.ts
@@ -49,10 +49,10 @@ export type TouchEvent =
   | GestureTouchEvent
   | NativeSyntheticEvent<GestureTouchEvent>;
 
-export type GestureUpdateEvent<THandlerData> = THandlerData;
+export type GestureEvent<THandlerData> = THandlerData;
 
 export type UnpackedGestureHandlerEvent<THandlerData> =
-  | GestureUpdateEvent<THandlerData>
+  | GestureEvent<THandlerData>
   | GestureTouchEvent;
 
 // This is not how Animated.event is typed in React Native. We add _argMapping in order to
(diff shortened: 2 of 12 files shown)


packages/react-native-gesture-handler/src/v3/types/EventTypes.ts
```diff
@@ -49,12 +49,9 @@ export type TouchEvent =
   | GestureTouchEvent
   | NativeSyntheticEvent<GestureTouchEvent>;
 
-export type GestureStateChangeEvent<THandlerData> = StateChangeEventPayload &
-  THandlerData;
-export type GestureUpdateEvent<THandlerData> = EventPayload & THandlerData;
+export type GestureUpdateEvent<THandlerData> = THandlerData;
```
m-bert: I'm not sure if this won't result in some name conflicts, but what about making it `GestureEvent` instead?

packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFlingGesture.ts
```diff
@@ -28,10 +27,7 @@ type FlingGestureInternalConfig = BaseDiscreteGestureConfig<
 export type FlingGestureConfig =
   ExcludeInternalConfigProps<FlingGestureInternalConfig>;
 
-export type FlingGestureStateChangeEvent =
-  GestureStateChangeEvent<FlingHandlerData>;
-
-export type FlingGestureUpdateEvent = GestureUpdateEvent<FlingHandlerData>;
+export type FlingGestureEvent = GestureUpdateEvent<FlingHandlerData>;
```
m-bert: If now `GestureUpdateEvent` acts as identity, can't we do

```tsx
export type FlingGestureEvent = FlingHandlerData;
```

Of course I can see why we would like to keep the former version, so I'll leave that up to you.

packages/react-native-gesture-handler/src/v3/types/EventTypes.ts
j-piasecki: Done in https://github.com/software-mansion/react-native-gesture-handler/pull/3844/commits/f30f2f010a2e1398bb147138e4376b97d7ae3821

packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/useFlingGesture.ts
j-piasecki: I'm not sure whether it won't be useful in the future. It might be useful to have the abstraction in place.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -95,11 +95,12 @@ - (void)dispatchGestureEvent:(RNGestureHandlerDetectorEventEmitter::OnGestureHan
...
-      .state = event.state,
-      .handlerTag = event.handlerTag,
-      .handlerData = event.handlerData,
-    });
+    std::dynamic_pointer_cast<const RNGestureHandlerDetectorEventEmitter>(_eventEmitter)
+        ->onGestureHandlerAnimatedEvent({
+            .state = event.state,
+            .handlerTag = event.handlerTag,
+            .handlerData = event.handlerData,
+        });
```
m-bert, line 103: It is only formatting, right? Do we want that in this PR?

j-piasecki: I didn't even notice that. It must have been added by lint-staged.

j-piasecki: I don't mind it, so I'll leave it up to you.

m-bert: I'd prefer not to have it here as it will affect blame. But if you leave it it will also be fine.

PR #3846, opened 2025-11-28
Add `rootContainerStyle` prop to `ReanimatedDrawerLayout`

jr-k:
**What this PR does**

This PR adds a new `mainContainerStyle` prop to the main component. Until now, it was not possible to apply custom styles to the wrapper since it was fully controlled internally. By exposing this prop, consumers can now override or extend the wrapper's styles and apply absolute positioning or any other layout adjustment when needed.

**Why**

I needed to apply absolute positioning to the top-level wrapper but there was no way to customize it. Adding this prop keeps the existing default behavior while allowing more flexibility for advanced layout cases.

**How**

- Added the `mainContainerStyle` prop
- Passed it directly to the `ReanimatedDrawerLayout` element
- Ensured it merges cleanly with the component's default styling


m-bert:
Hi @jr-k! Would you mind sharing an example where this prop is necessary?

jr-k:
Hi! To be honest, it has been a while and I do not remember the exact case that led me to add this React prop.

That said, I still think it is useful to expose it. It gives consumers more flexibility to customize the component when needed, without us having to anticipate every possible use case or block people later on.

packages/react-native-gesture-handler/src/components/ReanimatedDrawerLayout.tsx
```diff
@@ -197,6 +197,11 @@ export interface DrawerLayoutProps {
    */
   drawerContainerStyle?: StyleProp<ViewStyle>;
 
+  /**
+  * Style wrapping the main component.
+  */
+  mainContainerStyle?: StyleProp<ViewStyle>;
```
m-bert: Not sure if `main` fits here, maybe `background`?

m-bert (changes requested):
I guess it wouldn't hurt us to have such prop, though I'm still not sure about the use case.

This should also have its entry in [Drawer documentation](https://docs.swmansion.com/react-native-gesture-handler/docs/components/reanimated-drawer-layout).

m-bert (approved):
Hi @jr-k! Thank you again for this PR. I took the liberty of updating it since it had conflicts with `main` branch. We've also decided to use `rootContainerStyle` name instead. Let me know if this works for you, or do you have any other suggestions!

j-piasecki (approved):

jr-k:
Perfect, thanks!

PR #3849, opened 2025-11-28
[iOS] Fix pan not triggering `onFinalize` when blocked

j-piasecki:
## Description

I noticed that https://github.com/software-mansion/react-native-gesture-handler/pull/3756 restored calling `triggerAction` to all gestures that were affected by https://github.com/software-mansion/react-native-gesture-handler/pull/3740, but not to Pan. This is causing Pan to trigger `onBegin` without then triggering `onFinalize`.

This PR restores calling `triggerAction` in `reset`.

## Test plan

```jsx
import { View, ScrollView } from 'react-native';
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,

  const pan = Gesture.Pan()
    .runOnJS(true)
    .onBegin(() => {
      'worklet';
      console.log('begin');
    })
    .onFinalize(() => {
      'worklet';
      console.log('finalize');
    });
  return (
    <GestureHandlerRootView>
        <View style={{ flex: 1, backgroundColor: 'blue' }}>
          <ScrollViewExample pan={pan} />
        </View>
      </GestureDetector>
    </GestureHandlerRootView>
  );
}

function ScrollViewExample({ pan }: any) {
  const native = Gesture.Native().blocksExternalGesture(pan).runOnJS(true);
  return (
    <GestureDetector gesture={native}>
      <ScrollView
        horizontal
        pagingEnabled
        style={{
          flex: 1,
          backgroundColor: 'yellow',
        }}>
        <View
          style={{
            width: 250,
            height: '100%',
            backgroundColor: 'green',
          }}
        />
        <View
          style={{
            width: 250,
            height: '100%',
            backgroundColor: 'purple',
          }}
        />
      </ScrollView>
    </GestureDetector>
  );
}

```

### DIFF
--- a/packages/react-native-gesture-handler/apple/Handlers/RNPanHandler.m
+++ b/packages/react-native-gesture-handler/apple/Handlers/RNPanHandler.m
@@ -259,6 +259,7 @@
 
 - (void)reset
 {
+  [self triggerAction];
   [_gestureHandler.pointerTracker reset];
   [NSObject cancelPreviousPerformRequestsWithTarget:self selector:@selector(activateAfterLongPress) object:nil];
   self.enabled = YES;


akwasniewski (approved):

m-bert (changes requested):
Seems like it doesn't work on iOS 26 😞 

## 26

https://github.com/user-attachments/assets/10e4f70f-4012-43eb-adc5-7ab207b003ec

## 18.5

https://github.com/user-attachments/assets/c1bd7916-9ffd-45b9-8869-9e098eed68fa

j-piasecki:
@m-bert It seems like the iOS 26 bug is unrelated to this change. Do you think this one can be merged, or do you prefer both to be fixed in one go?

PR #3850, opened 2025-11-28
[android] fix scroll actions interpreted as touch

kosmydel:
## Description

The scroll action on the Meta Quest Joystick lead to triggering the Pan gesture when it shouldn't. This PR filters out this `ACTION_SCROLL` by applying whitelist in the `dispatchGenericMotionEvent` for the hover events only.


j-piasecki (approved):
Thanks ❤️!

PR #3854, opened 2025-12-01
Update mocks

akwasniewski:
## Description

Updates mocks to mock `HostGestureDetector`,  and have a separate file for mocking buttons. Moreover it adds tests for correct error throws in V3.

## Test plan

`yarn test`

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -595,6 +595,16 @@ open class GestureHandler {
       // generated faster than they can be treated by JS thread
       eventCoalescingKey = nextEventCoalescingKey++
     }
+
+    check(hostDetectorView != null || orchestrator != null) {
+      "Manually handled gesture had not been assigned to any detector"
+    }
+
+    if (orchestrator == null) {
+      // If the state is set manually, the handler may not have been fully recorded by the orchestrator.
+      hostDetectorView?.recordHandlerIfNotPresent(this)
+    }
+
     orchestrator!!.onHandlerStateChange(this, newState, oldState)
     onStateChange(newState, oldState)
   }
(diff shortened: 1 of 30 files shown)


packages/react-native-gesture-handler/jestSetup.js
```diff
@@ -7,7 +7,7 @@ jest.mock('./lib/commonjs/RNGestureHandlerModule', () =>
   require('./lib/commonjs/mocks/mocks')
```
j-piasecki: 1. Can this be renamed from `mocks`?
2. I don't think it should contain anything more than the module methods. The old gesture handlers, gesture-wrapped components, and host detector should be separate and mocked to the correct files, no?

I think the current setup dates back to the time when all of Gesture Handler was a single file, so it was correct to mock everything to a single file.

m-bert:
What's the state of this PR?

packages/react-native-gesture-handler/jestSetup.js
akwasniewski: > Can this be renamed from mocks?

I renamed it to module as now it only mocks the module file

>  don't think it should contain anything more than the module methods. The old gesture handlers, gesture-wrapped components, and host detector should be separate and mocked to the correct files, no?

Seperated them into different files. However, old gesture handlers are declared in the module so it only makes sense to mock them with the rest of the module.

I also removed the `Direction` and `State` consts, it seems to work without them.

packages/react-native-gesture-handler/src/mocks/module.tsx
```diff
@@ -70,7 +33,12 @@ export default {
...
-  Directions,
-  State,
+  NativeViewGestureHandler,
+  TapGestureHandler,
+  ForceTouchGestureHandler,
+  LongPressGestureHandler,
+  PinchGestureHandler,
+  RotationGestureHandler,
+  FlingGestureHandler,
+  PanGestureHandler,
```
j-piasecki: Do those actually do anything? This file is used to mock "RNGestureHandlerModule" which doesn't contain the old handlers, no?

akwasniewski: We don't really need to mock them anyway, removed in f2eca7fad29fb8b0dff479447944d7d702c5ef16

j-piasecki (approved):

PR #3725, opened 2025-10-01
v3 examples

akwasniewski:
## Description

This PR adds examples for v3 api. Most of them are rewritten v2 examples, but I added a few.

New Examples: 
- SVG example - to showcase LogicDetector
- Lock example - I got carried away debugging old transformations example

Rewritten Examples:
- [x] Bottom Sheet
- [x] Calculator
- [x] Camera
- [x] Chat Heads
- [x] Fling
- [x] Hover
- [x] Hoverable Icons
- [x] Overlap
- [x] Velocity Test

To be done in the future: 
- Manual Gestures - they are not yet implemented on v3

## Test plan

See if the examples in v3 section from common-example work

PR #3857, opened 2025-12-02
V3 examples

akwasniewski:
## Description

Made a new PR, as it is easier than attempt to merge the [old one](https://github.com/software-mansion/react-native-gesture-handler/pull/3725) with next.

This PR adds a toggle in the example app to switch between Legacy and V3 api examples. In the new api section I added split into few sections. Most of them are rewritten old V2 examples with more consistent styling.
 I also added a Feedback component to showcase features console free. 
The new examples are:
* basic examples for every gestures
* lock to showcase combining gestures
* sharedValue, nestedText, svg - to showcase VirtualDetector 
* animated - to showcase cooperation with animated

## Test plan

See if the examples in new api section from common-example work

### DIFF
--- a/packages/react-native-gesture-handler/src/index.ts
+++ b/packages/react-native-gesture-handler/src/index.ts
@@ -15,10 +15,13 @@ export type {
   HandlerStateChangeEventPayload,
   // Pointer events
   GestureTouchEvent,
+  GestureTouchEvent as SingleGestureTouchEvent,
   TouchData,
   // New api event types
   GestureUpdateEvent,
   GestureStateChangeEvent,
+  // Config types
+  ActiveCursor,
 } from './handlers/gestureHandlerCommon';
 export { MouseButton } from './handlers/gestureHandlerCommon';
 export type { GestureType } from './handlers/gestures/gesture';
@@ -152,8 +155,6 @@ export type {
   PressableStateCallbackType,
 } from './components/Pressable';
 export { default as LegacyPressable } from './components/Pressable';
-
-export type { GestureTouchEvent as SingleGestureTouchEvent } from './handlers/gestureHandlerCommon';
 
 export * from './v3';
 
--- a/packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
+++ b/packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
@@ -434,7 +434,7 @@ function getHandlerData(
     return {
       handlerType: componentOrGesture.type as HandlerNames,
       handlerTag: componentOrGesture.handlerTag,
-      enabled: maybeUnpackValue(componentOrGesture.config.enabled),
+      enabled: maybeUnpackValue<boolean>(componentOrGesture.config.enabled),
       emitEvent: (eventName, args) => {
         const { state, oldState, handlerTag, ...rest } = args.nativeEvent;
 
--- a/packages/react-native-gesture-handler/src/v3/hooks/gestures/pan/usePanGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/gestures/pan/usePanGesture.ts
@@ -76,8 +76,8 @@ function validateOffsetsArray(
     return;
   }
 
-  const offsetStart = maybeUnpackValue(offsets[0]);
-  const offsetEnd = maybeUnpackValue(offsets[1]);
+  const offsetStart = maybeUnpackValue<number>(offsets[0]);
+  const offsetEnd = maybeUnpackValue<number>(offsets[1]);
 
   if (offsetStart > 0 || offsetEnd < 0) {
     throw new Error(
@@ -125,7 +125,7 @@ function transformOffsetProp(
     config[`${propName}Start`] = propValue[0];
     config[`${propName}End`] = propValue[1];
   } else {
-    const offsetValue = maybeUnpackValue(propValue);
+    const offsetValue = maybeUnpackValue<number>(propValue);
 
     if (offsetValue < 0) {
       config[`${propName}Start`] = propValue;
--- a/packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
@@ -99,7 +99,9 @@ export function hasWorkletEventHandlers<THandlerData, TConfig>(
   );
 }
 
-export function maybeUnpackValue<T>(v: SharedValueOrT<T>) {
+export function maybeUnpackValue<T>(
+  v: SharedValueOrT<T, boolean> | undefined
+): T {
   'worklet';
 
   return (Reanimated?.isSharedValue(v) ? v.value : v) as T;
--- a/packages/react-native-gesture-handler/src/v3/types/ConfigTypes.ts
+++ b/packages/react-native-gesture-handler/src/v3/types/ConfigTypes.ts
@@ -68,7 +68,7 @@ export type CommonGestureConfig = {
     cancelsTouchesInView?: boolean;
     manualActivation?: boolean;
   },
-  HitSlop | ActiveCursor | MouseButton
+  ActiveCursor | MouseButton
 >;
 
 export type ComposedGestureConfig = {
--- a/packages/react-native-gesture-handler/src/v3/types/ReanimatedTypes.ts
+++ b/packages/react-native-gesture-handler/src/v3/types/ReanimatedTypes.ts
@@ -10,19 +10,31 @@ export type SharedValue<Value = unknown> = {
   ) => void;
 };
 
-export type SharedValueOrT<T> = T | SharedValue<Exclude<T, undefined>>;
+// Utility type that turns `T` into `T | SharedValue<T>`. `P` is used to avoid splitting union types.
+export type SharedValueOrT<T, P = never> =
+  // We always want to get `T` in the resulting type.
+  | T
+  // If `T` is one of the types in `P`, we don't want to split the union, so we return SharedValue<T>.
+  | (Exclude<T, undefined> extends P
+      ? SharedValue<Exclude<T, undefined>>
+      : // If `T` is not in `P`, we want to split the union and wrap each member with SharedValue, using Distributive Conditional Types (https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types).
+        T extends any
+        ? // Wrap each member of the union with SharedValue.
+          SharedValue<Exclude<T, undefined>>
+        : // If `T` is only one type, just return SharedValue<T>.
+          SharedValue<Exclude<T, undefined>>);
 
 // Utility type that decides whether to recurse for objects or apply SharedValue directly.
 export type WithSharedValue<T, P = never> = T extends object
   ? WithSharedValueRecursive<T, P>
-  : Simplify<SharedValueOrT<T>>;
+  : Simplify<SharedValueOrT<T, P>>;
 
 // Apply SharedValue recursively. P is used to make sure that composed types won't be expanded.
 // For example, if we pass `HoverEffect` as P, then resulting type will have HoverEffect | SharedValue<HoverEffect>,
 // not HoverEffect, SharedValue<HoverEffect.NONE>, ...
 type WithSharedValueRecursive<T extends object, P> = {
   [K in keyof T]: Exclude<T[K], undefined> extends P
-    ? Simplify<SharedValueOrT<T[K]>>
+    ? Simplify<SharedValueOrT<T[K], P>>
     : // Special case for boolean as passing `boolean` as P doesn't look ok.
       boolean extends T[K]
       ? boolean | SharedValue<boolean>


apps/common-app/src/new_api/showcase/shared_value/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  return (
+    <View style={styles.centerView}>
+      <View>
+        <GestureDetector gesture={gestures}>
+          <Animated.View style={[styles.box, animatedStyle]} />
+        </GestureDetector>
+      </View>
+      <Text style={styles.instructions}>
+        The ball has simultanous pan and longPress gestures. Upon update pan
+        changes minDistanceof longPress, such that longPress will fail if is
```
j-piasecki:
```suggestion
        changes maxDistance of longPress, such that longPress will fail if it's
```

apps/common-app/src/new_api/showcase/shared_value/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      });
+    },
+    onFinalize: () => {
+      'worklet';
+      colorProgress.value = withTiming(0, {
+        duration: 100,
+      });
+    },
+    minDuration: 1000,
+    maxDistance: maxLongPressDistance,
```
j-piasecki, line 56: On iOS, `maxDistance` doesn't seem to do anything. Could you investigate that?

apps/common-app/src/new_api/showcase/animated/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const rotateZ = rotation.interpolate({
+    inputRange: [-20, 20],
+    outputRange: ['-20deg', '20deg'],
+  });
+
+  return (
+    <View style={styles.container}>
+      <GestureDetector gesture={panGesture}>
+        <Animated.View
```
j-piasecki, line 127: I like this one! Can it also change background color on `onBegin`?

apps/common-app/src/new_api/components/buttons/index.tsx
```diff
@@ -31,9 +32,9 @@ function ButtonWrapper({ ButtonComponent, color }: ButtonWrapperProps) {
 export default function ButtonsExample() {
   return (
     <GestureHandlerRootView style={styles.container}>
-      <ButtonWrapper ButtonComponent={BaseButton} color="#2196F3" />
-      <ButtonWrapper ButtonComponent={RectButton} color="#f44336" />
-      <ButtonWrapper ButtonComponent={BorderlessButton} color="#ff9800" />
+      <ButtonWrapper ButtonComponent={BaseButton} color={COLORS.KINDA_BLUE} />
+      <ButtonWrapper ButtonComponent={RectButton} color={COLORS.NAVY} />
+      <ButtonWrapper ButtonComponent={BorderlessButton} color={COLORS.PURPLE} />
```
j-piasecki: They only `console.log` when pressed, should they also show some message on screen?

apps/common-app/App.tsx
```diff
@@ -273,129 +57,168 @@ export default function App() {
...
-      ([openLastExample, lastExample]) => {
-        if (openLastExample[1] === 'true' && lastExample[1]) {
+    useEffect(() => {
+      void AsyncStorage.multiGet([
+        OPEN_LAST_EXAMPLE_KEY,
+        LAST_EXAMPLE_KEY,
+        SHOW_LEGACY_EXAMPLES_KEY,
+      ]).then(([openLastExample, lastExample, showLegacyExamples]) => {
+        const showLegacy = showLegacyExamples[1] === 'true';
+        if (showLegacyExamples[1]) {
```
j-piasecki:
```suggestion
        if (showLegacyExamples[1] != null) {
```

I'm not sure whether this is what this check was supposed to do, or whether it should check `showLegacy` - imo it should be explicit.

apps/common-app/App.tsx
```diff
@@ -435,18 +262,15 @@ const styles = StyleSheet.create({
...
-  autoOpenSetting: {
-    margin: 16,
-    borderRadius: 16,
+    margin: 12,
+    borderRadius: 10,
     backgroundColor: '#eef0ff',
-    paddingHorizontal: 16,
-    justifyContent: 'space-between',
+    padding: 10,
+    justifyContent: 'center',
```
j-piasecki, line 272: Can we stick to multiples of 4?

Also, the gap between the buttons is larger than the gap between the buttons and the screen edge.

apps/common-app/src/common.tsx
```diff
@@ -25,9 +75,101 @@ export class LoremIpsum extends React.Component<Props> {
...
+
+type FeedbackProps = {
+  duration?: number;
+};
+
+export type FeedbackHandle = {
+  showMessage: (message: string) => void;
+};
+
+export const Feedback = forwardRef<FeedbackHandle, FeedbackProps>(
```
j-piasecki: Copying from linter:

> In React 19, 'forwardRef' is no longer necessary. Pass 'ref' as a prop instead

apps/common-app/src/new_api/complicated/camera/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+    },
+  });
+
+  const takeVideoGesture = useLongPressGesture({
+    shouldCancelWhenOutside: false,
+    maxDistance: 10000,
+    onActivate: () => {
+      'worklet';
+      runOnJS(startRecording)();
+      captureProgress.value = withTiming(1, { duration: VIDEO_DURATION });
```
j-piasecki: I think this is missing `easing: Easing.linear` - the animation speeds up and slows down.

apps/common-app/src/new_api/complicated/lock/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const snapThreshold = 0.4;
+  const scaleThreshold = 0.1;
+  const minScale = 0.5;
+  const maxScale = 1;
+  const TWO_PI = 2 * Math.PI;
+
+  // Tap to lock
+  const tap = useTapGesture({
+    onDeactivate: () => {
+      'worklet';
```
j-piasecki: If you use the latest `react-native-worklets`, this directive should be auto-included. Maybe it's worth it to update worklets and remove the directive?

apps/common-app/src/new_api/hover_mouse/hover/index.tsx
```diff
@@ -0,0 +1,156 @@
+import { COLORS, Feedback } from '../../../common';
```
j-piasecki: This is not platform-specific, it just requires additional hardware on mobile (mouse, stylus).

apps/common-app/src/new_api/showcase/overlap/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+      }
+    },
+    disableReanimated: true,
+  });
+
+  return (
+    <View style={styles.subcontainer}>
+      <View style={styles.row}>
+        <Text style={styles.text}>Overlap Siblings</Text>
+        <InterceptingGestureDetector>
```
j-piasecki, line 57: Out of curiosity, why `InterceptingGestureDetector` here?

apps/common-app/src/new_api/showcase/overlap/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+    },
+    disableReanimated: true,
+  });
+
+  return (
+    <View style={styles.subcontainer}>
+      <View style={styles.row}>
+        <Text style={styles.text}>Overlap Child</Text>
+
+        <InterceptingGestureDetector>
```
j-piasecki, line 102: Same

apps/common-app/src/new_api/showcase/shared_value/index.tsx
akwasniewski: cab8a09be19d1e86e5bfc24ee002c9be210bed2b

apps/common-app/src/new_api/showcase/animated/index.tsx
akwasniewski: Sure, I also made it turn red/green on tilt 68893ad438cb8e72ce847bacee4bf369d31b770d

apps/common-app/src/new_api/components/buttons/index.tsx
akwasniewski: They should 6b70961407cc0c8383a8bb1b561274890fed1fbb

apps/common-app/App.tsx
akwasniewski: thanks, b16d4f12b3b57fedc4f4697d9c3d35b2dea487d0

apps/common-app/App.tsx
akwasniewski: c300e2741af52752e56b63ffcf3999952374bee8

apps/common-app/src/common.tsx
akwasniewski: b967bef42b9c69c34797c5c947ac7c9d5ce9f316

apps/common-app/src/new_api/complicated/camera/index.tsx
akwasniewski: e2c93222e5f947318dca89edf70a40e4a15be66c

apps/common-app/src/new_api/showcase/overlap/index.tsx
akwasniewski: Then it won't change which square is on top, as overlap only supports this for siblings. We cannot apply elevated on the `GestureDetector`, thus I decided `VirtualDetector` works best here.

apps/common-app/src/new_api/complicated/lock/index.tsx
akwasniewski: Ohh, great. Thank you, removed in de7371abff6e0c6f3b6b1e7e4dc2f1168f401535

apps/common-app/src/new_api/complicated/camera/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+    // no-op
+  }
+
+  function stopRecording() {
+    alert("I didn't bother to implement this either :)");
+  }
+
+  const takePhotoGesture = useTapGesture({
+    onDeactivate: () => {
+      'worklet';
```
m-bert:
```suggestion
```

This should no longer be necessary.

apps/common-app/src/new_api/complicated/camera/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+    },
+  });
+
+  const pinchZoomGesture = usePinchGesture({
+    onUpdate: (e) => {
+      zoom.value = Math.max(
+        1,
+        Math.min(2, zoom.value * ((e.scaleChange - 1) * 0.2 + 1))
+      );
+      console.log(zoom.value);
```
m-bert: Actually, do we need that console.log?

apps/common-app/src/new_api/complicated/camera/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+  const changeCameraGesture = useTapGesture({
+    numberOfTaps: 2,
+    onDeactivate: () => {
+      setFacing((f) => (f === 'back' ? 'front' : 'back'));
+    },
+    disableReanimated: true,
+  });
+
+  const raceGesture = useCompetingGestures(
```
m-bert:
```suggestion
  const competingGesture = useCompetingGestures(
```

apps/common-app/src/new_api/complicated/camera/index.tsx
m-bert, line 1: I also thought about splitting this file into multiple files, as it contains quite a lot elements. Let me know what you think.

apps/common-app/src/new_api/complicated/chat_heads/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+const CHAT_HEADS = [
+  { imageUrl: 'https://avatars0.githubusercontent.com/u/379606?v=4&s=460' },
+  { imageUrl: 'https://avatars3.githubusercontent.com/u/90494?v=4&s=460' },
+  { imageUrl: 'https://avatars3.githubusercontent.com/u/726445?v=4&s=460' },
+  { imageUrl: 'https://avatars.githubusercontent.com/u/15989228?v=4&s=460' },
+];
+
+interface AnimatedOffset {
+  x: Animated.SharedValue<number>;
+  y: Animated.SharedValue<number>;
```
m-bert: As far as I can see, importing `SharedValue` form `Animated` namespace is deprecated.

Image

apps/common-app/src/new_api/complicated/lock/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const tap = useTapGesture({
+    onDeactivate: () => {
+      if (savedRotation.value === 0 && scale.value === maxScale) {
+        runOnJS(setLocked)(false);
+      }
+    },
+  });
+
+  // Long press to cancel tap
+  const longPress = useLongPressGesture({});
```
m-bert: Why do we need that?

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+export default function App() {
+  const aref = useAnimatedRef<View>();
+  const offsetX = useSharedValue(0);
+  const offsetY = useSharedValue(0);
+  const isPressed = useSharedValue(false);
+  const colorProgress = useSharedValue(0);
+
+  const pan = usePanGesture({
+    onBegin: () => {
+      'worklet';
```
m-bert:
```suggestion
```

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      isPressed.value = true;
+      colorProgress.value = withTiming(1, {
+        duration: 100,
+      });
+    },
+    onUpdate: (event) => {
+      'worklet';
```
m-bert:
```suggestion
```

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      offsetX.value += event.changeX;
+      offsetY.value += event.changeY;
+    },
+    onFinalize: (event) => {
+      'worklet';
```
m-bert:
```suggestion
```

apps/common-app/src/new_api/hover_mouse/mouse_buttons/index.tsx
```diff
@@ -0,0 +1,299 @@
...
+import { COLORS, Feedback } from '../../../common';
+
+const COLORS_ARR = [
+  COLORS.PURPLE,
+  COLORS.GREEN,
+  COLORS.NAVY,
+  COLORS.RED,
+  COLORS.KINDA_RED,
+];
+export default function Buttons() {
```
m-bert, line 15: i'm a bit confused, `Titles` is an arrow function, tests are standard functions, also everything is inside `Buttons`. Is there a specific reason to it? I agree that this example lacked info about which button is supposed to work, but I think that the structure was cleaner.

apps/common-app/src/new_api/showcase/bottom_sheet/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+const FULLY_OPEN_SNAP_POINT = SNAP_POINTS_FROM_TOP[0];
+const CLOSED_SNAP_POINT = SNAP_POINTS_FROM_TOP[SNAP_POINTS_FROM_TOP.length - 1];
+
+function Example() {
+  const [snapPoint, setSnapPoint] = useState(CLOSED_SNAP_POINT);
+  const translationY = useSharedValue(0);
+  const scrollOffset = useSharedValue(0);
+  const bottomSheetTranslateY = useSharedValue(CLOSED_SNAP_POINT);
+
+  const onHandlerEndOnJS = (point: number) => {
```
m-bert: Given that now we have `deactivate`, not `end` this may be a bit confusing. Maybe something like `updateSnapPoint` to reflect its purpose?

apps/common-app/src/new_api/showcase/nested_text/nested_text.tsx
```diff
@@ -0,0 +1,140 @@
...
+    <View style={styles.subcontainer}>
+      <Text style={styles.header}>
+        Native Detector example - this one should work
+      </Text>
+      <InterceptingGestureDetector gesture={tapAll}>
+        <Text style={{ fontSize: 18, textAlign: 'center' }}>
+          Some text example running with RNGH
+          <VirtualGestureDetector gesture={tapFirstPart}>
+            <Text style={{ fontSize: 24, color: COLORS.NAVY }}>
+              {' '}
```
m-bert:
```suggestion
```

Can we remove it?

apps/common-app/src/new_api/showcase/shared_value/index.tsx
m-bert, line 1: This one seems to work completely different on both platforms (or should I say, doesn't seem to work on android)

https://github.com/user-attachments/assets/9b375272-c699-47e5-b1f3-7f8857b78d28

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
m-bert, line 1: Why is this "complicated" example?

apps/common-app/src/new_api/showcase/svg/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    },
+    disableReanimated: true,
+  });
+
+  const containerTap = useTapGesture({
+    onActivate: () => {
+      feedbackRef.current?.showMessage('Tapped container!');
+    },
+    disableReanimated: true,
+  });
```
m-bert, line 33: This doesn't seem to work on android.

https://github.com/user-attachments/assets/4b53afd3-7291-47c1-8b34-e8737aec0dcf

m-bert: Note that I'm aware of problems with detector shadow node, but in that case I've removed the early return and it works as in the video.

apps/common-app/src/new_api/simple/longPress/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const animatedStyle = useAnimatedStyle(() => {
+    const backgroundColor = interpolateColor(
+      colorProgress.value,
+      [0, 1, 2],
+      [COLORS.NAVY, COLORS.PURPLE, COLORS.KINDA_BLUE]
+    );
+
+    return {
+      backgroundColor,
+    };
```
m-bert, line 27: Can't we simply do

```js
    return {
      backgroundColor: interpolateColor(
        colorProgress.value,
        [0, 1, 2],
        [COLORS.NAVY, COLORS.PURPLE, COLORS.KINDA_BLUE]
      );
    };
```

m-bert: But if you prefer this it is also fine.

apps/common-app/src/new_api/simple/tap/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    onBegin: () => {
+      colorProgress.value = withTiming(1, {
+        duration: 100,
+      });
+    },
+    onFinalize: () => {
+      colorProgress.value = withTiming(0, {
+        duration: 100,
+      });
+    },
```
m-bert, line 35: As far as I remember, one of the old examples (probably with long press) had this 3 color interpolation - based on `success`. I think It would be nice to have it here (i.e. if gesture ended properly than it goes through green color, otherwise through red). 

It is not necessary, but now this example does not distinguish between ending properly and failing.

apps/common-app/src/new_api/simple/tap/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  return (
+    <View style={styles.centerView}>
+      <GestureDetector gesture={tapGesture}>
+        <Animated.View style={[styles.box, animatedStyle]} />
```
m-bert: Also I thought about it and it seems that many examples use the same `View` with the same styles, acting as a `box`. So in that case, maybe we could extract it? It would also be easier if we were to use it in other examples.

apps/common-app/App.tsx
```diff
@@ -250,8 +34,8 @@ type RootStackParamList = {
 };
 
 const Stack = createStackNavigator<RootStackParamList>();
-
 export default function App() {
+  const [showLegacyVersion, setShowLegacyVersion] = React.useState(false);
```
m-bert: If we have `useEffect` below, can we also have `useState` instead of `React.useState`? 😅

apps/common-app/App.tsx
```diff
@@ -273,129 +57,168 @@ export default function App() {
...
+    dest: string
+  ) {
+    AsyncStorage.setItem(LAST_EXAMPLE_KEY, dest);
+    navigation.navigate(dest);
+  }
 
-function MainScreen({ navigation }: StackScreenProps<ParamListBase>) {
-  const insets = useSafeAreaInsets();
+  function MainScreen({ navigation }: StackScreenProps<ParamListBase>) {
+    const insets = useSafeAreaInsets();
```
m-bert, line 85: Are those made automatically by linter?

apps/common-app/src/new_api/showcase/nested_text/nested_text.tsx
```diff
@@ -0,0 +1,140 @@
...
+          <GestureDetector gesture={tapSecondPart}>
+            <Text style={{ fontSize: 28, color: COLORS.KINDA_BLUE }}>
+              {' '}
+              or on this part
+            </Text>
+          </GestureDetector>
+          this part is not special :(
+        </Text>
+      </GestureDetector>
+      <Feedback ref={feedbackRef} />
```
m-bert, line 103: Seems like it is not shown:

https://github.com/user-attachments/assets/e7e59cce-276a-4e90-bf35-c92bae9c624d

apps/common-app/src/new_api/showcase/animated/index.tsx
m-bert, line 1: Beside color blinking at entering, it seems to be broken on android.

https://github.com/user-attachments/assets/57a9ce83-ab2e-4864-8249-7aeed9f14b3e

apps/common-app/src/new_api/hover_mouse/context_menu/index.tsx
m-bert, line 1: I'll just leave a comment - it seems that it doesn't work, but I'm not sure if this is because `next` has not been merged yet.

apps/common-app/src/new_api/components/drawer/index.tsx
m-bert, line 1: Probably not for this PR, as it only changed colors, but it breaks when type is changed to `back`:

https://github.com/user-attachments/assets/37bccbae-99e2-47e4-a68a-3a5ce7de6fd3

apps/common-app/src/new_api/hover_mouse/hoverable_icons/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+// eslint-disable-next-line import-x/no-commonjs, @typescript-eslint/no-var-requires
+const FREEZE = require('../../../common_assets/hoverable_icons/freeze.png');
+// eslint-disable-next-line import-x/no-commonjs, @typescript-eslint/no-var-requires
+const REA = require('../../../common_assets/hoverable_icons/rea.png');
+// eslint-disable-next-line import-x/no-commonjs, @typescript-eslint/no-var-requires
+const GH = require('../../../common_assets/hoverable_icons/gh.png');
+// eslint-disable-next-line import-x/no-commonjs, @typescript-eslint/no-var-requires
+const SCREENS = require('../../../common_assets/hoverable_icons/screens.png');
+
+const images = [GH, REA, SCREENS, SVG, FREEZE];
```
m-bert: I think we should change how those are displayed 🫨 

Image

apps/common-app/src/new_api/showcase/nested_text/nested_text.tsx
akwasniewski: I'm not sure why it was there, copied it from old example. Works without it 478550e8b63cb4bbfcd7456f491e8b0a38f5d3b2

apps/common-app/src/new_api/showcase/bottom_sheet/index.tsx
akwasniewski: I think the function is redundant, we can just call it directly 441a4c717e8931be88782dcc330299ba045018c8

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
akwasniewski: 1828c73ae5bccabedf047ff4b67a046cc6be16f2

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
akwasniewski: 1828c73ae5bccabedf047ff4b67a046cc6be16f2

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
akwasniewski: 1828c73ae5bccabedf047ff4b67a046cc6be16f2

apps/common-app/src/new_api/complicated/lock/index.tsx
akwasniewski: As far as I remember, the idea was that longPress was supposed to cancel the locking and be handled differently. But in hindsight it is too confusing, simplified in 9831b5d38271e3eb12f46d8944b4aae19970abfc

apps/common-app/src/new_api/complicated/camera/index.tsx
akwasniewski: 2083fab43da4a5433bdffb61a2248141bad72a58

apps/common-app/src/new_api/complicated/camera/index.tsx
akwasniewski: not really, 46a053338f56672cf8b69585b8b716d1d5d0631d

apps/common-app/src/new_api/complicated/camera/index.tsx
akwasniewski: thanks 😅, 1828c73ae5bccabedf047ff4b67a046cc6be16f2

apps/common-app/src/new_api/hover_mouse/hover/index.tsx
akwasniewski: Yeah, i was trying to group the examples more logically then before. I think they should grouped together, but could not find a good section name. For now renamed it to 'Hover and mouse' 1679ca6f7125dae9c9547f148303805d19cf56f2. Let me now if you have a better name. Do you have better ideas, for grouping in general? cc @m-bert

apps/common-app/src/new_api/simple/longPress/index.tsx
akwasniewski: ok, sure, if you prefer it d80d8fef1c4871353826504815fc12066f877f80

apps/common-app/src/new_api/showcase/nested_text/nested_text.tsx
m-bert: > I'm not sure why it was there

Probably linter trying to be smart 🤓

apps/common-app/src/new_api/showcase/shared_value/index.tsx
akwasniewski: Fixed in https://github.com/software-mansion/react-native-gesture-handler/pull/3906

apps/common-app/src/new_api/hover_mouse/hoverable_icons/index.tsx
akwasniewski: Fixed in https://github.com/software-mansion/react-native-gesture-handler/pull/3857/commits/0232a841f150a9105a4a46f669ee185cb4a53d8e

apps/common-app/src/new_api/showcase/shared_value/index.tsx
akwasniewski: I think android behaviour is correct, it fails on iOS because minDistance is broken on iOS. I'm resolving it in https://github.com/software-mansion/react-native-gesture-handler/pull/3906

apps/common-app/src/new_api/complicated/camera/index.tsx
akwasniewski: Split somewhat in d700fbdee8195eda0b656ad3ea2a3e5013b3cf20. I don't think further splitting makes sense as the gestures are preety intertwined. Let me know what you think.

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
akwasniewski: I have no idea where to put it, but I don't want to cut it out. If you have idea for a new category where it would fit let me know,

apps/common-app/src/new_api/showcase/nested_text/nested_text.tsx
akwasniewski: The legacy detector one only works on web, I left it there to highlight that it did not work in v2. I can remove the bottom part if you wish.

apps/common-app/src/new_api/simple/tap/index.tsx
akwasniewski: I added it to `LongPress` in 5d2af0af07465417761f1f42f0caaa7bfeade12d. Tap rarely fails, do you think it is necessary to have it here?

apps/common-app/App.tsx
akwasniewski: 11543b5afa9bbc3bf6d999e83cd7c486fbc698ba

apps/common-app/App.tsx
akwasniewski: Yea, I've put everything underneath one component, as now we have `showLegacyVersion` useState.

apps/common-app/src/new_api/complicated/chat_heads/index.tsx
akwasniewski: 32337bc5f289101a55f6d30961b3f03dd8a2adfa

apps/common-app/src/new_api/hover_mouse/mouse_buttons/index.tsx
akwasniewski: I've overhauled this example in 0ef53e1b00efdcf9601a1e40b20bdce2f83ce8e4. Let me know what you think.

apps/common-app/src/new_api/simple/tap/index.tsx
akwasniewski: extracted common styles in c506a3ffcd466a6e5fcdbb1eba73b3544a8e7558

apps/common-app/src/new_api/showcase/animated/index.tsx
akwasniewski: I may have overcomplicated the colours which caused flickering. Simplified in 25d5517547fd1e6fa0f6218a45f06638f1281641. This seems to have fixed android.

apps/common-app/src/new_api/hover_mouse/mouse_buttons/index.tsx
m-bert: I'd keep data in array and then use `map` instead of specifying each `Tests` component. What do you think?

apps/common-app/src/new_api/simple/tap/index.tsx
m-bert: > Tap rarely fails, do you think it is necessary to have it here?

It would be nice, but it is not necessary. Also I can imagine why it makes less sense for `Tap`, so it is fine if we don't have that.

apps/common-app/src/new_api/showcase/nested_text/nested_text.tsx
m-bert: > I can remove the bottom part if you wish.

No need, now I see the info text about Legacy Detector. Though it was quite confusing at first.

apps/common-app/src/new_api/hover_mouse/hoverable_icons/index.tsx
m-bert: I'd put our logo in the middle. Other than that looks good!

apps/common-app/src/new_api/hover_mouse/mouse_buttons/index.tsx
m-bert: Also, I think the outer `View` lacks `{flex: 1;}`, as now I'm seeing only empty screen without it 😅

apps/common-app/src/new_api/showcase/state_manager/index.tsx
m-bert, line 1: I don't like the fact that everything is hardcoded to `4`, but let's say in such simple example it is fine. The problem here is that I can't get it to work on `android`. Have you also experienced that?

https://github.com/user-attachments/assets/31f8d978-ece9-449b-86d4-8f3e250a97a6

It also seems to work randomly on web:

https://github.com/user-attachments/assets/4aab990b-c589-4559-bbc8-2fc9f62f64fd

akwasniewski: It uses longPress, thus on android it won't work due to issues with shadow node. Web seems to be broken, I'll look into that. This is definetly an issue with `StateManager` not the example, thus I will fix it in a seperate PR.

m-bert: Oh right, I thought I had this "fixed" locally. So I guess we can assume that example is fine

apps/common-app/src/new_api/complicated/velocity_test/index.tsx
m-bert: For now we can leave it. We will have to reorganize this app structure anyways, so for now it can stay here.

apps/common-app/src/new_api/hover_mouse/mouse_buttons/index.tsx
akwasniewski: > I'd keep data in array and then use map instead of specifying each Tests component. What do you think?

Sure, good idea. c808fa0af9fd9d345723c32af8ecbdd554ab36d4

akwasniewski: > Also, I think the outer View lacks {flex: 1;}, as now I'm seeing only empty screen without it 😅

Yeah, it was broken on mobile. Fixed https://github.com/software-mansion/react-native-gesture-handler/pull/3857/changes/fad4aaca9372addddebbfaab3f4d83732619b43c

apps/common-app/src/new_api/showcase/svg/index.tsx
akwasniewski: The problem is with the detector shadow node, tap is broken on android

apps/common-app/src/new_api/hover_mouse/hover/index.tsx
j-piasecki: That sounds good to me

apps/common-app/src/new_api/hover_mouse/context_menu/index.tsx
akwasniewski: As far as I recall it used to work, context menu seems to have broken in the meantime. I'll look into it, but it's probably outside of the scope of this PR.

akwasniewski: Fixed in https://github.com/software-mansion/react-native-gesture-handler/pull/3939

apps/common-app/src/new_api/hover_mouse/hover/index.tsx
m-bert: The one idea that I have for now (and which I had for a while) is to make a separate group for tests that we check before releasing new version. This of course doesn't mean that we have two move examples there, those entries can be duplicated.

Nevertheless it can be done in separate PR so current state looks ok (although I'm not a fan of `complicated` group, but I don't have better name for now and we can still change that later, as this PR is pretty big now).

apps/common-app/src/new_api/showcase/svg/index.tsx
m-bert: Was it? I think fix for that was already merged and it seems it does not work anyway. Or is it something in my setup and it works on your end.

Also, while we're talking about this example, I'm seeing [the same warning](https://github.com/software-mansion/react-native-gesture-handler/pull/3857#pullrequestreview-3582056461) as @j-piasecki mentioned

j-piasecki: Was https://github.com/software-mansion/react-native-gesture-handler/pull/3930 already merged then?

akwasniewski: Ok, interesting. I was certain that It is the shadow node as the intercepting detector itself worked. It should work after c8fc6d402a4114469bba0aa0e113945f7d77f6c9. However, this example did work without any changes before (it is the same as in the PR introducing VirtualDetector).

m-bert: Was the problem with missing `onPress`, or the order of the components?

akwasniewski: The order of the components, the missing `onPress` was not necessary after all thus it was removed in the next commit 0bd5b3621552c0d65de6235c3646edc5cc06fd80

m-bert: So we cannot simply attach `InterceptingDetector` to `SVG`? It is important because now the migration guide states that it is possible.

j-piasecki: If so, we should investigate that, as well as the warning.

akwasniewski: Properly fixed in 4fd21f9de635c3b1d70d8492fa33486599a6c899, thanks @j-piasecki for finding the solution.

akwasniewski: > While we're talking about this example, I'm seeing https://github.com/software-mansion/react-native-gesture-handler/pull/3857#pullrequestreview-3582056461 as @j-piasecki mentioned

I don't see the error on any platform, the issue behind it should be fixed by now

PR #3858, opened 2025-12-02
Change exports

m-bert:
## Description

This PR changes `export *` in `index` files so that it doesn't export everything from modules.

## Test plan

- `yarn ts-check`
- `yarn lint-js`

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
+++ b/packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
@@ -20,7 +20,7 @@ import createNativeWrapper from '../createNativeWrapper';
 
 import { NativeWrapperProperties } from '../types/NativeWrapperType';
 import { NativeWrapperProps } from '../hooks/utils';
-import { DetectorType } from '../detectors';
+import { GestureDetectorType } from '../detectors';
 import { NativeGesture } from '../hooks/gestures/native/useNativeGesture';
 import { ghQueueMicrotask } from '../../ghQueueMicrotask';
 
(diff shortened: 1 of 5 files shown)


packages/react-native-gesture-handler/src/index.ts
```diff
@@ -149,18 +149,8 @@ export type {
...
-  VirtualGestureDetector,
-} from './v3/detectors';
-
-export * from './v3/hooks/composition';
-
-export type { ComposedGesture } from './v3/types';
 export type { GestureTouchEvent as SingleGestureTouchEvent } from './handlers/gestureHandlerCommon';
 
-export * from './v3/hooks/gestures';
+export * from './v3';
```
j-piasecki, line 157: Can you double-check that we're not exporting internal helpers (like `useGesture`, `useComposedGesture`, ...).

m-bert: I did, and looks like everything is fine.

However, if you prefer I can remove `export *` from `v3` directory and export everything explicitly from there.

j-piasecki: I think it may make more sense to keep that and list everything explicitly in https://github.com/software-mansion/react-native-gesture-handler/blob/0291abb7b2ab6154269904d34fa78a7dbeb5b99c/packages/react-native-gesture-handler/src/v3/index.ts

m-bert: Yes yes, that's what I meant by **from there** 😅

m-bert: Also, should we keep `DetectorType` in index? Or move it somewhere else? Maybe to `detectors/common.ts`? WDYT @j-piasecki?

m-bert: Added explicit exports in [88b9ede](https://github.com/software-mansion/react-native-gesture-handler/pull/3858/commits/88b9edec1adc8258714638fa79c8c17605fb6ae0).

PR #3860, opened 2025-12-02
Change `runCallback` types

m-bert:
## Description

This PR removes `...unknown[]` from `runCallback` function, as in V3 it is always `boolean`.

## Test plan

`yarn ts-check`

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/FlingProperties.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/gestures/fling/FlingProperties.ts
@@ -1,5 +1,3 @@
-import { Directions } from '../../../../Directions';
-
 export type FlingGestureNativeProperties = {
   /**
    * Expressed allowed direction of movement. It's possible to pass one or many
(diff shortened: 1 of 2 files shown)


j-piasecki:
Won't this break if we decide to bring the state manager back in its original form?

m-bert:
> Won't this break if we decide to bring the state manager back in its original form?

It will, we will decide what to do with this PR on the next meeting. I've left that opened for the time being.

j-piasecki:
I think we can go with that as-is, since state manager will change, no?

packages/react-native-gesture-handler/src/v3/hooks/utils/eventHandlersUtils.ts
```diff
@@ -84,16 +94,25 @@ export function touchEventTypeToCallbackType(
...
 
-  // TODO: add proper types (likely boolean)
-  // @ts-ignore It works, duh
-  handler?.(event, ...args);
+  if (!handler) {
+    return;
+  }
+
+  if (didSucceed === undefined) {
+    (handler as SingleParameterCallback<typeof event>)(event);
```
m-bert, line 114: I know that theoretically we have two possibilities - `GestureEventCallback`and `GestureTouchEventCallback`. However, since those calls will differ only by cast, the resulting JS will look like this:

```js
if(...) {
  handler(event)
} else {
  handler(event)
}
```

So I don't think it makes sense to keep them separated.

PR #3867, opened 2025-12-08
[Android] Fix `ScrollView`

m-bert:
## Description

I've noticed that `ScrollView` did not work on Android. There were two issues with that:

1. `GHScrollView` received wrong prop. For some reason linter did not catch that.
2. `RefreshControl` was blocking scroll at the top of the list.

This PR changes prop name to the correct one and adds check for scrolling direction so that `RefreshControl` no longer blocks scrolling.

## Test plan

**ScrollView** and **FlatList** examples.

### DIFF
--- a/.github/workflows/rngh-api-v3.yml
+++ b/.github/workflows/rngh-api-v3.yml
@@ -1,10 +1,11 @@
-name: Test relations traversal algorithm
+name: Test Gesture Handler 3 API
 
 on:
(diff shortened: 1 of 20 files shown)


m-bert:
Converted to draft as there's another problem on Android - scroll starts working when pointer is released.

j-piasecki (approved):

PR #3880, opened 2025-12-16
State Manager

akwasniewski:
## Description

Implements V3 State Manager, to manually manage gestures.

## Test plan

Tested on the following example, the long press should activate instantely, while the minDuration is extremely big.
<details>

```ts
import React from 'react';
import { View, Text, StyleSheet } from 'react-native';
import { GestureHandlerRootView, GestureDetector, useLongPressGesture, GestureStateManager } from 'react-native-gesture-handler';

    onTouchesDown: (e) => {
      'worklet';
      console.log("touches down")
      GestureStateManager.begin(e.handlerTag)
      GestureStateManager.activate(e.handlerTag)
    },
    onActivate: () => {
      'worklet';
      console.log("long pressed")

    },
    minDuration: 100000000,
    disableReanimated: true
  })
  return (
    <GestureHandlerRootView>
      <View style={styles.root}>
          <View style={styles.outer}>
            <Text style={styles.label}>Outer</Text>
          </View>
        </GestureDetector>
      </View>
    </GestureHandlerRootView>
  );
}

  root: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#f7f7f7',
  },
  outer: {
    padding: 20,
    backgroundColor: '#ddd',
    borderRadius: 12,
  },
  label: {
    fontSize: 18,
    marginBottom: 10,
  },
});
```
</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/gestureStateManager.ts
+++ b/packages/react-native-gesture-handler/src/v3/gestureStateManager.ts
@@ -1,6 +1,13 @@
 import { State } from '../State';
 import { tagMessage } from '../utils';
 
+export type GestureStateManagerType = {
+  begin(handlerTag: number): void;
+  activate(handlerTag: number): void;
+  fail(handlerTag: number): void;
+  end(handlerTag: number): void;
+};
+
 const setGestureState = (handlerTag: number, state: State) => {
   'worklet';
 
@@ -13,7 +20,7 @@ const setGestureState = (handlerTag: number, state: State) => {
   }
 };
 
-export const GestureStateManager = {
+export const GestureStateManager: GestureStateManagerType = {
   begin(handlerTag: number) {
     'worklet';
     setGestureState(handlerTag, State.BEGAN);
--- a/packages/react-native-gesture-handler/src/v3/gestureStateManager.web.ts
+++ b/packages/react-native-gesture-handler/src/v3/gestureStateManager.web.ts
@@ -1,7 +1,8 @@
 import { State } from '../State';
 import NodeManager from '../web/tools/NodeManager';
+import { GestureStateManagerType } from './gestureStateManager';
 
-export const GestureStateManager = {
+export const GestureStateManager: GestureStateManagerType = {
   begin(handlerTag: number): void {
     'worklet';
     NodeManager.getHandler(handlerTag).begin();


packages/react-native-gesture-handler/src/v3/gestureStateManager.ts
```diff
@@ -0,0 +1,38 @@
+import { State } from '../State';
+import { tagMessage } from '../utils';
+
+declare const globalThis: {
+  _setGestureStateSync?: (handlerTag: number, state: State) => void;
+  _setGestureStateAsync?: (handlerTag: number, state: State) => void;
+};
```
m-bert: Shouldn't this be added as [global.d.ts](https://github.com/software-mansion/react-native-gesture-handler/blob/%40mbert/imperative-state-change/packages/react-native-gesture-handler/src/global.d.ts)?

packages/react-native-gesture-handler/src/v3/gestureStateManager.web.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const activate = (handlerTag: number): void => {
+  'worklet';
+  const handler = NodeManager.getHandler(handlerTag);
+
+  // Force going from UNDETERMINED to ACTIVE through BEGAN to preserve
+  // the correct state transition flow.
+  if (handler.state === State.UNDETERMINED) {
+    handler.begin();
+  }
```
m-bert: Why don't we do it on native platforms?

packages/react-native-gesture-handler/src/v3/gestureStateManager.ts
j-piasecki: Wouldn't that leak those definitions outside of the library?

packages/react-native-gesture-handler/src/v3/gestureStateManager.web.ts
j-piasecki: https://github.com/software-mansion/react-native-gesture-handler/blob/480a4944481216de9894f14f4be56fe020a50f67/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt#L146-L150

packages/react-native-gesture-handler/src/v3/gestureStateManager.ts
m-bert: Actually you might be right. As far as I remember we did it with @tjzel to overcome some TS problems, but this may be a better solution.

j-piasecki: Can we maybe get rid of the file you linked with this approach?

m-bert: It doesn't exist on `next` 😅

tjzel: We made the `global.d.ts` file with @m-bert so it's read only when developing the package and it's not exposed to the library consumer. If you leave this `declare cost` here then it will leak to the user.

You should move these types to `global.d.ts`.

j-piasecki:
Hmm, `StateManager.forTag(tag).begin()` sounds nice, but we'd need to create a new state manager for each `forTag(tag)` invocation. Though it's possible that the impact of that is negligible.

Edit: Maybe let's stick to the current approach, and we will see what feedback we get on that. WDYT?

akwasniewski:
> Hmm, `StateManager.forTag(tag).begin()` sounds nice, but we'd need to create a new state manager for each `forTag(tag)` invocation. Though it's possible that the impact of that is negligible.
> 
> Edit: Maybe let's stick to the current approach, and we will see what feedback we get on that. WDYT?

To be honest I prefer the simpler approach, but if you both want it to be `forTag(tag).begin()` then by the virtues of democracy I will change it. Let's see what feedback we get, and then decide.

packages/react-native-gesture-handler/src/v3/gestureStateManager.ts
akwasniewski: Thanks, done in 30c8c82de965bcfa0870fd2930bfee535d89d8e4

packages/react-native-gesture-handler/src/v3/gestureStateManager.ts
```diff
@@ -13,22 +13,24 @@ const setGestureState = (handlerTag: number, state: State) => {
   }
 };
 
-export const begin = (handlerTag: number): void => {
-  'worklet';
-  setGestureState(handlerTag, State.BEGAN);
-};
+export const GestureStateManager = {
+  begin(handlerTag: number): void {
```
m-bert:
```suggestion
  begin(handlerTag: number) {
```

I don't think we need those type declarations here.

akwasniewski: https://github.com/software-mansion/react-native-gesture-handler/pull/3880/changes/a020b99b743e04155098bc82a331343322d775d3

akwasniewski:
> I know it might be an overkill, but since we have two different implementations, maybe we could introduce an interface to make sure that they both meet the same criteria? This probably won't change, but it won't hurt us to have one. Let me know what you think.

~~They are different by design, one is gesture specific, and the other global, accepting `handlerTag`~~

EDIT: as we discussed in the office, I misunderstood. Good idea, added interface in 7fa4cfaa951dcb99f51c97145e3d99efcf6a76fe

m-bert (approved):
Merry Christmas 🎄🎅

PR #3885, opened 2025-12-18
[General] Cleanup duplicated callbacks

j-piasecki:
## Description

After the optimization PRs, the event handlers for js/ui thread were collapsed to a single function, but throughout the code, those are still being passed like before those changes - as three separate functions. Those are the same exact functions but under different names. This PR changes that, so that only one function is passed. Ultimately, this function still ends up in three different events, but the split now happens at the leafs instead of the root.

## Test plan

Gestures should still work

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
@@ -217,7 +217,7 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
         };
 
   const jsEventHandler = useMemo(
-    () => createGestureEventHandler('defaultEventHandler'),
+    () => createGestureEventHandler('jsEventHandler'),
     [createGestureEventHandler]
   );
 
--- a/packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
+++ b/packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
@@ -445,12 +445,12 @@ function getHandlerData(
         };
 
         if (eventName === 'onGestureHandlerStateChange') {
-          componentOrGesture.detectorCallbacks.defaultEventHandler?.({
+          componentOrGesture.detectorCallbacks.jsEventHandler?.({
             oldState: oldState as State,
             ...event,
           });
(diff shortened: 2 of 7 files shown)


packages/react-native-gesture-handler/src/v3/detectors/VirtualDetector/InterceptingGestureDetector.tsx
```diff
@@ -213,40 +197,59 @@ export function InterceptingGestureDetector<THandlerData, TConfig>({
...
+          shouldUseReanimatedDetector
+            ? reanimatedHandlers.onGestureHandlerReanimatedStateChange
+            : undefined
         }
         // @ts-ignore This is a type mismatch between RNGH types and RN Codegen types
         onGestureHandlerReanimatedEvent={
-          shouldUseReanimatedDetector ? reanimatedEventHandler : undefined
+          shouldUseReanimatedDetector
+            ? reanimatedHandlers.onGestureHandlerReanimatedEvent
+            : undefined
```
m-bert, line 248: So this is the only callback used by the native side, right?

j-piasecki: Yes, it should handle all types of events.

m-bert:
Could you check this example?

<details>
<summary>Example code</summary>

```tsx
import {
  Gesture,
  GestureDetector,
  GestureHandlerRootView,
  useTapGesture,

const uiManager = global?.nativeFabricUIManager ? 'Fabric' : 'Paper';
console.log(`Using ${uiManager}`);

  const g = useTapGesture({
    onDeactivate: () => {
      'worklet';
      console.log('Tap!');
    },
  });

  return (
      <GestureDetector gesture={g}>
        <View style={styles.ball} />
      </GestureDetector>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
  },
  ball: {
    width: 300,
    height: 300,
    borderRadius: 150,
    backgroundColor: 'crimson',
  },
});
```

</details>

It seems to work on both `android` and `ios`, but not on `web`. I've also checked on `next` branch and it seems to work there.

j-piasecki:
Seems to be working (after rebasing on next):

https://github.com/user-attachments/assets/9b7437be-0c30-425c-87b1-7e56832b65a9

PR #3887, opened 2025-12-18
[General] Move the end flag to the event passed to `onDeactivate` and `onFinalize`

j-piasecki:
## Description

Up to this point, the API mirrored V2, where `success` was being passed as an additional parameter to the end callbacks.

This PR changes it, so the flag is being passed as part of the event to the callbacks. I also think that the approach to working with gestures is to assume that they will succeed, so I changed the flag to `canceled`.

This also seems to be the easiest one to explain in the docs - "the gesture was canceled due to failure to meet activation criteria, the activation of another gesture, by the OS".

## Test plan

Verify that `calceled` exists in events passed to `onDeactivate` and `onFinalize`.


m-bert (commented):
Been a while, sorry for that!

I like the idea, though I think we stray from the "all callbacks use the same event". But on the other hand, we already did that so it should be fine. 

Anyway, if you'd still like this change, can you merge `main`? 😅

packages/react-native-gesture-handler/jest.config.js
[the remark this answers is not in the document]
j-piasecki: no idea where this came from

packages/docs-gesture-handler/docs/gestures/_examples/RotationExample.tsx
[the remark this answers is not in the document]
j-piasecki: no idea where this came from

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
```diff
@@ -55,18 +55,18 @@ Called each time a pointer tracked by the gesture changes state, typically due t
 ### onDeactivate
 
 ```ts
-onDeactivate: (event: GestureEvent<HandlerData>, didSucceed: boolean) => void
+onDeactivate: (event: GestureEndEvent<HandlerData>) => void
 ```
 
-Called when handler stops recognizing gestures, but only if the handler activated. It is called before `onFinalize`. If the handler was interrupted, the `didSucceed` argument is set to `false`. Otherwise it is set to `true`.
+Called when handler stops recognizing gestures, but only if the handler activated. It is called before `onFinalize`. The event object contains a `canceled` property — if the handler was interrupted or failed to recognize the gesture, `canceled` is set to `true`. Otherwise it is set to `false`.
```
m-bert: > or failed to recognize

If `onDeactivate` was called, it means that `onActivate` was called earlier. Should we say that handler failed to recognize gesture?

packages/docs-gesture-handler/docs/guides/upgrading-to-3.mdx
```diff
@@ -92,6 +92,32 @@ code2={
 `}
 />
 
+### canceled instead of didSucceed
```
m-bert:
```suggestion
### canceled instead of success
```

packages/react-native-gesture-handler/src/v3/hooks/callbacks/stateChangeHandler.ts
m-bert, line 1: Should this be removed in this PR?

packages/react-native-gesture-handler/src/v3/hooks/callbacks/touchEventHandler.ts
m-bert, line 1: Same as above

packages/react-native-gesture-handler/src/v3/hooks/callbacks/updateHandler.ts
m-bert, line 1: Same as above

packages/react-native-gesture-handler/src/v3/hooks/callbacks/stateChangeHandler.ts
j-piasecki: Not necessarily in this one, but this file is not referenced anywhere. I'm not sure how it got to this state. Would you prefer to move this to a separate one?

packages/react-native-gesture-handler/src/v3/hooks/callbacks/touchEventHandler.ts
j-piasecki: Same as above

packages/react-native-gesture-handler/src/v3/hooks/callbacks/updateHandler.ts
j-piasecki: Same as above

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
j-piasecki: [eb1570c](https://github.com/software-mansion/react-native-gesture-handler/pull/3887/commits/eb1570ceee38ad5b0a66255f4de4c77ca96940d1)

m-bert (approved):
https://github.com/software-mansion/react-native-gesture-handler/pull/3887#discussion_r3086934709

> Would you prefer to move this to a separate one?

I think it would be better. However, since it is not referenced anywhere it should not break anything, so If you don't want to make another PR and keep it here to ship it faster it's fine.

PR #3905, opened 2026-01-08
[docs] State management

m-bert:
## Description

This PR handles state management in our docs:

- Moves information about states to "under the hood" section
- Rewrites entire **states & events** page to focus more on callbacks
- Updates `GestureStateManager` entry and moves it to **fundamentals**
- Merges manual gesture guide with manual gesture docs

## Test plan

Read docs 🤓


packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
```diff
@@ -0,0 +1,217 @@
...
+ 
+Called when activation criteria for handler are met. If `onActivate` was called, it is guaranteed that `onDeactivate` will be called later.
+
+### onUpdate
+
+```ts
+onUpdate: (event: GestureEvent<HandlerData>) => void
+```
+
+Called every time a gesture is updated after it has started.
```
j-piasecki: What does it mean that "gesture is updated"?

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
```diff
@@ -0,0 +1,79 @@
...
+  onActivate: () => {
+    console.log('LongPress activated!');
+  },
+});
+
+const pan = usePanGesture({
+  onBegin: () => {
+    GestureStateManager.activate(longPress.handlerTag);
+  },
+});
```
j-piasecki: This example doesn't make sense, why not in reverse? (long press activating pan)

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
```diff
@@ -0,0 +1,79 @@
...
+Triggers [`onActivate`](/docs/fundamentals/callbacks-events#onactivate) callback on gesture with specified `handlerTag`.
+
+### deactivate
+
+```tsx
+deactivate: (handlerTag: number) => void;
+```
+
+Triggers first [`onDeactivate`](/docs/fundamentals/callbacks-events#ondeactivate), then [`onFinalize`](/docs/fundamentals/callbacks-events#onfinalize) callbacks on gesture with specified `handlerTag`.
+
```
j-piasecki: What if I call it before the gesture activates?

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
```diff
@@ -0,0 +1,79 @@
...
+## Methods
+
+### begin
+
+```tsx
+begin: (handlerTag: number) => void;
+```
+
+Triggers [`onBegin`](/docs/fundamentals/callbacks-events#onbegin) callback on gesture with specified `handlerTag`.
```
j-piasecki: Do we actually need the explicit "begin" method here? Won't all gestures automatically begin when detecting the first touch?

packages/docs-gesture-handler/docs/gestures/use-long-press-gesture.mdx
```diff
@@ -31,7 +31,7 @@ import BaseGestureCallbacks from './\_shared/base-gesture-callbacks.mdx';
 import SharedValueInfo from './\_shared/shared-value-info.md';
 
 A discrete gesture that activates when the corresponding view is pressed for a sufficiently long time.
-This gesture's state will turn into [END](/docs/fundamentals/states-events#end) immediately after the finger is released.
```
j-piasecki, line 34: Do we want to get rid of all the state links? We could be linking to specific callbacks instead.

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
```diff
@@ -11,32 +11,193 @@ import BaseGestureCallbacks from './\_shared/base-gesture-callbacks.mdx';
...
+  },
+});
 ```
 
+</details>
+
+
+## Modifying existing gestures
+
+While manual gestures open great possibilities we are aware that reimplementing pinch or rotation from scratch just because you need to activate in specific circumstances or require position of the fingers, would be a waste of time as those gestures are already available. Therefore, you can use touch events with every gesture to extract more detailed information about the gesture than what the basic events alone provide. We also added a `manualActivation` modifier on all continuous gestures, which prevents the gesture it is applied to from activating automatically, giving you full control over its behavior.
```
j-piasecki, line 203: I'm not sure whether this is the right place for this information.

packages/docs-gesture-handler/docs/gestures/use-native-gesture.mdx
```diff
@@ -97,15 +97,15 @@ Do not use `Native` gesture with components exported by React Native Gesture Han
 shouldActivateOnStart: boolean | SharedValue<boolean>;
```
j-piasecki, line 97: Unrelated, but this name doesn't make sense with the current api.

packages/docs-gesture-handler/docs/gestures/use-pinch-gesture.mdx
```diff
@@ -33,7 +33,7 @@ import BaseContinuousGestureCallbacks from './\_shared/base-continuous-gesture-c
 import SharedValueInfo from './\_shared/shared-value-info.md';
 
 A continuous gesture that recognizes pinch gesture. It allows for tracking the distance between two fingers and use that information to scale or zoom your content.
-The gesture [activates](/docs/fundamentals/states-events#active) when fingers are placed on the screen and change their position.
+The gesture activates when fingers are placed on the screen and change their position.
```
j-piasecki:
```suggestion
The gesture activates when fingers are placed on the screen and move away from each other or pull closer together.
```

packages/docs-gesture-handler/docs/gestures/use-rotation-gesture.mdx
```diff
@@ -34,7 +34,7 @@ import SharedValueInfo from './\_shared/shared-value-info.md';
 
 A continuous gesture that can recognize a rotation gesture and track its movement.
 
-The gesture [activates](/docs/fundamentals/states-events#active) when fingers are placed on the screen and change position in a proper way.
+The gesture activates when fingers are placed on the screen and change position in a proper way.
```
j-piasecki:
```suggestion
The gesture activates when fingers are placed on the screen and rotate around a common point.
```

packages/docs-gesture-handler/docs/under-the-hood/state.md
```diff
@@ -4,88 +4,16 @@ title: Handler State
...
+| **`ACTIVE`**       | The gesture is recognized and activation criteria are met.                               | [`onActivate`](/docs/fundamentals/callbacks-events#onactivate) / [`onUpdate`](/docs/fundamentals/callbacks-events#onupdate) |
+| **`END`**          | The user successfully completed the gesture.                                             | [`onDeactivate`](/docs/fundamentals/callbacks-events#ondeactivate)                                                          |
+| **`FAILED`**       | The handler failed to recognize the gesture.                                             | [`onFinalize`](/docs/fundamentals/callbacks-events#onfinalize)                                                              |
+| **`CANCELLED`**    | The system interrupted the gesture.                                                      | [`onFinalize`](/docs/fundamentals/callbacks-events#onfinalize)                                                              |
```
j-piasecki: Failed and Cancelled will also trigger `onDeactivate` if the gesture was active.

packages/docs-gesture-handler/docs/gestures/use-native-gesture.mdx
m-bert: `shouldActivateOnActivate` 😄

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
m-bert: Okay, I agree that it might not make much sense. However, I'm not sure how to rephrase it. Maybe something like

"Called every time a gesture receives a `GestureEvent`, provided it has previously activated."

What do you think?

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
m-bert: Is it always the case? What about `pinch` and `rotation`? At first glance (at least on web), they won't begin unless interacting with at least 2 pointers.

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
m-bert: ## tl;dr

I think we can remove it. If we really want to keep this, then I believe it should be rephrased - in that case we have to decide what we actually want to say here.

## Whole answer

Probably not. However, I have mixed feelings about this message. It looks like it gathers information from few sections and tries to merge them into one, but I'm not sure if it works.

1. It mentions `TouchEvent`, but only to say that "you don't have to use manual, you have more info on other gestures already"
2. It introduces concept of `continuous gestures`, which is not explained in the docs
3. It also describes `manualActivation`, which is already described when looking at specific handlers' configs.

I'm also not sure where do we want to put this if we were to move it. Overall it just sounds like "hey, we have manual handlers, but probably you could use the ones that are already configured" and that doesn't make much sense, especially after whole "tutorial" of how you use them.

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
m-bert: Changed in [268a9b1](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/268a9b19cddec654ccd4b8ad81297f76d29474e7)

packages/docs-gesture-handler/docs/under-the-hood/state.md
m-bert: Added in [f64e319](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/f64e319781b1ce3ab51b1bcf73f87bff6263ebee). Although I'm not sure if it is clear enough, maybe we should explicitly say that gesture has to activate in order to trigger them. Let me know.

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
m-bert: You're right, then `onDeactivate` won't be called. [f64e319](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/f64e319781b1ce3ab51b1bcf73f87bff6263ebee)

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
j-piasecki: > When the gesture is active, `onUpdate` gets called each time a pointer tracked by it changes its state - most likely moves.

Still feels a bit off though :/

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
j-piasecki: Is this consistent with other platforms? Maybe it's something that needs to be standardized.

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
j-piasecki: What about adding a hint at the top saying something along the lines of `If you need to change the activation criteria of a gesture, check its configuration for existing properties or use `manualActivation` to control it yourself.`?

m-bert: Sounds, good. Should we wrap it with `tip` admonition? Or leave it as a plain text? I think `tip` looks ok:

image

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
m-bert: > Is this consistent with other platforms?

No. At least from what I can see, this is how it works:

- `android` - begins immediately after you place **first** pointer.
- `web` - begins immediately after you place **second** pointer.
- `iOS` - begins when handler tracks 2 pointers and you move them a bit.

packages/docs-gesture-handler/docs/gestures/use-long-press-gesture.mdx
m-bert: As discussed, we can get rid of them for now. Later on, when we read docs we can put them at necessary places.

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
j-piasecki: It looks good, but the wording is rough. I didn't really write it with the intention of it ending up in the docs. Maybe something like:

> If you need to change the activation criteria for a gesture, check its configuration for properties that influence it, or use `manualActivation` to control its activation imperatively.

packages/docs-gesture-handler/docs/under-the-hood/state.md
j-piasecki: Why not put more context in the table? Like:

> `onActivate` when it first transitions into the `ACTIVE` state.
`onUpdate` when it has new data about the gesture.

> `onDeactivate` if the gesture was in `ACTIVE` state before.
`onFinalize`

packages/docs-gesture-handler/docs/fundamentals/state-manager.md
m-bert: I guess this is something that we would like to standardize, but not in the PR with docs. I've added it to flaws section in docs checklist so that we won't forget about it.

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
m-bert: Sure, I had some other versions too, like 

"If you need to modify the activation criteria of a gesture, check its configuration for relevant properties or use `manualActivation` to manage it directly."

Or 

"If you need to change how a gesture gets activated, check its properties for any existing options or use `manualActivation` to handle it on your own."

packages/docs-gesture-handler/docs/under-the-hood/state.md
m-bert: Added in [9ad871b](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/9ad871b547f7457f6392600e8f1ffa9f45f5d768). Though I'm not sure if it is clear that `FAIL` and `CANCEL` will result in `didSucceed` set to `false`.

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
m-bert: Let me know what you think: [c0afea2](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/c0afea2c09accea1adbabcf1ac3e37430c616068)

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
m-bert: "Called every time a gesture updates after it has been activated. This callback is triggered each time a pointer tracked by the gesture changes state, typically due to movement."

WDYT?

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
j-piasecki: "If you need to modify the activation criteria of gestures other than `Manual`, check their configuration for relevant properties or use `manualActivation` to manage their state directly."?

packages/docs-gesture-handler/docs/under-the-hood/state.md
j-piasecki: > Though I'm not sure if it is clear that FAIL and CANCEL will result in didSucceed set to false.

It's not even mentioned, so I'd say it's not 😄

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
j-piasecki: It essentially duplicates "on every update" across both sentences. What about:

> Called each time a pointer tracked by the gesture changes state, typically due to movement, after the gesture has been activated.

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+## Callbacks
+
+### onBegin
+
+```ts
+onBegin: (event: GestureEvent<HandlerData>) => void
+```
+
+Called when a handler begins to recognize gestures. If `onBegin` was called, it is guaranteed that `onFinalize` will be called later.
```
akwasniewski, line 38:
```suggestion
Called when a handler begins to recognize gestures.
```
This had been already said earlier.

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+### onActivate
+
+```ts
+onActivate: (event: GestureEvent<HandlerData>) => void
+```
+ 
+Called when activation criteria for handler are met. If `onActivate` was called, it is guaranteed that `onDeactivate` will be called later.
```
akwasniewski, line 46:
```suggestion
Called when activation criteria for handler are met,
```
same as above

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+### onUpdate
+
+```ts
+onUpdate: (event: GestureEvent<HandlerData>) => void
+```
+
+Called every time a gesture is updated after it has started.
```
akwasniewski:
```suggestion
Called every time a gesture is updated after it has activated.
```

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+When `Reanimated` is installed, all callbacks are automatically workletized. For more details, refer to the [Integration with Reanimated](/docs/fundamentals/reanimated-interactions#automatic-workletization-of-gesture-callbacks) section.
+
+## Callbacks flow
+
+### GestureEvent callbacks
+
+<GestureEventFlowChart />
+
+Note that some of these callbacks are complementary:
```
akwasniewski, line 21:
```suggestion
Note that some of these callbacks are paired:
```

packages/docs-gesture-handler/src/examples/CallbacksFlowCharts/TouchEventChart.jsx
akwasniewski, line 1: is it possible to make it so that the onUpdate loop does not intersect with the edge above it?

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
m-bert: Yes, but if someone was redirected directly to `onBegin` they probably won't scroll whole page to find that information. Redundancy in docs is ok in most cases (cc @j-piasecki)

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
m-bert: Same as above

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
m-bert: Why?

packages/docs-gesture-handler/src/examples/CallbacksFlowCharts/TouchEventChart.jsx
m-bert: I've spent some time on those charts and I have not figured out easy way how to do this so I wanted to focus more on writing content than spending more time on layout fixes. Right now it uses in-built layout engine which automatically sets position for all nodes. We could try to change it, but then we would probably have to manually set position to each node.

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
m-bert: I think we got it: [bd65c24](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/bd65c24ff32f5f701d2154750a046ec274b940c3)

packages/docs-gesture-handler/docs/under-the-hood/state.md
m-bert: Added in [efea146](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/efea146e336ed2d51473181fc57f1c61cbccbc82). Do you think we should also change how `onDeactivate` and `onFinalize` are described? 

Namely the part:

"If the handler was interrupted, the didSucceed argument is set to `false`. Otherwise it is set to `true`."

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
m-bert: We've changed `onUpdate` description in [2cd0bb5](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/2cd0bb5d085527b60a2067171ee8b08bfd439884)

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
m-bert: Ok, I think that works. [2cd0bb5](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/2cd0bb5d085527b60a2067171ee8b08bfd439884)

packages/docs-gesture-handler/docs/gestures/_manual_gesture_steps/step7.md
```diff
@@ -0,0 +1,10 @@
+```tsx
+const gesture = useManualGesture({
+  ...
+  onActivate: () => {
+    active.value = true;
+  },
+  onDeactivate: () => {
+    active.value = false;
+  },
```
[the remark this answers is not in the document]
m-bert: That's why we have `...`

packages/docs-gesture-handler/docs/gestures/use-manual-gesture.mdx
```diff
@@ -11,32 +11,197 @@ import BaseGestureCallbacks from './\_shared/base-gesture-callbacks.mdx';
...
+  return <Animated.View style={[styles.pointer, animatedStyle]} />;
+}
+
+export default function Example() {
+  const trackedPointers: SharedValue<Pointer>[] = [];
+  const active = useSharedValue(false);
+
+  for (let i = 0; i < 10; i++) {
+    trackedPointers[i] = useSharedValue<Pointer>({
+      x: 0,
```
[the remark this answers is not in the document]
m-bert: Well, 10 should be enough

packages/docs-gesture-handler/src/examples/CallbacksFlowCharts/FlowChart.jsx
```diff
@@ -0,0 +1,124 @@
...
+  SWM_NAVY_DARK_100: '#001a72',
+};
+
+const mobileThreshold = 996;
+
+export default function FlowChart({ nodes, edges }) {
+  const containerRef = useRef(null);
+  const [isMobile, setIsMobile] = useState(
+    typeof window !== 'undefined' ? window.innerWidth <= mobileThreshold : false
+  );
```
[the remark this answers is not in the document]
m-bert: [8e580e4](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/8e580e4c4a7b0ac6c502d68e2d7b9760f3c7ea8d)

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
```diff
@@ -0,0 +1,181 @@
+---
+id: callbacks-events
+title: Gesture callbacks & events
+sidebar_label: Gesture callbacks & events
+sidebar_position: 7
+---
+
+import {GestureEventFlowChart, TouchEventFlowChart} from '@site/src/examples/CallbacksFlowCharts'
+import CollapsibleCode from '@site/src/components/CollapsibleCode';
```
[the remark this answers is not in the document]
m-bert: [8e580e4](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/8e580e4c4a7b0ac6c502d68e2d7b9760f3c7ea8d)

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
```diff
@@ -0,0 +1,181 @@
...
+} as const;
+
+export const TouchEventType = {
+  UNDETERMINED: 0,
+  TOUCHES_DOWN: 1,
+  TOUCHES_MOVE: 2,
+  TOUCHES_UP: 3,
+  TOUCHES_CANCEL: 4,
+} as const;
+
```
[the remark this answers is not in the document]
m-bert: [8e580e4](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/8e580e4c4a7b0ac6c502d68e2d7b9760f3c7ea8d)

packages/docs-gesture-handler/docs/fundamentals/state-manager.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+### Outside gesture definition
+
+If you want to control gesture lifecycle outside of it, you can use `handlerTag` from created gesture object.
+
+<CollapsibleCode 
+label="Show full example"
+expandedLabel="Hide full example"
+lineBounds={[10, 21]}
+src={`
```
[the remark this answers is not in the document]
m-bert: [8e580e4](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/8e580e4c4a7b0ac6c502d68e2d7b9760f3c7ea8d)

packages/docs-gesture-handler/docs/fundamentals/state-manager.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+### Inside gesture definition
+
+If you want to manipulate gesture's state in its callbacks, you can get `handlerTag` from event parameter.
+
+<CollapsibleCode 
+label="Show full example"
+expandedLabel="Hide full example"
+lineBounds={[9, 17]}
+src={`
```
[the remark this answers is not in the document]
m-bert: [8e580e4](https://github.com/software-mansion/react-native-gesture-handler/pull/3905/commits/8e580e4c4a7b0ac6c502d68e2d7b9760f3c7ea8d)

packages/docs-gesture-handler/src/examples/CallbacksFlowCharts/TouchEventChart.jsx
akwasniewski: Ok, it's not important, I mentioned it hoping it is gonna be an easy fix, let's leave it for now.

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
akwasniewski: If you prefer 'complementary' you can leave it, but I'm not sure it precisely conveys what relation they have. Complementary usually means that they simply work well together, it doesn't imply they are bound together.

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
j-piasecki: A while ago, Kacper showed me a slide they've been using during workshops to explain it:

image

I'd love for it to be obvious from the docs. Currently, it kind of does by explicitly stating it on every callback. The callback flow is dense, and by looking at it, it's not a guaranteed conclusion.

Maybe it could be its own section where we put some graph akin to the one above?

packages/docs-gesture-handler/docs/fundamentals/callbacks-events.mdx
j-piasecki: I always thought about it in a way that the callbacks complement each other (as in a set's complement or an angle's complement).

akwasniewski: Ok, math reference changed my mind, I see now why it fits.

packages/docs-gesture-handler/docs/under-the-hood/state.md
j-piasecki: IMO it's fine

PR #3906, opened 2026-01-08
[ios] fix hit frame

akwasniewski:
## Description

The `minDistance` property did not work on ios. The reason was that the `GestureDetector` view is sometimes getting collapsed so that it bounds are {x: 0, y:0, width:0, height:0}, then the function `isWithinBounds` will always return false. The solution is to check whether the view is collapsed, and if so, get the bounds of its subview instead.

## Test plan

Tested on the following example, `minDistance` works!

<details>

```ts
import React from 'react';
import {
  GestureDetector,
  GestureHandlerRootView,
  useLongPressGesture,
  usePanGesture,
  useSimultaneousGestures,
import Animated, {
  useSharedValue,
  useAnimatedStyle,
  interpolateColor,
  withTiming,

export default function PanExample() {
  const translateY = useSharedValue(0);
  const offsetX = useSharedValue(0);
  const offsetY = useSharedValue(0);

  const panGesture = usePanGesture({
    onBegin: () => {
      'worklet';
      colorProgress.value = withTiming(1, { duration: 150 });
    },
    onUpdate: (event) => {
      'worklet';
      translateX.value = offsetX.value + event.translationX;
      translateY.value = offsetY.value + event.translationY;
    },
    onFinalize: () => {
      'worklet';
      offsetX.value = translateX.value;
      offsetY.value = translateY.value;
    },
  });

  const longPressGesture = useLongPressGesture({
    onBegin: () => {
      'worklet';
      colorProgress.value = withTiming(1, {
        duration: 100,
      });
    },
    onActivate: () => {
      'worklet';
      colorProgress.value = withTiming(2, {
        duration: 100,
      });
    },
    onFinalize: () => {
      'worklet';
      colorProgress.value = withTiming(0, {
        duration: 100,
      });
    },
    minDuration: 1000,
    maxDistance: 10000000000000,
  });

  const gestures = useSimultaneousGestures(longPressGesture, panGesture);
      colorProgress.value,
      [0, 1, 2],
      ["navy", "purple", "blue"]
    );
    return {
      transform: [
        { translateX: translateX.value },
        { translateY: translateY.value },
      ],
      backgroundColor,
    };
  });

  return (
    <GestureHandlerRootView>
      <View style={styles.centerView}>
        <View>
          <GestureDetector gesture={gestures}>
          </GestureDetector>
        </View>
      </View>
    </GestureHandlerRootView>
  );
}

  centerView: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
  box: {
    height: 120,
    width: 120,
    borderRadius: 60,
    marginBottom: 100,
  },
  instructions: {
    marginTop: 8,
    textAlign: 'center',
    paddingHorizontal: 16,
  },
});
```


akwasniewski:
I'm not a big fan of this solution, I think it may introduce other bugs, but after a lot of tinkering around this proved to be the only working solution.

j-piasecki:
Hmm, weird that the detector has the wrong dimensions while its children are sized correctly. Could you check [this part](https://github.com/software-mansion/react-native-gesture-handler/blob/59c3de78714121da34de92eaa5142fa6d3c61c3b/packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerDetectorShadowNode.cpp#L54) to see if the values there seem correct?

j-piasecki:
Can you change if the issue is still there with https://github.com/software-mansion/react-native-gesture-handler/pull/3930 applied?

akwasniewski:
> Can you change if the issue is still there with #3930 applied?

I'm surprised to report that fixing shadow node problem did not help. It is still broken without the fix in this PR.

akwasniewski:
Superseded by https://github.com/software-mansion/react-native-gesture-handler/pull/3942

PR #3913, opened 2026-01-09
Fix state manager on unregistered gestures

akwasniewski:
## Description

The new `StateManager` is global and given `handlerTag` it can manually set the states of an arbitrary gesture. This causes errors, when the gesture, which state is being set, has not been yet recorded in the orchestrator. Recording gestures in the orchestrator on android is done lazily, thus if it never received touches it is not recorded.

It also adds explicit error when trying to manually handled a gesture not attached to any detector on all platforms.

## Test plan

Tested on the following example
<details>

```ts
import React from 'react';
import { View, Text, StyleSheet } from 'react-native';
import { GestureHandlerRootView, GestureDetector, useLongPressGesture, GestureStateManager, usePanGesture, useSimultaneousGestures, useTapGesture } from 'react-native-gesture-handler';

    onTouchesDown: (e) => {
      'worklet';
      console.log("touches down")
    },
    onActivate: () => {
      'worklet';
      console.log("long pressed")
    },
    minDuration: 100000000,
    disableReanimated: true
  })
  const pan = useTapGesture({
    onTouchesDown: () => {
      'worklet';
      console.log("tap")

      GestureStateManager.activate(longPress.tag)
    },
    disableReanimated: true
  });
  return (
    <GestureHandlerRootView>
      <View style={styles.root}>
          <View style={styles.outer}>
            <Text style={styles.label}>Long Press</Text>
          </View>
        </GestureDetector>
          <View style={styles.outer}>
            <Text style={styles.label}>Pan</Text>
          </View>
        </GestureDetector>

      </View>
    </GestureHandlerRootView>
  );
}

  root: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#f7f7f7',
  },
  outer: {
    padding: 20,
    backgroundColor: '#ddd',
    borderRadius: 12,
    marginBottom: 50
  },
  label: {
    fontSize: 18,
    marginBottom: 10,
  },
})
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
@@ -33,6 +33,7 @@ export class GestureHandlerWebDelegate
   };
 
   private areContextMenuListenersAdded = false;
+  private wasContextMenuEnabled = false;
 
   init(viewRef: number, handler: IGestureHandler): void {
     if (!viewRef) {
@@ -158,9 +159,11 @@ export class GestureHandlerWebDelegate
     }
 
     if (this.shouldDisableContextMenu()) {
+      this.wasContextMenuEnabled = false;
       this.view.addEventListener('contextmenu', this.disableContextMenu);
       this.areContextMenuListenersAdded = true;
     } else if (this.gestureHandler.enableContextMenu) {
+      this.wasContextMenuEnabled = true;
       this.view.addEventListener('contextmenu', this.enableContextMenu);
       this.areContextMenuListenersAdded = true;
     }
@@ -173,10 +176,14 @@ export class GestureHandlerWebDelegate
 
     this.ensureView(this.view);
 
-    if (this.shouldDisableContextMenu()) {
+    if (!this.areContextMenuListenersAdded) {
+      return;
+    }
+
+    if (!this.wasContextMenuEnabled) {
       this.view.removeEventListener('contextmenu', this.disableContextMenu);
       this.areContextMenuListenersAdded = false;
-    } else if (this.gestureHandler.enableContextMenu) {
+    } else {
       this.view.removeEventListener('contextmenu', this.enableContextMenu);
       this.areContextMenuListenersAdded = false;
     }
@@ -220,11 +227,16 @@ export class GestureHandlerWebDelegate
   }
 
   private setContextMenu() {
-    if (this.gestureHandler.enabled) {
-      this.addContextMenuListeners();
-    } else {
+    if (!this.gestureHandler.enabled) {
+      this.removeContextMenuListeners();
+      return;
+    }
+
+    if (!this.wasContextMenuEnabled) {
       this.removeContextMenuListeners();
     }
+
+    this.addContextMenuListeners();
   }
 
   onEnabledChange(): void {


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -175,6 +175,26 @@ open class GestureHandler {
     interactionController = controller
   }
 
+  fun prepareForManual(orchestrator: GestureHandlerOrchestrator?) {
+    check(this.orchestrator == null) {
```
m-bert: Shouldn't it be `this.orchestrator != null`? Especially that the condition that starts all this process is `if (orchestrator == null)`.

Also, the only way to end up here is from `orchestrator` itself, right?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -175,6 +175,26 @@ open class GestureHandler {
     interactionController = controller
   }
 
+  fun prepareForManual(orchestrator: GestureHandlerOrchestrator?) {
```
m-bert: I'm not a fan of this name. First of all, if it has the same purpose we could overload `prepare` (maybe with little comment when it is useful). If we want to keep it different, then I guess it should be more descriptive, like `prepareForManualStateChange`, `prepareForManualManagement`, or some other form.

Also, it seems to share most of the logic with `prepare` - we could extract it into separate function if needed so that we don't have this redundancy.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -441,7 +441,17 @@ class GestureHandlerOrchestrator(
...
+    if (gestureHandlers.contains(handler)) {
+      return
+    }
 
+    gestureHandlers.add(handler)
+    handler.isActive = false
+    handler.isAwaiting = false
+    handler.activationIndex = Int.MAX_VALUE
+    handler.prepareForManual(this)
+  }
```
m-bert: Same here, most of the logic is repeated.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -593,6 +613,12 @@ open class GestureHandler {
       // generated faster than they can be treated by JS thread
       eventCoalescingKey = nextEventCoalescingKey++
     }
+
+    if (orchestrator == null) {
+      // If the state is set manually, the handler may not have been fully recorded by the orchestrator.
+      hostDetectorView?.recordHandlerIfNotPresentForManual(this)
+    }
```
m-bert, line 606: I think we should throw error when both, `orchestrator` and `hostDetectorView` are `null`. This may happen if gesture is defined, but is not assigned to any of the detectors. In that case I believe it will still crash, but with other, less descriptive error (probably `NullPointerException`).

m-bert (commented):
As said in the first comment, I'm not a big fan of `ForManual` suffix. I think we could either use overloading (where possible) or add more descriptive suffix.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
akwasniewski: The `check` [works](https://kotlinandroid.org/kotlin/stdlib/kotlin-check/) differently than `if`, this throws the provided message when the orchestrator was already set.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
akwasniewski: Cleaned up in 27b2aa0459bd71dcde780b171428cc466cba3bd2. `prepare` works fine if we pass `null` as view, so we can just do that.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
akwasniewski: As above

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
akwasniewski: Thanks, f3c1e939d4933a11a32231a4faa76d0afae8830d, I will check whether this works on other platforms

akwasniewski: On web the error is more fitting, `Expected delegate view to be HTMLElement` and on iOS it fails silently

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -202,7 +203,23 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
+
+    while (parent != null) {
+      if (parent is RNGestureHandlerRootView) {
+        gestureHandlerRootView = parent
+      }
+      parent = parent.parent
+    }
 
+    return gestureHandlerRootView
+  }
```
j-piasecki: Don't we have this exact method already defined somewhere? Maybe it's worth moving it to some utils, or making it an extension on View?

akwasniewski: Sure, I moved it to be companion object for the rootView in 8a4ffc99f203d12e12f358e82323aea1198740a6. Let me know what you think.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
m-bert: > On web the error is more fitting, Expected delegate view to be HTMLElement

I don't think it is more fitting in that case. It doesn't tell you what you've done wrong, so it won't be obvious for someone who tries to change state of the gesture. 

> on iOS it fails silently

I don't think it is good either. Such behaviour should be unified between platforms.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
j-piasecki: Hmm, I'm not sure whether `this.findGestureHandlerRootView()?...` wouldn't have been cleaner, but I'll leave the final decision to you. Both work for me.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
akwasniewski: I added explicit throws.
web: 744b4a681dbb5cbae010ff7767a924985559ac19, ios: 8775c6f641603ef6f73260c6c5a86a456473d0bc

Issue #3904, opened 2026-01-07
[Pressable] pointerEvents "box-only" and "box-none" broken in latest version


gigobyte:
### Description

Only web and iOS are affected, Android looks to be fine.

## **box-only**

```tsx
<Pressable
  style={{ width: 100, height: 80, backgroundColor: 'red' }}
  onPress={() => Alert.alert('outer')}
  pointerEvents="box-only"
>
  <Pressable
    style={{ width: 50, height: 50, backgroundColor: 'green' }}
    onPress={() => Alert.alert('inner')}
  />
</Pressable>
```

Expected result: clicking on the inner alerts `"outer"`, clicking on the outer alerts `"outer"`
Actual result: clicking on the inner alerts nothing, clicking on the outer alerts nothing

## **box-none**
```tsx
 <Pressable
  style={{ width: 100, height: 80, backgroundColor: 'red' }}
  onPress={() => Alert.alert('outer')}
  pointerEvents="box-none"
>
  <Pressable
    style={{ width: 50, height: 50, backgroundColor: 'green' }}
    onPress={() => Alert.alert('inner')}
  />
</Pressable>
```

Expected result: clicking on the inner alerts `"inner"`, clicking on the outer alerts nothing
Actual result: clicking on the inner alerts `"inner"`, clicking on the outer alerts`"outer"`

### Steps to reproduce

In the description

### A link to a [Gist](https://gist.github.com/), an [Expo Snack](https://snack.expo.io/) or a link to a repository based on [this template](https://github.com/react-native-community/reproducer-react-native) that reproduces the bug.

n/a

### Gesture Handler version

2.30.0

### React Native version

0.82.0

### Platforms

Web, iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

New Architecture (Fabric)

### Build type

None

### Device

None

### Acknowledgements

Yes

Titozzz:
duplicate of #3891 ?

m-bert:
Hi @gigobyte! Thanks for reporting this issue. Have you checked #3925? Also, have you tried this on `android`? I know it says that 

> Android and Web look to be fine.

but from my tests it looks like it is also broken.

gigobyte:
@m-bert 
> but from my tests it looks like it is also broken.

You are correct, it's also broken on Android but in a different way, which is why I assumed it was working when I was testing the sample code. For example, `box-only` works like `none` on iOS but like `auto/undefined` on Android.

[...]

Issue #3891, opened 2025-12-22
bug(pressable, iOS): pointerEvents box-none not working


Titozzz:
### Description

On iOS, setting pointerEvents="box-none" on Pressable seems to have no effect. (the onPress still occurs and the view captures the press from pressable behind)
It works with Pressable from React Native.
Setting "none" does work properly (the touch is handled by views behind)

### Steps to reproduce

Paste the gist in your App.tsx to reproduce.

The black square should not capture events when box-none is set on it.

Please note that this is a simplified version of our code, we use box-none because we need children of this pressable to do stuff (here there are none)

### A link to a [Gist](https://gist.github.com/), an [Expo Snack](https://snack.expo.io/) or a link to a repository based on [this template](https://github.com/react-native-community/reproducer-react-native) that reproduces the bug.

https://gist.github.com/Titozzz/031d92d3bc149ede284320dee796ace4

```tsx
 <GestureHandlerRootView>
      <View
        style={[
          StyleSheet.absoluteFill,
          {
            backgroundColor: 'blue',
            alignItems: 'center',
            justifyContent: 'center',
          },
        ]}
      >
        <Pressable
          onPress={() => {
            console.log('pressed red');
          }}
          style={{
            backgroundColor: 'red',
            position: 'absolute',
            top: 0,
            left: 0,
            bottom: 0,
            right: 0,
          }}
        />

        <Pressable
          onPress={() => {
            console.log('pressed black');
          }}
          style={{
            backgroundColor: 'rgba(0, 0, 0, 0.5)',
            position: 'absolute',
            width: 100,
            height: 100,
            right: 200,
          }}
          pointerEvents="box-none"
        />
      </View>
    </GestureHandlerRootView>
```

### Gesture Handler version

2.30.0

### React Native version

0.81.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Acknowledgements

Yes

m-bert:
Hi @Titozzz! Sorry for late response, we've been quite busy recently. I can see that there's a PR that attempts to solve it, have you checked that (#3925)? I tried it and it seems to solve this issue. 

Also, have you tried `android`? I've also looked at it and it seems to be broken as well 😞

PR #3925, opened 2026-01-18
fix(iOS): handles `pointerEvents` for `Pressable` component

huextrat:
## Description

fixes: https://github.com/software-mansion/react-native-gesture-handler/issues/3891
fixes: https://github.com/software-mansion/react-native-gesture-handler/issues/3904

This PR implements `pointerEvents` support for `Pressable` component from `react-native-gesture-handler` on iOS. Previously, setting `pointerEvents="box-none"` (or other modes) had no effect on iOS, while it worked correctly <s>on Android</s> and with React Native's `Pressable` on iOS.

Android PR: https://github.com/software-mansion/react-native-gesture-handler/pull/3927

### Implementation Details

The implementation follows React Native's `hitTest` logic for `pointerEvents`:
- For `box-only`: Returns `self` if point is inside (respecting `hitSlop`), `nil` otherwise
- For `box-none`: Checks subviews only, returns the hit subview or `nil`
- For `none`: Always returns `nil`
- For `auto`: Uses standard hit testing with `shouldHandleTouch` logic

The implementation respects `hitTestEdgeInsets` (hitSlop) for all modes, ensuring consistent behavior with React Native's `Pressable`.

## Test plan

Tested all `pointerEvents` modes on iOS:
- ✅ `pointerEvents="none"` - View and subviews don't receive touches
- ✅ `pointerEvents="box-none"` - View doesn't receive touches, subviews do
- ✅ `pointerEvents="box-only"` - View receives touches, subviews don't
- ✅ `pointerEvents="auto"` - Default behavior works as expected

I've used https://github.com/huextrat/repro-pressable-gh to test scenarios

Tested on both old architecture (Paper) and new architecture (Fabric).

edit: `pointerEvents` with RNGH Pressable is not working on Android

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerPointerEvents.h
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerPointerEvents.h
@@ -1,3 +1,5 @@
+#import <Foundation/Foundation.h>
+
 typedef NS_ENUM(NSInteger, RNGestureHandlerPointerEvents) {
   RNGestureHandlerPointerEventsNone,
   RNGestureHandlerPointerEventsBoxNone,


packages/react-native-gesture-handler/apple/RNGestureHandlerButtonComponentView.mm
```diff
@@ -207,8 +223,37 @@ - (void)updateProps:(const Props::Shared &)props oldProps:(const Props::Shared &
...
+{
+  if (![self pointInside:point withEvent:event]) {
+    return nil;
+  }
+
+  CGPoint buttonPoint = [self convertPoint:point toView:_buttonView];
+
+  UIView *hitView = [_buttonView hitTest:buttonPoint withEvent:event];
+
+  return hitView;
```
m-bert:
```suggestion
  return [_buttonView hitTest:buttonPoint withEvent:event];
```

packages/react-native-gesture-handler/apple/RNGestureHandlerButtonComponentView.mm
```diff
@@ -9,11 +9,27 @@
...
+  switch (pointerEvents) {
+    case facebook::react::PointerEventsMode::None:
+      return @"none";
+    case facebook::react::PointerEventsMode::BoxNone:
+      return @"box-none";
+    case facebook::react::PointerEventsMode::BoxOnly:
+      return @"box-only";
+    case facebook::react::PointerEventsMode::Auto:
+    default:
+      return @"auto";
```
m-bert: Could we make this an `enum` instead of using strings? 😅 e.g. `RNGestureHandlerPointerEvent`.

huextrat: Yep ✅  39f08f9b64bbe0e878c6a5e2886d5b49ab6ffcac

packages/react-native-gesture-handler/apple/RNGestureHandlerButton.h
```diff
@@ -22,12 +22,21 @@
...
 @interface RNGestureHandlerButton : UIControl
 #endif // TARGET_OS_OSX
+
+typedef NS_ENUM(NSInteger, RNGestureHandlerPointerEvents) {
+  RNGestureHandlerPointerEventsNone,
+  RNGestureHandlerPointerEventsBoxNone,
+  RNGestureHandlerPointerEventsBoxOnly,
+  RNGestureHandlerPointerEventsAuto
+};
+
```
m-bert: I'd prefere to have it in separate file, like other enums 😅

huextrat: 👍 b82c1860dca96fc0d9ed16b09a659bad481f5d34

PR #3927, opened 2026-01-20
fix(android): handles `pointerEvents` for `Pressable` component

huextrat:
## Description

fixes: https://github.com/software-mansion/react-native-gesture-handler/issues/3891
fixes: https://github.com/software-mansion/react-native-gesture-handler/issues/3904

This PR fixes `pointerEvents` support for `Pressable` component from `react-native-gesture-handler` on Android.

Waiting for iOS PR: https://github.com/software-mansion/react-native-gesture-handler/pull/3925 as codegen is involved and a small iOS changes is needed

## Test plan

Tested all `pointerEvents` modes on Android:
- ✅ `pointerEvents="none"` - View and subviews don't receive touches
- ✅ `pointerEvents="box-none"` - View doesn't receive touches, subviews do
- ✅ `pointerEvents="box-only"` - View receives touches, subviews don't
- ✅ `pointerEvents="auto"` - Default behavior works as expected

I've used https://github.com/huextrat/repro-pressable-gh to test scenarios

Tested on both old architecture (Paper) and new architecture (Fabric).

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerButtonComponentView.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerButtonComponentView.mm
@@ -223,6 +223,9 @@ static RNGestureHandlerPointerEvents RCTPointerEventsToEnum(facebook::react::Poi
   _buttonView.hitTestEdgeInsets = UIEdgeInsetsMake(
       -newProps.hitSlop.top, -newProps.hitSlop.left, -newProps.hitSlop.bottom, -newProps.hitSlop.right);
 
+  // We need to cast to ViewProps to access the pointerEvents property with the correct type.
+  // This is necessary because pointerEvents is redefined in the spec,
+  // which shadows the base property with a different, incompatible type.
   const auto &newViewProps = static_cast<const ViewProps &>(newProps);
   if (!oldProps) {
     _buttonView.pointerEvents = RCTPointerEventsToEnum(newViewProps.pointerEvents);


packages/react-native-gesture-handler/src/specs/RNGestureHandlerButtonNativeComponent.ts
```diff
@@ -6,6 +6,7 @@ import type {
 } from 'react-native/Libraries/Types/CodegenTypes';
 import type { ViewProps, ColorValue } from 'react-native';
 
+// @ts-ignore - Redefining pointerEvents with WithDefault for codegen, conflicts with ViewProps type but codegen needs it
```
huextrat, line 9: `Omit` is not working with codegen so I had to ignore the error

If anyone has a better idea, don't hesitate 👍

huextrat:
@m-bert Following the fix for iOS, here is the fix for Android 👍

packages/react-native-gesture-handler/apple/RNGestureHandlerButtonComponentView.mm
```diff
@@ -223,12 +223,14 @@ - (void)updateProps:(const Props::Shared &)props oldProps:(const Props::Shared &
   _buttonView.hitTestEdgeInsets = UIEdgeInsetsMake(
       -newProps.hitSlop.top, -newProps.hitSlop.left, -newProps.hitSlop.bottom, -newProps.hitSlop.right);
 
+  const auto &newViewProps = static_cast<const ViewProps &>(newProps);
```
m-bert, line 229: Could we add a small comment about why are these changes necessary?

packages/react-native-gesture-handler/src/specs/RNGestureHandlerButtonNativeComponent.ts
m-bert: I think we can leave that

packages/react-native-gesture-handler/apple/RNGestureHandlerButtonComponentView.mm
huextrat: Done ✅  1af36768eee44ae6fb6597fb12fc6803a0ac1298

PR #3935, opened 2026-01-27
Move `Reanimated` availability check to `JS` side

m-bert:
## Description

Currently if `Reanimated` is not installed, running apps on `iOS` results in redbox that says that it `Reanimated` module doesn't exist. 

It happens because in [this line](https://github.com/software-mansion/react-native-gesture-handler/blob/6825dbbd07bb73730a7c23ab140f96c00b93bb8d/packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm#L125) in module we call `[self.moduleRegistry moduleForName:"ReanimatedModule"]`. This goes through [RCTModuleRegistry](https://github.com/facebook/react-native/blob/ca520ee4c19dd921f9225c71a9ad872c4bbd7cb9/packages/react-native/React/Base/RCTModuleRegistry.m#L50) to [RCTTurboModuleManager](https://github.com/facebook/react-native/blob/ca520ee4c19dd921f9225c71a9ad872c4bbd7cb9/packages/react-native/ReactCommon/react/nativemodule/core/platform/ios/ReactCommon/RCTTurboModuleManager.mm#L980), where this method is forced to throw redbox if module is not found.

To fix this, we move check for `Reanimated` availability to JS side.

## Test plan

Tested on basic-example app with and without `Reanimated`

### DIFF
--- a/packages/react-native-gesture-handler/src/mocks/mocks.tsx
+++ b/packages/react-native-gesture-handler/src/mocks/mocks.tsx
@@ -25,6 +25,7 @@ const setGestureHandlerConfig = NOOP;
 const updateGestureHandlerConfig = NOOP;
 const flushOperations = NOOP;
 const configureRelations = NOOP;
+const setReanimatedAvailable = NOOP;
 const install = NOOP;
 const NativeViewGestureHandler = View;
 const TapGestureHandler = View;
@@ -66,6 +67,7 @@ export default {
   setGestureHandlerConfig,
   updateGestureHandlerConfig,
   configureRelations,
+  setReanimatedAvailable,
   flushOperations,
   install,
   // Probably can be removed


packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
j-piasecki, line 96: This check is ignored by the new logic

packages/react-native-gesture-handler/src/specs/NativeRNGestureHandlerModule.ts
```diff
@@ -24,6 +24,7 @@ export interface Spec extends TurboModule {
   configureRelations: (handlerTag: Double, relations: Object) => void;
   dropGestureHandler: (handlerTag: Double) => void;
   flushOperations: () => void;
+  setReanimatedAvailability: () => void;
```
j-piasecki: Isn't this method also needed on the web module?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -124,6 +124,11 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
   @ReactMethod
   override fun flushOperations() = Unit
 
+  @ReactMethod
+  override fun setReanimatedAvailability() {
```
j-piasecki:
```suggestion
  override fun setReanimatedAvailable() {
```

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -191,6 +186,11 @@ - (void)flushOperations
   }];
 }
 
+- (void)setReanimatedAvailability
```
j-piasecki:
```suggestion
- (void)setReanimatedAvailable
```

m-bert: Done in [c3e699d](https://github.com/software-mansion/react-native-gesture-handler/pull/3935/commits/c3e699def2cf3f9500a2c6d4719ce485ff4bbe99)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
m-bert: Done in [c3e699d](https://github.com/software-mansion/react-native-gesture-handler/pull/3935/commits/c3e699def2cf3f9500a2c6d4719ce485ff4bbe99)

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
m-bert: Good catch, done in [c3e699d](https://github.com/software-mansion/react-native-gesture-handler/pull/3935/commits/c3e699def2cf3f9500a2c6d4719ce485ff4bbe99)

packages/react-native-gesture-handler/src/specs/NativeRNGestureHandlerModule.ts
m-bert: Added in [1b1b318](https://github.com/software-mansion/react-native-gesture-handler/pull/3935/commits/1b1b318f6996b861f85b549fc5e1a0738d553f6c) & [bd5bdfb](https://github.com/software-mansion/react-native-gesture-handler/pull/3935/commits/bd5bdfb9e73be2fcbfccd1bfc47dba67a5c06ffb)

j-piasecki (approved):

akwasniewski (approved):

PR #3936, opened 2026-01-28
[iOS] Prevent gesture recognizers from beginning, when they should not

j-piasecki:
## Description

Gesture detection on complex views (like Text) relies on `gestureRecognizerShouldBegin` checking for the virtual react tag. The logic itself is fine, but iOS runs it when it wants to see if the gesture should activate, at which point all gestures attached to the composed view have fired their `onBegin` callbacks.

This PR explicitly calls this method inside `onTouches(Interactions)Began` to immediately fail handlers that don't pass the test.

## Test plan

Test the `basic-example/Text` with `onBegin` instead of `onActivate`.

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandler.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandler.mm
@@ -301,6 +301,12 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromReset
 {
+  // Don't dispatch state changes from undetermined when resetting handler. There will be no follow-up
+  // since the handler is being reset, so these events are wrong.
+  if (fromReset && _lastState == RNGestureHandlerStateUndetermined) {
+    return;
+  }
+
   RNGHUIView *view = [self chooseViewForInteraction:recognizer];
 
   // it may happen that the gesture recognizer is reset after it's been unbound from the view,
@@ -308,12 +314,6 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
   // would be nil.
   if (view.reactTag == nil && _actionType != RNGestureHandlerActionTypeNativeDetector &&
       _actionType != RNGestureHandlerActionTypeVirtualDetector) {
-    return;
-  }
-
-  // Don't dispatch state changes from undetermined when resetting handler. There will be no follow-up
-  // since the handler is being reset, so these events are wrong.
-  if (fromReset && _lastState == RNGestureHandlerStateUndetermined) {
     return;
   }
 


packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
```diff
@@ -30,6 +30,12 @@ - (void)touchesBegan:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
   _lastPoint = [[[touches allObjects] objectAtIndex:0] locationInView:_gestureHandler.recognizer.view];
   [_gestureHandler reset];
   [super touchesBegan:touches withEvent:event];
+
+  if (self.state == UIGestureRecognizerStatePossible && ![self.delegate gestureRecognizerShouldBegin:self]) {
+    self.state = UIGestureRecognizerStateFailed;
+    return;
+  }
+
```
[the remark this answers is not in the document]
j-piasecki: Intentional - other way doesn't work

packages/react-native-gesture-handler/apple/Handlers/RNLongPressHandler.m
```diff
@@ -81,6 +81,12 @@ - (void)touchesBegan:(NSSet<RNGHUITouch *> *)touches withEvent:(UIEvent *)event
 {
   [_gestureHandler setCurrentPointerType:event];
   [super touchesBegan:touches withEvent:event];
+
+  if (self.state == UIGestureRecognizerStatePossible && ![self.delegate gestureRecognizerShouldBegin:self]) {
+    self.state = UIGestureRecognizerStateFailed;
+    return;
+  }
+
```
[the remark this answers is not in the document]
j-piasecki: Intentional - other way doesn't work

packages/react-native-gesture-handler/apple/Handlers/RNRotationHandler.m
```diff
@@ -43,6 +43,10 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromReset
 {
+  if (self.state == UIGestureRecognizerStatePossible && ![self.delegate gestureRecognizerShouldBegin:self]) {
+    self.state = UIGestureRecognizerStateFailed;
```
m-bert: Does setting `state` here work? As far as I remember, recognizers are quite strict when it comes to setting state and doing so outside `onTouches*` doesn't work.

j-piasecki: You're right. Fixed in [bf33c8b](https://github.com/software-mansion/react-native-gesture-handler/pull/3936/commits/bf33c8be6e41ac3f18491d2e5cb7b24cf7ec8b05)

j-piasecki:
@m-bert Can you recheck it now?

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -311,6 +311,12 @@ - (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromRese
     return;
   }
 
+  // Don't dispatch state changes from undetermined when resetting handler. There will be no follow-up
+  // since the handler is being reset, so these events are wrong.
+  if (fromReset && _lastState == RNGestureHandlerStateUndetermined) {
+    return;
+  }
+
```
m-bert: We could also do this check earlier. In that case we would avoid calling `chooseViewForInteraction`

m-bert (approved):
Works now 🎉

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
j-piasecki: Done in [6c68fb0](https://github.com/software-mansion/react-native-gesture-handler/pull/3936/commits/6c68fb0779eb64bfb18466b4c161147f0c9470d4)

PR #3939, opened 2026-01-29
[web] fix context menu

akwasniewski:
## Description

Context menu was broken on v3, this PR fixes it. 

WebDelegate never handled changing contextMenu. when context had been disabled the `areContextMenuListenersAdded` was set to true, later when it was enabled it would not change the listener from disabled to enabled as the `addContextMenuListeners` returned after seeing that the listeners had already been added. 

## Test plan

Tested on the following example:

<details>

```tsx

import React from 'react';
import {
  GestureDetector,
  MouseButton,
  usePanGesture,

export default function ContextMenuExample() {
  const p1 = usePanGesture({ mouseButton: MouseButton.RIGHT });
  const p2 = usePanGesture({});
  const p3 = usePanGesture({});
  return (
      <GestureDetector gesture={p1}>
        <View style={[styles.box, styles.grandParent]}>
          <GestureDetector gesture={p2} enableContextMenu={true}>
            <View style={[styles.box, styles.parent]}>
              <GestureDetector gesture={p3} enableContextMenu={false}>
                <View style={[styles.box, styles.child]} />
              </GestureDetector>
            </View>
          </GestureDetector>
        </View>
      </GestureDetector>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
  },

  grandParent: {
    width: 300,
    height: 300,
    backgroundColor: "navy",
  },

  parent: {
    width: 200,
    height: 200,
    backgroundColor: "purple",
  },

  child: {
    width: 100,
    height: 100,
    backgroundColor: "blue",
  },

  box: {
    display: 'flex',
    alignItems: 'center',
    borderRadius: 20,
  },
});

```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/android/noreanimated/src/main/java/com/swmansion/gesturehandler/ReanimatedProxy.kt
+++ b/packages/react-native-gesture-handler/android/noreanimated/src/main/java/com/swmansion/gesturehandler/ReanimatedProxy.kt
@@ -9,8 +9,4 @@ class ReanimatedProxy {
   fun <T : Event<T>> sendEvent(event: T, reactApplicationContext: ReactContext) {
     // no-op
   }
-
-  companion object {
-    const val REANIMATED_INSTALLED = false
-  }
 }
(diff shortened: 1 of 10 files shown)


akwasniewski:
> I'm not a fan of having 2 similar flags, but on the other hand I don't see better solution for now 😅

I thought about making a 3 value a enum out of it but it would just add unnecessary typology.

m-bert:
> I thought about making a 3 value a enum out of it but it would just add unnecessary typology.

I'm not sure if this would work as those flags represent different overlapping states 😞 Either way, it is fine, maybe we will find something better in the future.

PR #3952, opened 2026-02-04
Fix intercepting detector props

akwasniewski:
## Description

 `InterceptingDetectorProps` was missing some types from `GestureDetectorProps` this PR makes both types extend `CommonGestureDetectorProps`

## Test plan

`yarn ts-check`

### DIFF
--- a/.github/workflows/rngh-api-v3.yml
+++ b/.github/workflows/rngh-api-v3.yml
@@ -17,7 +17,7 @@ jobs:
 
     runs-on: ubuntu-latest
     concurrency:
-      group: jest-${{ github.ref }}
+      group: rngh-api-v3-${{ github.ref }}
       cancel-in-progress: true
 
     steps:


packages/react-native-gesture-handler/src/v3/detectors/common.ts
```diff
@@ -11,16 +11,20 @@ export enum GestureDetectorType {
...
 
-export interface InterceptingGestureDetectorProps<THandlerData, TConfig> {
-  children?: React.ReactNode;
+export interface NativeDetectorProps<THandlerData, TConfig>
+  extends CommonGestureDetectorProps {
+  gesture: Gesture<THandlerData, TConfig>;
+}
+
+export interface InterceptingGestureDetectorProps<THandlerData, TConfig>
+  extends CommonGestureDetectorProps {
```
m-bert, line 27: Why not 

```ts
export interface NativeDetectorProps<THandlerData, TConfig> {
  children?: React.ReactNode;
  gesture: Gesture<THandlerData, TConfig>;
  userSelect?: UserSelect;
  touchAction?: TouchAction;
  enableContextMenu?: boolean;
}

export interface InterceptingGestureDetectorProps<THandlerData, TConfig> = Partial<NativeDetectorProps<THandlerData, TConfig>>
```

?

j-piasecki: It's less explicit about intent - currently, it's immediately visible exactly what the difference between the two is. With `Partial`, you need to thinkabout what it changes. It's also less future-proof.

m-bert:
Also, shouldn't these props be added to the `VirtualGestureDetector` as well?

PR #3959, opened 2026-02-06
[docs] Update docs

m-bert:
## Description

In this PR, I go through each section and make sure that everything is in the same style, has correct information and there are no broken links.

## Test plan

Read docs 🤓

### DIFF
--- a/packages/docs-gesture-handler/docs/fundamentals/installation.mdx
+++ b/packages/docs-gesture-handler/docs/fundamentals/installation.mdx
@@ -19,9 +19,9 @@ import { GestureHandlerCompatibility } from '../../components/Compatibility';
 
 ### Running gestures on the UI thread
 
-In order to respond to gestures on UI thread, you also have to install [`react-native-reanimated`](https://docs.swmansion.com/react-native-reanimated/) along with [`react-native-worklets`](https://docs.swmansion.com/react-native-worklets/).
+Using Reanimated is the recommended method of handling gesture-driven interactions on the UI thread. In order to use it, you need to install [`react-native-reanimated`](https://docs.swmansion.com/react-native-reanimated/) along with [`react-native-worklets`](https://docs.swmansion.com/react-native-worklets/). Another approach is to use React Native's [Animated API](https://reactnative.dev/docs/animated).
 
-While using Reanimated is the recommended method for responding to gestures on the UI thread, you can also use the [Animated API](https://reactnative.dev/docs/animated). For more details on how to implement these, refer to the dedicated sections for [Reanimated](/docs/fundamentals/reanimated-interactions) and [Animated](/docs/fundamentals/animated-interactions).
+For more details on how to implement these, refer to the dedicated sections for [Reanimated](/docs/fundamentals/reanimated-interactions) and [Animated](/docs/fundamentals/animated-interactions).
 
 ## Setup
 


packages/docs-gesture-handler/docs/components/reanimated_swipeable.md
```diff
@@ -139,10 +139,10 @@ const panGesture = Gesture.Pan();
 
 <GestureDetector gesture={panGesture}>
   <ReanimatedSwipeable simultaneousWithExternalGesture={panGesture} />
-</GestureDetector>
+</GestureDetector>;
```
j-piasecki: Why?

packages/docs-gesture-handler/docs/fundamentals/_examples/hooks/Simultaneous.tsx
```diff
@@ -0,0 +1,95 @@
+import { StyleSheet } from 'react-native';
```
j-piasecki, line 1: It may make sense to extract all of the examples in this case.

packages/docs-gesture-handler/docs/fundamentals/gesture-detector.mdx
```diff
@@ -10,9 +10,9 @@ import CollapsibleCode from '@site/src/components/CollapsibleCode';
 
 ## Gesture Detector
 
-The `GestureDetector` is a key component of `react-native-gesture-handler`. It supports gestures created either using the hooks API or the builder pattern. Additionally, it allows for the recognition of multiple gestures through [gesture composition](/docs/fundamentals/gesture-composition). `GestureDetector` interacts closely with [`Reanimated`](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/getting-started/). For more details, refer to the [Integration with Reanimated](/docs/fundamentals/reanimated-interactions) section.
+The `GestureDetector` is a key component of `react-native-gesture-handler`. It supports gestures created either using the hook API or the builder pattern. Additionally, it allows for the recognition of multiple gestures through [gesture composition](/docs/fundamentals/gesture-composition). `GestureDetector` interacts closely with [`Reanimated`](https://docs.swmansion.com/react-native-reanimated/). For more details, refer to the [Integration with Reanimated](/docs/fundamentals/reanimated-interactions) section.
```
j-piasecki:
```suggestion
The `GestureDetector` is a key component of `react-native-gesture-handler`. It supports gestures created either using the hook-based API or the builder pattern. Additionally, it allows for the recognition of multiple gestures through [gesture composition](/docs/fundamentals/gesture-composition). `GestureDetector` interacts closely with [`Reanimated`](https://docs.swmansion.com/react-native-reanimated/). For more details, refer to the [Integration with Reanimated](/docs/fundamentals/reanimated-interactions) section.
```

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
```diff
@@ -11,11 +11,17 @@ import { GestureHandlerCompatibility } from '../../components/Compatibility';
...
+### Compatibility with React Native versions
+
 `react-native-gesture-handler` supports the three latest minor releases of `react-native`.
 
 <GestureHandlerCompatibility spacerAfterIndex={1} />
 
-In order to fully utilize the [touch events](/docs/gestures/touch-events/) you also need to use `react-native-reanimated` 2.3.0 or newer.
+### Running gestures on the UI thread
+
+In order to run gestures on the UI thread, you also have to install [`react-native-reanimated`](https://docs.swmansion.com/react-native-reanimated/) along with [`react-native-worklets`](https://docs.swmansion.com/react-native-worklets/).
```
j-piasecki: That's not entirely true - gestures always run on the ui thread. Reanimated (and Animated with native driver) allows responding to events from gestures on the UI thread.

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
```diff
@@ -52,17 +58,22 @@ export default function App() {
...
 
-If you don't provide anything to the `styles` prop, it will default to `flex: 1`. If you want to customize the styling of the root view, don't forget to also include `flex: 1` in the custom style, otherwise your app won't render anything. Keep `GestureHandlerRootView` as close to the actual root of the app as possible. It's the entry point for all gestures and all gesture relations. The gestures won't be recognized outside of the root view, and relations only work between gestures mounted under the same root view.
+If you don't provide anything to the `styles` prop, it will default to `flex: 1`. If you want to customize the styling of the root view, don't forget to also include `flex: 1` in the custom style, otherwise your app won't render anything. 
 
 If you're unsure if one of your dependencies already renders `GestureHandlerRootView` on its own, don't worry and add one at the root anyway. In case of nested root views, Gesture Handler will only use the top-most one and ignore the nested ones.
```
j-piasecki, line 67:
```suggestion
If you're unsure if one of your dependencies already renders `GestureHandlerRootView` on its own, don't worry and add one at the root anyway. In case of nested root views, Gesture Handler will only use the top-most one and ignore the nested ones by default.
```

Do we want to mention the `unstable_forceActive` prop?

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
```diff
@@ -52,17 +58,22 @@ export default function App() {
...
 
-:::tip
-If you're using gesture handler in your component library, you may want to wrap your library's code in the `GestureHandlerRootView` component. This will avoid extra configuration for the user.
+:::note
+Remember that you need to wrap each screen that you use in your app with `GestureHandlerRootView` as with native navigation libraries each screen maps to a separate root view. It will not be enough to wrap the main screen only.
```
j-piasecki, line 70: You don't need to wrap each screen with a separate root view; wrapping the navigator is enough (when it comes to `react-navigation` at least)

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
j-piasecki: We may want to update that.

packages/docs-gesture-handler/docs/fundamentals/state-manager.mdx
```diff
@@ -7,15 +7,15 @@ sidebar_position: 6
 
 import CollapsibleCode from '@site/src/components/CollapsibleCode';
 
-RNGH3 allows to manually control gestures lifecycle by using `GestureStateManager`.
+RNGH3 allows to manually control [gestures lifecycle](/docs/under-the-hood/state) by using [`GestureStateManager`](#gesturestatemanager).
 
 ## State management
 
-Manual state management is based on `handlerTag`. There are two ways of manual state control.
+Manual state management is based on `handlerTag`. There are two ways of manual state control. Some gestures also support [`manualActivation`](#manualactivation) property, which blocks their automatic activation, event if they met specific criteria.
```
j-piasecki:
```suggestion
Manual state management is based on `handlerTag`. There are two ways of manual state control. Some gestures also support [`manualActivation`](#manualactivation) property, which blocks their automatic activation, even if they meet their activation criteria.
```

packages/docs-gesture-handler/docs/fundamentals/state-manager.mdx
```diff
@@ -7,15 +7,15 @@ sidebar_position: 6
...
 
 ### Inside gesture definition
 
-If you want to manipulate gesture's state in its callbacks, you can get `handlerTag` from event parameter.
+If you want to manipulate gesture's state in its callbacks, you can get `handlerTag` from event parameter.
```
j-piasecki:
```suggestion
If you want to manipulate gesture's state in its callbacks, you can get `handlerTag` from event parameter.
```

packages/docs-gesture-handler/docs/gestures/use-fling-gesture.mdx
```diff
@@ -103,19 +106,33 @@ const styles = StyleSheet.create({
 ### direction
 
 ```ts
-direction: Directions | SharedValue<Directions>;
 ```
```
j-piasecki: Was leaving that intentional?

packages/docs-gesture-handler/docs/gestures/use-native-gesture.mdx
```diff
@@ -5,21 +5,30 @@ sidebar_label: Native gesture
...
 import BaseContinuousGestureCallbacks from './\_shared/base-continuous-gesture-callbacks.mdx';
 import SharedValueInfo from './\_shared/shared-value-info.md';
 
 A gesture that allows other touch handling components to work within RNGH's gesture system. This streamlines interactions between gestures and the native component, allowing it to form [relations](/docs/fundamentals/gesture-composition) with other gestures.
 
-When used, the native component should be the direct child of a `GestureDetector`.
+When used, the native component should be the direct child of a [`GestureDetector`](/docs/fundamentals/gesture-detectors#gesture-detector).
+
+:::danger
+Do not use `Native` gesture with components exported by React Native Gesture Handler, as they already have it preapplied. Attaching `Native` gesture twice will likely result in the components not working as intended.
```
j-piasecki:
```suggestion
Do not use `Native` gesture with components exported by React Native Gesture Handler, as they already have it pre-applied. Attaching `Native` gesture twice will result in undefined behavior.
```

packages/docs-gesture-handler/docs/gestures/use-rotation-gesture.mdx
```diff
@@ -25,18 +25,24 @@ import RotationGestureBasicSrc from '!!raw-loader!@site/static/examples/Rotation
...
-The gesture activates when fingers are placed on the screen and rotate around a common point.
+import RotationExample from '!!raw-loader!./_examples/RotationExample';
 
-Gesture callback can be used for continuous tracking of the rotation gesture. It provides information about the gesture such as the amount rotated, the focal point of the rotation (anchor), and its instantaneous velocity.
+Gesture that can recognize rotation and track its movement.
+
+The gesture activates when fingers are placed on the screen and rotate around a common point. It provides information such as the amount [rotated](#rotation), the focal point of the rotation ([anchor](#anchorx)), and its instantaneous [velocity](#velocity).
+
+:::note
+When implementing rotation based on `anchor` point, make sure to use it after the gesture had activated, i.e. in `onActivate` or `onUpdate` callbacks. Using it in `onBegin` may lead to unexpected behavior.
```
j-piasecki, line 44: How hard would it be to hide those fields from `onBegin`?

packages/docs-gesture-handler/docs/components/reanimated_swipeable.md
m-bert: I'm not sure why it was here, but should be gone after [e9e88a5](https://github.com/software-mansion/react-native-gesture-handler/pull/3959/commits/e9e88a5b813e794f2307a0cd4b38f57005a1c32f)

packages/docs-gesture-handler/docs/fundamentals/_examples/hooks/Simultaneous.tsx
m-bert: tl;dr

It's done on purpose because of string interpolation in this example.

<hr />

Yes and no. This one was extracted because it contains string interpolation, so passing it as `string` in `src` doesn't really work. Also I didn't want to change the problematic fragment to 

```ts
 { rotateZ: rotation.value + "rad" }
```

as this is not how you would normally write it. 

There are many examples in the docs, so I don't think extracting each one makes sense. If you're talking just about composition then maybe, but I wouldn't do it with all.

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
m-bert: So maybe:

"In order to respond to gestures on UI thread"

Also, do we want to mention `Animated` in this case?

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
m-bert: I think we can leave root view for #3970. If there's something that should be changed in it, please leave a comment there.

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
m-bert: Okay, this also might be to #3970. Maybe we could rephrase that so it says that wrapping navigator should be. enough, but if it isn't, then wrap all your screens?

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
m-bert: Sure 😶 [f3500a4](https://github.com/software-mansion/react-native-gesture-handler/pull/3959/commits/f3500a4ab3f753a2bc881709691897dc62016209)

packages/docs-gesture-handler/docs/gestures/use-fling-gesture.mdx
m-bert: If you mean the `ts` block, then no 😄 [f927bf5](https://github.com/software-mansion/react-native-gesture-handler/pull/3959/commits/f927bf516e976adad45ede9e27404521d3d139f1)

packages/docs-gesture-handler/docs/gestures/use-native-gesture.mdx
m-bert: But still, we will have to update docs for `Native` gesture 😶

packages/docs-gesture-handler/docs/gestures/use-rotation-gesture.mdx
m-bert: It depends 😄 Do you mean _hide_ (ts) or _remove_? It is not hard to to this, question is how to do it right 😅 I can do this, but first let's decide whether to remove them, or just hide.

j-piasecki: Ideally, we would remove them (and the focal point from pinch), but then we end up with `onUpdate` with a different event type than other callbacks. I'm not sure whether that's an ok tradeoff.

packages/docs-gesture-handler/docs/fundamentals/_examples/hooks/Simultaneous.tsx
j-piasecki: Yeah, I meant in composition, there are a few of them here and are quite long.

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
j-piasecki: > Also, do we want to mention Animated in this case?

We can mention it, but Reanimated is the recommended way.

packages/docs-gesture-handler/docs/gestures/use-rotation-gesture.mdx
m-bert: Either way we end up with different type 😞 Doesn't matter if we remove them or hide - it is still different

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
m-bert: Okay, take a look at [4fd4d8c](https://github.com/software-mansion/react-native-gesture-handler/pull/3959/commits/4fd4d8c3757ac136e028f6877121b57aadea4972). I'm not sure if we should write "you have to install (...)" and later say that "hey, btw. you can use Animated", but those are quite different and maybe it will work. WDYT?

packages/docs-gesture-handler/docs/fundamentals/state-manager.mdx
```diff
@@ -125,7 +125,59 @@ const styles = StyleSheet.create({
 });
 `}/>
 
-## Methods
+
+### manualActivation
+
+When `manualActivation` property is set to `true` on a gesture, it will not activate automatically even if its activation criteria are met. You will need to use [`GestureStateManager`](#gesturestatemanager) to manipulate its state manually.
```
akwasniewski: You `will need` kinda implies that it's not intended.
```suggestion
When `manualActivation` property is set to `true` on a gesture, it will not activate automatically even if its activation criteria are met. You can then use [`GestureStateManager`](#gesturestatemanager) to manipulate its state manually.
```

packages/docs-gesture-handler/docs/fundamentals/_examples/hooks/Simultaneous.tsx
m-bert: Okay, done in [a09df73](https://github.com/software-mansion/react-native-gesture-handler/pull/3959/commits/a09df737fe78218a3a238b128b7e4c704b2b3d5e)

packages/docs-gesture-handler/docs/gestures/use-pan-gesture.mdx
```diff
@@ -136,15 +143,15 @@ Minimum distance the finger (or multiple finger) need to travel before the gestu
...
+A number of fingers that is required to be placed before gesture can activate. Should be a higher or equal to 0.
 
 ### maxPointers
 
 ```ts
 maxPointers: number | SharedValue<number>;
 ```
 
-When the given number of fingers is placed on the screen and gesture hasn't yet activated it will fail recognizing the gesture. Should be a higher or equal to 0 integer.
+When the given number of fingers is placed on the screen and gesture hasn't yet activated it will fail recognizing the gesture. Should be a higher or equal to 0.
```
akwasniewski:
```suggestion
When the given number of fingers is placed on the screen and gesture hasn't yet activated it will fail recognizing the gesture. Should be non-negative.
```
or
```suggestion
When the given number of fingers is placed on the screen and gesture hasn't yet activated it will fail recognizing the gesture. Should be higher or equal to 0
```

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
m-bert: Added to #3970 in [this commit]([47b6daf](https://github.com/software-mansion/react-native-gesture-handler/pull/3970/commits/47b6daffa12ca921ed4a55ffd4147458c243278f)). Let me know what you think (preferably in mentioned PR).

packages/docs-gesture-handler/docs/fundamentals/state-manager.mdx
m-bert: > You will need kinda implies that it's not intended.

I'm not sure why it implies that it is not intended, given this context. Also, "You can" sounds like there are other possibilities, but that's not true in that case. 

If you really want this change, maybe this would work:

"Use `GestureStateManager` to manipulate its state manually."

Let me know what you think.

packages/docs-gesture-handler/docs/gestures/use-pan-gesture.mdx
m-bert: I'll go with second one as this is more common in this documentation.

m-bert: Done in [f94da57](https://github.com/software-mansion/react-native-gesture-handler/pull/3959/commits/f94da573a8cd005c07ecd8745fcd996ec6696257)

packages/docs-gesture-handler/docs/fundamentals/state-manager.mdx
akwasniewski: I would go with it "Use GestureStateManager...", but if you want to you can leave it as it is.

m-bert: [c9c429e](https://github.com/software-mansion/react-native-gesture-handler/pull/3959/commits/c9c429ebcf032783e8aa6da01bf19b3a9c7528d0)

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
```diff
@@ -11,11 +11,19 @@ import { GestureHandlerCompatibility } from '../../components/Compatibility';
...
 `react-native-gesture-handler` supports the three latest minor releases of `react-native`.
 
 <GestureHandlerCompatibility spacerAfterIndex={1} />
 
-In order to fully utilize the [touch events](/docs/gestures/touch-events/) you also need to use `react-native-reanimated` 2.3.0 or newer.
+### Running gestures on the UI thread
+
+In order to respond to gestures on UI thread, you also have to install [`react-native-reanimated`](https://docs.swmansion.com/react-native-reanimated/) along with [`react-native-worklets`](https://docs.swmansion.com/react-native-worklets/).
+
+While using Reanimated is the recommended method for responding to gestures on the UI thread, you can also use the [Animated API](https://reactnative.dev/docs/animated). For more details on how to implement these, refer to the dedicated sections for [Reanimated](/docs/fundamentals/reanimated-interactions) and [Animated](/docs/fundamentals/animated-interactions).
```
j-piasecki:
```suggestion
Using Reanimated is the recommended method of handling gesture-driven interactions on the UI thread. In order to use it, you need to install [`react-native-reanimated`](https://docs.swmansion.com/react-native-reanimated/) along with [`react-native-worklets`](https://docs.swmansion.com/react-native-worklets/). Another approach is to use React Native's [Animated API](https://reactnative.dev/docs/animated).

For more details on how to implement these, refer to the dedicated sections for [Reanimated](/docs/fundamentals/reanimated-interactions) and [Animated](/docs/fundamentals/animated-interactions).
```

packages/docs-gesture-handler/docs/gestures/use-rotation-gesture.mdx
j-piasecki: Let's leave it as is in this PR to reduce scope creep and continue the discussion offline.

packages/docs-gesture-handler/docs/fundamentals/installation.mdx
m-bert: ❤️

PR #3961, opened 2026-02-06
Test examples

akwasniewski:
## Description

This PR ports some legacy examples to v3, which I didn't rewrite previosuly in https://github.com/software-mansion/react-native-gesture-handler/pull/3857. Particularly those marked as "release tests" and those related to pressable.

It also adds two new examples:
* reattach, for testing whether gesture reattaching works, because I often find myself testing it
* timer, inspired by old multitap example, which I found confusing.

## Test plan

Check if the examples work

### DIFF
--- a/packages/react-native-gesture-handler/src/index.ts
+++ b/packages/react-native-gesture-handler/src/index.ts
@@ -85,7 +85,6 @@ export {
   LegacyBaseButton,
   LegacyRectButton,
   LegacyBorderlessButton,
-  LegacyPureNativeButton,
 } from './components/GestureButtons';
 
 export type {
--- a/packages/react-native-gesture-handler/src/v3/index.ts
+++ b/packages/react-native-gesture-handler/src/v3/index.ts
@@ -63,7 +63,6 @@ export {
   BaseButton,
   RectButton,
   BorderlessButton,
-  PureNativeButton,
   Pressable,
   ScrollView,
   Switch,


apps/common-app/src/new_api/index.tsx
```diff
@@ -95,4 +106,17 @@ export const NEW_EXAMPLES: ExamplesSection[] = [
       { name: 'Reanimated Drawer Layout', component: ReanimatedDrawerLayout },
     ],
   },
+  {
+    sectionTitle: 'Tests',
+    data: [
+      { name: 'RectButton', component: RectButtonExample },
+      { name: 'Two Finger Pan', component: TwoFingerPanExample },
```
j-piasecki: Can we add that "two finger" refers to the trackpad?

apps/common-app/src/new_api/tests/reattaching/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <View
+            style={[
+              commonStyles.box,
+              { backgroundColor: isTopActive ? COLORS.NAVY : COLORS.RED },
+            ]}
+          />
+        </GestureDetector>
+        <Feedback ref={feedbackRef} />
+      </View>
+    </View>
```
j-piasecki, line 53: This one could use some explanation/hints. It's confusing just looking at the two squares on the screen.

Maybe the clickable one could have a "tap me" text? The color scheme doesn't suggest which one is active.

apps/common-app/src/new_api/index.tsx
```diff
@@ -95,4 +106,17 @@ export const NEW_EXAMPLES: ExamplesSection[] = [
...
+      { name: 'Web Styles Reset', component: WebStylesResetExample },
```
j-piasecki: This crashes on mobile. If it's supposed to be web-only, then I believe we had some way of limiting examples to certain platforms.

PR #3968, opened 2026-02-09
[docs] Buttons

m-bert:
## Description

This PR updates Buttons entry in docs. 

## Test plan

Read docs 🤓


skills/gesture-handler-3-migration/SKILL.md
```diff
@@ -160,6 +167,8 @@ Don't suggest replacing buttons from Gesture Handler with components from React
 
 The implementation of buttons has been updated, resolving most button-related issues. They have also been internally rewritten to utilize the new hook API. The legacy JS implementations of button components are still accessible but have been renamed with the prefix `Legacy`, e.g., `RectButton` is now available as `LegacyRectButton`. Those still use the new native component under the hood.
 
+`PureNativeButton` has been removed. If encountered, inform the user that it has been removed and let them decide how to handle that case. They can achieve similar funcionality with other buttons.
```
akwasniewski: We probably want to suggest possible solutions it can propose to the user

packages/docs-gesture-handler/docs/components/buttons.mdx
```diff
@@ -11,111 +11,178 @@ import GifGallery from '@site/components/GifGallery';
...
-- [`BaseButton`](/docs/components/buttons/#basebutton)
-- [`RectButton`](/docs/components/buttons/#rectbutton)
-- [`BorderlessButton`](/docs/components/buttons/#borderlessbutton)
+- [`RawButton`](#rawbutton)
+- [`BaseButton`](#basebutton)
+- [`RectButton`](#rectbutton)
+- [`BorderlessButton`](#borderlessbutton)
 
-On top of that all the buttons are wrapped with `NativeViewGestureHandler` and therefore allow for all the [common gesture handler properties](/docs/gestures/use-native-gesture#properties-common-to-all-gestures) and `NativeViewGestureHandler`'s [extra properties](/docs/gestures/use-native-gesture#properties-specific-to-nativegesture) to be applied to them.
+On top of that all the buttons are wrapped with [`Native`](/docs/gestures/use-native-gesture) gesture and therefore allow for all its [properties](/docs/gestures/use-native-gesture#config) to be applied to them.
```
j-piasecki:
```suggestion
On top of that all the buttons are wrapped with [`Native gesture`](/docs/gestures/use-native-gesture) and therefore allow for all its [properties](/docs/gestures/use-native-gesture#config) to be applied to them.
```

packages/docs-gesture-handler/docs/components/buttons.mdx
```diff
@@ -11,111 +11,178 @@ import GifGallery from '@site/components/GifGallery';
...
+The Gesture Handler library offers native components that function as buttons, serving as alternatives to `TouchableHighlight` or `TouchableOpacity` from the core React Native framework. These buttons process touch recognition natively, which ensures a deterministic response. This capability significantly enhances performance; for example, it allows for immediate ripple effects on Android, unlike `TouchableNativeFeedback`, which requires a touch event roundtrip to JavaScript that can cause delays, especially noticeable on older devices. Additionally, these components handle default platform interactions natively, particularly in scrollable containers where interactions are smartly delayed to prevent unintended highlighting during a fling.
```
j-piasecki, line 14: Do we mention pressable anywhere? If not, then this may be the right place to do so.

skills/gesture-handler-3-migration/SKILL.md
m-bert: I think it depends on their use case. Using `PureNativeButton` should be rare case anyway and I'm not sure if simply saying "change it to `RawButton`" will work, as it has `Native gesture` pre-applied. So I'm not sure if there's a panacea for that specific case. cc @j-piasecki

packages/docs-gesture-handler/docs/components/buttons.mdx
m-bert: [2418206](https://github.com/software-mansion/react-native-gesture-handler/pull/3968/commits/241820604e492b18dacce09c0babfa337c27a875)

packages/docs-gesture-handler/docs/components/buttons.mdx
m-bert: It has it's own docs page 😅

packages/docs-gesture-handler/docs/guides/upgrading-to-3.mdx
```diff
@@ -259,6 +259,7 @@ Although the legacy JS implementation of the buttons is still available, they al
 | `BaseButton`          | `LegacyBaseButton`          |
 | `RectButton`          | `LegacyRectButton`          |
 | `BorderlessButton`    | `LegacyBorderlessButton`    |
+| `PureNativeButton`    | *Not available in Gesture Handler 3*    |
```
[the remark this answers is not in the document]
m-bert: There's already PR that removes it (#3972)

packages/docs-gesture-handler/docs/guides/upgrading-to-3.mdx
```diff
@@ -248,7 +248,7 @@ code2={
 
 The implementation of buttons has been updated, resolving most button-related issues. They have also been internally rewritten to utilize the new hook API. The original button components are still accessible but have been renamed with the prefix `Legacy`, e.g., `RectButton` is now available as `LegacyRectButton`.
 
-Although the legacy JS implementation of the buttons is still available, they also use the new host component internally.
+Although the legacy JS implementation of the buttons is still available, they also use the new host component internally. Because of that, `PureNativeButton` is no longer available in Gesture Handler 3.
```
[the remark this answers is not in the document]
m-bert: There's already PR that removes it (https://github.com/software-mansion/react-native-gesture-handler/pull/3972)

skills/gesture-handler-3-migration/SKILL.md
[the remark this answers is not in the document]
m-bert: There's already PR that removes it (https://github.com/software-mansion/react-native-gesture-handler/pull/3972)

packages/docs-gesture-handler/docs/components/buttons.mdx
```diff
@@ -11,111 +11,153 @@ import GifGallery from '@site/components/GifGallery';
...
+      <Text>Bar</Text>
+    </View>
+  </RectButton>
+);
+```
 
 ## Design patterns
 
-Components listed here were not designed to behave and look in the same way on both platforms but rather to be used for handling similar behaviour on iOS and Android taking into consideration their's design concepts.
+Components listed here were not designed to behave and look in the same way on both platforms but rather to be used for handling similar behaviour on iOS and Android taking into consideration theirs design concepts.
```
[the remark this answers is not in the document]
m-bert: [698a100](https://github.com/software-mansion/react-native-gesture-handler/pull/3968/commits/698a100cbd6579db9a62c1943f744e3898c75f9a)

packages/docs-gesture-handler/docs/components/buttons.mdx
```diff
@@ -11,111 +11,153 @@ import GifGallery from '@site/components/GifGallery';
...
+
+```ts
+activeOpacity?: number;
+```
+
+Opacity applied to the underlay when button is in active state.
+
+## BorderlessButton
 
 This type of button component should be used with simple icon-only or text-only buttons. The interaction will be different depending on platform: on Android a borderless ripple will be rendered (it means that the ripple will animate into a circle that can span outside of the view bounds), whereas on iOS the button will be dimmed (similar to how `TouchableOpacity` works). In addition to the props of [`BaseButton`](/docs/components/buttons#basebutton), it accepts the following:
```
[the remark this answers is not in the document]
m-bert: [698a100](https://github.com/software-mansion/react-native-gesture-handler/pull/3968/commits/698a100cbd6579db9a62c1943f744e3898c75f9a)

skills/gesture-handler-3-migration/SKILL.md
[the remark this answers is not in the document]
m-bert: [698a100](https://github.com/software-mansion/react-native-gesture-handler/pull/3968/commits/698a100cbd6579db9a62c1943f744e3898c75f9a)

packages/docs-gesture-handler/docs/components/buttons.mdx
akwasniewski: Yes, but I think it still deserves to be mentioned in this introduction, after all it also functions as a button.

skills/gesture-handler-3-migration/SKILL.md
j-piasecki: I'm not sure there is a clear migration path here - I think the only use case for `PureNativeButton` was to pass it into `createAnimatedComponent` and wrap with a native gesture later. `createAnimatedComponent` now won't work with the new implementation due to the "sandwich" structure of the new button, so this use case is gone.

You'd most likely need to wrap the button with a view and animate that, while having the button styled in a way to adapt to the view changes. We could add that to the skill, with some note that the agent should ask or at least inform the user that it needs double checking.

packages/docs-gesture-handler/docs/components/buttons.mdx
j-piasecki: Hmm, maybe it's worth mentioning that there is a `Pressable` component built on top of the buttons that already has a lot of abstraction. But also, the `Pressable` section is literally below the `Buttons` section in the docs, so I don't have a strong preference.

m-bert: Well, we've already discussed it to leave it as it is. I can mention it if you want, but I don't think it makes much sense. If we do not make it as infobox (like note/info) it will be easy to miss it, especially if we add this information at the bottom. 

Also this might sound like "Hey, we have buttons, but maybe you should use `Pressable`".

My opinion is that it is not necessary, but if you both agree that it should be added then I can do it.

PR #3974, opened 2026-02-12
[Native] Fix ButtonWrapper Shadow Node custom layout logic

akwasniewski:
## Description

After https://github.com/software-mansion/react-native-gesture-handler/pull/3634 the Pressable did not work on iOS. 
The reason lied in the shadow node of the button wrapper introduced in aforementioned PR which took the layout of the button and instead of changing its own layout to the button's it changed the child view's layout. As a result, the detector got bounds (0,0,0,0) thus all the non-native were considered out of bounds. This broke pressable as it relies on longPress to register touches.

## Test plan

Tested on the following code

<details>

```tsx

import React from 'react';
import { Text, StyleSheet, View } from 'react-native';
import { Pressable, GestureHandlerRootView } from 'react-native-gesture-handler'

const PressableExample = () => {
  return (
      <Pressable onPress={() => setCount((c) => c + 1)}>
        <View style={styles.pressable}>
          <Text>{count}</Text>
        </View>
      </Pressable>
    </GestureHandlerRootView >
  );
};

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
  pressable: {
    alignItems: 'center', justifyContent: 'center', backgroundColor: 'green', width: 100, height: 100
  },
  wrapperCustom: {
    borderRadius: 8,
    padding: 16,
    minWidth: 150,
    alignItems: 'center',
  },
  text: {
    fontSize: 18,
    color: 'white',
    fontWeight: '600',
  },
});

export default PressableExample;

```

</details>

### DIFF
--- a/apps/common-app/src/common.tsx
+++ b/apps/common-app/src/common.tsx
@@ -32,6 +32,21 @@ export interface ExamplesSection {
   data: Example[];
 }
 
+export const COLORS = {
+  offWhite: '#f8f9ff',
+  headerSeparator: '#eef0ff',
+  PURPLE: '#b58df1',
+  NAVY: '#001A72',
+  RED: '#A41623',
+  YELLOW: '#F2AF29',
+  GREEN: '#0F956F',
+  GRAY: '#ADB1C2',
+  KINDA_RED: '#FFB2AD',
+  KINDA_YELLOW: '#FFF096',
+  KINDA_GREEN: '#C4E7DB',
+  KINDA_BLUE: '#A0D5EF',
+};
+
 /* eslint-disable react-native/no-unused-styles */
 export const commonStyles = StyleSheet.create({
   centerView: {
@@ -44,6 +59,8 @@ export const commonStyles = StyleSheet.create({
     width: 150,
     borderRadius: 20,
     marginBottom: 30,
+    justifyContent: 'center',
+    alignItems: 'center',
   },
   ball: {
     height: 120,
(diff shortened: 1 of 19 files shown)


packages/react-native-gesture-handler/shared/shadowNodes/react/renderer/components/rngesturehandler_codegen/RNGestureHandlerButtonWrapperShadowNode.cpp
```diff
@@ -88,10 +88,12 @@ void RNGestureHandlerButtonWrapperShadowNode::layout(
...
   previousGrandChildLayoutMetrics_ = metrics;
 
-  mutableChild->setLayoutMetrics(metrics);
+  setLayoutMetrics(metrics);
 
   auto metricsNoOrigin = grandChild->getLayoutMetrics();
   metricsNoOrigin.frame.origin = Point{};
+
+  mutableChild->setLayoutMetrics(metricsNoOrigin);
   mutableGrandChild->setLayoutMetrics(metricsNoOrigin);
```
j-piasecki, line 99: You also need to handle that in `if (shouldSkipCustomLayout)` branch above.

akwasniewski: merci, 5e6f6b7498fe00df44057cf62492d794e996853a

PR #3975, opened 2026-02-13
[docs] Add platform badges

m-bert:
## Description

This PR adds badges for platform specific features.

| darkmode | lightmode |
| ---------- | ---------- |
|image | image|

## Test plan

Read docs 🤓

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandler.h
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandler.h
@@ -85,6 +85,7 @@
 @property (nonatomic) BOOL dispatchesReanimatedEvents;
 @property (nonatomic, weak, nullable) RNGHUIView *hostDetectorView;
 @property (nonatomic, nullable, assign) NSNumber *virtualViewTag;
+@property (nonatomic, copy, nullable) NSNumber *viewTag;
 
 - (BOOL)isViewParagraphComponent:(nullable RNGHUIView *)view;
 - (nonnull RNGHUIView *)chooseViewForInteraction:(nonnull UIGestureRecognizer *)recognizer;
--- a/packages/react-native-gesture-handler/apple/RNGestureHandler.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandler.mm
@@ -262,6 +262,8 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
 
   [recognizerView addGestureRecognizer:self.recognizer];
   [self bindManualActivationToView:recognizerView];
+
+  self.viewTag = view.reactTag;
 }
 
 - (void)unbindFromView
@@ -271,6 +273,7 @@ static NSHashTable<RNGestureHandler *> *allGestureHandlers;
 
   self.hostDetectorView = nil;
   self.virtualViewTag = nil;
+  self.viewTag = nil;
 
   [self unbindManualActivation];
 }
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.h
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.h
@@ -42,4 +42,6 @@
 - (nullable RNGestureHandler *)handlerWithTag:(nonnull NSNumber *)handlerTag;
 
 - (nullable RNGHUIView *)viewForReactTag:(nonnull NSNumber *)reactTag;
+
+- (void)reattachHandlersIfNeeded;
 @end
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerManager.mm
@@ -171,6 +171,21 @@ constexpr int NEW_ARCH_NUMBER_OF_ATTACH_RETRIES = 25;
   }
 
   [_attachRetryCounter removeObjectForKey:viewTag];
+  [self maybeBindHandler:handlerTag toViewWithTag:viewTag withActionType:actionType withHostDetector:hostDetector];
+}
+
+// Resolves a view tag to its native UIView (including contentView unwrapping),
+// sets reactTag, and binds the gesture handler to it. No-op if the handler is
+// already attached to the correct view.
+- (void)maybeBindHandler:(nonnull NSNumber *)handlerTag
+           toViewWithTag:(nonnull NSNumber *)viewTag
+          withActionType:(RNGestureHandlerActionType)actionType
+        withHostDetector:(nullable RNGHUIView *)hostDetector
+{
+  RNGHUIView *view = [_viewRegistry viewForReactTag:viewTag];
+  if (view == nil || view.superview == nil) {
+    return;
+  }
 
   // I think it should be moved to RNNativeViewHandler, but that would require
   // additional logic for setting contentView.reactTag, this works for now
@@ -181,12 +196,17 @@ constexpr int NEW_ARCH_NUMBER_OF_ATTACH_RETRIES = 25;
     }
   }
 
+  RNGestureHandler *handler = [_registry handlerWithTag:handlerTag];
+
+  // Already attached to the correct native view, nothing to do.
+  if (handler != nil && handler.recognizer.view == view && handler.actionType == actionType) {
+    return;
+  }
+
   view.reactTag = viewTag; // necessary for RNReanimated eventHash (e.g. "42onGestureHandlerEvent"), also will be
                            // returned as event.target
 
   [_registry attachHandlerWithTag:handlerTag toView:view withActionType:actionType withHostDetector:hostDetector];
-
-  // register view if not already there
   [self registerViewWithGestureRecognizerAttachedIfNeeded:view];
 }
 
@@ -222,6 +242,24 @@ constexpr int NEW_ARCH_NUMBER_OF_ATTACH_RETRIES = 25;
 - (id)handlerWithTag:(NSNumber *)handlerTag
 {
   return [_registry handlerWithTag:handlerTag];
+}
+
+- (void)reattachHandlersIfNeeded
+{
+  // Re-bind handlers to their current native views. On Fabric, when a parent view has
+  // display:none and siblings change, the native UIView backing a component may be recycled
+  // and replaced. maybeBindHandler is a no-op if the view is nil or unchanged. This is only
+  // needed for handlers using the old api.
+  for (RNGestureHandler *handler in _registry.handlers.objectEnumerator) {
+    if (handler.viewTag == nil || [handler usesNativeOrVirtualDetector]) {
+      continue;
+    }
+
+    [self maybeBindHandler:handler.tag
+             toViewWithTag:handler.viewTag
+            withActionType:handler.actionType
+          withHostDetector:nil];
+  }
 }
 
 #pragma mark Root Views Management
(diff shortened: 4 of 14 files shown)


m-bert:
>Can you merge/rebase main? The diff seems wrong now. 

Done, could you re-check if everything is fine? @j-piasecki 

> Looks nice! Though badges seem to be missing from components/buttons

Yes, those are missing from buttons because buttons docs are rewritten in other PR and I think that this one will be merged earlier 😅 (@akwasniewski)

packages/docs-gesture-handler/docs/fundamentals/gesture-composition.mdx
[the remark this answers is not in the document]
m-bert: It has already been reverted ([a997a49](https://github.com/software-mansion/react-native-gesture-handler/pull/3975/commits/a997a49a6aa3db7349483daec0368253026387f8))

packages/docs-gesture-handler/docs/components/reanimated_swipeable.mdx
```diff
@@ -269,7 +275,9 @@ enum MouseButton {
 
 Allows users to choose which mouse button should handler respond to. Arguments can be combined using `|` operator, e.g. `mouseButton(MouseButton.LEFT | MouseButton.RIGHT)`. Default value is set to `MouseButton.LEFT`.
 
-### enableContextMenu (Web only)
+<HeaderWithBadge platforms={[ 'web']}>
```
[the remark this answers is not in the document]
m-bert: [b6eead3](https://github.com/software-mansion/react-native-gesture-handler/pull/3975/commits/b6eead3ef27d862fae4f035b1df3560d254df571)

packages/docs-gesture-handler/src/components/HeaderWithBadge/index.tsx
```diff
@@ -0,0 +1,68 @@
+import { StyleSheet } from 'react-native';
+
+interface HeaderWithBadgeProps {
+  platforms: ('android' | 'iOS' | 'web')[];
+  children?: React.ReactNode;
```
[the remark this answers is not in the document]
m-bert: [b6eead3](https://github.com/software-mansion/react-native-gesture-handler/pull/3975/commits/b6eead3ef27d862fae4f035b1df3560d254df571)

packages/docs-gesture-handler/docs/components/pressable.mdx
```diff
@@ -87,31 +89,39 @@ cancelable?: null | boolean;
 
 Whether a press gesture can be interrupted by a parent gesture such as a scroll event. Defaults to `true`.
 
-### onHoverIn (Web only)
+<HeaderWithBadge platforms={['web']}>
+### onHoverIn
+</HeaderWithBadge>
```
j-piasecki: Doesn't this also work on native when using mouse/stylus?

packages/docs-gesture-handler/docs/components/pressable.mdx
```diff
@@ -87,31 +89,39 @@ cancelable?: null | boolean;
...
 ```ts
 onHoverIn?: null | ((event: PressableEvent) => void);
 ```
 
 Called when pointer is hovering over the element.
 
-### onHoverOut (Web only)
+<HeaderWithBadge platforms={['web']}>
+### onHoverOut
+</HeaderWithBadge>
```
j-piasecki: same

packages/docs-gesture-handler/docs/components/pressable.mdx
```diff
@@ -87,31 +89,39 @@ cancelable?: null | boolean;
...
 ```ts
 onHoverOut?: null | ((event: PressableEvent) => void);
 ```
 
 Called when pointer stops hovering over the element.
 
-### delayHoverIn (Web only)
+<HeaderWithBadge platforms={['web']}>
+### delayHoverIn
+</HeaderWithBadge>
```
j-piasecki, line 114: This one might be web only, but it would be best to double-check as well.

packages/docs-gesture-handler/docs/components/pressable.mdx
```diff
@@ -87,31 +89,39 @@ cancelable?: null | boolean;
...
 ```ts
 delayHoverIn?: number | null;
 ```
 
 Duration to wait after hover in before calling `onHoverIn`.
 
-### delayHoverOut (Web only)
+<HeaderWithBadge platforms={['web']}>
+### delayHoverOut
+</HeaderWithBadge>
```
j-piasecki, line 124: same

packages/docs-gesture-handler/docs/components/reanimated_swipeable.mdx
```diff
@@ -240,15 +242,19 @@ blocksExternalGesture?: AnyGesture | AnyGesture[];
...
 ```ts
 enableTrackpadTwoFingerGesture?: boolean;
 ```
 
 Enables two-finger gestures on supported devices, for example iPads with trackpads. If not enabled the gesture will require click + drag, with `enableTrackpadTwoFingerGesture` swiping with two fingers will also trigger the gesture.
 
-### mouseButton (Web & Android only)
+<HeaderWithBadge platforms={['android', 'web']}>
+### mouseButton
+</HeaderWithBadge>
```
j-piasecki, line 257: Unrelated, but don't drawer and swipeable share these props with pan gesture? If so, it may make sense to unify those in some way.

packages/docs-gesture-handler/docs/gestures/use-pan-gesture.mdx
```diff
@@ -209,23 +211,29 @@ When the finger moves outside this range (in points) along X axis and gesture ha
 If range is set as an array, first value must be lower or equal to 0, a the second one higher or equal to 0.
 If only one number `p` is given a range of `(-inf, p)` will be used if `p` is higher or equal to 0 and `(-p, inf)` otherwise.
 
-### averageTouches (Android only)
+<HeaderWithBadge platforms={['android']}>
+### averageTouches
+</HeaderWithBadge>
```
j-piasecki, line 216: This doesn't affect web?

packages/docs-gesture-handler/docs/gestures/use-tap-gesture.mdx
```diff
@@ -153,7 +155,9 @@ maxDistance: number | SharedValue<number>;
 
 Maximum distance, expressed in points, that defines how far the finger is allowed to travel during a tap gesture. If the finger travels further than the defined distance and the gesture hasn't yet activated, it will fail to recognize the gesture.
 
-### mouseButton (Web & Android only)
+<HeaderWithBadge platforms={['android', 'web']}>
+### mouseButton
+</HeaderWithBadge>
```
j-piasecki, line 160: We don't have a way to share common props between the pages?

packages/docs-gesture-handler/docs/components/pressable.mdx
m-bert: I also thought about it. I'm not sure about iOS, I can test it on android though.

packages/docs-gesture-handler/docs/gestures/use-tap-gesture.mdx
m-bert: We do, but not for this case. I agree that we can extract it. As far as I remember some of descriptions are slightly different, but maybe there's a way to unify them.

packages/docs-gesture-handler/docs/gestures/use-pan-gesture.mdx
m-bert: As far as I know - it doesn't.

packages/docs-gesture-handler/docs/components/reanimated_swipeable.mdx
m-bert: Actually, now I've checked that and it is not present in `Swipeable` since #3820. Now, we can either go back to all pan props and handle them in the implementation (which in hook-based API will be easier than in builder), or remove them from docs. 

I don't think it should be done in this PR, but I'd like to know what you prefer @j-piasecki.

j-piasecki: I'd say it depends on which props those are. `mouseButton` likely doesn't make much sense on a swipeable.

packages/docs-gesture-handler/docs/gestures/use-pan-gesture.mdx
```diff
@@ -209,23 +211,29 @@ When the finger moves outside this range (in points) along X axis and gesture ha
...
 ```ts
 averageTouches: boolean | SharedValue<boolean>;
 ```
 
 Android, by default, will calculate translation values based on the position of the leading pointer (the first one that was placed on the screen). This modifier allows that behavior to be changed to the one that is default on iOS - the averaged position of all active pointers will be used to calculate the translation values.
 
-### enableTrackpadTwoFingerGesture (iOS only)
+<HeaderWithBadge platforms={['iOS']}>
+### enableTrackpadTwoFingerGesture
+</HeaderWithBadge>
```
j-piasecki: This also works on web

packages/docs-gesture-handler/src/components/HeaderWithBadge/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const platformBadge =
+    platform === 'android'
+      ? styles.androidBadge
+      : platform === 'iOS'
+        ? styles.iosBadge
+        : platform === 'web'
+          ? styles.webBadge
+          : {};
+
+  return <div style={{ ...styles.badge, ...platformBadge }}>{platform}</div>;
```
j-piasecki: Can this be case-insensitive? `Android` and `Web` should be capitalized when displayed, and `iOS` should do its own thing.

packages/docs-gesture-handler/docs/components/pressable.mdx
m-bert: Yes, this was web - only (at least it didn't work on android

packages/docs-gesture-handler/docs/components/pressable.mdx
m-bert: Same as above

packages/docs-gesture-handler/docs/components/pressable.mdx
m-bert: Same as above

packages/docs-gesture-handler/docs/components/reanimated_swipeable.mdx
m-bert: Still, I don't think it belongs to this PR. But we can remove it, true.

packages/docs-gesture-handler/docs/gestures/use-tap-gesture.mdx
m-bert: Similar to [Swipeable discussion](https://github.com/software-mansion/react-native-gesture-handler/pull/3975#discussion_r2815391421), I think it should be done in different PR.

m-bert:
> How hard would adding colored dots in the table of contents on the right?

I guess you know the answer 😅 

#3985

packages/docs-gesture-handler/src/components/HeaderWithBadge/index.tsx
m-bert: Here you go: [0cd3c28](https://github.com/software-mansion/react-native-gesture-handler/pull/3975/commits/0cd3c2838c5c22cdbb89e77b7e8c85cdad680ead)

packages/docs-gesture-handler/docs/gestures/use-pan-gesture.mdx
m-bert: True, [e15ee60](https://github.com/software-mansion/react-native-gesture-handler/pull/3975/commits/e15ee602375c395d53c58e251f41f19e9366db03)

PR #3976, opened 2026-02-13
[docs] Redirects

m-bert:
## Description

This PR adds redirects for our docs so that old gesture links will point to new version.

## Test plan

Read docs 🤓

### DIFF
--- a/packages/docs-gesture-handler/redirects.json
+++ b/packages/docs-gesture-handler/redirects.json
@@ -1,40 +1,44 @@
 {
   "redirects": [
     {
-      "to": "/docs/gestures/use-pan-gesture",
+      "to": "/docs/legacy-gestures/pan-gesture",
       "from": "/docs/gestures/pan-gesture"
     },
     {
-      "to": "/docs/gestures/use-tap-gesture",
+      "to": "/docs/legacy-gestures/tap-gesture",
       "from": "/docs/gestures/tap-gesture"
     },
     {
-      "to": "/docs/gestures/use-long-press-gesture",
+      "to": "/docs/legacy-gestures/long-press-gesture",
       "from": "/docs/gestures/long-press-gesture"
     },
     {
-      "to": "/docs/gestures/use-rotation-gesture",
+      "to": "/docs/legacy-gestures/rotation-gesture",
       "from": "/docs/gestures/rotation-gesture"
     },
     {
-      "to": "/docs/gestures/use-pinch-gesture",
+      "to": "/docs/legacy-gestures/pinch-gesture",
       "from": "/docs/gestures/pinch-gesture"
     },
     {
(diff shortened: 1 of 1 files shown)


packages/docs-gesture-handler/docusaurus.config.js
```diff
@@ -7,6 +7,9 @@ const darkCodeTheme = require('./src/theme/CodeBlock/highlighting-dark.js');
 // @ts-check
 const webpack = require('webpack');
 
+const redirectsData = require('./redirects.json');
+const { red } = require('@mui/material/colors');
```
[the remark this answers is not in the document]
m-bert: [643477b](https://github.com/software-mansion/react-native-gesture-handler/pull/3976/commits/643477bd25f1151f706f7eca15c69add2372e4eb)

packages/docs-gesture-handler/redirects.json
m-bert, line 1: If you prefer other form, let me know.

packages/docs-gesture-handler/docs/legacy-gestures/_shared/base-continuous-gesture-callbacks.md
```diff
@@ -0,0 +1,9 @@
+### Callbacks common to all continuous gestures:
```
j-piasecki, line 1: Where do these files come from?

packages/docs-gesture-handler/redirects.json
```diff
@@ -0,0 +1,48 @@
+{
+  "redirects": [
+    {
+      "to": "/docs/gestures/use-pan-gesture",
+      "from": "/docs/gestures/pan-gesture"
```
j-piasecki: Hmm, shouldn't we redirect to 2.x docs when someone is linked to the old api?

packages/docs-gesture-handler/docs/legacy-gestures/_shared/base-continuous-gesture-callbacks.md
m-bert: Copied from versioned docs to add legacy section.

packages/docs-gesture-handler/redirects.json
m-bert: I thought we agreed to link to new API as those are not that different in case of properties. We can redirect to v2, but in that case to `legacy-gestures` section. WDYT?

j-piasecki: I think that's fine, at least for the beta period.

m-bert: Done in [4b515f3](https://github.com/software-mansion/react-native-gesture-handler/pull/3976/commits/4b515f36b596274f095870d2fa3cc936ce8dacff). I've also brought back missing `ForceTouch`

PR #3978, opened 2026-02-13
[docs] Add version badges

m-bert:
## Description

This PR, similarly to #3975, adds badges for versions for the features.

### Darkmode

image

### Lightmode

image

## Test plan

Read docs 🤓


packages/docs-gesture-handler/src/components/HeaderWithBadges/index.tsx
```diff
@@ -18,17 +33,20 @@ export function Badge({ platform }: { platform: 'android' | 'iOS' | 'web' }) {
...
     <div style={styles.container}>
       {children}
 
-      {platforms.map((platform) => (
-        <Badge key={platform} platform={platform} />
+      {platforms?.map((platform) => (
+        <PlatformBadge key={platform} platform={platform} />
       ))}
+
+      {version && <VersionBadge version={version} />}
```
j-piasecki, line 85: IMO, the version badge should come before the platform ones.

We can also go the Android way and have it always to the right

Image

We may also want to default to showing "3.0" if no version is provided to be consistent. So there's no ambiguity about which version something is available from.

m-bert: > I'm not sold on the styling. I don't think the version badge should be that attention-grabbing.

I agree. Android convention looks good, but do we want it as a plain text, or have it in a badge like platform, with with different color (like white badge with black text)?

m-bert: Styles updated in [36b7fe0](https://github.com/software-mansion/react-native-gesture-handler/pull/3978/commits/36b7fe01bf62a4ce9be9746460abf385ede9e0a4). I've also removed unused import from `FlowChart` (been there to steal `useColorMode`)

m-bert: So if styling looks good, the question is where do we want to put them. Props like `disableReanimated` and so on? Or do we wait for something new.

j-piasecki: > So if styling looks good, the question is where do we want to put them. Props like disableReanimated and so on? Or do we wait for something new.

That's a good question - on one hand, I like the way Android does it, so everything is clearly labeled, but on the other, that's a lot of boxes. It could be ok if we reduce it only to the version instead of `Available from ...`, but it's not as clean then.

TL;DR: I don't know

packages/docs-gesture-handler/src/components/HeaderWithBadges/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+const styles = StyleSheet.create({
+  container: {
+    display: 'flex',
+    flexDirection: 'row',
+    gap: 10,
+    marginBottom: 20,
+  },
+
+  badge: {
+    borderRadius: 10,
```
j-piasecki:
```suggestion
    borderRadius: 16,
```

IMO it looks nicer.

packages/docs-gesture-handler/src/components/HeaderWithBadges/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+      {version && <VersionBadge version={version} />}
+    </div>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    display: 'flex',
+    flexDirection: 'row',
```
j-piasecki:
```suggestion
    flexDirection: 'row',
    flexWrap: 'wrap',
```

I think it looks nicer this way on narrow screens.
Image

vs

Image

m-bert: I agree. I also moved it to the left on smaller screens:

image

m-bert: [fbdd4d6](https://github.com/software-mansion/react-native-gesture-handler/pull/3978/commits/fbdd4d6a71412cff6757e614f2c8de1bc5b0fdd9)

m-bert:
> Can you please merge this after the platform badges?

Sure, that's my plan

> I forgot about it yesterday, but when I was checking it, I was experimenting with reducing the font size for the version badge (I think 12px). I'm fine with both, so I'll leave it up to you.

`12px` looks way too small (at least when looking from distance):

image

But I guess your point is to reduce it. To be honest, I'd leave that as it is now.

packages/docs-gesture-handler/src/components/HeaderWithBadges/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return (
+    <div className={clsx(styles.headerWithBadges)}>
+      {children}
+
+      {platforms
+        ?.map((platform) => platform.toLowerCase() as Platform)
+        .sort()
+        .map((platform) => (
+          <PlatformBadge key={platform} platform={platform} />
+        ))}
```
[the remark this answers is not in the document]
m-bert: We want this to be case insensitive

PR #3979, opened 2026-02-16
[ios] Force going through begin in state manager

akwasniewski:
## Description

The state manager was not functioning correctly on iOS because the state change from begin to activate did not go through begin. `handleStateChange` doesn't call `handleGesture` for activate directly as it bypasses gesture interaction system. Instead it relies on the activate gesture being called from reset.

It worked previously, but after commit c6051cae923fff1467313dff8eee2454c74d654c introduced a direct check in `handleGesture` to determine whether it was invoked from reset. As a result, manually handled gestures were never activated.

## Test plan

Tested on the StateManager example.


j-piasecki:
Are you sure this doesn't remove all gesture relations on iOS?

akwasniewski:
> Are you sure this doesn't remove all gesture relations on iOS?

Ok, I trusted xcode, which said nobody calls those functions and they don't overload anything... I assumed they were obsolete, but I see they are indeed overloads from some iOS functions. My example which used gesture relations was also unaffected, neither was pressable but I see the lock example does not work now.

akwasniewski:
> Are you sure this doesn't remove all gesture relations on iOS?

Ok, in a89a631b6309f90399c9d73f93ecadc8da1e08b8 I fixed StateManager by simply forcing states to go through the `begin` state. It still relies on activate being handled in reset, which does not seem clean, but I don't see a better way around it.

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -222,6 +222,10 @@ - (void)setGestureStateSync:(int)state forHandler:(int)handlerTag
     } else if (state == 3) { // CANCELLED
       handler.recognizer.state = RNGHGestureRecognizerStateCancelled;
     } else if (state == 4) { // ACTIVE
+      if (handler.recognizer.state == 0) {
+        // Force going from UNDETERMINED to ACTIVE through BEGAN to preserve the correct state transition flow.
+        [self setGestureStateSync:2 forHandler:handlerTag];
```
j-piasecki:
```suggestion
        handler.recognizer.state = RNGHGestureRecognizerStatePossible;
```

Wouldn't this work? I don't see anything else done for the "2" path, so instead of going through the entire flow of finding the gesture and ensuring it's set up correctly twice, we can just change the state directly.

akwasniewski: It would work, but we still need to call `handleGesture` for begin, I thought that using recursion is cleaner here, but I agree having to find handler again adds unnecessary complexity.

akwasniewski: 68b0164836b90a316e1f38ae6da9f9d1fbf8dabb

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -222,6 +222,11 @@ - (void)setGestureStateSync:(int)state forHandler:(int)handlerTag
     } else if (state == 3) { // CANCELLED
       handler.recognizer.state = RNGHGestureRecognizerStateCancelled;
     } else if (state == 4) { // ACTIVE
+      if (handler.recognizer.state == 0) {
+        // Force going from UNDETERMINED to ACTIVE through BEGAN to preserve the correct state transition flow.
+        handler.recognizer.state = RNGHGestureRecognizerStatePossible;
```
m-bert: Does setting state here work? As far as I remember, recognizers are quite strict when it comes to setting state and doing so outside onTouches* doesn't work.

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -222,6 +222,11 @@ - (void)setGestureStateSync:(int)state forHandler:(int)handlerTag
     } else if (state == 3) { // CANCELLED
       handler.recognizer.state = RNGHGestureRecognizerStateCancelled;
     } else if (state == 4) { // ACTIVE
+      if (handler.recognizer.state == 0) {
```
m-bert: Can we please check for appropriate value from `UIGestureRecognizerState`?

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
akwasniewski: Yeah, it does. This method always relied on setting the state directly.

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
akwasniewski: 149e469973c8308aec1cc4046546eaeceb0645ee.

PR #3855, opened 2025-12-01
[iOS] Fix `onFinalize` callbacks on iOS 26

m-bert:
## Description

This PR aims to fix callback differences between `iOS` 26 and 18. Most of them were already resolved, but there were some issues with `Pan` and `onFinalize` callback. 

Calling `reset` has been removed from `Tap` as it is called automatically by recognizer and leaving it resulted in double callbacks.

> [!NOTE]
> For related changes, see #3740, #3756 and #3849.

## Test plan

<details>
<summary>Tested on the following code:</summary>

```tsx
import { StyleSheet, View, Text } from 'react-native';

import {
  GestureHandlerRootView,
  Gesture,
  GestureDetector,
  GestureType,
} from 'react-native-gesture-handler';

function TestBox({
  gestureType,
  bgColor,
}: {
  gestureType: GestureType;
  bgColor: string;
}) {
  const handlerName = gestureType.handlerName;

  const gesture = gestureType
    .onBegin(() => {
      console.log(`[${handlerName}] onBegin`);
    })
    .onEnd((_e, s) => {
      console.log(`[${handlerName}] onEnd (${s})`);
    })
    .onFinalize((_e, s) => {
      console.log(`[${handlerName}] onFinalize (${s})`);
    })
    .runOnJS(true);

  try {
    // @ts-ignore this is exactly why we have the try-catch block
    gesture.onUpdate(() => {
      console.log(`[${handlerName}] onUpdate`);
    });
  } catch {
    /* empty */
  }

  return (
    <View style={styles.center}>
      <Text>{handlerName}</Text>
      <GestureDetector gesture={gesture}>
        <View style={[styles.box, { backgroundColor: bgColor }]} />
      </GestureDetector>
    </View>
  );
}

export default function App() {
  return (
    <GestureHandlerRootView style={[{ flex: 1, padding: 50 }, styles.center]}>
      <TestBox gestureType={Gesture.Pan()} bgColor="#b58df1" />
      <TestBox gestureType={Gesture.LongPress()} bgColor="#f1a85d" />
      <TestBox gestureType={Gesture.Fling()} bgColor="#5df1a8" />
      <TestBox gestureType={Gesture.Tap()} bgColor="#5d8ef1" />
    </GestureHandlerRootView>
  );
}

const styles = StyleSheet.create({
  center: {
    display: 'flex',
    justifyContent: 'space-around',
    alignItems: 'center',
  },
  box: {
    height: 100,
    width: 100,
    backgroundColor: '#b58df1',
[...]

PR #3983, opened 2026-02-18
[iOS] Add wrapper for `handleGesture:fromReset:`

m-bert:
## Description

`Pinch` and `Rotation` are broken because they fail on [this line](https://github.com/software-mansion/react-native-gesture-handler/blob/0e74233db6a30dee547c3782aefc4f90f0551141/packages/react-native-gesture-handler/apple/RNGestureHandler.mm#L306). This happens because `handleGesture:fromReset` is called automatically via [`action` property](https://github.com/software-mansion/react-native-gesture-handler/blob/0e74233db6a30dee547c3782aefc4f90f0551141/packages/react-native-gesture-handler/apple/Handlers/RNPinchHandler.m#L36), but when UIKit calls it, it only passes `recognizer` as first parameter - `fromReset` is some garbage from memory. It gets treated as truthy value, even though it shouldn't be.

Now, `handleGesture:fromRecognizer` will not be called automatically when recognizer resets its state, so it should be safe to assume that default `fromReset` value is `NO`.

## Test plan

Tested on the **transformations** example and example code from #3855 on both, iOS 18.5 and iOS 26.0


packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
```diff
@@ -138,9 +138,9 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
   return self;
 }
 
-- (void)handleGesture:(NSPanGestureRecognizer *)gestureRecognizer fromReset:(BOOL)fromReset
+- (void)handleGesture:(NSPanGestureRecognizer *)gestureRecognizer
 {
-  [_gestureHandler handleGesture:self fromReset:fromReset];
+  [_gestureHandler handleGesture:self];
```
m-bert, line 143: This just reverts changes from #3855, we will have to look at macos anyway so for now it should be sufficient.

j-piasecki (approved):

PR #3986, opened 2026-02-19
[ios] not calling activation callback on inactive gestures

akwasniewski:
## Description

Gestures always call activation callback upon finalizing, regardless of the previous state. This is intended for normal gestures, however for manually handled gestures we don't want to call it. This PR unifies the behaviour across platforms, GestureManager.finalize() won't call activation callback.

## Test plan

Tested on the state manager example

### DIFF
--- a/apps/common-app/src/legacy/release_tests/gesturizedPressable/testingBase.tsx
+++ b/apps/common-app/src/legacy/release_tests/gesturizedPressable/testingBase.tsx
@@ -3,26 +3,26 @@ import {
   StyleSheet,
   Text,
   View,
-  Pressable,
+  Pressable as RNPressable,
   PressableProps as RNPressableProps,
 } from 'react-native';
 import {
-  LegacyPressable as GesturizedPressable,
+  LegacyPressable as GHPressable,
   LegacyPressableProps as GHPressableProps,
 } from 'react-native-gesture-handler';
 
 const TestingBase = (props: GHPressableProps & RNPressableProps) => (
   <>
-    <GesturizedPressable {...props}>
+    <GHPressable {...props}>
       <View style={styles.textWrapper}>
-        <Text style={styles.text}>Gesturized pressable!</Text>
+        <Text style={styles.text}>RNGH pressable!</Text>
       </View>
-    </GesturizedPressable>
-    <Pressable {...props}>
+    </GHPressable>
+    <RNPressable {...props}>
       <View style={styles.textWrapper}>
-        <Text style={styles.text}>Legacy pressable!</Text>
+        <Text style={styles.text}>RN pressable!</Text>
       </View>
-    </Pressable>
+    </RNPressable>
   </>
 );
 
--- a/apps/common-app/src/legacy/release_tests/nestedPressables/index.tsx
+++ b/apps/common-app/src/legacy/release_tests/nestedPressables/index.tsx
@@ -1,15 +1,12 @@
 import React from 'react';
 import {
-  Pressable as LegacyPressable,
+  Pressable as RNPressable,
   PressableStateCallbackType,
   StyleSheet,
   Text,
   View,
 } from 'react-native';
-import {
-  ScrollView,
-  Pressable as GesturizedPressable,
-} from 'react-native-gesture-handler';
+import { ScrollView, LegacyPressable } from 'react-native-gesture-handler';
 
 export default function Example() {
   return (
@@ -50,21 +47,21 @@ const outerStyle = ({ pressed }: PressableStateCallbackType) => [
 
 function GesturizedBoxes() {
   return (
-    <GesturizedPressable
+    <LegacyPressable
       style={outerStyle}
       testID="outer"
       onPressIn={() => console.log('[outer] onPressIn')}
       onPressOut={() => console.log('[outer] onPressOut')}
       onPress={() => console.log('[outer] onPress')}
       onLongPress={() => console.log('[outer] onLongPress')}>
-      <GesturizedPressable
+      <LegacyPressable
         style={middleStyle}
         testID="middle"
         onPressIn={() => console.log('[middle] onPressIn')}
         onPressOut={() => console.log('[middle] onPressOut')}
         onPress={() => console.log('[middle] onPress')}
         onLongPress={() => console.log('[middle] onLongPress')}>
-        <GesturizedPressable
+        <LegacyPressable
           style={innerStyle}
           testID="inner"
           onPressIn={() => console.log('[inner] onPressIn')}
@@ -72,34 +69,34 @@ function GesturizedBoxes() {
           onPress={() => console.log('[inner] onPress')}
           onLongPress={() => console.log('[inner] onLongPress')}
         />
-      </GesturizedPressable>
-    </GesturizedPressable>
+      </LegacyPressable>
+    </LegacyPressable>
   );
 }
 
 function LegacyBoxes() {
   return (
-    <LegacyPressable
+    <RNPressable
       style={outerStyle}
       onPressIn={() => console.log('[outer] onPressIn')}
       onPressOut={() => console.log('[outer] onPressOut')}
       onPress={() => console.log('[outer] onPress')}
       onLongPress={() => console.log('[outer] onLongPress')}>
-      <LegacyPressable
+      <RNPressable
         style={middleStyle}
         onPressIn={() => console.log('[middle] onPressIn')}
         onPressOut={() => console.log('[middle] onPressOut')}
         onPress={() => console.log('[middle] onPress')}
         onLongPress={() => console.log('[middle] onLongPress')}>
-        <LegacyPressable
+        <RNPressable
           style={innerStyle}
           onPressIn={() => console.log('[inner] onPressIn')}
           onPressOut={() => console.log('[inner] onPressOut')}
           onPress={() => console.log('[inner] onPress')}
           onLongPress={() => console.log('[inner] onLongPress')}
         />
-      </LegacyPressable>
-    </LegacyPressable>
+      </RNPressable>
+    </RNPressable>
   );
 }
 
--- a/apps/common-app/src/legacy/v2_api/pressable/index.tsx
+++ b/apps/common-app/src/legacy/v2_api/pressable/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { StyleSheet, Text, View } from 'react-native';
-import { Pressable } from 'react-native-gesture-handler';
+import { LegacyPressable } from 'react-native-gesture-handler';
 
 const SECTION_RADIUS = 40;
 const BASE_SIZE = 120;
@@ -33,7 +33,7 @@ export default function PressableExample() {
     <View style={{ flex: 1, backgroundColor: 'white' }}>
       <View style={styles.pressRectContainer}>
         <View style={styles.hitRectContainer}>
-          <Pressable
+          <LegacyPressable
             style={({ pressed }) =>
               pressed ? styles.highlight : styles.pressable
             }
@@ -48,7 +48,7 @@ export default function PressableExample() {
             <View style={styles.textWrapper}>
               <Text style={styles.text}>Pressable!</Text>
             </View>
-          </Pressable>
+          </LegacyPressable>
           <Text style={styles.rectText}>Hit Rect</Text>
         </View>
         <Text style={styles.rectText}>Press Rect</Text>
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -644,11 +644,22 @@ open class GestureHandler {
     onStateChange(newState, oldState)
   }
 
-  fun wantsEvent(event: MotionEvent): Boolean = isEnabled &&
-    state != STATE_FAILED &&
-    state != STATE_CANCELLED &&
-    state != STATE_END &&
-    isTrackingPointer(event.getPointerId(event.actionIndex))
+  fun wantsEvent(event: MotionEvent): Boolean {
+    if (!isEnabled || state == STATE_FAILED || state == STATE_CANCELLED || state == STATE_END) {
+      return false
+    }
+
+    if (event.actionMasked == MotionEvent.ACTION_MOVE) {
+      for (i in 0 until event.pointerCount) {
+        if (isTrackingPointer(event.getPointerId(i))) {
+          return true
+        }
+      }
+      return false
+    } else {
+      return isTrackingPointer(event.getPointerId(event.actionIndex))
+    }
+  }
 
   open fun shouldRequireToWaitForFailure(handler: GestureHandler): Boolean {
     if (handler === this) {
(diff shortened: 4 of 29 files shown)


packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -332,10 +339,17 @@ - (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromRese
...
 }
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer inState:(RNGestureHandlerState)state
+{
+  [self handleGesture:recognizer inState:state manualActivation:NO];
+}
+
+- (void)handleGesture:(UIGestureRecognizer *)recognizer
+              inState:(RNGestureHandlerState)state
+     manualActivation:(Boolean)manualActivation
```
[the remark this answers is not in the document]
akwasniewski: c7d67aa3203fd421f935a9493d7bdd86dcf35d5f

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -391,7 +413,9 @@ - (void)sendEventsInState:(RNGestureHandlerState)state
       static uint16_t nextEventCoalescingKey = 0;
       self->_eventCoalescingKey = nextEventCoalescingKey++;
 
-    } else if (state == RNGestureHandlerStateEnd && _lastState != RNGestureHandlerStateActive && !_manualActivation) {
+    } else if (
+        state == RNGestureHandlerStateEnd && _lastState != RNGestureHandlerStateActive && !manualActivation &&
+        !_manualActivation) {
```
m-bert: 2x "manualActivation" is quite confusing, I'd rename it.

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
```diff
@@ -140,7 +140,14 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
 
 - (void)handleGesture:(NSPanGestureRecognizer *)gestureRecognizer fromReset:(BOOL)fromReset
 {
-  [_gestureHandler handleGesture:self fromReset:fromReset];
+  [_gestureHandler handleGesture:self fromReset:fromReset manualActivation:NO];
```
m-bert: Other gestures call this on `self` instead of `_gestureHandler`.

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
```diff
@@ -138,9 +138,16 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
   return self;
 }
 
+- (void)handleGesture:(NSPanGestureRecognizer *)gestureRecognizer fromReset:(BOOL)fromReset
+{
+  [_gestureHandler handleGesture:self fromReset:fromReset manualActivation:NO];
+}
```
j-piasecki: Is this override needed at all? It doesn't change anything compared to the base implementation

packages/react-native-gesture-handler/apple/Handlers/RNLongPressHandler.m
```diff
@@ -49,22 +53,23 @@ @implementation RNBetterLongPressGestureRecognizer {
...
   return self;
 }
 
-- (void)handleGesture:(UIGestureRecognizer *)recognizer
+- (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromReset
 {
-  previousTime = CACurrentMediaTime();
-  [_gestureHandler handleGesture:recognizer fromReset:NO];
+  [self handleGesture:recognizer fromReset:fromReset manualActivation:NO];
 }
```
j-piasecki, line 61: Same

packages/react-native-gesture-handler/apple/Handlers/RNPinchHandler.m
```diff
@@ -42,8 +42,14 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
 #endif
   return self;
 }
+- (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromReset
+{
+  [self handleGesture:recognizer fromReset:fromReset manualActivation:NO];
+}
```
j-piasecki, line 49: Same

packages/react-native-gesture-handler/apple/Handlers/RNRotationHandler.m
```diff
@@ -41,12 +41,19 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
   return self;
 }
 
+- (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromReset
+{
+  [self handleGesture:recognizer fromReset:fromReset manualActivation:NO];
+}
```
j-piasecki, line 47: Same

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -399,7 +413,9 @@ - (void)sendEventsInState:(RNGestureHandlerState)state
...
```
j-piasecki, line 440: What's the end result here? `onFinalize` is invoked with `didSucceed: true` without going through `activate/deactivate`?

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
akwasniewski: Yes we need the default value one so that the selector works correctly (as in https://github.com/software-mansion/react-native-gesture-handler/pull/3983) and we need to overload the one with all the arguements so that when called from stateManager it initializes the handler specific values (like previousTime in LongPress or scale in pinch).

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
akwasniewski: Yes, this unifies the behaviour across platforms. Previously we forced calling activation callback. Do we want to handle it differently?

akwasniewski: I think if the user called `onFinalize`, rather than `onFail` he wants to have the `onSuccess` flag set to true (for whaterver reason), but if he wanted the activation callback he would have called StateManager.activate directly.

m-bert: What do you mean by

> I think if the user called `onFinalize`, rather than `onFail`

I don't recall having `onFail` in that context.

akwasniewski: Sorry, I meant when they called  GestureStateManager.finalize() rather than GestureStateManager,fail()

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
j-piasecki: In the PR you linked, the base one (`(void)handleGesture:(nonnull id)recognizer;`) is implemented once in the base class. Here, handlers have both - the base one, and the extended one - defined. The base one is exactly the same as the implementation in `GestureHandler`. Hover doesn't seem to override either, and I assume the selector works there, no?

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
akwasniewski: renamed toi `fromManual` in 0a7f0d7584a38ed1beff071d8659b18b29793e0c

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
akwasniewski: Removed overload from fling alltogether and fix similar issue in longpress https://github.com/software-mansion/react-native-gesture-handler/pull/3986/changes/8be794c564531de23e9cc7861fac62ec60226c27

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
akwasniewski: ah right, I assume the override in Fling was used at some, but not anumore. Removed altogether in 1246b9dc0a016ed711052c6db2557a09453d43d6. The other overrides are needed though.

packages/react-native-gesture-handler/apple/Handlers/RNLongPressHandler.m
```diff
@@ -57,24 +61,25 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer
 {
-  previousTime = CACurrentMediaTime();
-  [_gestureHandler handleGesture:recognizer fromReset:NO];
+  [_gestureHandler handleGesture:recognizer fromReset:NO manualActivation:NO];
```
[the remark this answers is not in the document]
akwasniewski: https://github.com/software-mansion/react-native-gesture-handler/pull/3986/changes/8be794c564531de23e9cc7861fac62ec60226c27

packages/react-native-gesture-handler/apple/RNGestureHandler.h
```diff
@@ -97,7 +97,9 @@
 - (void)updateRelations:(nonnull NSDictionary *)relations;
 - (void)handleGesture:(nonnull id)recognizer;
 - (void)handleGesture:(nonnull id)recognizer fromReset:(BOOL)fromReset;
+- (void)handleGesture:(nonnull id)recognizer fromReset:(BOOL)fromReset fromManual:(BOOL)fromManual;
 - (void)handleGesture:(nonnull id)recognizer inState:(RNGestureHandlerState)state;
+- (void)handleGesture:(nonnull id)recognizer inState:(RNGestureHandlerState)state fromManual:(BOOL)fromManual;
```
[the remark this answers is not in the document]
akwasniewski: @m-bert what do you think?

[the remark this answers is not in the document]
m-bert: I agree with @copilot , my first thought was `fromManualStateChange`, but second one is also ok.

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
j-piasecki: Why are they needed? They don't differ from the default implementation.

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
j-piasecki: Is `didSucceed: true` consistent across platforms here?

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
akwasniewski: Without them corresponding gestures crash on activation

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -404,7 +429,12 @@ - (void)sendEventsInState:(RNGestureHandlerState)state
       static uint16_t nextEventCoalescingKey = 0;
       self->_eventCoalescingKey = nextEventCoalescingKey++;
```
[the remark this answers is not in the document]
akwasniewski: be4c295db0209779ebaa1a8772c845816293b4ac

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
```diff
@@ -138,11 +138,6 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
...
 }
 
-- (void)handleGesture:(NSPanGestureRecognizer *)gestureRecognizer
-{
-  [_gestureHandler handleGesture:self];
-}
-
 - (void)mouseDown:(NSEvent *)event
 {
   [super mouseDown:event];
```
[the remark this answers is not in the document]
akwasniewski: 4e353a1de4df4fd7c20a9a5eca79edc1e04d5d40

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
akwasniewski: It's currently s the same on web, it's false on android. I will change it on android as well

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
j-piasecki: Ok, i see - this method is defined in the recognizer not the handler. I'm not sure whether that's right - can you try moving them to the handler?

akwasniewski: Correct me if I am wrong, but it's by design. In those few gestures it is defined in the recognizer (rather than the handler) as it sets some recognizer fields.

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
akwasniewski: I just noticed that android finalize works yet again differently than on other platforms, I'll have to take a good look at it, thus android changes will be done in a seperate pr

akwasniewski: PR which syncs android https://github.com/software-mansion/react-native-gesture-handler/pull/4011. The logic is in fact the same, the didSucceed flag is set to true there, however there was an issue that canceled manually activated gestures.

packages/react-native-gesture-handler/apple/Handlers/RNFlingHandler.m
j-piasecki: Possibly, but it's confusing as hell when they share the exact same definitions and handler/recognizer is in the same file.

Can you look into refactoring that (after this PR)?

j-piasecki (approved):

packages/react-native-gesture-handler/apple/Handlers/RNPinchHandler.m
```diff
@@ -44,6 +44,11 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
 }
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer
+{
+  [self handleGesture:recognizer fromReset:NO fromManualStateChange:NO];
+}
+
+- (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromReset fromManualStateChange:(BOOL)fromManualStateChange
```
m-bert, line 51: Why? Can't we do the same in `handleGesture`?

packages/react-native-gesture-handler/apple/Handlers/RNRotationHandler.m
```diff
@@ -42,11 +42,16 @@ - (id)initWithGestureHandler:(RNGestureHandler *)gestureHandler
 }
 
 - (void)handleGesture:(UIGestureRecognizer *)recognizer
+{
+  [self handleGesture:recognizer fromReset:NO fromManualStateChange:NO];
+}
+
+- (void)handleGesture:(UIGestureRecognizer *)recognizer fromReset:(BOOL)fromReset fromManualStateChange:(BOOL)fromManualStateChange
```
m-bert, line 49: Same as above

packages/react-native-gesture-handler/apple/Handlers/RNPinchHandler.m
m-bert: Okay, I've just read [this discussion](https://github.com/software-mansion/react-native-gesture-handler/pull/3986#discussion_r2839434800)

m-bert (approved):

PR #3991, opened 2026-02-23
[iOS] Distinguish between mouse and stylus when hovering

### LINKED ISSUE #3977: [iOS] Hover gesture reports wrong pointerType
### Description

The hover gesture reports the wrong type of pointer.

The eventExtraData method of the hover handler uses `UITouchTypePencil` instead of the RNGH internal `RNGestureHandlerStylus`, which causes the enum mapping to become MOUSE instead.

https://github.com/software-mansion/react-native-gesture-handler/blob/v2.30.0/packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m#L160

The same method also causes the pointer type to be reported as the same, whether using a stylus or a mouse. There doesn't seem to be a clear way of separating the types on iOS, but checking if `zOffset` property of `UIHoverGestureRecognizer` is `> 0` might suffice? 

https://developer.apple.com/documentation/uikit/uihovergesturerecognizer/4098402-zoffset?language=objc

### Steps to reproduce

1. Open Snack reproducer on an iOS device capable of stylus hover.
2. Hover salmon coloured square with Stylus (e.g. Apple Pencil Pro).
3. Tap the same area with the same Stylus.
4. Notice Tap Pointer Type label indicates STYLUS, while Hover Pointer Type indicates MOUSE.

https://snack.expo.dev/@andreashogstromnavigraph/rngh-hover-pointertype

### Gesture Handler version

2.29.0

### React Native version

0.83.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

New Architecture (Fabric)

### Build type

None

### Device

Real device

### Device model

iPad Pro 13-inch (M4)

### Acknowledgements

Yes

AndreasHogstrom:
## Description

Improves the hover gesture on iOS by distinguishing between a mouse and a stylus, if possible.

With these changes the handler will now assume a mouse (just like previously with the enum bug), but if `zOffset` is available (iOS 16.1) and it is over 0.0 it will instead be recognised as styus.

~*Maybe it makes more sense to invert the logic and assume a mouse as default, and stylus if > 0. Let me know what you think.*~

See #3977 for more context.

Fixes #3977 

## Test plan

Here's a small test page that can be added to the basic-example app for testing;
```jsx
import {
  GestureDetector,
  GestureHandlerRootView,
  useHoverGesture,
  useSimultaneousGestures,
  useTapGesture,
import { runOnJS } from 'react-native-worklets';

const POINTER_TYPES: Record<number, string> = {
  0: "TOUCH",
  1: "STYLUS",
  2: "MOUSE",
  3: "KEY",
  4: "OTHER",
}

export default function Hover() {
  const [hoverPointerType, setHoverPointerType] = useState(-1);
  const [tapPointerType, setTapPointerType] = useState(-1);
  const hoverGesture = useHoverGesture({
    onUpdate: (event) => {
      runOnJS(setHoverPointerType)(event.pointerType);
    },
  });
  const tapGesture = useTapGesture({
    onFinalize: (event) => {
      runOnJS(setTapPointerType)(event.pointerType);
    },
  });
  const gesture = useSimultaneousGestures(hoverGesture, tapGesture);
  return (
    <GestureHandlerRootView
      style={{
        flex: 1,
        backgroundColor: 'white',
        justifyContent: 'center',
        alignItems: 'center',
      }}>
      <Text>Hover pointer type: {POINTER_TYPES[hoverPointerType] ?? "N/A"}</Text>
      <Text>Tap pointer type: {POINTER_TYPES[tapPointerType] ?? "N/A"}</Text>
        <View
          style={{ width: 200, height: 200, backgroundColor: 'salmon' }}></View>
      </GestureDetector>
    </GestureHandlerRootView>
  );
}
```

Mouse is testable in the iPad simulator by enabling `I/O -> Input -> Send Pointer to Device` in the top menus.
Mouse is also testable on a real iPad by Linking keyboard and mouse to it from a mac in the Displays settings.
Apple Pencil is only testable with an actual pencil on a real iPad.

### DIFF
--- a/packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
+++ b/packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
@@ -193,6 +193,7 @@ API_AVAILABLE(ios(13.4))
 {
   if ((self = [super initWithTag:tag])) {
     _recognizer = [NSGestureRecognizer alloc];
+    _pointerType = RNGestureHandlerMouse;
   }
 
   return self;


akwasniewski:
Looks good! Though I think the logic should be inverted and assume default as mouse for ios<16.1. Due to the bug where we returned `UITouchTypePencil` rather than `RNGestureHandlerStylus`, mouse pointer used to be the de facto default. I think it's most logical to keep mouse as default. Moreover using hover with a mouse is probably more common than with a stylus.

AndreasHogstrom:
> Looks good! Though I think the logic should be inverted and assume default as mouse for ios<16.1. Due to the bug where we returned `UITouchTypePencil` rather than `RNGestureHandlerStylus`, mouse pointer used to be the de facto default. I think it's most logical to keep mouse as default. Moreover using hover with a mouse is probably more common than with a stylus.

Sounds good to me. Flipped it around.

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
```diff
@@ -153,11 +153,21 @@ - (void)updateConfig:(NSDictionary *)config
 #endif
 }
 
-- (RNGestureHandlerEventExtraData *)eventExtraData:(UIGestureRecognizer *)recognizer
+- (RNGestureHandlerEventExtraData *)eventExtraData:(UIHoverGestureRecognizer *)recognizer
 {
+  RNGestureHandlerPointerType pointerType = RNGestureHandlerMouse;
+
+#if CHECK_TARGET(16_1)
+  if (@available(iOS 16.1, *)) {
```
m-bert: Why this checks for `16.1`, while other checks are for `13.4`?

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
```diff
@@ -153,11 +153,21 @@ - (void)updateConfig:(NSDictionary *)config
 #endif
 }
 
-- (RNGestureHandlerEventExtraData *)eventExtraData:(UIGestureRecognizer *)recognizer
+- (RNGestureHandlerEventExtraData *)eventExtraData:(UIHoverGestureRecognizer *)recognizer
 {
+  RNGestureHandlerPointerType pointerType = RNGestureHandlerMouse;
```
m-bert: Handlers already have [`pointerType`](https://github.com/software-mansion/react-native-gesture-handler/blob/2e34b85ec14d5725b8e8873288a105ccc850bcc4/packages/react-native-gesture-handler/apple/RNGestureHandler.h#L67), along with [`setCurrentPointerType`](https://github.com/software-mansion/react-native-gesture-handler/blob/2e34b85ec14d5725b8e8873288a105ccc850bcc4/packages/react-native-gesture-handler/apple/RNGestureHandler.h#L124). Now, I know that it accepts event, but maybe if we can't use it we could change `pointerType` somewhere inside recognizer and reset it when gesture ends?

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
AndreasHogstrom: `UIHoverGestureRecognizer.zOffset` is only available in 16.1+.

Ref: https://developer.apple.com/documentation/uikit/uihovergesturerecognizer/4098402-zoffset

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
AndreasHogstrom: That's a good point and it would make sense to try to conform. Can't do it exactly the same way as some of the other gestures due to the hover one being a bit special it seems.

I tried with overriding the `handleGesture` method in `RNHoverGestureHandler` to be able to detect and set the pointer type when the gesture begins. And as you said, the `setCurrentPointerType` method expects an UIEvent (with touches available), which we do not have here. So instead we set the `_pointerType` directly.

```objc
- (void)handleGesture:(UIHoverGestureRecognizer *)recognizer
{
  if (recognizer.state == UIGestureRecognizerStateBegan) {
    _pointerType = [self detectPointerType:recognizer];
  }

  [super handleGesture:recognizer];
}
```

The detection method is the same as before;
```objc
- (RNGestureHandlerPointerType)detectPointerType:(UIHoverGestureRecognizer *)recognizer
{
  RNGestureHandlerPointerType pointerType = RNGestureHandlerMouse;

#if CHECK_TARGET(16_1)
  if (@available(iOS 16.1, *)) {
    if (recognizer.zOffset > 0.0) {
      pointerType = RNGestureHandlerStylus;
    }
  }
#endif

  return pointerType;
}
```

And then obviously in the `eventExtraData` method we only need to update the argument to send `_pointerType`.

```objc
// ...
withPointerType:_pointerType];
// ...
```

I will test a bit more during the day when I have some time.


Any feedback on this?

m-bert: ### tl;dr

We could use `handleGesture` along with `setCurrentPointerType`. If you're willing to do so, you can find details below. If you don't have time (or something else) let me know, I'll introduce necessary changes into this PR 😅 

### `handleGesture`

> I tried with overriding the `handleGesture` method in `RNHoverGestureHandler` to be able to detect and set the pointer type when the gesture begins.

In that case you'd have to change `target` when initializing recognizer:

```diff
- if ((self = [super initWithTarget:_gestureHandler action:@selector(handleGesture:)])) {
+ if ((self = [super initWithTarget:self action:@selector(handleGesture:)])) {
```

Also if you decide to handle it in `handleGesture` it would be nice to check recognizer's state to only set it once, when gesture begins.

```obj-c
if(self.state == UIGestureRecognizerStateBegan){
  [_gestureHandler setCurrentPointerType:nil];
}
```

And remember to call this method on `_gestureHandler`:

```obj-c
[_gestureHandler handleGesture:self];
```

### `setCurrentPointerType`

The other thing is, `_pointerType` is not easily accessible from recognizer. What we could do is to handle this inside already existing `setCurrentPointerType`. Since we do not need `UIEvent`, we can make it `nullable`. Also handlers' have access to their recognizers, so we could do something like this:

```obj-c
if([self.recognizer isKindOfClass:[UIHoverGestureRecognizer class]]) {
  // _pointerType = ...
  return;
}
```

and for safety we could assume default value if for some reason we enter this function from other recognizer without event:

```obj-c
if(event == nil){
  _pointerType = RNGestureHandlerTouch;
  return;
}
```

AndreasHogstrom: I pushed a new commit with this approach in mind.

> In that case you'd have to change target when initializing recognizer:

I don't think this is the case, self would be the recognizer here, right? The handleGesture method is on the handler. Which is where i've added an override for handleGesture, setting the pointerType (now via setCurrentPointerType instead of directly). Well, at least it works, maybe I'm missing another detail here?

But I must say it feels a bit off to have hover specific logic in the generic GestureHandler instead of inside the HoverGesture file. Welcoming any refactoring suggestions.

@m-bert Feel free to make further modifications if you have any good ideas, and I'll test it on my devices.

m-bert: > I don't think this is the case, self would be the recognizer here, right?

True, I though about adding `handleGesture` to recognizer like we do in other handlers 😅 I've done it in [4c0271e](https://github.com/software-mansion/react-native-gesture-handler/pull/3991/commits/4c0271ebae2dea8b597f0a42511980c5b0f1cab2).  

> But I must say it feels a bit off to have hover specific logic in the generic GestureHandler instead of inside the HoverGesture file. Welcoming any refactoring suggestions.

I agree. I've moved it to Hover in [1b69398](https://github.com/software-mansion/react-native-gesture-handler/pull/3991/commits/1b693987c47341d1cd1d653b7caa9d643468969e).

It would be great if you could confirm that it works on your end 😅

AndreasHogstrom: Seems fine to me!

Tested with a apple pencil pro, connected mouse, simulated mouse and touch. All reporting the expected pointer type :)

Do you need anything else from me at this stage?

m-bert: Glad to hear that!

> Do you need anything else from me at this stage?

No, thank you very much for your time ❤️

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -207,6 +207,11 @@ - (void)setEnabled:(BOOL)enabled
 #if !TARGET_OS_OSX
 - (void)setCurrentPointerType:(UIEvent *)event
 {
+  if (event == nil) {
+    _pointerType = RNGestureHandlerTouch;
+    return;
+  }
```
j-piasecki: This is a bit misleading. I'd expect `setCurrentPointerType` to accept `RNGestureHandlerPointerType`, and another method - `setCurrentPointerTypeForEvent` that would accept an event.

Then the one accepting event can stay as nonnull, and call the other one. The hover gesture can then set whatever pointer type it needs directly.

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
```diff
@@ -40,13 +40,22 @@ @implementation RNBetterHoverGestureRecognizer {
...
     _gestureHandler = gestureHandler;
     _hoverEffect = RNGestureHandlerHoverEffectNone;
   }
   return self;
 }
 
+- (void)handleGesture:(UIHoverGestureRecognizer *)recognizer
+{
+  if (recognizer.state == UIGestureRecognizerStateBegan) {
+    [_gestureHandler setCurrentPointerType:nil];
```
j-piasecki: Will this work on macOS?

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
m-bert: Or actually, there might be. When I do some changes based on reviews, I'll ask you again to test before merging if we didn't break anything. Is that ok? 😅

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
m-bert: Let me know what you think: [19010f2](https://github.com/software-mansion/react-native-gesture-handler/pull/3991/commits/19010f2418dd667fad153d59f25fe2b68b9221c1)

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
m-bert: Works, but I've changed default pointer type to mouse: [b102e96](https://github.com/software-mansion/react-native-gesture-handler/pull/3991/commits/b102e96ff3867059dc9da035fb197797267d9944)

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
m-bert: Could you please check it one more time @AndreasHogstrom? I think now we are done with changes 😅

AndreasHogstrom: Works perfectly fine on all my devices and hover-capable peripherals. So it at least covers my use cases :)

Thanks a lot for input and changes :+1:

PR #3994, opened 2026-02-24
Use shared release action

j-piasecki:
## Description

Removes release scripts and related tests and uses [shared publish action](https://github.com/software-mansion-labs/npm-package-publish), which uses the same scripts.

## Test plan

Tested it on a fork:
- https://github.com/j-piasecki/react-native-gesture-handler/actions/runs/22343887406/job/64653421948
- https://github.com/j-piasecki/react-native-gesture-handler/actions/runs/22344178991/job/64654420340


.github/workflows/publish-release.yml
```diff
@@ -42,17 +42,35 @@ jobs:
...
       - name: Publish automatic nightly release
         if: ${{ github.event_name == 'schedule' }}
-        uses: ./.github/actions/publish-npm-package
+        uses: software-mansion-labs/npm-package-publish@main
         with:
+          package-name: 'react-native-gesture-handler'
+          package-json-path: 'packages/react-native-gesture-handler/package.json'
+          install-dependencies-command: 'yarn install --immutable'
           release-type: 'nightly'
           dry-run: false
```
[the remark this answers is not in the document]
j-piasecki: It does

.github/workflows/publish-release.yml
```diff
@@ -42,17 +42,35 @@ jobs:
...
+          registry-url: https://registry.npmjs.org/
+
       - name: Publish manual release
         if: ${{ github.event_name == 'workflow_dispatch' }}
-        uses: ./.github/actions/publish-npm-package
+        uses: software-mansion-labs/npm-package-publish@main
         with:
+          package-name: 'react-native-gesture-handler'
+          package-json-path: 'packages/react-native-gesture-handler/package.json'
+          install-dependencies-command: 'yarn install --immutable'
```
m-bert, line 63:
```suggestion
```

We don't have to specify this command as it is default value, right?

.github/workflows/publish-release.yml
```diff
@@ -42,17 +42,35 @@ jobs:
...
           dry-run: ${{ inputs.dry-run }}
 
       - name: Publish automatic nightly release
         if: ${{ github.event_name == 'schedule' }}
-        uses: ./.github/actions/publish-npm-package
+        uses: software-mansion-labs/npm-package-publish@main
         with:
+          package-name: 'react-native-gesture-handler'
+          package-json-path: 'packages/react-native-gesture-handler/package.json'
+          install-dependencies-command: 'yarn install --immutable'
```
m-bert, line 74: Same here

.github/workflows/publish-release.yml
j-piasecki: It is the default, but I'd rather have it defined here in case it changes upstream in the future.

.github/workflows/publish-release.yml
j-piasecki: Same

PR #3997, opened 2026-02-25
Fix reanimated hot reload

akwasniewski:
## Description

On hot reload, with reanimated turned on. The first time hot reloads happens when the worklet callback did not change, but the file it was defined it has changed the gesture stops working. The next reload fixes the gesture and all subsequent reloads don't break it.

The reason was that callbacks object was rebuild on first hot reload while worklets were not, which caused the worklets to reference callback which no longer existed. This PR fixes the issue by always rebuilding when the callbacks/handlers object has changed.

## Test plan

Tested on the same code [this PR](https://github.com/software-mansion/react-native-gesture-handler/pull/3984), only commenting the first line.


j-piasecki:
I'm not a fan of adding workarounds like that unless necessary. You mentioned some time ago that you found a bug in worklets. Did it lead anywhere?

akwasniewski:
> I'm not a fan of adding workarounds like that unless necessary. You mentioned some time ago that you found a bug in worklets. Did it lead anywhere?

I've created a [PR](https://github.com/software-mansion/react-native-reanimated/pull/9075) to Reanimated with a bugfix in `useHandler` from their side. However, it had been currently blocked as they will want to rewrite it entirely.

j-piasecki:
I'll close this due to https://github.com/software-mansion/react-native-gesture-handler/pull/4096

PR #4001, opened 2026-02-26
Revert "Export new `createNativeWrapper` (#3971)"

m-bert:
This reverts commit bdf353db24013d30e2b4baa5ab0dd17556c015de.

## Description

We decided that it will be better not to export new `createNativeWrapper` as it might be confusing when to use it. 

## Test plan

Try to import `createNativeWrapper`
Read docs 🤓

### DIFF
--- a/packages/docs-gesture-handler/docs/guides/upgrading-to-3.mdx
+++ b/packages/docs-gesture-handler/docs/guides/upgrading-to-3.mdx
@@ -283,7 +283,7 @@ Other components have also been internally rewritten using the new hook API but 
 
 ### createNativeWrapper
 
-`createNativeWrapper` is now exported as `legacy_createNativeWrapper`.
+`createNativeWrapper` is deprecated and it is now exported as `legacy_createNativeWrapper`.
 
 ## Replaced types
 


packages/react-native-gesture-handler/src/index.ts
```diff
@@ -52,7 +52,7 @@ export { PanGestureHandler } from './handlers/PanGestureHandler';
 export { PinchGestureHandler } from './handlers/PinchGestureHandler';
 export { RotationGestureHandler } from './handlers/RotationGestureHandler';
 export { FlingGestureHandler } from './handlers/FlingGestureHandler';
-export { default as legacy_createNativeWrapper } from './handlers/createNativeWrapper';
+export { default as createNativeWrapper } from './handlers/createNativeWrapper';
```
m-bert: Even though we're still in beta, I can see the reason why we would want to keep `legacy` prefix here. WDYT @j-piasecki @akwasniewski?

j-piasecki: Fine by me, we should mention that in the migration guide and llm skill

m-bert: [aa95b97](https://github.com/software-mansion/react-native-gesture-handler/pull/4001/commits/aa95b97ca60a30dd35d4059c42b745702abedab4), let me know what you think.

j-piasecki: Do you see any reason agains deprecating it?

m-bert: No, deprecated in https://github.com/software-mansion/react-native-gesture-handler/pull/4001/changes/ca857a7ed3b83ca229a6aeb8d5f1f7a8d179eebf.

Though this is how it looks like in vsc:

image

m-bert: I've also added it to docs: [2c6a3c9](https://github.com/software-mansion/react-native-gesture-handler/pull/4001/commits/2c6a3c937408962f77736b4bbf369eecf0698b8b)

PR #4011, opened 2026-03-03
[android] dont cancel manually activated gestures

akwasniewski:
## Description

Manually handled gestures were being cancelled when another gesture received an event. This PR fixes the issue.

## Test plan

Tested on slightly modified version of StateManager example:

<details>

```tsx

import { COLORS, commonStyles } from '../../../common';
import React from 'react';
import { LongPressGestureConfig } from 'react-native-gesture-handler';
import {
  GestureHandlerRootView,
  GestureDetector,
  useLongPressGesture,
  GestureStateManager,
  LongPressGesture,
import Animated, {
  useAnimatedStyle,
  useSharedValue,
  withTiming,

  const isActivated = [
    useSharedValue(0),
    useSharedValue(0),
    useSharedValue(0),
    useSharedValue(0),
  ];
  const gestures: LongPressGesture[] = [];

  const createGestureConfig = (index: number): LongPressGestureConfig => ({
    onActivate: () => {
      'worklet';
      isActivated[index].value = 1;
      console.log(`Box ${index}: long pressed`);

      const nextIndex = index + 1;
      if (nextIndex < gestures.length) {
        const nextGesture = gestures[nextIndex];
        if (nextGesture) {
          GestureStateManager.activate(nextGesture.handlerTag);
        }
      }
    },
    onFinalize: (_, success) => {
      'worklet';
      isActivated[index].value = 0;
      console.log(`Box ${index}: finalised gesture, `, success);
      const nextIndex = index + 1;
      if (nextIndex < gestures.length) {
        const nextGesture = gestures[nextIndex];
        if (nextGesture) {
          GestureStateManager.deactivate(nextGesture.handlerTag);
        }
      }
    },
    disableReanimated: true,
  });

  const g0 = useLongPressGesture(createGestureConfig(0));
  const g1 = useLongPressGesture(createGestureConfig(1));
  const g2 = useLongPressGesture(createGestureConfig(2));
  const g3 = useLongPressGesture(createGestureConfig(3));

  gestures[0] = g0;
  gestures[1] = g1;
  gestures[2] = g2;
  gestures[3] = g3;

  const colors = [COLORS.PURPLE, COLORS.NAVY, COLORS.GREEN, COLORS.RED];

  function Box({ index }: { index: number }) {
      opacity: isActivated[index].value === 1 ? 0.5 : 1,
      transform: [
        { scale: withTiming(isActivated[index].value === 1 ? 0.95 : 1) },
      ],
    }));

    return (
      <GestureDetector gesture={gestures[index]}>
        <Animated.View
          style={[
            commonStyles.box,
            { backgroundColor: colors[index] },
            animatedStyle,
          ]}
        />
      </GestureDetector>
    );
  }
  return (
    <GestureHandlerRootView>
      <View style={commonStyles.centerView}>
        <Box index={0} />
        <Box index={1} />
        <Box index={2} />
        <Box index={3} />
      </View>
    </GestureHandlerRootView>
  );
}

```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
@@ -287,7 +287,7 @@ class GestureHandlerOrchestrator(
   }
 
   private fun deliverEventToGestureHandler(handler: GestureHandler, sourceEvent: MotionEvent) {
-    if (!isViewAttachedUnderWrapper(handler.view)) {
+    if (!isViewAttachedUnderWrapper(handler.view ?: handler.hostDetectorView)) {
       handler.cancel()
       return
     }


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -160,8 +160,8 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
       }
 
       when (newState) {
-        GestureHandler.STATE_ACTIVE -> handler.activate(force = true)
-        GestureHandler.STATE_BEGAN -> handler.begin()
+        GestureHandler.STATE_ACTIVE -> handler.activate(force = true, fromManualStateChange = true)
+        GestureHandler.STATE_BEGAN -> handler.begin(fromManualStateChange = true)
```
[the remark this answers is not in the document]
akwasniewski: merci, https://github.com/software-mansion/react-native-gesture-handler/pull/4011/changes/6f6dfeb8c0f6184c2f14514b63569e57a7952ff3

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -287,7 +295,7 @@ class GestureHandlerOrchestrator(
   }
 
   private fun deliverEventToGestureHandler(handler: GestureHandler, sourceEvent: MotionEvent) {
-    if (!isViewAttachedUnderWrapper(handler.view)) {
+    if (!isViewAttachedUnderWrapper(handler.view) && !handler.isHandledManually) {
```
j-piasecki: Generally, I'd rather track it to the source - why is `handler.view` `null` here? Should it be?

This will only affect the v3 API, where we will always know where the handler ends up in the hierarchy. Something like this would be sufficient, no?
```suggestion
    if (!isViewAttachedUnderWrapper(handler.view ?: handler.hostDetectorView)) {
```

akwasniewski:
>You probably noticed, but the handlers activated manually in the repro never reset on their own. I haven't looked into this, but that seems like something we'd want to address.

That's actually intended, this unifies the behaviour across platforms. If we wanted them to reset on their own, when would they do it? I think if it is handled manually it should wait for the user to explicitly cancel or finalise the gesture. Otherwise what is the point of allowing manual cancel/finalise at all if it's gonna be reset anyway?

j-piasecki:
> If we wanted them to reset on their own, when would they do it?

I was thinking when all pointers leave the screen.

> I think if it is handled manually it should wait for the user to explicitly cancel or finalise the gesture.

If you use the state manager to activate a gesture that's currently in progress (began, but not active yet, e.g., you manually activate a pan gesture before it would do so by itself), it will behave as a standard gesture afterwards (meeting all failure criteria).

> Otherwise what is the point of allowing manual cancel/finalise at all if it's gonna be reset anyway?

So you can stop the gesture when certain criteria are met (moved too much, gesture took too much time, animation stopped in another part of the screen, ...)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
akwasniewski: 3b7ae588cec65a9cc8f1e1d428b9ff92ee614b36. Ok, thanks for noticing this. If we want the state manager finalise to work how you described in the other comment, then this fix should be enough. I'll later add another PR which will implement SM to work that way.

PR #4016, opened 2026-03-05
Fix `LegacyPressable`

m-bert:
## Description

`LegacyPressable` does not work as underlying button implementation had changed. Because previous button is still accessible, I've changed `LegacyPressable` to use it instead of new one, bringing back old behavior. 

On web it required some changes as buttons are exported as `export default`, so simply changing non-web button and import in `Pressable` caused crash on web. 

I've also removed `forwardedRef` from web button and fixed button example so it doesn't always indicate `onLongPress`.

## Test plan

Tested on expo-example, on examples with `Pressable` and `Buttons` example

### DIFF
--- a/packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
+++ b/packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
@@ -1,10 +1,12 @@
 import * as React from 'react';
 import { View, ViewProps } from 'react-native';
 
-export const ButtonComponent = (
-  props: ViewProps & { ref?: React.Ref<React.ComponentRef<typeof View>> }
-) => {
-  return <View accessibilityRole="button" {...props} />;
+type ButtonProps = ViewProps & {
+  ref?: React.Ref<React.ComponentRef<typeof View>>;
 };
+
+export const ButtonComponent = (props: ButtonProps) => (
+  <View accessibilityRole="button" {...props} />
+);
 
 export default ButtonComponent;


packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
```diff
@@ -70,7 +70,7 @@ export interface ButtonProps extends ViewProps, AccessibilityProps {
   /**
    * Invoked on mount and layout changes.
    */
-  onLayout?: (event: LayoutChangeEvent) => void;
+  onLayout?: ((event: LayoutChangeEvent) => void) | undefined;
```
m-bert, line 73: It is necessary because of #4012

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -1,6 +1,12 @@
...
-export default React.forwardRef<React.ComponentRef<typeof View>>(
-  (props, ref) => <View ref={ref} accessibilityRole="button" {...props} />
+type ButtonProps = ViewProps & {
+  ref?: React.Ref<React.ComponentRef<typeof View>>;
+};
+
+export const ButtonComponent = (props: ButtonProps) => (
+  <View accessibilityRole="button" {...props} />
 );
+
```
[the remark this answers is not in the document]
m-bert: Forwarded ref is not needed since React 19. 

> (...); also remove the manual ref field from the props type since it’s not a real prop.

TS don't like that, otherwise it wouldn't be here.

packages/react-native-gesture-handler/src/components/Pressable/Pressable.tsx
```diff
@@ -19,7 +19,7 @@ import {
   StyleProp,
   ViewStyle,
 } from 'react-native';
-import NativeButton from '../GestureHandlerButton';
+import { ButtonComponent as NativeButton } from '../GestureHandlerButton';
```
akwasniewski, line 22: Just a remark, you don't have to change it if you don't think it's necessary: I think it would be sensible to simply change the name of the ButtonComponent to NativeButton. Also after changes in the ../GestureHandlerButton, the Button imported in the v3 version is the wrapped one, thus aliasing it as PureNativeButton seems weird, we can just remove the alias.

m-bert: I think we can leave it as it is for now and leave those renames for other PRs. This one already grew bigger than I expected 😅

PR #4017, opened 2026-03-06
[android] fix talkback on pressable

akwasniewski:
## Description

When TalkBack is activated, pressable cannot be clicked on android.  This PR fixes the issue.

The reason was that Pressable requires a specific order of events. This order of events is hardcoded and found empirically. However, with talback enabled the order of events is different, I added a function which queries native side whether or not talback is enabled and adjusts expected event order accordingly.  
Moreover with talback enabled longPress gets both onPointerUp and onPointerDown before native side receives its events, thus we must not reset the state machine state on onPointerUp. 

## Test plan
Tested on the following example:

<details>

```tsx

import React from 'react';
import { Text, StyleSheet, View } from 'react-native';
import { Pressable, GestureHandlerRootView } from 'react-native-gesture-handler'

const PressableExample = () => {
  return (
      <Pressable onPress={() => setCount((c) => c + 1)}>
        <View style={styles.pressable}>
          <Text>{count}</Text>
        </View>
      </Pressable>
    </GestureHandlerRootView >
  );
};

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
  pressable: {
    alignItems: 'center', justifyContent: 'center', backgroundColor: 'green', width: 100, height: 100
  },
  wrapperCustom: {
    borderRadius: 8,
    padding: 16,
    minWidth: 150,
    alignItems: 'center',
  },
  text: {
    fontSize: 18,
    color: 'white',
    fontWeight: '600',
  },
});

export default PressableExample;

```

</details>


packages/react-native-gesture-handler/src/components/Pressable/stateDefinitions.ts
```diff
@@ -112,6 +132,10 @@ export function getStatesConfig(
   handlePressOut: (event: PressableEvent) => void
 ): StateDefinition[] {
   if (Platform.OS === 'android') {
+    if (RNGestureHandlerModule.isAccessibilityEnabled()) {
+      console.log('accessible!!!');
```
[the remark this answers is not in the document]
akwasniewski: bca566a8b6103965da63d7242a6564aa8f46297f

packages/react-native-gesture-handler/src/RNGestureHandlerModule.web.ts
```diff
@@ -92,4 +92,7 @@ export default {
   setReanimatedAvailable(_isAvailable: boolean) {
     // No-op on web
   },
+  isAccessibilityEnabled() {
+    // TODO
```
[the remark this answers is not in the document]
akwasniewski: ddc78744a90cd9f729a9ea28a7439f8019f146ba

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -129,6 +131,15 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
     isReanimatedAvailable = isAvailable
   }
 
+  @ReactMethod
+  override fun isAccessibilityEnabled(): Boolean {
+    val accessibilityManager = reactApplicationContext.getSystemService(
+      Context.ACCESSIBILITY_SERVICE,
+    ) as AccessibilityManager?
+
+    return accessibilityManager?.isEnabled ?: false
```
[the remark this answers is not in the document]
akwasniewski: 61bee1fd00508a8c150a4357d5e42268f71f65a6

packages/react-native-gesture-handler/src/specs/NativeRNGestureHandlerModule.ts
```diff
@@ -25,6 +25,7 @@ export interface Spec extends TurboModule {
   dropGestureHandler: (handlerTag: Double) => void;
   flushOperations: () => void;
   setReanimatedAvailable: (isAvailable: boolean) => void;
+  isAccessibilityEnabled: () => boolean;
```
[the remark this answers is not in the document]
akwasniewski: 35814ae6e4042f5afe9a5341305adccc7c48e4c9

packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
```diff
@@ -257,7 +258,10 @@ const Pressable = (props: PressableProps) => {
       );
     },
     onTouchesUp: () => {
-      if (Platform.OS === 'android') {
+      if (
+        Platform.OS === 'android' &&
+        !RNGestureHandlerModule.isAccessibilityEnabled()
+      ) {
```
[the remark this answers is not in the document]
akwasniewski: 03d6451ae746dffbfe6fd0b0216e730ee3048f6c

packages/react-native-gesture-handler/src/utils.ts
```diff
@@ -93,3 +96,16 @@ export function deepEqual(obj1: any, obj2: any) {
...
+
+AccessibilityInfo.addEventListener('screenReaderChanged', (enabled) => {
+  isScreenReaderEnabledCache = enabled;
+});
+
+export function isScreenReaderEnabled(): boolean {
+  if (isScreenReaderEnabledCache === null) {
+    isScreenReaderEnabledCache = RNGestureHandlerModule.isScreenReaderEnabled();
+  }
+  return isScreenReaderEnabledCache;
```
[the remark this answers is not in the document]
akwasniewski: f5fb8411b94381979cf7b057e16c5d74dbd8ece5

packages/react-native-gesture-handler/src/components/Pressable/stateDefinitions.ts
```diff
@@ -112,6 +132,9 @@ export function getStatesConfig(
   handlePressOut: (event: PressableEvent) => void
 ): StateDefinition[] {
   if (Platform.OS === 'android') {
+    if (isScreenReaderEnabled()) {
+      return getAndroidAccessibilityStatesConfig(handlePressIn, handlePressOut);
+    }
     return getAndroidStatesConfig(handlePressIn, handlePressOut);
```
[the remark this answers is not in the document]
akwasniewski: 11f7445c39277064fca1272c232ad6f409b7c961

packages/react-native-gesture-handler/src/utils.ts
```diff
@@ -93,3 +96,16 @@ export function deepEqual(obj1: any, obj2: any) {
 }
 
 export const INT32_MAX = 2 ** 31 - 1;
+
+let isScreenReaderEnabledCache: boolean | null = null;
+
+AccessibilityInfo.addEventListener('screenReaderChanged', () => {
+  isScreenReaderEnabledCache = RNGestureHandlerModule.isScreenReaderEnabled();
```
j-piasecki: Why do we need a new method in the module when [there's one in React Native](https://reactnative.dev/docs/accessibilityinfo#isscreenreaderenabled), and you can also just get the value from the event here?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -129,6 +131,15 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
...
   }
 
+  @ReactMethod
+  override fun isScreenReaderEnabled(): Boolean {
+    val accessibilityManager = reactApplicationContext.getSystemService(
+      Context.ACCESSIBILITY_SERVICE,
+    ) as AccessibilityManager?
+
+    return accessibilityManager?.isTouchExplorationEnabled ?: false
+  }
```
j-piasecki: Don't we already have this defined in some utils or extensions file?

m-bert: [Here](https://github.com/software-mansion/react-native-gesture-handler/blob/0ec1cd2ab186103c5178ec9b454f8ede6da62fe5/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/Extensions.kt#L12)

packages/react-native-gesture-handler/src/RNGestureHandlerModule.windows.ts
```diff
@@ -56,4 +56,7 @@ export default {
   flushOperations() {
     // NO-OP
   },
+  isScreenReaderEnabled() {
+    // NO-OP
+  },
```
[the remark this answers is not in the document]
m-bert: Yep, we might change that to `false`. It doesn't cost us much and if for some reason someone was using windows (and we had [such cases](https://github.com/software-mansion/react-native-gesture-handler/issues/2266)) it won't break.

packages/react-native-gesture-handler/src/utils.ts
m-bert: I'm a bit confused. The docs say:

> Fires when the state of the screen reader changes. The argument to the event handler is a boolean. The boolean is `true` when a screen reader is enabled and `false` otherwise.

so why don't we use this value?

packages/react-native-gesture-handler/src/utils.ts
```diff
@@ -93,3 +96,16 @@ export function deepEqual(obj1: any, obj2: any) {
...
+});
+
+export function isScreenReaderEnabled(): boolean {
+  if (isScreenReaderEnabledCache === null) {
+    isScreenReaderEnabledCache = RNGestureHandlerModule.isScreenReaderEnabled();
+  }
```
m-bert: What is the point of this "cache" if we recalculate it anyway before using it?

packages/react-native-gesture-handler/src/components/Pressable/stateDefinitions.ts
```diff
@@ -112,6 +132,9 @@ export function getStatesConfig(
   handlePressOut: (event: PressableEvent) => void
 ): StateDefinition[] {
   if (Platform.OS === 'android') {
+    if (isScreenReaderEnabled()) {
+      return getAndroidAccessibilityStatesConfig(handlePressIn, handlePressOut);
+    }
```
m-bert:
```js
return isScreenReaderEnabled() ? ... : ...
```

packages/react-native-gesture-handler/src/utils.ts
akwasniewski: > Why do we need a new method in the module when [there's one in React Native](https://reactnative.dev/docs/accessibilityinfo#isscreenreaderenabled)

The one from RN is a promise so it caused some issues when I tried to apply it directly without caching so I decided for implementing my own. I might have overcomplicated it though, in 733dd7468c2bcc4edfb0a3561eecd1b59aff4f8b I remade it to use RN function in a custom hook.

> so why don't we use this value?

I didn't notice that the listener returns the value, we can just use it directly 733dd7468c2bcc4edfb0a3561eecd1b59aff4f8b

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
akwasniewski: Obsolete after 733dd7468c2bcc4edfb0a3561eecd1b59aff4f8b

packages/react-native-gesture-handler/src/utils.ts
akwasniewski: Obsolete after https://github.com/software-mansion/react-native-gesture-handler/commit/733dd7468c2bcc4edfb0a3561eecd1b59aff4f8b

packages/react-native-gesture-handler/src/components/Pressable/stateDefinitions.ts
akwasniewski: f843e4734250be35fb6843ad3c4b82840231ab9d

packages/react-native-gesture-handler/src/utils.ts
```diff
@@ -93,3 +96,32 @@ export function deepEqual(obj1: any, obj2: any) {
...
+  const [isEnabled, setIsEnabled] = useState(false);
+
+  useEffect(() => {
+    const checkStatus = async () => {
+      try {
+        const res = await AccessibilityInfo.isScreenReaderEnabled();
+        setIsEnabled(res);
+      } catch (error) {
+        console.warn('Could not read accessibility info: defaulting to false');
+      }
```
m-bert: Is there a scenario where this method fails, but previously didn't, so after `catch` it will remain `true`?

I'm asking out of curiosity, it seems like very subtle edge case 😅

packages/react-native-gesture-handler/src/utils.ts
```diff
@@ -93,3 +96,32 @@ export function deepEqual(obj1: any, obj2: any) {
 }
 
 export const INT32_MAX = 2 ** 31 - 1;
+
+export function useIsScreenReaderEnabled() {
```
m-bert: Since it it is dedicated hook I'd make it a separate file (cc @j-piasecki)

akwasniewski: 7521519fe7135afdb78821bddc1a594cd4736a39

packages/react-native-gesture-handler/src/utils.ts
akwasniewski: I don't think so, the hook should run once when the component mounts, set the initial value and register the listener. Any changes are handled by the listener, thus after catching it is always false.

j-piasecki:
The code itself looks good, but I'll repeat the question from before:

> Have you tried to approach this from the other side? i.e., why are the events in a different order when talkback is enabled? Maybe it's a problem with the native implementation (especially since the changes seems targeted to Android)

akwasniewski:
>Have you tried to approach this from the other side? i.e., why are the events in a different order when talkback is enabled? Maybe it's a problem with the native implementation (especially since the changes seems targeted to Android)

The order of the events is dedicated by the order in which screenReader sends them. First it sends touchEvents and only then it calls the 'performClick' method which will activate native handlers. From what I've tried, it seems that the only way to make it work from the native side without changing android itself would be to cache and resend the `touchEvent` in performClick. However, I reckon that this solution would be more troublesome than the current one.

tastydev:
Hey, i have similar logic in my application to observe accessibility states via events and storing them inside a Zustand store outside of the "react" realm to prevent useEffects etc. and then consuming the a11y zustand states reactivly where needed. 

i.e, something like this just for a hint if you need to abserve more a11y things in the future
```ts
AccessibilityInfo.addEventListener("screenReaderChanged", (enabled: boolean) => {
    if(__DEV__) {
      console.warn(`screenReaderStateChanged -> ${!!enabled}`)
     }
    useAccessibilityStore.getState().setReaderState(enabled)
  }),
```

But besides that i am also syncing up a11y states for the app state change background -> active, to prevent stale states, so leaving this information here if you might consider it important to be implemented aswell:

```ts
AppState.addEventListener("change", (state) => {
    if (state !== "active") {
      return
    }
   // sync states by getters i.e AccessibilityInfo.isScreenReaderEnabled
})
```

PR #4036, opened 2026-03-23
Make buttons activate immediately

m-bert:
## Description

`Native` gesture is specific and its behavior differs across platforms. This leads to strange workarounds in our codebase (e.g. [buttons](https://github.com/software-mansion/react-native-gesture-handler/blob/4a7639d8c83d5d467403cac702ee14ca0a479c4e/packages/react-native-gesture-handler/src/components/GestureButtons.tsx#L71)).

In this PR unifies buttons behavior by changing Native gesture.

## Test plan

Tested on expo-example app (buttons / Pressable)

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
+++ b/packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
@@ -310,7 +310,6 @@ const Pressable = (props: PressableProps) => {
     onActivate: () => {
       if (Platform.OS !== 'android' && Platform.OS !== 'ios') {
         // Native.onActivate is broken with Android + hitSlop
-        // On iOS, onActivate fires on drag (not touch down), so we use onBegin + LONG_PRESS_TOUCHES_DOWN instead
         stateMachine.handleEvent(StateMachineEvent.NATIVE_START);
       }
     },


packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -184,7 +190,7 @@ - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
     }
   }
 
-  [self sendEventsInState:RNGestureHandlerStateActive
+  [self sendEventsInState:RNGestureHandlerStateBegan
            forViewWithTag:sender.reactTag
             withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES
                                                        withNumberOfTouches:event.allTouches.count
```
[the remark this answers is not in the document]
m-bert: [c73ec16](https://github.com/software-mansion/react-native-gesture-handler/pull/4036/commits/c73ec16f8b7369f608b240b238d0f1313bdc186f)

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -146,6 +146,12 @@ - (void)bindToView:(UIView *)view
                   action:@selector(handleTouchUpInside:forEvent:)
         forControlEvents:UIControlEventTouchUpInside];
     [control addTarget:self action:@selector(handleDragExit:forEvent:) forControlEvents:UIControlEventTouchDragExit];
+    [control addTarget:self
+                  action:@selector(handleDragInside:forEvent:)
+        forControlEvents:UIControlEventTouchDragInside];
+    [control addTarget:self
+                  action:@selector(handleDragOutside:forEvent:)
+        forControlEvents:UIControlEventTouchDragOutside];
```
j-piasecki, line 154: So this means that ios buttons will also spam events when pointer moves when pressed?

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -184,7 +190,7 @@ - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
     }
   }
 
-  [self sendEventsInState:RNGestureHandlerStateActive
+  [self sendEventsInState:RNGestureHandlerStateBegan
            forViewWithTag:sender.reactTag
             withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES
```
j-piasecki, line 207: Hmm, maybe the "correct" approach would be to send `begin` and `active` immediately when pressed down (and do the same on other platforms)?

This may be weird that the button isn't "active" unless the pointer moves.

packages/react-native-gesture-handler/src/v3/components/GestureButtons.tsx
```diff
@@ -38,26 +38,22 @@ export const BaseButton = (props: BaseButtonProps) => {
   };
 
   const onBegin = (e: CallbackEventType) => {
-    if (Platform.OS === 'android' && e.pointerInside) {
-      longPressDetected.current = false;
-      if (onLongPress) {
-        longPressTimeout.current = setTimeout(wrappedLongPress, delayLongPress);
-      }
+    // iOS, macOS. Web has its own implementation of button.
```
j-piasecki: Why remove the `pointerInside` check?

packages/react-native-gesture-handler/src/v3/components/GestureButtons.tsx
```diff
@@ -38,26 +38,22 @@ export const BaseButton = (props: BaseButtonProps) => {
...
-    onActiveStateChange?.(true);
-
-    if (Platform.OS !== 'android' && e.pointerInside) {
-      longPressDetected.current = false;
-      if (onLongPress) {
-        longPressTimeout.current = setTimeout(wrappedLongPress, delayLongPress);
-      }
-
-      props.onBegin?.(e);
+    if (Platform.OS === 'android') {
```
j-piasecki: same

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
m-bert: Basically yes.

packages/react-native-gesture-handler/src/v3/components/GestureButtons.tsx
m-bert: [cb58495](https://github.com/software-mansion/react-native-gesture-handler/pull/4036/commits/cb5849582be0220901fbaa42f80d8375af3b479b)

packages/react-native-gesture-handler/src/v3/components/GestureButtons.tsx
m-bert: [cb58495](https://github.com/software-mansion/react-native-gesture-handler/pull/4036/commits/cb5849582be0220901fbaa42f80d8375af3b479b)

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
m-bert: I also thought about it. This would also match current `web` logic.

j-piasecki: So are we doing this or no?

m-bert: Yes, we are ([ad05004](https://github.com/software-mansion/react-native-gesture-handler/pull/4036/commits/ad05004f6f63f6c333921d92e7028fbdca26b58f))

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -146,6 +146,12 @@ - (void)bindToView:(UIView *)view
                   action:@selector(handleTouchUpInside:forEvent:)
         forControlEvents:UIControlEventTouchUpInside];
     [control addTarget:self action:@selector(handleDragExit:forEvent:) forControlEvents:UIControlEventTouchDragExit];
+    [control addTarget:self
+                  action:@selector(handleDragInside:forEvent:)
+        forControlEvents:UIControlEventTouchDragInside];
+    [control addTarget:self
+                  action:@selector(handleDragOutside:forEvent:)
+        forControlEvents:UIControlEventTouchDragOutside];
     [control addTarget:self action:@selector(handleDragEnter:forEvent:) forControlEvents:UIControlEventTouchDragEnter];
```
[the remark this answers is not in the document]
m-bert: [288c458](https://github.com/software-mansion/react-native-gesture-handler/pull/4036/commits/288c45825ad39d2ddbb9d17af2ef9726c95d0328)

packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
```diff
@@ -308,8 +308,9 @@ const Pressable = (props: PressableProps) => {
       stateMachine.handleEvent(StateMachineEvent.NATIVE_BEGIN);
     },
     onActivate: () => {
-      if (Platform.OS !== 'android') {
+      if (Platform.OS !== 'android' && Platform.OS !== 'ios') {
         // Native.onActivate is broken with Android + hitSlop
+        // On iOS, onActivate fires on drag (not touch down), so we use onBegin + LONG_PRESS_TOUCHES_DOWN instead
         stateMachine.handleEvent(StateMachineEvent.NATIVE_START);
       }
```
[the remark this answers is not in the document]
m-bert: [1f1d2df](https://github.com/software-mansion/react-native-gesture-handler/pull/4036/commits/1f1d2df6db5ff901661d6e0b481b6c2bff2f4649)

packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
```diff
@@ -305,11 +307,15 @@ const Pressable = (props: PressableProps) => {
...
+      stateMachine.handleEvent(
+        StateMachineEvent.NATIVE_BEGIN,
+        lastTouchEvent.current ?? undefined
+      );
     },
     onActivate: () => {
-      if (Platform.OS !== 'android') {
+      if (Platform.OS !== 'android' && Platform.OS !== 'ios') {
         // Native.onActivate is broken with Android + hitSlop
+        // On iOS, onActivate fires on drag (not touch down), so we use onBegin + LONG_PRESS_TOUCHES_DOWN instead
```
j-piasecki: Is this still true?

packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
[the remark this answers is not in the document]
j-piasecki: I don't understand:

> If NATIVE_BEGIN arrives before LONG_PRESS_TOUCHES_DOWN (as it does on Android), the state machine will consume the begin step with no payload and handlePressIn will never run

If NATIVE_BEGIN arrives before LONG_PRESS_TOUCHES_DOWN then storing the event from the latter doesn't change anything.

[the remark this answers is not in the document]
m-bert: Actually I've looked into it and seems that it shouldn't be a problem. The difference comes from different orders of callbacks between gestures across platforms and as far as I know this is deterministic.

Reverted in [aa6769f](https://github.com/software-mansion/react-native-gesture-handler/pull/4036/commits/aa6769fc4a354f560dd71ccefd95e75826ce5e90)

packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
m-bert: The condition is necessary, yes. Though this comment was quite misleading, I removed it in [e6adcb4](https://github.com/software-mansion/react-native-gesture-handler/pull/4036/commits/e6adcb44b55d94cd69afd0f21141dcde216e811a)

PR #4043, opened 2026-03-27
fix(android): compile RNGestureHandlerModule with Kotlin 2.3

### LINKED ISSUE #4042: Android: Kotlin 2.3 compile error in RNGestureHandlerModule (rootViewTag vs getRootViewTag)
## Summary

When compiling the Android library with **Kotlin 2.3.x** (e.g. as required by other dependencies such as Stripe Android SDK), `RNGestureHandlerModule.kt` fails to compile in `findRootHelperForViewAncestor`:

```
Function invocation 'getRootViewTag()' expected
```

at the expression using `it.rootView.rootViewTag` next to a local variable also named `rootViewTag`.

## Environment

- `react-native-gesture-handler` **2.30.1** (also reproducible on current 2.30.x line)
- Kotlin **2.3.0+** (project/toolchain aligned with RN 0.83 + libraries compiled with Kotlin 2.3 metadata)
- Android Gradle Plugin **8.12** (typical RN 0.83 setup)

## Root cause

Inside `findRootHelperForViewAncestor`, the local `val rootViewTag = ...` shadows the synthetic Kotlin property `rootViewTag` on `ReactRootView`. The compiler then fails to resolve the property access on the left-hand side of the comparison (Kotlin 2.3 is stricter here than 2.1.x).

`ReactRootView` exposes the tag via **`getRootViewTag()`** in Java; the fix is to use an explicit receiver and call **`getRootViewTag()`** (or rename the local variable).

## Suggested fix

```kotlin
synchronized(roots) {
  return roots.firstOrNull {
    val rv = it.rootView
    rv is ReactRootView && rv.getRootViewTag() == rootViewTag
  }
}
```

## Workaround

Consumers can apply the one-line change locally (e.g. `patch-package`) until an upstream release includes the fix.

---

Related downstream context: projects that bump Kotlin to **2.3** for Stripe or other SDKs hit this as a hard compile failure on `:react-native-gesture-handler:compileDebugKotlin`.

christian-apollo:
## Summary

Fixes Android compilation when using **Kotlin 2.3.x**. The previous code compared `it.rootView.rootViewTag` to a local `rootViewTag`; the local name shadows the synthetic property accessor, and Kotlin **2.3** reports:

> Function invocation `getRootViewTag()` expected

Use an explicit receiver and `ReactRootView#getRootViewTag()` instead.

## Context

- Downstream projects bump Kotlin (e.g. to **2.3.0**) for dependencies such as **Stripe Android SDK** and hit a hard failure in `:react-native-gesture-handler:compileDebugKotlin`.
- Closes #4042.

## Test plan

- [ ] `./gradlew :packages:react-native-gesture-handler:compileDebugKotlin` (or project-equivalent) with Kotlin **2.3.x**

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
@@ -170,7 +170,8 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
     }
     synchronized(roots) {
       return roots.firstOrNull {
-        it.rootView is ReactRootView && it.rootView.rootViewTag == rootViewTag
+        val rv = it.rootView
+        rv is ReactRootView && rv.getRootViewTag() == rootViewTag
       }
     }
   }


m-bert (approved):
Thank you ❤️

PR #4046, opened 2026-03-30
[General] Update button's animation duration props

j-piasecki:
## Description

Native platforms delay touches when the button is inside a ScrollView. This may cause situations where press-out is triggered immediately after press-in, effectively running with no animation.

This PR:
- changes the easing on web to better align with native platforms
- changes `animationDuration` to `pressAndHoldAnimationDuration`
- adds `tapAnimationDuration`

`pressAndHoldAnimationDuration` defaults to `tapAnimationDuration` when unspecified; `tapAnimationDuration` defaults to 100ms.

Press out handler now has three branches:
1. pressDuration >= `pressAndHoldAnimationDuration` -> press out animation is started immediately with the duration of `pressAndHoldAnimationDuration`
2. pressDuration * 2 in [`tapAnimationDuration`, `pressAndHoldAnimationDuration`] (*2 so there's at least half of `tapAnimationDuration` for the remaining animation) -> press out animation is started immediately with the duration of the press so far
3. pressDuration < `tapAnimationDuration / 2` -> pressIn animation is played fully in the remaining time in `tapAnimationDuration`, and pressOut is scheduled after with the duration of `tapAnimationDuration`

## Test plan

Updated the button underlay example

### DIFF
--- a/packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
+++ b/packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
@@ -61,7 +61,7 @@ export interface ButtonProps extends ViewProps, AccessibilityProps {
   /**
    * Duration of the press-in animation when the button is held down, in
    * milliseconds. Defaults to `tapAnimationDuration` when not set (or set
-   * to -1).
+   * to any negative value).
    */
   pressAndHoldAnimationDuration?: number | undefined;
 


packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
```diff
@@ -59,9 +59,18 @@ export interface ButtonProps extends ViewProps, AccessibilityProps {
   touchSoundDisabled?: boolean | undefined;
 
   /**
-   * Duration of the animation when the button is pressed.
+   * Duration of the press-in animation when the button is held down, in
+   * milliseconds. Defaults to `tapAnimationDuration` when not set (or set
+   * to -1).
    */
-  animationDuration?: number | undefined;
+  pressAndHoldAnimationDuration?: number | undefined;
```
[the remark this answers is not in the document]
j-piasecki: This hasn't been released yet

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -28,17 +30,63 @@ export const ButtonComponent = ({
...
+      // elapsed * 2 to ensure there is at least half of the tapAnimationDuration left for the animation to play
+    } else if (elapsed * 2 >= tapAnimationDuration) {
+      setCurrentDuration(elapsed);
+      setPressed(false);
+    } else {
+      // Let the in-progress CSS press-in transition continue; schedule press-out after remaining time
+      const remaining = tapAnimationDuration - elapsed;
+      pressOutTimer.current = setTimeout(() => {
+        pressOutTimer.current = null;
+        setCurrentDuration(tapAnimationDuration);
```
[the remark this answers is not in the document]
j-piasecki: I don't think it's possible to restart the css transition but from the point it currently is at.

packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
```diff
@@ -59,9 +59,18 @@ export interface ButtonProps extends ViewProps, AccessibilityProps {
   touchSoundDisabled?: boolean | undefined;
 
   /**
-   * Duration of the animation when the button is pressed.
+   * Duration of the press-in animation when the button is held down, in
+   * milliseconds. Defaults to `tapAnimationDuration` when not set (or set
+   * to -1).
```
m-bert:
```suggestion
   * Duration of the press-in animation when the button is held down, in
   * milliseconds. Defaults to `tapAnimationDuration` when not set.
```

tbh I don't know whether or not we should mention it. The only use-case I can see is when someone is alternating between values and they alternate between `x` and `-1`. Is there anything else?

packages/react-native-gesture-handler/apple/RNGestureHandlerButton.mm
```diff
@@ -96,6 +103,31 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+  if (newWindow == nil) {
+    if (_pendingPressOutBlock) {
+      dispatch_block_cancel(_pendingPressOutBlock);
+      _pendingPressOutBlock = nil;
+    }
+    RNGHUIView *target = self.animationTarget ?: self;
+    [target.layer removeAllAnimations];
+    [_underlayLayer removeAllAnimations];
+  }
+}
```
m-bert, line 132: Maybe ot would be cleaner to extract it to separate function and then use `if` only to call `super` and this `helper`. 

We can keep that, but headers inside `if` are not the prettiest to read 😅

packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
m-bert: Also, if I can see correctly, all platforms default to `tapAnimationDuration` in case of any negative number, not only `-1`.

j-piasecki: Yeah, any negative number is going to be treated as "no value". I'd like to keep it somehow in the description, so it's possible to reset it to the default after changing it. I'm not sure what the best way is to handle it.

packages/react-native-gesture-handler/apple/RNGestureHandlerButton.mm
j-piasecki: Done in [dccd561](https://github.com/software-mansion/react-native-gesture-handler/pull/4046/commits/dccd56158b4403ff2c15ee171c393920581c6e64)

packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
m-bert: Well, if we want to keep it a number we can simply say "negative number" instead of `-1` and it should be fine.

j-piasecki: Updated in [453b5cb](https://github.com/software-mansion/react-native-gesture-handler/pull/4046/commits/453b5cbf07b92e9f92c58a8af9987fa7babfb943)

PR #4048, opened 2026-03-31
Mocks: fix button content missing

alpha0010:
## Description

Lacking button content in the provided mocks makes it impossible to write tests such as "is the button with 'foo' rendered" or "press the button with 'bar'".

## Test plan

```tsx
import {fireEvent, render} from '@testing-library/react-native';
import {Text} from 'react-native';
import {RectButton} from 'react-native-gesture-handler';

test('Trigger press by text', () => {
  const onPress = jest.fn();
  const {getByText} = render(
    <RectButton onPress={onPress}>
      <Text>Press Me</Text>
    </RectButton>,
  );

  fireEvent.press(getByText('Press Me'));

  expect(onPress).toHaveBeenCalled();
});
```

### DIFF
--- a/packages/react-native-gesture-handler/src/mocks/mocks.tsx
+++ b/packages/react-native-gesture-handler/src/mocks/mocks.tsx
@@ -33,7 +33,7 @@ const RotationGestureHandler = View;
 const FlingGestureHandler = View;
 export const RawButton = ({ enabled, children, ...rest }: any) => (
   <TouchableNativeFeedback disabled={enabled === false} {...rest}>
-    {children}
+    {children ?? <View />}
   </TouchableNativeFeedback>
 );
 export const BaseButton = RawButton;


packages/react-native-gesture-handler/src/mocks/mocks.tsx
```diff
@@ -31,9 +31,9 @@ const LongPressGestureHandler = View;
 const PinchGestureHandler = View;
 const RotationGestureHandler = View;
 const FlingGestureHandler = View;
-export const RawButton = ({ enabled, ...rest }: any) => (
+export const RawButton = ({ enabled, children, ...rest }: any) => (
   <TouchableNativeFeedback disabled={enabled === false} {...rest}>
-    <View />
+    {children}
```
m-bert:
```suggestion
    {children ?? <View />}
```

Could we please keep `View` in case where `children` is `undefined`?

PR #4078, opened 2026-04-10
[Web] Fix `Pinch` and `Rotation`

m-bert:
## Description

I've noticed that `Pinch` and `Rotation` behave weirdly when it comes to ending gesture on web. I've took a look and fixed some minor issues.

> [!NOTE]
> This PR may introduce mismatch between `android` and `web` versions of `Rotation`. I still think that this is how it is supposed to work, so we may want to [make required changes to `android` as well](https://github.com/software-mansion/react-native-gesture-handler/pull/4079) (or drop this PR if you prefer to).

### Pinch

I've noticed that `Pinch` is not reset when pointers are placed on view and then released. This was because for some reason `Pinch` simply ignored finished states if it had not activated earlier. 

### Rotation 

`Rotation`, always finalized with `true`. This was because we always tried to move it to `end` state. Our internal event handler hadn't been sending `onDeactivate`, but `true` was still passed into `onFinalize`.

## Test plan

<details>
<summary>Tested on Transformations example and code below:</summary>

```tsx
import {
  GestureDetector,
  GestureHandlerRootView,
  usePinchGesture,
  useRotationGesture,

  const g = useRotationGesture({
    onBegin: () => console.log('onBegin'),
    onActivate: () => console.log('onActivate'),
    onUpdate: (e) => console.log('onUpdate', e.rotation),
    onDeactivate: () => console.log('onDeactivate'),
    onFinalize: (_, s) => console.log('onFinalize', s),
  });

  return (
      <GestureDetector gesture={g}>
        <View style={styles.box} />
      </GestureDetector>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  box: {
    width: 200,
    height: 200,
    backgroundColor: 'blue',
    borderRadius: 12,
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/web/handlers/PinchGestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/PinchGestureHandler.ts
@@ -100,6 +100,7 @@ export default class PinchGestureHandler extends GestureHandler {
     this.tracker.removeFromTracker(event.pointerId);
 
     if (this.state === State.ACTIVE) {
+      // We don't have to call it in the else branch as it would simply return `true`.
       this.scaleGestureDetector.onTouchEvent(event, this.tracker);
 
       this.end();
--- a/packages/react-native-gesture-handler/src/web/handlers/PinchGestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/PinchGestureHandler.ts
@@ -98,12 +98,10 @@ export default class PinchGestureHandler extends GestureHandler {
   protected override onPointerUp(event: AdaptedEvent): void {
     super.onPointerUp(event);
     this.tracker.removeFromTracker(event.pointerId);
-    if (this.state !== State.ACTIVE) {
-      return;
-    }
-    this.scaleGestureDetector.onTouchEvent(event, this.tracker);
 
     if (this.state === State.ACTIVE) {
+      this.scaleGestureDetector.onTouchEvent(event, this.tracker);
+
       this.end();
     } else {
       this.fail();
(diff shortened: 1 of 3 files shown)


packages/react-native-gesture-handler/src/web/handlers/PinchGestureHandler.ts
```diff
@@ -98,12 +98,10 @@ export default class PinchGestureHandler extends GestureHandler {
   protected override onPointerUp(event: AdaptedEvent): void {
     super.onPointerUp(event);
     this.tracker.removeFromTracker(event.pointerId);
-    if (this.state !== State.ACTIVE) {
-      return;
-    }
-    this.scaleGestureDetector.onTouchEvent(event, this.tracker);
 
     if (this.state === State.ACTIVE) {
+      this.scaleGestureDetector.onTouchEvent(event, this.tracker);
```
j-piasecki, line 104: This means that the scale detector will not be receiving up events in the active state. Is that intended?

m-bert: > Is that intended?

Yes, it was. If you look at the removed code it didn't make any sense and now it resembles the same logic as it had previously. I've also checked if it should be moved up, but if you look at [ScaleGestureDetector](https://github.com/software-mansion/react-native-gesture-handler/blob/b09afc0465f36303b6677e9f7a375afd03826db0/packages/react-native-gesture-handler/src/web/detectors/ScaleGestureDetector.ts#L60) it will simply return `true` in case of inactive handler, so we don't have to call it. I've added comment in [02ed360](https://github.com/software-mansion/react-native-gesture-handler/pull/4078/commits/02ed360deaf6f1a0e2bcd0a3cc01652c22d56c21).

As for moving this logic to `onScaleEnd` it is not as easy as with `Rotation` gesture, as it is only called if handler was active.

j-piasecki (approved):

PR #4094, opened 2026-04-16
Add `cancelsJSResponder ` prop to the gesture handlers

coado:
## Description

Adds a per-gesture `cancelsJSResponder?: boolean` prop on both platforms to the v3 handler hooks. It defaults to `true` to preserve the behavior of cancelling RN JS responder system. 

On `Android` it moves the root-view dispatch off `onCancel` onto new orchestrator callbacks `onCancelJSResponderRequested` and `onCancelJSResponderReleased`. The orchestrator only files the "released" callback when the last active handler with `cancelsJSResponder: true` finishes.

on `iOS` the `RNRootViewGestureRecognizer` checks the handler prop before notifying the delegate.   

## Test plan

Added new test example with multiple handlers that might or might not cancel the JS responder. 

Simulator Screenshot - iPhone 16 Pro Max - 2026-04-17 at 12 59 42


coado:
@copilot review?

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -26,6 +28,7 @@ export function NativeDetector<
       : HostGestureDetector;
 
   ensureNativeDetectorComponent(NativeDetectorComponent);
+
```
j-piasecki:
```suggestion
```

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -1,31 +1,32 @@
...
+import { onGestureHandlerEvent } from '../eventReceiver';
+import {
+  GestureHandlerNativeEvent,
+  PropsRef,
+  ResultEvent,
+} from '../../../web/interfaces';
 
 export const ALLOWED_PROPS = [
   ...baseGestureHandlerWithDetectorProps,
+  'cancelsJSResponder',
```
j-piasecki: I don't think this is needed, no? We're not exposing this prop to the v2 API.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -42,6 +42,9 @@ class GestureHandlerOrchestrator(
   private var finishedHandlersCleanupScheduled = false
   private var activationIndex = 0
 
+  var onCancelJSResponderRequested: ((GestureHandler) -> Unit)? = null
+  var onCancelJSResponderReleased: ((GestureHandler) -> Unit)? = null
```
j-piasecki: Can we change that to be an interface implemented by the root helper, which could then be injected into the orchestrator?

This would align with existing patterns

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
```diff
@@ -41,6 +41,18 @@ class RNGestureHandlerRootHelper(private val context: ReactContext, wrappedView:
       rootView,
     ).apply {
       minimumAlphaForTraversal = MIN_ALPHA_FOR_TOUCH
+      onCancelJSResponderRequested = { _ ->
+        shouldIntercept = true
```
j-piasecki: Moving `shouldIntercept` assignment here makes it so every handler with `cancelsJSRecognizer` set to false will not prevent host components from handling touch after it activates.

We could experiment with that in the future, but that behavior shouldn't be affected by this new prop.

packages/docs-gesture-handler/docs/gestures/_shared/base-gesture-config.mdx
```diff
@@ -137,12 +139,27 @@ useAnimated: boolean;
...
+  ### cancelsJSResponder
+</HeaderWithBadges>
+
+```ts
+cancelsJSResponder?: boolean;
+```
+
+Controls whether activating a Gesture Handler recognizer should cancel React Native JS responders.
+
+This option is configured per gesture handler (for example in `usePanGesture({ cancelsJSResponder: ... })`).
```
j-piasecki: I don't think this needs explicit mention given that it's in the handler config section

j-piasecki (commented):
Can you update the title and description with the final prop name?

Also, I don't have an iOS device on hand, so can you check the following scenario?

There are two handlers and js recognizer on the same view. Gesture A with `cancelsJSRecognizer: false`, and a config that prevents it from activating (like `minimumDistance: 100000` on pan). Gesture B has `cancelsJSRecognizer: false` and requires A to fail before it activates (`useExclusiveGestures(A, B)`), but should try to activate almost immediately (like pan with `minimumDistance: 10`). Now, in that scenario, does B trying to activate but getting blocked on A, cancel the JS recognizer, or not?

I've checked Android and there it works as expected (until A fails, the JS recognizer is not interrupted).

packages/react-native-gesture-handler/apple/RNRootViewGestureRecognizer.m
```diff
@@ -53,10 +53,10 @@ - (BOOL)canPreventGestureRecognizer:(UIGestureRecognizer *)preventedGestureRecog
 - (BOOL)canBePreventedByGestureRecognizer:(UIGestureRecognizer *)preventingGestureRecognizer
 {
   // When this method is called it means that one of handlers has activated, in this case we want
-  // to send an info to JS so that it cancells all JS responders, as long as the preventing
+  // to send an info to JS so that it cancels all JS responders, as long as the preventing
   // recognizer is from Gesture Handler, otherwise we might break some interactions
```
j-piasecki: This comment should also be updated to mention `cancelsJSResponder`

apps/common-app/src/new_api/tests/rnResponderCancellation/index.tsx
```diff
@@ -0,0 +1,433 @@
...
+const SECTION_MIN_HEIGHT = 640;
+
+export default function RNResponderCancellationExample() {
+  return (
+    <ScrollView
+      style={scrollStyles.scroll}
+      contentContainerStyle={scrollStyles.content}
+      showsVerticalScrollIndicator>
+      <View style={scrollStyles.section}>
+        <SingleHandlerExample />
```
m-bert: What about extracting this (and `Multi*`) to separate files?

packages/docs-gesture-handler/docs/gestures/_shared/base-gesture-config.mdx
```diff
@@ -137,12 +139,27 @@ useAnimated: boolean;
...
 
 ```ts
 activeCursor: ActiveCursor | SharedValue<ActiveCursor>;
 ```
 
 This parameter allows specifying which cursor should be used when the gesture activates. Supports all [CSS cursor values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/cursor#keyword) (e.g. `"grab"`, `"zoom-in"`). Default value is set to `"auto"`.
+
+<HeaderWithBadges platforms={['ios', 'android']}>
+  ### cancelsJSResponder
+</HeaderWithBadges>
```
m-bert: As far as I remember this is now `Badges` and `HeaderWithBadges` no longer exists (and if it does, it shouldn't)

packages/docs-gesture-handler/docs/gestures/_shared/base-gesture-config.mdx
m-bert: Agree. Maybe we could keep

"This option is configured per gesture handler"

But it doesn't seem necessary either.

packages/docs-gesture-handler/docs/gestures/_shared/base-gesture-config.mdx
```diff
@@ -137,12 +139,27 @@ useAnimated: boolean;
...
+```ts
+cancelsJSResponder?: boolean;
+```
+
+Controls whether activating a Gesture Handler recognizer should cancel React Native JS responders.
+
+This option is configured per gesture handler (for example in `usePanGesture({ cancelsJSResponder: ... })`).
+
+- `true` (default): when this gesture activates, it cancels React Native JS responders in the same root view.
+- `false`: disables that cancellation, so Gesture Handler callbacks and RN responder callbacks can run at the same time.
```
m-bert: I know this PR is not really about docs, but imo this one would be hard to read. If you quickly look at `false` to check what it does you see "disables that cancellation (...)" - I'd change it to be more specific.

Also we tend to be more descriptive, e.g.:

>Setting this property to `true` ensures that the [Animated API](/docs/fundamentals/animated-interactions) functions correctly when `useNativeDriver` is set to `false`. The default value is set to `false`.

So maybe something like this would work?

> Setting this property to true ensures that the gesture handler cancels any active React Native JS responders in the same root view upon activation. When set to false, both Gesture Handler and React Native responder callbacks can run simultaneously. The default value is set to true.

m-bert (commented):
Seems like @j-piasecki caught everything

apps/common-app/src/new_api/tests/rnResponderCancellation/index.tsx
coado: Yeah I don't actually like this example so will change that a bit.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -19,6 +19,7 @@ class GestureHandlerOrchestrator(
   private val handlerRegistry: GestureHandlerRegistry,
   private val viewConfigHelper: ViewConfigurationHelper,
   private val rootView: ViewGroup,
+  private val onJSResponderCancelListener: OnJSResponderCancelListener? = null,
```
j-piasecki: Is there a place where we don't want to pass `onJSResponderCancelListener`? If not, then this can be non-nullable.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
```diff
@@ -34,11 +35,26 @@ class RNGestureHandlerRootHelper(private val context: ReactContext, wrappedView:
...
+        val time = SystemClock.uptimeMillis()
+        val event = MotionEvent.obtain(time, time, MotionEvent.ACTION_CANCEL, 0f, 0f, 0)
+        if (rootView is RootView) {
+          rootView.onChildStartedNativeGesture(rootView, event)
+        }
+        event.recycle()
+      }
+
+      override fun onCancelJSResponderReleased(handler: GestureHandler) {
+        shouldIntercept = false
```
j-piasecki, line 49: Why do we need to reset `shouldIntercept` here?

coado: When the last active handler with `cancelsJSReponder=true` deactivates, this method is called so that RN touch events are no longer intercepted.

coado:
> Can you update the title and description with the final prop name?
> 
> Also, I don't have an iOS device on hand, so can you check the following scenario?
> 
> There are two handlers and js recognizer on the same view. Gesture A with `cancelsJSRecognizer: false`, and a config that prevents it from activating (like `minimumDistance: 100000` on pan). Gesture B has `cancelsJSRecognizer: false` and requires A to fail before it activates (`useExclusiveGestures(A, B)`), but should try to activate almost immediately (like pan with `minimumDistance: 10`). Now, in that scenario, does B trying to activate but getting blocked on A, cancel the JS recognizer, or not?
> 
> I've checked Android and there it works as expected (until A fails, the JS recognizer is not interrupted).

Not sure if you meant `cancelsJSRecognizer: true` on B, but in both cases seems like it works as expected.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
j-piasecki: Can you check how this behaves with iOS system recognizers? i.e., two simultaneous gestures: A with `cancelsJSReponder=true`, B with `cancelsJSReponder=false`, and a JS responder/native view. Both A and B activate, then A finishes while B stays active. Does the JS responder/native view work in this scenario?

I think this is the case where this would matter on Android, but it would also allow native components to start handling touch in addition to the JS responder. Ideally, both platforms would be aligned on this.

coado: The JS responder doesn't work when the A is activated and starts working when A deactivates (regardless if B is activated or not).

j-piasecki: The same thing is happening on Android in the current state, no?

j-piasecki (approved):

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
coado: On Android after the first pointer is down, the RN JS responder system is not deciding the target for another pointers, so it doesn't behave the same way. For other cases, where only one pointer is checked it works as expected.

PR #4100, opened 2026-04-20
[Web] Don't cancel buttons when pointer moves inside

j-piasecki:
## Description

Prevents the buttons from being canceled when the pointer is moved inside the button.

Note that this only works with a mouse, since the browser will cancel any events with no `touchAction` during any drag gesture.

## Test plan

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/08b342f8-24cc-43e9-a240-844641f97552" />|<video src="https://github.com/user-attachments/assets/5cdc15bf-2518-4602-a180-ffd4ae48745e" />|

<details>
<summary>Expand</summary>

```jsx
import { Touchable } from 'react-native-gesture-handler';

  const [isActive, setIsActive] = useState(false);

  return (
      <Touchable
        style={[styles.button, isActive && styles.buttonActive]}
        onPress={() => console.log('pressed')}
        onPressIn={() => setIsActive(true)}
        onPressOut={() => setIsActive(false)}
        activeScale={1.2}>
        <Text style={styles.label}>{isActive ? 'Highlighted' : 'Press me'}</Text>
      </Touchable>
      <Text style={styles.hint}>
        Press the button, then slowly drag your pointer away from it.
      </Text>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    padding: 24,
    gap: 24,
  },
  button: {
    backgroundColor: '#ddd',
    paddingHorizontal: 32,
    paddingVertical: 16,
    borderRadius: 8,
  },
  buttonActive: {
    backgroundColor: '#f97316',
  },
  label: {
    fontSize: 20,
    fontWeight: '600',
  },
  hint: {
    textAlign: 'center',
    opacity: 0.6,
    fontSize: 14,
  },
});

```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/web/tools/PointerEventManager.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/PointerEventManager.ts
@@ -162,6 +162,18 @@ export default class PointerEventManager extends EventManager<HTMLElement> {
     const adaptedEvent: AdaptedEvent = this.mapEvent(event, EventTypes.LEAVE);
 
     this.onPointerMoveOut(adaptedEvent);
+
+    // When the view is not capturing the pointer (e.g. when `role="button"`),
+    // `pointermove` stops firing once the pointer leaves the view's bounds, so
+    // the out-of-bounds detection in `pointerMoveCallback` never runs. Fall back
+    // to the DOM `pointerleave` event for any tracked, in-bounds pointer.
+    if (
+      this.trackedPointers.has(event.pointerId) &&
+      this.pointersInBounds.indexOf(event.pointerId) >= 0
+    ) {
+      this.onPointerLeave(adaptedEvent);
+      this.markAsOutOfBounds(event.pointerId);
+    }
   };
 
   private lostPointerCaptureCallback = (event: PointerEvent) => {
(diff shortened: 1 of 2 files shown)


packages/react-native-gesture-handler/src/web/tools/PointerEventManager.ts
```diff
@@ -162,6 +162,18 @@ export default class PointerEventManager extends EventManager<HTMLElement> {
...
+
+    // When the view is not capturing the pointer (e.g. when `role="button"`),
+    // `pointermove` stops firing once the pointer leaves the view's bounds, so
+    // the out-of-bounds detection in `pointerMoveCallback` never runs. Fall back
+    // to the DOM `pointerleave` event for any tracked, in-bounds pointer.
+    if (
+      this.trackedPointers.has(event.pointerId) &&
+      this.pointersInBounds.indexOf(event.pointerId) >= 0
+    ) {
+      this.onPointerLeave(adaptedEvent);
```
m-bert, line 174: If this happens when `onPointerMove` is blocked (as in buttons), won't this be called twice for other elements?

j-piasecki: The only other path that could call `onPointerLeave` is guarded by the pointer being present inside `pointersInBound`: https://github.com/software-mansion/react-native-gesture-handler/blob/2154bfd56382a4897dee061623f3837ae6b57da1/packages/react-native-gesture-handler/src/web/tools/PointerEventManager.ts#L120-L140

So those calls should make the other one a noop (no matter the order).

m-bert (approved):

PR #4102, opened 2026-04-20
[General] Don't emit duplicate events for `NativeViewGestureHandler`

j-piasecki:
## Description

Updates `NativeViewGestureHandler` not to emit update event on every pointer move unless the payload would change. This significantly reduces the number of events dispatched by the native gesture handlers, which should have a positive impact on performance without causing any information loss.

Supersedes https://github.com/software-mansion/react-native-gesture-handler/pull/3348

## Test plan

Add `onUpdate={console.log}` to the internal `Touchable` implementation and check any example using `Touchable`.

### DIFF
--- a/packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
+++ b/packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
@@ -113,6 +113,7 @@
 @implementation RNNativeViewGestureHandler {
   BOOL _shouldActivateOnStart;
   BOOL _disallowInterruption;
+  RNGestureHandlerEventExtraData *_lastActiveExtraData;
 }
 
 - (instancetype)initWithTag:(NSNumber *)tag
@@ -180,6 +181,16 @@
   [super unbindFromView];
 }
 
+- (void)sendActiveStateEventIfChangedForView:(UIView *)sender extraData:(RNGestureHandlerEventExtraData *)extraData
+{
+  if ([_lastActiveExtraData.data isEqualToDictionary:extraData.data]) {
+    return;
+  }
+
+  _lastActiveExtraData = extraData;
+  [self sendEventsInState:RNGestureHandlerStateActive forViewWithTag:sender.reactTag withExtraData:extraData];
+}
+
 - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
 {
   [self setCurrentPointerTypeForEvent:event];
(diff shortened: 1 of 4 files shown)


packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -202,11 +213,11 @@ - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
...
 
-  [self sendEventsInState:RNGestureHandlerStateActive
-           forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES
-                                                       withNumberOfTouches:event.allTouches.count
-                                                           withPointerType:_pointerType]];
+  _lastActiveExtraData = nil;
+  [self sendActiveStateEventIfChangedForView:sender
+                                   extraData:[RNGestureHandlerEventExtraData forPointerInside:YES
+                                                                          withNumberOfTouches:event.allTouches.count
```
m-bert, line 219: What happened here? 🤔

packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
```diff
@@ -199,4 +203,23 @@ export default class NativeViewGestureHandler extends GestureHandler {
       ),
     };
   }
+
+  protected override shouldSuppressActiveUpdate(
+    handlerData: HandlerData<unknown>
+  ): boolean {
+    const current = handlerData as Record<string, unknown>;
```
m-bert:
```suggestion
    const current = handlerData;
```

I think it is not necessary. And if we remove this cast we could use `handlerData` instead.

Or have I missed something?

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
j-piasecki: This is how clang-format handles it - it aligns the nested method call around the `:` symbol.

packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
j-piasecki: [ab0dc12](https://github.com/software-mansion/react-native-gesture-handler/pull/4102/commits/ab0dc1230615b332262d9944d58d1230ac4aa3e7) and [b5d49ff](https://github.com/software-mansion/react-native-gesture-handler/pull/4102/commits/b5d49ff6e09e99f6320898128444048aa8d78e94)

m-bert (approved):

PR #4106, opened 2026-04-22
[Native] Fix `Touchable` behavior when nested within gesture handlers

j-piasecki:
## Description

See https://github.com/software-mansion/react-native-gesture-handler/discussions/4097

Fixes interactions of nested touchables/gestures:
1. Touchable was missing `disallowInterruption: true` prop, which caused it to misbehave when used alongside other gestures
2. Android keeps the list of touch targets internally in [each `ViewGroup`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/ViewGroup.java#220). Since RNGH can start intercepting events mid-stream, this list could contain stale values. The OS, trying to clean it up, would dispatch a synthetic `ACTION_CANCEL` event, which resulted in the touchable working every other time it's been pressed. To fix that, we do the cleanup by ourselves, by dispatching a synthetic `ACTION_CANCEL` event before the new event stream starts.

## Test plan

Added a new example

This doesn't work correctly on web, but it didn't work before either.

### DIFF
--- a/apps/common-app/src/new_api/index.tsx
+++ b/apps/common-app/src/new_api/index.tsx
@@ -35,6 +35,7 @@ import RotationExample from './simple/rotation';
 import TapExample from './simple/tap';
 import NestedPressablesExample from './tests/nestedPressables';
 import NestedRootViewExample from './tests/nestedRootView';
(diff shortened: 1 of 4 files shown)


m-bert (approved):
LGTM!

PR #4107, opened 2026-04-22
Do not check for `worklets` if babel plugin was not used

m-bert:
## Description

In some environments babel plugin is not used (e.g. it can be omitted when using `Next.js`). Our detector checks if callbacks are `worklets` by checking `__workletHash` field. However, if plugin wasn't used, this will be `undefined`. In that case, our warning will wrongly suggest to use `runOnJS`. This may be misleading, if plugin was skipped on `web`, but used on native platforms.

In this PR, we check if plugin was applied by checking empty worklet function. If it has `__workletHash` field it means that plugin was used. If it wasn't, we won't suggest using `runOnJS`.

## Test plan

Confirmed by the original reporter

### DIFF
--- a/packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
+++ b/packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
@@ -36,6 +36,11 @@ export const ALLOWED_PROPS = [
   ...nativeViewGestureHandlerProps,
 ];
 
+// This is used to check if babel plugin was used.
+function emptyWorklet() {
+  'worklet';
+}
+
 function convertToHandlerTag(ref: GestureRef): number {
   if (typeof ref === 'number') {
     return ref;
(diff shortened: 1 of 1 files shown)


packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -36,6 +36,11 @@ export const ALLOWED_PROPS = [
   ...nativeViewGestureHandlerProps,
 ];
 
+// This is used to check if babel plugin was used.
```
m-bert: Let me know if this need further clarification

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
```diff
@@ -103,9 +108,11 @@ export function checkGestureCallbacksForWorklets(gesture: GestureType) {
   }
 
   const areAllNotWorklets = !areSomeWorklets && areSomeNotWorklets;
+  // @ts-ignore if callback is a worklet, the property will be available, if not then the check will return false
+  const wasBabelPluginEnabled = emptyWorklet.__workletHash !== undefined;
```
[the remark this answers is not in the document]
m-bert: I can change comment, but I don't think doing this cast makes sense here.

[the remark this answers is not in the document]
j-piasecki: I don't think it differs that much. Maybe `ts-expect-error` would fit more in this case.

j-piasecki: I don't know if microoptimizing it that much makes sense, but I think this can be hoisted to top-level

j-piasecki (approved):

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
m-bert: Sure, [ac74816](https://github.com/software-mansion/react-native-gesture-handler/pull/4107/commits/ac748166c1cf36d232448228f05a751399bd0e48)

packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts
m-bert: [ac74816](https://github.com/software-mansion/react-native-gesture-handler/pull/4107/commits/ac748166c1cf36d232448228f05a751399bd0e48)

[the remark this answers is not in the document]
m-bert: [ac74816](https://github.com/software-mansion/react-native-gesture-handler/pull/4107/commits/ac748166c1cf36d232448228f05a751399bd0e48)

PR #4113, opened 2026-04-23
[iOS] Handle `0` duration animation correctly on `Touchable`

j-piasecki:
## Description

On iOS, setting `tapAnimationDuration` on Touchable didn't turn off the animation entirely, instead making it take 0.25s. CoreAnimation treats `0` in this case as `not set` and uses the default duration.

This PR adds an explicit check for the `0` duration and applies the end state immediately for the underlay.

## Test plan

```jsx
      <Touchable
        style={{
          width: 220,
          paddingVertical: 20,
          borderRadius: 12,
          backgroundColor: '#A0D5EF',
          alignItems: 'center',
          justifyContent: 'center',
          overflow: 'hidden',
        }}
        underlayColor={COLORS.RED}
        activeUnderlayOpacity={0.6}
        tapAnimationDuration={0}
        onPress={() => pushLog('onPress')}>
        <Text>Hold me</Text>
      </Touchable>
```


m-bert (approved):
It compiled in my head 👀

PR #4115, opened 2026-04-24
[General] Allow to create gestures with no config

j-piasecki:
## Description

Updates gesture hook signatures to include the default, empty config for each gesture. This allows for creating gesture objects without passing a config, like so:

```jsx
const nativeGesture = useNativeGesture()
```

`Native` gesture would be the most useful here, since its config is often empty when pulling in a host component into RNGH's gesture system. Other gestures may be a nice DX improvement, though, since now it's simpler to create a "placeholder" gesture.

The default config is referentially stable, so it shouldn't cause unnecessary updates between renders as opposed to passing an empty object (with no compiler and memoization).

## Test plan

Static checks


m-bert (approved):
❤️

PR #4120, opened 2026-04-24
[General] Don't activate buttons on start

j-piasecki:
## Description

Changes `shouldActivateOnStart` on buttons to `false` - `true` breaks nested buttons on Android.

cc @m-bert - I don't recall why it was `true` in the first place, any chance that after https://github.com/software-mansion/react-native-gesture-handler/pull/4116 it wouldn't have been needed anyway?

## Test plan

|Before|Update|
|-|-|
|<video src="https://github.com/user-attachments/assets/e6fd41f8-98dc-43e5-9d31-442621eb8e08" />|<video src="https://github.com/user-attachments/assets/c65a4fa2-2f3a-4052-8e7f-edcf96b7256d" />|

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/components/GestureButtons.tsx
+++ b/packages/react-native-gesture-handler/src/v3/components/GestureButtons.tsx
@@ -23,7 +23,7 @@ export const RawButton = createNativeWrapper<
   RawButtonProps
 >(GestureHandlerButton, {
   shouldCancelWhenOutside: false,
-  shouldActivateOnStart: true,
+  shouldActivateOnStart: false,
 });
 
 /**


m-bert:
> cc @m-bert - I don't recall why it was true in the first place

This was changed in #4036 in order to get rid of strange platform dependent workarounds in buttons. 

> any chance that after https://github.com/software-mansion/react-native-gesture-handler/pull/4116 it wouldn't have been needed anyway?

Probably. I think in that case we should also check if we should swap `onBegin` with `onActivate` in buttons (as you did in `Touchable`).

j-piasecki:
> I think in that case we should also check if we should swap onBegin with onActivate in buttons (as you did in Touchable).

Do we want to do that in the buttons given, which are deprecated (and this will likely result in a behavior change)?

m-bert (approved):
> Do we want to do that in the buttons given, which are deprecated (and this will likely result in a behavior change)?

If they work correctly after this, then we don't have to change it. I've looked at examples and they seem work (at least in case of callbacks) so I guess we can ship it :shipit:

PR #4128, opened 2026-04-28
[Web] Cancel button gesture when other handlers are extracted before it

j-piasecki:
## Description

Applies the same solution for the problem described in https://github.com/software-mansion/react-native-gesture-handler/pull/4127.

There's no native gesture hooks mechanism on web, so the check happens directly in `NativeViewGestureHandler`. Since the button feedback animation is independent of the gesture logic currently, I added custom DOM events dispatched by `NativeViewGestureHandler` when it begins and gets canceled. This way, the button can block/enable the feedback depending on gesture state.

## Test plan

Tested on the example from https://github.com/software-mansion/react-native-gesture-handler/pull/4127

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/1304c59d-4464-410d-a9f2-8508a298d946" />|<video src="https://github.com/user-attachments/assets/863c454f-ec8f-471e-bfad-8cfec85eb504" />|

### DIFF
--- a/packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
+++ b/packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
@@ -72,6 +72,11 @@ export const ButtonComponent = ({
     };
     const handleGestureCanceled = () => {
       gestureEnabledRef.current = false;
+      if (pressOutTimer.current != null) {
+        clearTimeout(pressOutTimer.current);
+        pressOutTimer.current = null;
+      }
+      pressInTimestamp.current = 0;
       setPressed(false);
     };
 
--- a/packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
+++ b/packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import type { ColorValue, NativeSyntheticEvent, ViewProps } from 'react-native';
 import { View } from 'react-native';
 
+import { GestureLifecycleEvent } from '../web/tools/GestureLifecycleEvents';
+
 type ButtonProps = ViewProps & {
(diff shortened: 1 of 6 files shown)


packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -46,9 +49,47 @@ export const ButtonComponent = ({
...
   React.useEffect(() => {
+    const node = viewRef.current;
+
+    const handleGestureBegan = () => {
+      gestureEnabledRef.current = true;
+    };
+    const handleGestureCanceled = () => {
+      gestureEnabledRef.current = false;
+      setPressed(false);
+    };
```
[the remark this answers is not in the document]
j-piasecki: [3b7c367](https://github.com/software-mansion/react-native-gesture-handler/pull/4128/commits/3b7c3675a1d1f4dc436ae71fd4f221eb807cc9ae)

packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
```diff
@@ -208,6 +217,31 @@ export default class NativeViewGestureHandler extends GestureHandler {
...
   }
 
+  public override shouldBeginWithRecorded(
+    recorded: IGestureHandler[]
+  ): boolean {
+    if (!this.isButton()) {
+      return true;
+    }
+
+    const self = this as unknown as IGestureHandler;
```
m-bert: Is `as unknown` necessary?

j-piasecki: No - [a0374bc](https://github.com/software-mansion/react-native-gesture-handler/pull/4128/commits/a0374bc32f83ca3789460cfee331f1626e3b792a)

m-bert (approved):

PR #4129, opened 2026-04-29
[iOS] Activate buttons on pointer up

j-piasecki:
## Description

Changes the moment the `ACTIVE` event is dispatched on iOS from pointer down to pointer up, aligning it with Android and Web implementations.

Updates existing checks to account for this change.

Updates the v3 button implementation to do cleanup in `onFinalize` instead of `onDeactivate` - setup happens in `onBegin`, so there's no guarantee that `onDeactivate` will trigger.

## Test plan

Tested existing examples, including legacy ones

### DIFF
--- a/packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
+++ b/packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
@@ -253,11 +253,6 @@
             withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES
                                                        withNumberOfTouches:event.allTouches.count
                                                            withPointerType:_pointerType]];
-
-  [self sendActiveStateEventIfChangedForView:sender
-                                   extraData:[RNGestureHandlerEventExtraData forPointerInside:YES
-                                                                          withNumberOfTouches:event.allTouches.count
-                                                                              withPointerType:_pointerType]];
 }
 
 - (void)handleTouchUpOutside:(UIView *)sender forEvent:(UIEvent *)event
--- a/packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
+++ b/packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
@@ -275,11 +275,12 @@
 
 - (void)handleTouchUpInside:(UIView *)sender forEvent:(UIEvent *)event
 {
-  [self sendEventsInState:RNGestureHandlerStateEnd
-           forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES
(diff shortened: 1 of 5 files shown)


packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -275,11 +275,12 @@ - (void)handleTouchUpOutside:(UIView *)sender forEvent:(UIEvent *)event
...
-           forViewWithTag:sender.reactTag
-            withExtraData:[RNGestureHandlerEventExtraData forPointerInside:YES
-                                                       withNumberOfTouches:event.allTouches.count
-                                                           withPointerType:_pointerType]];
+  RNGestureHandlerEventExtraData *extraData = [RNGestureHandlerEventExtraData forPointerInside:YES
+                                                                           withNumberOfTouches:event.allTouches.count
+                                                                               withPointerType:_pointerType];
+
+  [self sendActiveStateEventIfChangedForView:sender extraData:extraData];
+  [self sendEventsInState:RNGestureHandlerStateEnd forViewWithTag:sender.reactTag withExtraData:extraData];
```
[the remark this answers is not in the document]
j-piasecki: Messed up rebase - [276beec](https://github.com/software-mansion/react-native-gesture-handler/pull/4129/commits/276beec223a989b8afcbacad00118e28207ef886)

packages/react-native-gesture-handler/src/components/GestureButtons.tsx
```diff
@@ -50,7 +50,8 @@ class InnerBaseButton extends React.Component<BaseButtonWithRefProps> {
     nativeEvent,
   }: HandlerStateChangeEvent<NativeViewGestureHandlerPayload>) => {
     const { state, oldState, pointerInside } = nativeEvent;
-    const active = pointerInside && state === State.ACTIVE;
+    const active =
+      pointerInside && (state === State.BEGAN || state === State.ACTIVE);
```
m-bert: 🫨

m-bert (approved):

packages/react-native-gesture-handler/src/components/GestureButtons.tsx
j-piasecki: Yeah, I renamed it to `isPressed` in [575f6ee](https://github.com/software-mansion/react-native-gesture-handler/pull/4129/commits/575f6eee87e9cd25d8fd5c4a7d12b3d0472f9d04) to better tell what it does. It's still used in `onActiveStateChange` but I don't think we want to change the api here as the consumers are deprecated.

Let me know what you think

m-bert: Looks good!

PR #4135, opened 2026-04-30
[General] Add `yieldsToNativeGestures` property to native gesture

j-piasecki:
## Description

Currently, `NativeViewGestureHandler` exposes `disallowInterruption` property, which causes it to cancel all other gestures upon activation. This is great for the `Touchable` component, since it shouldn't allow other gesture handlers to activate when it's pressed. The issue with that was the fact that `disallowInterruption` means that all native gestures are cancelled as well, including `ScrollView` or any other custom container, making `Touchable` not usable in real use-cases.

This PR adds `yieldsToNativeGestures` property, which works only when `disallowInterruption` is `true`. It defaults to `false`, where it behaves as `disallowInterruption` does currently. When set to `true`, it allows the gesture to be canceled by other native gestures but not any other type of gesture.

Touchable is now using this new config to make it work inside scrollable containers.

## Test plan

Changed `RectButtons` with `Touchables` in the common app

|-|Before|After|
|-|-|-|
|Android|<video src="https://github.com/user-attachments/assets/d5299802-e6f9-4987-aacb-ab41562fc660" />|<video src="https://github.com/user-attachments/assets/5f012d5b-7da7-4d1a-8a32-d380eb09906d" />|
|iOS|<video src="https://github.com/user-attachments/assets/3e0ef50b-c928-4187-99f8-b3e4ea57a8cd" />|<video src="https://github.com/user-attachments/assets/13343460-93a7-47f6-9c90-8a547cb04c34" />|

Since scrolling on web dispatches `pointercancel` event, this prop is effectively a no-op for `Touchable`.

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
@@ -79,11 +79,7 @@ class NativeViewGestureHandler : GestureHandler() {
         return false
       }
     }
-    val canBeInterrupted = !disallowInterruption ||
-      (
-        yieldsToNativeGestures &&
-          (handler is NativeViewGestureHandler || handler is RNGestureHandlerRootHelper.RootViewGestureHandler)
-        )
+    val canBeInterrupted = canBeInterruptedBy(handler)
     val otherState = handler.state
     return if (state == STATE_ACTIVE && otherState == STATE_ACTIVE && canBeInterrupted) {
       // if both handlers are active and the current handler can be interrupted it we return `false`
@@ -98,10 +94,16 @@ class NativeViewGestureHandler : GestureHandler() {
     // otherwise we can only return `true` if already in an active state
   }
 
-  override fun shouldBeCancelledBy(handler: GestureHandler): Boolean = !disallowInterruption ||
+  override fun shouldBeCancelledBy(handler: GestureHandler): Boolean = canBeInterruptedBy(handler)
+
+  /**
+   * Whether this handler permits [other] to take over the touch stream, given its
+   * `disallowInterruption` and `yieldsToNativeGestures` configuration.
+   */
+  fun canBeInterruptedBy(other: GestureHandler): Boolean = !disallowInterruption ||
     (
       yieldsToNativeGestures &&
-        (handler is NativeViewGestureHandler || handler is RNGestureHandlerRootHelper.RootViewGestureHandler)
+        (other is NativeViewGestureHandler || other is RNGestureHandlerRootHelper.RootViewGestureHandler)
       )
 
   override fun shouldBeginWithRecordedHandlers(recorded: List<GestureHandler>): Boolean =
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
@@ -29,6 +29,14 @@ class NativeViewGestureHandler : GestureHandler() {
   var disallowInterruption = false
     private set
 
+  /**
+   * Composes with [disallowInterruption]. When both are `true`, the handler still resists
+   * generic gesture peers (Pan, Tap, etc.) but yields to other [NativeViewGestureHandler] peers
+   * such as a wrapping ScrollView. No-op when [disallowInterruption] is `false`.
+   */
+  var yieldsToNativeGestures = false
+    private set
+
   private var hook: NativeViewGestureHandlerHook = defaultHook
 
   private data class ActiveUpdateSnapshot(val pointerInside: Boolean, val numberOfPointers: Int, val pointerType: Int)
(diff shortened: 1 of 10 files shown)


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -81,7 +98,8 @@ class NativeViewGestureHandler : GestureHandler() {
     // otherwise we can only return `true` if already in an active state
   }
 
-  override fun shouldBeCancelledBy(handler: GestureHandler): Boolean = !disallowInterruption
+  override fun shouldBeCancelledBy(handler: GestureHandler): Boolean = !disallowInterruption ||
+    (yieldsToNativeGestures && handler is NativeViewGestureHandler)
```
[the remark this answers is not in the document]
j-piasecki: [4735e63](https://github.com/software-mansion/react-native-gesture-handler/pull/4135/commits/4735e639cb19acb40a96c2ade70e234d72571fd7)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
```diff
@@ -81,7 +98,11 @@ class NativeViewGestureHandler : GestureHandler() {
     // otherwise we can only return `true` if already in an active state
   }
 
-  override fun shouldBeCancelledBy(handler: GestureHandler): Boolean = !disallowInterruption
+  override fun shouldBeCancelledBy(handler: GestureHandler): Boolean = !disallowInterruption ||
+    (
+      yieldsToNativeGestures &&
+        (handler is NativeViewGestureHandler || handler is RNGestureHandlerRootHelper.RootViewGestureHandler)
+      )
```
m-bert, line 107: wdyt about extracting it to helper, as it is repeated twice?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerInteractionManager.kt
```diff
@@ -50,7 +50,11 @@ class RNGestureHandlerInteractionManager : GestureHandlerInteractionController {
 
   override fun shouldHandlerBeCancelledBy(handler: GestureHandler, otherHandler: GestureHandler): Boolean {
     if (otherHandler is NativeViewGestureHandler) {
-      return otherHandler.disallowInterruption
+      return otherHandler.disallowInterruption &&
+        !(
+          otherHandler.yieldsToNativeGestures &&
+            (handler is NativeViewGestureHandler || handler is RNGestureHandlerRootHelper.RootViewGestureHandler)
+          )
```
m-bert: It would be nice to also extract this one if possible

packages/react-native-gesture-handler/apple/Handlers/RNNativeViewHandler.mm
```diff
@@ -239,9 +241,17 @@ - (void)handleTouchDown:(UIView *)sender forEvent:(UIEvent *)event
...
     // When `disallowInterruption` is set we cancel all gesture handlers when this UIControl
-    // gets DOWN event
+    // gets DOWN event. When `yieldsToNativeGestures` is also set we leave alone:
+    //   - non-RNGH recognizers (e.g. UIScrollView's pan), so native containers can take over
+    //   - peer NativeViewGestureHandler recognizers (RNDummyGestureRecognizer), so wrapping
+    //     RNGH-managed scrollables/native handlers can still take over
     for (RNGHUITouch *touch in [event allTouches]) {
       for (UIGestureRecognizer *recogn in [touch gestureRecognizers]) {
+        if (_yieldsToNativeGestures &&
+            (recogn.gestureHandler == nil || [recogn isKindOfClass:[RNDummyGestureRecognizer class]])) {
```
m-bert: I know it's been there, but wdyt about changing `recogn` to `recognizer`?

j-piasecki: [Rename to recognizer](https://github.com/software-mansion/react-native-gesture-handler/pull/4135/commits/f660f5c897d1c8773eab43809f0e418da4daf9c0)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerInteractionManager.kt
j-piasecki: [Extract logic to canBeInterruptedBy](https://github.com/software-mansion/react-native-gesture-handler/pull/4135/commits/4a3da154fe96c9b7d8559ed8ed1f2c4f46739f12)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/NativeViewGestureHandler.kt
j-piasecki: [Extract logic to canBeInterruptedBy](https://github.com/software-mansion/react-native-gesture-handler/pull/4135/commits/4a3da154fe96c9b7d8559ed8ed1f2c4f46739f12)

m-bert (approved):
LGTM

PR #4136, opened 2026-04-30
[Example] Use `Touchable` instead of `RectButton`

j-piasecki:
## Description

Replaces `RectButton` with `Touchable` in the common app

## Test plan

Run the app

### DIFF
--- a/apps/common-app/App.tsx
+++ b/apps/common-app/App.tsx
@@ -13,7 +13,6 @@ import { useEffect, useState } from 'react';
 import { Dimensions, Platform, StyleSheet, Text, View } from 'react-native';
 import {
   GestureHandlerRootView,
-  RectButton,
   Switch,
   Touchable,
 } from 'react-native-gesture-handler';
@@ -159,7 +158,9 @@ export default function App() {
     }
     return (
       <View style={styles.settings}>
-        <RectButton
+        <Touchable
+          androidRipple={{}}
+          activeUnderlayOpacity={Platform.OS !== 'android' ? 0.1 : 0}
           style={styles.settingsButton}
           onPress={() => {
             updateKeepSetting(!openLastExample);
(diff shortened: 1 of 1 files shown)


apps/common-app/App.tsx
```diff
@@ -11,8 +13,8 @@ import { useEffect, useState } from 'react';
 import { Dimensions, Platform, StyleSheet, Text, View } from 'react-native';
 import {
   GestureHandlerRootView,
-  RectButton,
   Switch,
+  Touchable,
 } from 'react-native-gesture-handler';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
```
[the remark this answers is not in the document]
j-piasecki: [d8337f9](https://github.com/software-mansion/react-native-gesture-handler/pull/4136/commits/d8337f9e400a818c6f7d238f3e0f34c263c1e717)

apps/common-app/App.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @eslint-react/no-nested-component-definitions */
+/* eslint-disable @eslint-react/no-nested-components */
```
m-bert, line 2: I actually thought about this rule when doing some `eslint`-related stuff. I don't think having it as `warn` makes sense, maybe we should disable it (or make it en `error` instead)

m-bert (approved):
https://github.com/software-mansion/react-native-gesture-handler/pull/4133

apps/common-app/App.tsx
j-piasecki: Yeah, keeping it as an error makes the most sense IMO.

m-bert: Okay, this one we can do in other PR, feel free to merge it.

PR #4146, opened 2026-05-06
[Web] Add hover effects to the `Touchable` component

j-piasecki:
## Description

Adds `hover*` props to the Touchable component, which are currently web-only. Those take effect when the pointer moves over the button without pressing it. `active` props have higher precedence than `hover`.

## Test plan

Updated touchables example

### DIFF
--- a/packages/docs-gesture-handler/docs/components/touchable.mdx
+++ b/packages/docs-gesture-handler/docs/components/touchable.mdx
@@ -172,6 +172,22 @@ defaultOpacity?: number;
 
 Defines the opacity of the whole component when the button is active. By default set to `1`.
 
+### activeScale
+
+```ts
+activeScale?: number;
+```
+
+Defines the scale of the whole component when the button is active.
+
+### defaultScale
+
+```ts
+defaultScale?: number;
+```
+
+Defines the scale of the whole component when the button is inactive. By default set to `1`.
+
 ### activeUnderlayOpacity
 
 ```ts
@@ -187,6 +203,46 @@ defaultUnderlayOpacity?: number;
 ```
 
 Defines the initial opacity of underlay when the button is inactive. By default set to `0`.
+
+<Badges platforms={['web']}>
+### hoverOpacity
+</Badges>
+
+```ts
+hoverOpacity?: number;
+```
+
+Defines the opacity of the whole component when the button is hovered. By default falls back to [`defaultOpacity`](#defaultopacity).
+
+<Badges platforms={['web']}>
+### hoverScale
+</Badges>
+
+```ts
+hoverScale?: number;
+```
+
+Defines the scale of the whole component when the button is hovered. By default falls back to [`defaultScale`](#defaultscale).
+
+<Badges platforms={['web']}>
+### hoverUnderlayOpacity
+</Badges>
+
+```ts
+hoverUnderlayOpacity?: number;
+```
+
+Defines the opacity of the underlay when the button is hovered. By default falls back to [`defaultUnderlayOpacity`](#defaultunderlayopacity).
+
+<Badges platforms={['web']}>
+### hoverAnimationDuration
+</Badges>
+
+```ts
+hoverAnimationDuration?: number;
+```
+
+Duration of the hover animation, in milliseconds.
 
 ### underlayColor
 
--- a/packages/docs-gesture-handler/docs/components/touchable.mdx
+++ b/packages/docs-gesture-handler/docs/components/touchable.mdx
@@ -242,7 +242,7 @@ Defines the opacity of the underlay when the button is hovered. By default falls
 hoverAnimationDuration?: number;
 ```
 
-Duration of the hover animation, in milliseconds.
+Duration of the hover animation, in milliseconds. By default falls back to [`tapAnimationDuration`](#tapanimationduration).
 
 ### underlayColor
 
--- a/packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
+++ b/packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
@@ -89,24 +89,32 @@ export interface ButtonProps extends ViewProps, AccessibilityProps {
   activeUnderlayOpacity?: number | undefined;
 
   /**
+   * Web only.
+   *
    * Opacity applied to the button when it is hovered. Defaults to
    * `defaultOpacity` when not set.
    */
   hoverOpacity?: number | undefined;
 
   /**
+   * Web only.
+   *
    * Scale applied to the button when it is hovered. Defaults to
    * `defaultScale` when not set.
    */
   hoverScale?: number | undefined;
 
   /**
+   * Web only.
+   *
    * Opacity applied to the underlay when the button is hovered. Defaults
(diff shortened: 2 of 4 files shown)


packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -190,10 +252,11 @@ export const ButtonComponent = ({
           ...(hasUnderlay && { overflow: 'hidden' }),
         },
       ]}
+      onPointerEnter={handlePointerEnter}
       onPointerDown={pressIn}
       onPointerUp={pressOut}
       onPointerCancel={pressOut}
-      onPointerLeave={pressOut}
+      onPointerLeave={handlePointerLeave}
       {...rest}>
```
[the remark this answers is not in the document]
j-piasecki: [Spread rest before other props](https://github.com/software-mansion/react-native-gesture-handler/pull/4146/commits/5b8d1559f631e3105aac396cd50ff5bdb0fd2d2f)

packages/docs-gesture-handler/docs/components/touchable.mdx
```diff
@@ -188,6 +204,46 @@ defaultUnderlayOpacity?: number;
...
+
+<Badges platforms={['web']}>
+### hoverAnimationDuration
+</Badges>
+
+```ts
+hoverAnimationDuration?: number;
+```
+
+Duration of the hover animation, in milliseconds.
```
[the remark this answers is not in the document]
j-piasecki: [Update docs](https://github.com/software-mansion/react-native-gesture-handler/pull/4146/commits/14f5b6501eae783e88fd98dcfc6abd6dee5f58a6)

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -156,14 +174,58 @@ export const ButtonComponent = ({
     [pressAndHoldAnimationDuration, tapAnimationDuration]
   );
 
+  const handlePointerEnter = React.useCallback(
+    (event: NativeSyntheticEvent<{ pointerType?: string }>) => {
+      if (!enabled || event.nativeEvent.pointerType === 'touch') {
```
m-bert, line 179: Shouldn't this check against `mouse` and `pen` rather than `touch`? 
https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType#value

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -156,14 +174,58 @@ export const ButtonComponent = ({
...
+      }
+      setHovered(true);
+    },
+    [enabled, pressed, hoverAnimationDuration]
+  );
+
+  const handlePointerLeave = React.useCallback(
+    (event: NativeSyntheticEvent<{ pointerType?: string }>) => {
+      pressOut(event);
+      if (event.nativeEvent.pointerType === 'touch') {
```
m-bert, line 194: Same as above

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -156,14 +174,58 @@ export const ButtonComponent = ({
...
+        setCurrentDuration(hoverAnimationDuration);
+      }
+      setHovered(true);
+    },
+    [enabled, pressed, hoverAnimationDuration]
+  );
+
+  const handlePointerLeave = React.useCallback(
+    (event: NativeSyntheticEvent<{ pointerType?: string }>) => {
+      pressOut(event);
```
m-bert, line 193: Should `pointerLeave` manipulate `pressed` flag?

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -40,8 +48,18 @@ export const ButtonComponent = ({
     pressAndHoldAnimationDurationProp < 0
       ? tapAnimationDuration
       : pressAndHoldAnimationDurationProp;
+  const hoverAnimationDuration =
+    hoverAnimationDurationProp < 0
+      ? tapAnimationDuration
```
m-bert, line 53: What about some general `animationDuration` and using it in both cases?

packages/docs-gesture-handler/docs/components/touchable.mdx
m-bert, line 1: I wouldn't link to other properties to describe default values. Looking at `hoverAnimationDuration`:

> By default falls back to [`tapAnimationDuration`](#tapanimationduration).

It doesn't tell much. It is more implementation detail that end-user does not care about. Also it looks weird (why would it resolve to tap duration?). Having just number makes it easier to look at and you don't have to jump over docs to find the actual value.

packages/react-native-gesture-handler/src/components/GestureHandlerButton.tsx
m-bert, line 1: Same as in docs (though this one is not that visible)

m-bert (commented):
Love the idea! 🚀

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
j-piasecki: This is an early return - do nothing for `touch`.

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
j-piasecki: Same

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
j-piasecki: I believe so - the press animation shouldn't be active after leaving the bouds of the button.

packages/docs-gesture-handler/docs/components/touchable.mdx
j-piasecki: It's not falling back to the default of `tapAnimationDuration`, it's falling back to `tapAnimationDuration`.

If you set `tapAnimationDuration` to `500` and leave `hoverAnimationDuration` undefined, it will default to `500`.

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
j-piasecki: Yeah, the props situation can definitely be improved. Do you have anything specific in mind?

packages/docs-gesture-handler/docs/components/touchable.mdx
m-bert: Okay, I see it now

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
m-bert: Had one, but seems like it is not relevant as I misunderstood the concept.

m-bert (approved):

PR #4156, opened 2026-05-07
[Android] Remove getChildInDrawingOrderAtIndex

javache:
## Summary
  - Remove the `getChildInDrawingOrderAtIndex` abstraction from `ViewConfigurationHelper` and its implementation in `RNViewConfigurationHelper`
  - Inline `viewGroup.getChildAt(i)` at the two call sites (`GestureHandlerOrchestrator` and `HoverGestureHandler`)
  - `ReactViewGroup.getZIndexMappedChildIndex` is a legacy architecture API that has been a no-op (returns its input unchanged) and has been removed in https://github.com/facebook/react-native/pull/56717

  ## Test plan
  - Android build via `cd apps/basic-example && yarn android`


j-piasecki (approved):
Thank you ❤️

PR #4158, opened 2026-05-08
Fix handling keyboard should persists taps

### LINKED ISSUE #3926: [ScrollView] `keyboardShouldPersistTaps="handled"` doesn't work
### Description

The RNGH ScrollView doesn't count `Pressable` taps or `Gesture.Tap()` taps as handled in the context of `keyboardShouldPersistTaps`. This is both on iOS and Android.

For the purpose of demonstration I set up the same example with RN components and RNGH components:

https://github.com/user-attachments/assets/f567b6a3-3780-4a46-87eb-bf928e0ac60e

Here is the code:

```tsx
import {
  Pressable as RNPressable,
  ScrollView as RNScrollView,
  TextInput as RNTextInput,
} from 'react-native'
import {
  Pressable as RNGHPressable,
  ScrollView as RNGHScrollView,
  TextInput as RNGHTextInput,
} from 'react-native-gesture-handler'

 <RNScrollView
  keyboardShouldPersistTaps="handled"
  contentContainerStyle={{ height: 200, backgroundColor: 'red' }}
>
  <RNPressable
    onPress={() => {}}
  >
    <RNTextInput value="RN input" />
  </RNPressable>
</RNScrollView>

<RNGHScrollView
  keyboardShouldPersistTaps="handled"
  contentContainerStyle={{ height: 200, backgroundColor: 'blue' }}
>
  <RNGHPressable
    onPress={() => {}}
  >
    <RNGHTextInput value="RNGH input" />
  </RNGHPressable>
</RNGHScrollView>
```

### Steps to reproduce

1. Run the code from the description
2. Tap the text input to focus it
3. Tap on the green Pressable, the keyboard shouldn't hide

-

### Gesture Handler version

2.30.0

### React Native version

0.81

### Platforms

iOS, Android

### JavaScript runtime

None

### Workflow

None

### Architecture

New Architecture (Fabric)

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

coado:
## Description

The `keyboardShouldPersistTaps="handled"` on the ScrollView persists the keyboard if there is a target that could claim the JS responder event in the bubble phase (bottom-up). The diff adds `onStartShouldSetResponder` method that returns `true` (indicating that the target is willing to claim touch event) to `GHPressable` and to the `NativeDetector` whenever the single tap gesture is passed.

For the `never` case, the keyboard is always dropped by the touch event captured by the React Native ScrollView. This is done on the JS side, hence the touch is still handled by the Gesture Handler Pressable, which is opposite to the React Native Pressable for which the touch dismissing the keyboard is dropped. 

Fixes #3926 

## Test plan

https://github.com/user-attachments/assets/b1a78a67-749e-41ec-aa2d-95dbfc6d123d


packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
```diff
@@ -365,9 +365,21 @@ const Pressable = (props: PressableProps) => {
...
+  // Claim the JS responder in the bubble phase so an ancestor RN ScrollView
+  // with keyboardShouldPersistTaps='handled' or 'always' cannot run its
+  // release-time TextInput blur. The GH tap still fires via the native
+  // recognizer; when it does, UIKit cancels touches.
+  const handleStartShouldSetResponder = useCallback(() => {
+    if (__DEV__) {
+      console.log('[GH Pressable] onStartShouldSetResponder -> true');
+    }
+    return true;
+  }, []);
```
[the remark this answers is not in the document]
coado: Fantastic catch

coado:
@copilot one more round?

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -55,8 +66,18 @@ export function NativeDetector<
...
+
+  const handleStartShouldSetResponder = useCallback(() => {
+    return !isComposedGesture(gesture) && isGestureEnabled(gesture);
+  }, [gesture]);
+
   return (
     <NativeDetectorComponent
+      onStartShouldSetResponder={
+        isTapGesture ? handleStartShouldSetResponder : undefined
+      }
```
j-piasecki: Won't this conflict with https://github.com/software-mansion/react-native-gesture-handler/pull/4094?

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -55,8 +66,18 @@ export function NativeDetector<
             gesture.detectorCallbacks.reanimatedEventHandler,
         };
 
+  const isTapGesture =
+    !isComposedGesture(gesture) && gesture.type === SingleGestureName.Tap;
+
+  const handleStartShouldSetResponder = useCallback(() => {
+    return !isComposedGesture(gesture) && isGestureEnabled(gesture);
+  }, [gesture]);
```
j-piasecki: This will not react to changes if a shared value is passed to the config. Also, I think reading the value of SVs on the JS thread will result in a warning.

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -55,8 +66,18 @@ export function NativeDetector<
             gesture.detectorCallbacks.reanimatedEventHandler,
         };
 
+  const isTapGesture =
+    !isComposedGesture(gesture) && gesture.type === SingleGestureName.Tap;
```
j-piasecki: Why restrict it only to a tap gesture? And why skip composed gestures?

Even if we only allow tap, then tap + double tap should be fine.

j-piasecki (commented):
Do we need to integrate with the js responder here?

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
coado: Shouldn't because for `cancelsJSResponder` to do anything the gesture has to be recognized first and `onStartShouldSetResponder` is invoked before that happens. Will verify that though.

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
coado: What would you recommend in this scenario? It seems to be necessary to check if the gesture is enabled.

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
coado: I wasn't sure for which gestures we want it so I started with the simplest one and hoped for some discussion here 😄

coado:
> Do we need to integrate with the js responder here?

Discussed offline:
  - the JS responder event needs to be handled before the ScrollView becomes the responder,
  - we cannot handle the event directly in GH components, because that would prevent React Native components higher in the tree from receiving the touch in some use cases,
  - we cannot disable the JS responder handling on ScrollView and handle this natively, because that would change the behavior of `keyboardShouldPersistTaps`.

Instead, we add a context provider inside `ScrollView`. `NativeDetector` and RNGH `Pressable` mark this context from their `onStartShouldSetResponder` callbacks, but they do not claim the responder themselves.

 If there is a React Native responder target between the GH component and the ScrollView, it can still handle the event first. Otherwise, the event continues bubbling until it reaches the dummy responder component rendered directly inside the ScrollView. That component checks the context marker and, for `keyboardShouldPersistTaps="handled"`, consumes the event before it reaches the ScrollView responder logic.

The solution was measured by rendering 500 gesture detectors and 500 Pressables on gesture handler with and without this change. No regression has been observed. 

<details>
<summary>Benchmark code</summary>


```ts
import { render } from '@testing-library/react-native';
import { act } from 'react';

import GestureHandlerRootView from '../components/GestureHandlerRootView';
import { Pressable, ScrollView } from '../v3/components';
import { GestureDetector } from '../v3/detectors';
import { useTapGesture } from '../v3/hooks/gestures';

declare const process: {
  env: Record<string, string | undefined>;
};

const COUNT = Number(process.env.RNGH_PERF_COUNT ?? 500);
const WARMUP = Number(process.env.RNGH_PERF_WARMUP ?? 10);
const SAMPLES = Number(process.env.RNGH_PERF_SAMPLES ?? 50);
const VARIANT = process.env.RNGH_PERF_VARIANT ?? 'unknown';

const flushImmediate = () =>
  new Promise((resolve) => {
    setImmediate(() => resolve(undefined));
  });

const now = () => performance.now();

function TapItem() {
  const tap = useTapGesture({ disableReanimated: true });

  return <GestureDetector gesture={tap} />;
}

function ManyGestureDetectors() {
  return (
    <GestureHandlerRootView>
      <ScrollView keyboardShouldPersistTaps="handled">
        {Array.from({ length: COUNT }, (_, index) => (
          <TapItem key={`detector-${index}`} />
        ))}
        {Array.from({ length: COUNT }, (_, index) => (
          <Pressable key={`pressable-${index}`} />
        ))}
      </ScrollView>
    </GestureHandlerRootView>
  );
}

function percentile(sortedValues: number[], percent: number) {
  const index = (sortedValues.length - 1) * percent;
  const lower = Math.floor(index);
  const upper = Math.ceil(index);

  if (lower === upper) {
    return sortedValues[lower];
  }

  return (
    sortedValues[lower] +
    (sortedValues[upper] - sortedValues[lower]) * (index - lower)
  );
}

function stats(values: number[]) {
  const sorted = [...values].sort((a, b) => a - b);
  const mean = values.reduce((sum, value) => sum + value, 0) / values.length;
  const variance =
    values.reduce((sum, value) => sum + (value - mean) ** 2, 0) /
    (values.length - 1);

  return {
    count: values.length,
    mean,
    median: percentile(sorted, 0.5),
    p95: percentile(sorted, 0.95),
    min: sorted[0],
    max: sorted[sorted.length - 1],
    sd: Math.sqrt(variance),
  };
}

describe('RNGH context mount performance', () => {
  test('mounts hundreds of GestureDetectors inside ScrollView', async () => {
    const measurements: number[] = [];

    for (let iteration = 0; iteration < WARMUP + SAMPLES; iteration += 1) {
      const startedAt = now();
      const screen = render(<ManyGestureDetectors />);

      await act(flushImmediate);

      const duration = now() - startedAt;
      screen.unmount();

      if (iteration >= WARMUP) {
        measurements.push(duration);
      }
    }

    console.log(
      'RNGH_CONTEXT_PERF_RESULT ' +
        JSON.stringify({
          variant: VARIANT,
          gestureDetectors: COUNT,
          pressables: COUNT,
          warmup: WARMUP,
          samples: SAMPLES,
          mountMs: stats(measurements),
        })
    );
  });
});
``` 

</details>

apps/common-app/src/new_api/tests/keyboardShouldPersistTaps/index.tsx
```diff
@@ -0,0 +1,369 @@
...
+const EXAMPLES: Example[] = ['pressable', 'tap'];
+
+const EXAMPLE_LABELS: Record<Example, string> = {
+  pressable: 'GH Pressable',
+  tap: 'useTapGesture',
+};
+
+const MODE_DESCRIPTIONS: Record<Mode, string> = {
+  never:
+    "RN: first tap outside the input dismisses the keyboard AND is swallowed — press doesn't fire. GH: keyboard still dismisses (ScrollView captures the responder normally), but the press ALSO fires because GH's native recognizer runs in parallel to the JS responder system.",
```
j-piasecki, line 35: Can we handle that in RNGH as well?

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -49,12 +52,63 @@ const GHScrollView = createNativeWrapper<
...
+        collapsable={false}
+        onStartShouldSetResponderCapture={resetRNGHResponderEvent}
+        onStartShouldSetResponder={handleStartShouldSetResponder}
+        pointerEvents="box-none"
+        style={styles.logicalResponder}>
+        {children}
+      </View>
+    </JSResponderContext>
+  );
+};
```
j-piasecki: Can we keep the logic related to this feature in a single place? Maybe rename `JSResponderContext.ts`?

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -55,8 +69,29 @@ export function NativeDetector<
...
+    if (shouldHandleResponderEvent) {
+      const responderEventRef = jsResponderContext?.isRNGHResponderEvent;
+
+      if (responderEventRef) {
+        responderEventRef.current = true;
+      }
+    }
+
+    return false;
+  }, [gesture, jsResponderContext]);
```
j-piasecki: This can also be extracted to a single hook, along with a call to `use(JSResponderContext);`

Also, can it be updated so that if called outside of the context, it returns some stable no-op function to avoid allocation?

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
j-piasecki: I don't see why it should be limited only to some gestures.

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -1,13 +1,25 @@
...
 import HostGestureDetector from './HostGestureDetector';
 import { ReanimatedNativeDetector } from './ReanimatedNativeDetector';
 import { configureRelations, ensureNativeDetectorComponent } from './utils';
 
+function isGestureEnabled<
+  TConfig,
+  THandlerData,
+  TExtendedHandlerData extends THandlerData,
+>(gesture: SingleGesture<TConfig, THandlerData, TExtendedHandlerData>) {
+  return maybeUnpackValue(gesture.config.enabled) !== false;
```
j-piasecki: Have you tried this with an actual shared value? Did you gen any warnings?

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
coado: I think for some gestures you may want to not have this behavior, while keeping the keyboard opened for presses only.

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
coado: I didn't get any warning.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -16,8 +16,9 @@
 @interface RNGestureHandlerDetector () <RCTRNGestureHandlerDetectorViewProtocol>
 
 @property (nonatomic, nonnull) NSMutableSet *nativeHandlers;
+@property (nonatomic, nonnull) NSMutableSet *subscribedHandlers;
```
j-piasecki, line 19: I don't think this should be here

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -55,8 +73,25 @@ export function NativeDetector<
...
+    if (shouldHandleJSResponderEvent()) {
+      const responderEventRef = jsResponderContext?.isRNGHResponderEvent;
+
+      if (responderEventRef) {
+        responderEventRef.current = true;
+      }
+    }
+
+    return false;
+  }, [jsResponderContext, shouldHandleJSResponderEvent]);
```
j-piasecki: I think that this, along with `const jsResponderContext = use(JSResponderContext);` should be its own hook.

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -55,8 +73,25 @@ export function NativeDetector<
             gesture.detectorCallbacks.reanimatedEventHandler,
         };
 
+  const shouldHandleJSResponderEvent = useCallback(() => {
+    return isGestureEnabled(gesture);
+  }, [gesture]);
+
+  const handleStartShouldSetResponder = useCallback(() => {
+    if (shouldHandleJSResponderEvent()) {
```
j-piasecki: This should be called only when jsResponderContext exists

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
j-piasecki: Only reading during render shows a warning. Reading in a callback is fine, so no issue with that. To listen to SV changes you can do something like this: https://github.com/software-mansion/react-native-gesture-handler/blob/f2159b8aee070e192025161cd88fa6808fb89848/packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts#L47

packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
j-piasecki: I'm not sure, but if so, I'd probably stick to discrete (tap, longpress, fling, native) and cancel for continuous (pan, pinch, rotate, manual). Hover is the weird one - I'd assume it shouldn't be hiding keyboard, so it would go with discrete ones.

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
coado: oh, not this PR 😄

apps/common-app/src/new_api/tests/keyboardShouldPersistTaps/index.tsx
coado: Yes, just like the description says it has a slightly different side-effect though.

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
```diff
@@ -75,6 +75,9 @@ let Reanimated:
       useComposedEventHandler<T>(
         handlers: (((event: T) => void) | null)[]
       ): (event: T) => void;
+      runOnJS<A extends unknown[], R>(
```
j-piasecki, line 79: `runOnJS` (and `runOnUI` for that matter) are deprecated in Reanimated. Can you add a TODO comment to remove the usages of these?

j-piasecki (approved):
LGTM! I'd love to have 3.0.2 released sometime this week with the fixes applied in recent days, and keep this one for 3.1.0 since it's a bigger change. Please wait with merging it.

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+  THandlerData,
+  TExtendedHandlerData extends THandlerData,
+>(gesture: Gesture<TConfig, THandlerData, TExtendedHandlerData>): boolean {
+  if (isComposedGesture(gesture)) {
+    // For composed gestures, we need to check if at least one of the composed gestures is enabled
+    return gesture.gestures.some(isGestureEnabled);
+  }
+
+  return maybeUnpackValue(gesture.config.enabled) !== false;
+}
```
m-bert: Looks like it could be in [`configUtils.ts`](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/v3/hooks/utils/configUtils.ts)

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+    return [];
+  }
+
+  if (isComposedGesture(gesture)) {
+    return gesture.gestures.flatMap(getEnabledSharedValues);
+  }
+
+  const enabled = gesture.config.enabled;
+  return Reanimated.isSharedValue<boolean>(enabled) ? [enabled] : [];
+}
```
m-bert: This also could be in [`configUtils.ts`](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/v3/hooks/utils/configUtils.ts) or [`reanimatedUtils.ts`](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts).

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
coado: Added in https://github.com/software-mansion/react-native-gesture-handler/pull/4158/commits/97d894b2f35eda6d6b94f44273913ba4923aefd2

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+  const [enabledSharedValueRevision, setEnabledSharedValueRevision] =
+    useState(0);
+  const listenerIdRef = useRef<number | null>(null);
+
+  if (listenerIdRef.current === null) {
+    listenerIdRef.current = nextJSResponderContextListenerId++;
+  }
+
+  useEffect(() => {
+    const reanimated = Reanimated;
```
m-bert, line 54: Could you check if we can use `Reanimated` directly, as in other places?

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+      notifyEnabledChanged
+    );
+
+    return () => {
+      reanimated.runOnUI(detachListeners)(enabledSharedValues, listenerId);
+    };
+  }, [gesture]);
+
+  const shouldHandleJSResponderEvent = useCallback(() => {
+    void enabledSharedValueRevision;
```
m-bert, line 103: why?

packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
```diff
@@ -392,11 +395,27 @@ const Pressable = (props: PressableProps) => {
...
+    if (!disabled) {
+      const responderEventRef = jsResponderContext?.isRNGHResponderEvent;
+
+      if (responderEventRef) {
+        responderEventRef.current = true;
+      }
+    }
+
+    return false;
+  }, [disabled, jsResponderContext]);
```
m-bert, line 410: Isn't it the same as in `useJSResponderHandler`? If so, maybe we could extract it?

m-bert: Just a thought, let me know what you think

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -53,8 +54,11 @@ export const ScrollView = (
   props: RNScrollViewProps & NativeWrapperProperties<RNScrollView | null>
 ) => {
   const {
+    children,
     refreshControl,
     onGestureUpdate_CAN_CAUSE_INFINITE_RERENDER,
+    horizontal,
```
m-bert: Does it need to be unpacked?

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -85,8 +91,12 @@ export const ScrollView = (
               scrollGesture ? { block: scrollGesture } : {}
             )
           : undefined
-      }
-    />
+      }>
+      <ScrollViewResponderInterceptor
+        keyboardShouldPersistTaps={keyboardShouldPersistTaps}>
```
m-bert, line 94: Does it have to be on both, `ScrollView` and `Interceptor`?

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
coado: haaland-whynot

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
coado: done in https://github.com/software-mansion/react-native-gesture-handler/pull/4158/changes/e33d436b633129939bd3a1958873194cabdae1be

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
coado: done in [24a339b](https://github.com/software-mansion/react-native-gesture-handler/pull/4158/commits/24a339b9972ba9b4054766fe697dd914bd0d00fa)

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
coado: It is for type narrowing. The `Reanimated` type is back to possibly being `undefined` in the cleanup callback.

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
coado: It is only to annotate that the `enabledSharedValueRevision` is needed in the callback (in the dep list to rerender) but is not used directly. It's just to make linter happy.

packages/react-native-gesture-handler/src/v3/components/Pressable.tsx
coado: Yeah, we can extract that https://github.com/software-mansion/react-native-gesture-handler/pull/4158/changes/7dd2dd23b6c130596e4d73f9c5639dcc75426bb2

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
coado: Right, fixed in [4ddf227](https://github.com/software-mansion/react-native-gesture-handler/pull/4158/commits/4ddf22706b2ffa11a45f3d73e9bd256b214688ac)

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
coado: Yeah, `ScrollViewResponderInterceptor` needs to know its value and we also pass it to `ScrollView` to preserve normal behavior.

packages/react-native-gesture-handler/src/v3/hooks/useJSResponderHandler.ts
m-bert: Okay, we tend to use `?` with `Reanimated`, but I'm fine with that approach.

m-bert (approved):

PR #4160, opened 2026-05-08
Fix `RNRenderer` import for React Native 0.86+

zoontek:
## Description

React Native 0.86 removed `Libraries/Renderer/shims/ReactNative`, breaking the `RNRenderer` import. This PR switch to `RendererProxy` and fall back to the legacy shim for compatibility with RN < 0.86.


m-bert (changes requested):
HI @zoontek! Thank you for this PR! 

I've checked that and it seems that it doesn't work 😞 I've also did this on standalone RN 0.86 app outside of our monorepo with the same result. Nesting second `require` in `try/catch` block fixes this problem.

What we've found out is that `RNRenderer` is used only in [one place](https://github.com/software-mansion/react-native-gesture-handler/blob/40bbfa0cda57a40ba1565ed241d9f4cd23776344/packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/utils.ts#L158), to validate `GestureDetector` children. However, this is already covered [earlier](https://github.com/software-mansion/react-native-gesture-handler/blob/40bbfa0cda57a40ba1565ed241d9f4cd23776344/packages/react-native-gesture-handler/src/handlers/gestures/GestureDetector/Wrap.tsx#L28) in the flow. So our conclusion is that it probably can be safely removed instead of changed.

Would you like to do this, or should we handle it? 😅

zoontek:
@m-bert Done ✅

m-bert (approved):
Thank you!

PR #4173, opened 2026-05-13
[Native] Fix exiting animation crash

j-piasecki:
## Description

@tomekzaw reported a crash with the `Touchable` component when used with `createAnimatedComponent` and layout animations. The specific prop was the `exiting` animation. It caused the native detector view to outlive the relevant gesture hook, which drops the native recognizer on unmount. Due to this, the exiting detector may have tried to reattach a non-existent recognizer, causing the crash.

This PR adds a check not to try reattaching already attached recognizers, which will effectively "freeze" the detector for the duration of the animation, unless its props change (which shouldn't happen, as its react node is gone at that point).

Android wasn't crashing (`attachHandlerToView` no-ops on non-existent handlers), but reattaching it would cause it to be canceled. For symmetry, I added the same guard there.

## Test plan

Rechecked reproducer from https://github.com/software-mansion/react-native-gesture-handler/pull/3672 to make sure it didn't regress.

Tested on the crashing app.


tomekzaw (approved):
Thank you 🥹

m-bert (approved):

PR #4174, opened 2026-05-13
Delay dropping native gesture handler on StrictMode

### LINKED ISSUE #4172: [v3] App crashes when using React.StrictMode
### Description

Gesture Handler v3 seems incompatible with `React.StrictMode`. When the app is wrapped in strict mode:

- On iOS the app instantly crashes. In the XCode logs, I see `*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Handler for tag 1 does not exists'`
- On Android, I don't see a relevant log or error, but the gestures don't work

https://github.com/user-attachments/assets/5ae17952-8d12-433e-a4d3-307e58c57f56

https://github.com/user-attachments/assets/232913b5-5f2d-42b7-97b8-fef8d0f01a67

Removing `React.StrictMode` makes the gestures work both on Android & iOS.

PS: The repro uses Community CLI, but I'm also had an error on Expo in my existing code. Though the error was different:

Image

### Steps to reproduce

1. Clone `git@github.com:satya164/GestureHandlerStrictModeRepro.git`
2. Build and run the app on Android emulator and iOS simulator
3. Observe gesture not working/app crashing
4. Edit `App.tsx` to remove `React.StrictMode` and observe that the gestures work

https://github.com/satya164/GestureHandlerStrictModeRepro

### Gesture Handler version

3.0.0-beta.4

### React Native version

0.85.3

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

None

### Build type

Debug mode

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

coado:
## Description

In the strict mode React does a simulated unmount/remount of effects (without fully unmounting/remounting the component instance). This causes [dropping](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts#L102) the native gesture handler and rerunning the [useEffect](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts#L107-L113) with the same `handlerTag` that was already dropped. 

The solution is to delay dropping the gesture handler with `ghQueueMicrotask` and increment a helper token on each effect setup and cleanup. During StrictMode’s simulated unmount/remount, the replayed setup increments the token before the delayed drop runs. The delayed callback then sees that its captured token no longer matches the current token and skips the drop.

On a real unmount, no setup runs afterward for that hook instance, so the token remains unchanged and the delayed drop proceeds. 

Fixes #4172 

## Test plan

I've used the repro from the #4172

### DIFF
--- a/packages/react-native-gesture-handler/src/__tests__/api_v3.test.tsx
+++ b/packages/react-native-gesture-handler/src/__tests__/api_v3.test.tsx
@@ -1,12 +1,27 @@
 import { render, renderHook } from '@testing-library/react-native';
-import { act } from 'react';
+import React, { act } from 'react';
 
 import GestureHandlerRootView from '../components/GestureHandlerRootView';
 import { fireGestureHandler, getByGestureTestId } from '../jestUtils';
+import RNGestureHandlerModule from '../RNGestureHandlerModule';
 import { State } from '../State';
 import { RectButton, Touchable } from '../v3/components';
 import { usePanGesture } from '../v3/hooks/gestures';
 import type { SingleGesture } from '../v3/types';
+
+async function flushQueuedOperations() {
+  await act(async () => {
+    await new Promise<void>((resolve) => {
+      setImmediate(() => resolve());
+    });
+  });
+}
+
+afterEach(async () => {
+  await flushQueuedOperations();
+  await flushQueuedOperations();
+  jest.restoreAllMocks();
+});
 
 describe('[API v3] Hooks', () => {
   test('Pan gesture', () => {
@@ -30,6 +45,32 @@ describe('[API v3] Hooks', () => {
 
     expect(onBegin).toHaveBeenCalledTimes(1);
     expect(onStart).toHaveBeenCalledTimes(1);
+  });
+
+  test('does not drop native handler during StrictMode effect replay', async () => {
+    const dropGestureHandlerSpy = jest.spyOn(
+      RNGestureHandlerModule,
+      'dropGestureHandler'
+    );
+
+    const StrictModeWrapper = ({ children }: React.PropsWithChildren) => (
+      <React.StrictMode>{children}</React.StrictMode>
+    );
+
+    const { unmount } = renderHook(
+      () => usePanGesture({ disableReanimated: true }),
+      { wrapper: StrictModeWrapper }
+    );
+
+    await flushQueuedOperations();
+
+    expect(dropGestureHandlerSpy).not.toHaveBeenCalled();
+
+    unmount();
+    await flushQueuedOperations();
+    await flushQueuedOperations();
+
+    expect(dropGestureHandlerSpy).toHaveBeenCalledTimes(1);
   });
 });
 
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
@@ -105,7 +105,12 @@ export function useGesture<
       const dropGestureHandlerToken = ++dropGestureHandlerTokenRef.current;
 
       ghQueueMicrotask(() => {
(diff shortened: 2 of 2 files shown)


coado:
@copilot review please

j-piasecki (changes requested):
I don't think this is the way to go - it's a workaround for strict mode, while it doesn't solve the underlying issue. When the hook is rendered using concurrent features it will either try to create the same gesture twice or end up with no gesture on the native side.

I think the correct approach should be moving management of the native gesture fully into the react lifecycle, so it's always created/deteled in an effect. One issue is that creation is synchronous, while deletion is not (and I don't think it can be - ui operations should run on main queue, and at this point dropping gesture likely means detaching it from a view), so the tag most likely needs to be advanced. Something along the lines of
```diff
diff --git a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
index 6548575b2..092880dc0 100644
--- a/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
+++ b/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useRef } from 'react';
+import { useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 
 import { getNextHandlerTag } from '../../handlers/getNextHandlerTag';
 import {
@@ -29,7 +29,8 @@ export function useGesture<
   type: SingleGestureName,
   config: BaseGestureConfig<TConfig, THandlerData, TExtendedHandlerData>
 ): SingleGesture<TConfig, THandlerData, TExtendedHandlerData> {
-  const handlerTag = useMemo(() => getNextHandlerTag(), []);
+  const [handlerTag, setHandlerTag] = useState(getNextHandlerTag);
+  const createdHandlerTag = useRef<number | undefined>(handlerTag);
   const disableReanimated = useMemo(() => config.disableReanimated, []);
 
   if (config.disableReanimated !== disableReanimated) {
@@ -61,15 +62,6 @@ export function useGesture<
     [handlerTag, config.simultaneousWith, config.requireToFail, config.block]
   );
 
-  const currentGestureRef = useRef({ type: '', handlerTag: -1 });
-  if (
-    currentGestureRef.current.handlerTag !== handlerTag ||
-    currentGestureRef.current.type !== (type as string)
-  ) {
-    currentGestureRef.current = { type, handlerTag };
-    NativeProxy.createGestureHandler(type, handlerTag, {});
-  }
-
   const gesture = useMemo(
     () => ({
       handlerTag,
@@ -93,14 +85,23 @@ export function useGesture<
     ]
   );
 
-  useEffect(() => {
-    return () => {
-      if (currentGestureRef.current.handlerTag === handlerTag) {
-        currentGestureRef.current = { type: '', handlerTag: -1 };
-      }
+  useLayoutEffect(() => {
+    if (createdHandlerTag.current !== handlerTag) {
+      const nextHandlerTag = getNextHandlerTag();
+      setHandlerTag(nextHandlerTag);
+      createdHandlerTag.current = nextHandlerTag;
+      return;
+    }
 
+    NativeProxy.createGestureHandler(type, handlerTag, {});
+    console.log('Created gesture handler with tag', handlerTag, 'and type', type);
+
+    return () => {
+      console.log('Dropping gesture handler with tag', handlerTag, 'and type', type);
       NativeProxy.dropGestureHandler(handlerTag);
       scheduleFlushOperations();
+
+      createdHandlerTag.current = undefined;
     };
   }, [type, handlerTag]);
```

Which works on iOS, doesn't on web, and I don't have an android build on hand to check.

Can you look more in this direction?

coado:
Closing this one as we discussed offline to proceed with different workaround in #4179 and #4180.

PR #4179, opened 2026-05-18
[General] Change how gestures are attached to the native detector to fix StrictMode

j-piasecki:
## Description

1. Changes `Module.createGestureHandler` to be asynchronous
2. Changes `NativeProxy.createGestureHandler` to use the same scheduling as drop to preserve relative ordering
3. Moves the creation of the native handler to the effect from the render phase
4. Updates the native implementations of `GestureDetector` not to rely on synchronous creation of gesture handlers. Instead, introduces the ability to register a callback to be fired once a gesture is created. This way, the detector can attach gestures as they are created on the native side with no dependency on JS (the only required thing is the relative ordering of create and drop operations, so `create -> drop -> create` isn't run as `create -> create -> drop`.

## Test plan

Use the example app.

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
@@ -14,19 +14,24 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
   private val observers = mutableMapOf<Int, MutableMap<Any, (GestureHandler) -> Unit>>()
 
   fun registerHandler(handler: GestureHandler) {
-    val callbacks = synchronized(this) {
+    val hasObservers = synchronized(this) {
       handlers.put(handler.tag, handler)
-      observers[handler.tag]?.values?.toList().orEmpty()
+      observers[handler.tag]?.isNotEmpty() == true
     }
 
-    if (callbacks.isEmpty()) {
+    if (!hasObservers) {
       return
     }
 
     // `createGestureHandler` runs on the JS thread, but observer callbacks read detector
     // view state (childCount, getChildAt) and may attach native handlers, so they must run
-    // on the UI thread.
+    // on the UI thread. Re-resolve the observer list on the UI thread so a cancellation that
+    // happens between this post and `notify` running (e.g. detector detach) actually prevents
+    // the callback.
     val notify = {
+      val callbacks = synchronized(this) {
+        observers[handler.tag]?.values?.toList().orEmpty()
+      }
       for (callback in callbacks) {
         callback(handler)
       }
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
@@ -19,8 +19,23 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
       observers[handler.tag]?.values?.toList().orEmpty()
     }
 
-    for (block in blocks) {
-      block(handler)
+    if (blocks.isEmpty()) {
+      return
+    }
+
+    // `createGestureHandler` runs on the JS thread, but observer callbacks read detector
+    // view state (childCount, getChildAt) and may attach native handlers, so they must run
+    // on the UI thread.
+    val notify = {
+      for (block in blocks) {
+        block(handler)
+      }
+    }
+
+    if (UiThreadUtil.isOnUiThread()) {
+      notify()
+    } else {
+      UiThreadUtil.runOnUiThread(notify)
     }
   }
 
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
@@ -14,12 +14,12 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
   private val observers = mutableMapOf<Int, MutableMap<Any, (GestureHandler) -> Unit>>()
 
   fun registerHandler(handler: GestureHandler) {
-    val blocks = synchronized(this) {
+    val callbacks = synchronized(this) {
       handlers.put(handler.tag, handler)
       observers[handler.tag]?.values?.toList().orEmpty()
     }
 
-    if (blocks.isEmpty()) {
+    if (callbacks.isEmpty()) {
       return
     }
 
@@ -27,8 +27,8 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
     // view state (childCount, getChildAt) and may attach native handlers, so they must run
     // on the UI thread.
     val notify = {
-      for (block in blocks) {
-        block(handler)
+      for (callback in callbacks) {
+        callback(handler)
       }
     }
 
@@ -55,9 +55,9 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
 
   @Synchronized
   fun cancelObservation(tag: Int, owner: Any) {
-    val table = observers[tag] ?: return
-    table.remove(owner)
-    if (table.isEmpty()) {
+    val observersForTag = observers[tag] ?: return
+    observersForTag.remove(owner)
+    if (observersForTag.isEmpty()) {
       observers.remove(tag)
     }
   }
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
@@ -11,10 +11,52 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
   private val handlers = SparseArray<GestureHandler>()
   private val attachedTo = SparseArray<Int?>()
   private val handlersForView = SparseArray<ArrayList<GestureHandler>>()
+  private val observers = mutableMapOf<Int, MutableMap<Any, (GestureHandler) -> Unit>>()
+
+  fun registerHandler(handler: GestureHandler) {
+    val blocks = synchronized(this) {
+      handlers.put(handler.tag, handler)
+      observers[handler.tag]?.values?.toList().orEmpty()
+    }
+
+    for (block in blocks) {
+      block(handler)
+    }
+  }
+
+  // Invokes `block` every time a handler with `tag` is registered, and synchronously once now if
+  // the handler already exists. The observation persists until explicitly cancelled: the registry
+  // holds both `owner` and `block` strongly, so callers MUST call `cancelObservation` or
+  // `cancelAllObservationsForOwner` when the owner is going away (typically in detach / dispose
+  // paths) to avoid leaking the owner. Observing the same tag twice with the same `owner` replaces
+  // the previous block.
+  fun observeHandler(tag: Int, owner: Any, block: (GestureHandler) -> Unit) {
+    val existing = synchronized(this) {
+      observers.getOrPut(tag) { mutableMapOf() }[owner] = block
+      handlers[tag]
+    }
+    existing?.let { block(it) }
+  }
 
   @Synchronized
-  fun registerHandler(handler: GestureHandler) {
-    handlers.put(handler.tag, handler)
+  fun cancelObservation(tag: Int, owner: Any) {
+    val table = observers[tag] ?: return
+    table.remove(owner)
+    if (table.isEmpty()) {
+      observers.remove(tag)
+    }
+  }
+
+  @Synchronized
+  fun cancelAllObservationsForOwner(owner: Any) {
+    val iterator = observers.entries.iterator()
+    while (iterator.hasNext()) {
+      val entry = iterator.next()
+      entry.value.remove(owner)
+      if (entry.value.isEmpty()) {
+        iterator.remove()
+      }
+    }
   }
 
   @Synchronized
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
@@ -16,8 +16,9 @@
 @interface RNGestureHandlerDetector () <RCTRNGestureHandlerDetectorViewProtocol>
 
 @property (nonatomic, nonnull) NSMutableSet *nativeHandlers;
+@property (nonatomic, nonnull) NSMutableSet *subscribedHandlers;
 @property (nonatomic, nonnull) NSMutableSet *attachedHandlers;
-@property (nonatomic) std::unordered_map<int, NSMutableSet *> attachedVirtualHandlers;
+@property (nonatomic) std::unordered_map<int, NSMutableSet *> subscribedVirtualHandlers;
 
 @end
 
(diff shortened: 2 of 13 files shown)


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
```diff
@@ -11,10 +11,52 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
...
-  @Synchronized
   fun registerHandler(handler: GestureHandler) {
-    handlers.put(handler.tag, handler)
+    val blocks = synchronized(this) {
+      handlers.put(handler.tag, handler)
+      observers[handler.tag]?.values?.toList().orEmpty()
+    }
+
+    for (block in blocks) {
+      block(handler)
```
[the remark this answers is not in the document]
j-piasecki: [8486c87](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/8486c878445afb3dbb92f357ed161cba45fd5e0b)

packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts
```diff
@@ -94,11 +85,10 @@ export function useGesture<
   );
 
   useEffect(() => {
-    return () => {
-      if (currentGestureRef.current.handlerTag === handlerTag) {
-        currentGestureRef.current = { type: '', handlerTag: -1 };
-      }
+    NativeProxy.createGestureHandler(type, handlerTag, {});
+    scheduleFlushOperations();
```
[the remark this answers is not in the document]
j-piasecki: I know, I wanted to handle it in a follow-up to keep changes as small as possible

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
```diff
@@ -11,10 +11,67 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
...
+    val existing = synchronized(this) {
+      observers.getOrPut(tag) { mutableMapOf() }[owner] = block
+      handlers[tag]
+    }
+    existing?.let { block(it) }
+  }
+
+  @Synchronized
+  fun cancelObservation(tag: Int, owner: Any) {
+    val table = observers[tag] ?: return
```
m-bert: What about other name?

```suggestion
    val observersForTag = observers[tag] ?: return
```

```suggestion
    val tagObservers = observers[tag] ?: return
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
```diff
@@ -11,10 +11,67 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
   private val handlers = SparseArray<GestureHandler>()
   private val attachedTo = SparseArray<Int?>()
   private val handlersForView = SparseArray<ArrayList<GestureHandler>>()
+  private val observers = mutableMapOf<Int, MutableMap<Any, (GestureHandler) -> Unit>>()
 
-  @Synchronized
   fun registerHandler(handler: GestureHandler) {
-    handlers.put(handler.tag, handler)
+    val blocks = synchronized(this) {
```
m-bert: `blocks` sounds weird (especially `block(handler)`), maybe `callbacks` instead?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -258,9 +271,9 @@ - (void)updateProps:(const Props::Shared &)propsBase oldProps:(const Props::Shar
   _moduleId = newProps.moduleId;
 
   [self attachHandlers:newProps.handlerTags
-            actionType:RNGestureHandlerActionTypeNativeDetector
-               viewTag:-1
-      attachedHandlers:_attachedHandlers];
+              actionType:RNGestureHandlerActionTypeNativeDetector
+                 viewTag:-1
+      subscribedHandlers:_subscribedHandlers];
```
m-bert, line 276: Shouldn't these be aligned to `:`?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
```diff
@@ -284,20 +297,24 @@ - (void)updateVirtualChildren:(const std::vector<RNGestureHandlerDetectorVirtual
...
+    if (_subscribedVirtualHandlers.find(child.viewTag) == _subscribedVirtualHandlers.end()) {
+      _subscribedVirtualHandlers[child.viewTag] = [NSMutableSet set];
     }
     [self attachHandlers:child.handlerTags
-              actionType:RNGestureHandlerActionTypeVirtualDetector
-                 viewTag:child.viewTag
-        attachedHandlers:_attachedVirtualHandlers[child.viewTag]];
+                actionType:RNGestureHandlerActionTypeVirtualDetector
+                   viewTag:child.viewTag
+        subscribedHandlers:_subscribedVirtualHandlers[child.viewTag]];
```
m-bert, line 317: Same here

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.h
```diff
@@ -20,6 +22,18 @@
...
 - (void)dropAllHandlers;
 
+// Invokes `block` every time a handler with `handlerTag` is registered. If the handler already
+// exists at the time this method is called, `block` is also invoked synchronously before returning.
+// The observation persists until explicitly cancelled (or until `owner` is deallocated, provided
+// the block does not strongly capture `owner`). `owner` is held weakly and acts as the cancellation
+// key; observing the same tag twice with the same `owner` replaces the previous block.
+- (void)observeHandlerWithTag:(nonnull NSNumber *)handlerTag
+                        owner:(nonnull id)owner
+                   usingBlock:(nonnull RNGestureHandlerReadyBlock)block;
```
m-bert: Here I'd also go with something like `withCallback` instead of `usingBlock`

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.h
```diff
@@ -20,6 +22,18 @@
 - (void)dropHandlerWithTag:(nonnull NSNumber *)handlerTag;
 - (void)dropAllHandlers;
 
+// Invokes `block` every time a handler with `handlerTag` is registered. If the handler already
+// exists at the time this method is called, `block` is also invoked synchronously before returning.
+// The observation persists until explicitly cancelled (or until `owner` is deallocated, provided
+// the block does not strongly capture `owner`). `owner` is held weakly and acts as the cancellation
```
m-bert, line 28: If owner is held `weakly` then is the case in parenthesis valid?

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.m
```diff
@@ -36,8 +38,66 @@ - (RNGestureHandler *)handlerWithTag:(NSNumber *)handlerTag
...
+- (void)cancelAllObservationsForOwner:(id)owner
+{
+  @synchronized(_handlers) {
+    NSArray<NSNumber *> *tags = [_observers allKeys];
+    for (NSNumber *tag in tags) {
+      NSMapTable *table = _observers[tag];
+      [table removeObjectForKey:owner];
+      if (table.count == 0) {
+        [_observers removeObjectForKey:tag];
+      }
```
m-bert: Can't we call `cancelObservationForTag`?

packages/react-native-gesture-handler/src/v3/detectors/HostGestureDetector.web.tsx
```diff
@@ -31,81 +32,227 @@ export interface VirtualChildrenWeb {
...
+  refs.virtualChildren.clear();
+}
 
 const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
   const { handlerTags, children } = props;
 
-  const handlerTagsSet = useMemo(() => new Set(handlerTags), [...handlerTags]);
+  const handlerTagsSet = useMemo(
+    () => new Set(handlerTags),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
m-bert: Is this needed?

packages/react-native-gesture-handler/src/v3/detectors/HostGestureDetector.web.tsx
```diff
@@ -115,80 +262,97 @@ const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
...
+      for (const virtualHandlers of refs.subscribedVirtualHandlers.values()) {
+        for (const tag of virtualHandlers) {
+          claimedByVirtual.add(tag);
+        }
+      }
+
+      for (const tag of refs.subscribedHandlers) {
+        if (!refs.attachedHandlers.has(tag) || claimedByVirtual.has(tag)) {
+          continue;
+        }
```
m-bert: We could also use `Set` operations, right?

```ts
const claimedByVirtual = Array.from(refs.subscribedVirtualHandlers.values())
  .reduce((acc, currentSet) => acc.union(currentSet), new Set<number>());

const validHandlers = refs.subscribedHandlers
  .intersection(refs.attachedHandlers)
  .difference(claimedByVirtual);

// 3. Simply iterate over the clean results
for (const tag of validHandlers) {
  ...
}
```

packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
```diff
@@ -50,6 +68,48 @@ export default abstract class NodeManager {
     this.gestures[handlerTag].detach();
   }
 
+  // Invokes `block` every time a handler with `tag` is created, and synchronously once now if
```
m-bert: "once now" sounds a bit weird 🫨

packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
```diff
@@ -50,6 +68,48 @@ export default abstract class NodeManager {
...
+      this.observers.delete(tag);
+    }
+  }
+
+  public static cancelAllObservationsForOwner(owner: object): void {
+    for (const [tag, table] of this.observers) {
+      table.delete(owner);
+      if (table.size === 0) {
+        this.observers.delete(tag);
+      }
```
m-bert: We could call `cancelObservation` right?

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: This is what the apple autoformat produces 🤷

packages/react-native-gesture-handler/apple/RNGestureHandlerDetector.mm
j-piasecki: This is what the apple autoformat produces 🤷

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.h
j-piasecki: Yes, the key is held weakly, but if the value holds a strong reference to the key, then it will keep itself in the table.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
j-piasecki: [Rename](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/a513ba965b76c358b1bb6be447212c0e2c7fbd11)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
j-piasecki: [Rename](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/a513ba965b76c358b1bb6be447212c0e2c7fbd11)

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.h
j-piasecki: [Rename](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/a513ba965b76c358b1bb6be447212c0e2c7fbd11)

packages/react-native-gesture-handler/src/v3/detectors/HostGestureDetector.web.tsx
j-piasecki: Not really, I removed it since it doesn't relate to this PR's goal, but we may want to revisit that in the future.

[Remove suppression](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/8c03cf95c29e3c35ba478ad1bb79757d9a727678)

packages/react-native-gesture-handler/src/v3/detectors/HostGestureDetector.web.tsx
j-piasecki: Yes - [Use set functions](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/d89d89b3210c65dab66d0a102072e3c42d9774de)

packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
j-piasecki: [Update comment](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/4ab34a579095f7dc24ec06213ad6750e35b64298)

packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
j-piasecki: Yes - [Use cancelObservation](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/af812757e7368cb285185634db9be415d772eee1)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
```diff
@@ -11,10 +11,67 @@ class RNGestureHandlerRegistry : GestureHandlerRegistry {
...
+        callback(handler)
+      }
+    }
+
+    if (UiThreadUtil.isOnUiThread()) {
+      notify()
+    } else {
+      UiThreadUtil.runOnUiThread(notify)
+    }
+  }
```
coado, line 45: Would it be possible that the callback is scheduled but the detector is cleaned right after that leaving stale `attachReadyHandler` calls?

packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
```diff
@@ -29,6 +38,15 @@ export default abstract class NodeManager {
 
     this.gestures[handlerTag] = handler;
     this.gestures[handlerTag].handlerTag = handlerTag;
+
+    const pending = this.observers.get(handlerTag);
+    if (pending) {
+      // Snapshot before iterating — blocks may call back into observeHandler / cancelObservation
+      // and mutate the underlying map.
+      for (const block of Array.from(pending.values())) {
+        block(handler as IGestureHandler);
```
m-bert: Web should also be renamed 😅 
```suggestion
        callback(handler as IGestureHandler);
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
j-piasecki: If I understood you correctly, then I don't think so (unless you really want to, and pass a non-existent tag as a prop to the detector).

Otherwise, the detector drops subscriptions for the tags that are no longer in props, and it drops all subscriptions when detached from the window.

packages/react-native-gesture-handler/src/web/tools/NodeManager.ts
j-piasecki: [5713715](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/5713715ce2c21d9c2a9714f6135db54421739669)

packages/react-native-gesture-handler/apple/RNGestureHandlerRegistry.m
j-piasecki: We can - [Use cancelObservationForTag](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/3679c8a23bb1e929e49d54d2a434340752fc2018)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRegistry.kt
coado: I was thinking about the case when the correct tag is passed to the detector and the notification is scheduled on the UI thread. Then the detector is dropped with its subscriptions but the scheduled callbacks haven't been invoked yet.

j-piasecki: You're right, it should be fixed by [503db25](https://github.com/software-mansion/react-native-gesture-handler/pull/4179/commits/503db258f5c4fdebc56c99439e038dadd5a51d82)

coado (approved):
👍

PR #4180, opened 2026-05-18
[General] Use handler observer API for relation updates

j-piasecki:
## Description

https://github.com/software-mansion/react-native-gesture-handler/pull/4179 changed the timing of gesture creation and attachment, after that change `configureRelations` was being fired before the gesture was created.

This PR updates the relations configuration flow to use the observers API.

## Test plan

Example app


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
```diff
@@ -106,10 +106,16 @@ class RNGestureHandlerModule(reactContext: ReactApplicationContext?) :
...
-    interactionManager.dropRelationsForHandlerWithTag(handlerTag)
-    interactionManager.configureInteractions(handler, relations)
+    // Observe the registry rather than looking up the handler directly - the JS-side detector may
+    // call `configureRelations` before `useGesture`'s effect has invoked `createGestureHandler`,
+    // so the handler doesn't exist yet.
+    val owner = Any()
+    registry.observeHandler(handlerTag, owner) { handler ->
+      interactionManager.dropRelationsForHandlerWithTag(handlerTag)
+      interactionManager.configureInteractions(handler, relations)
+      registry.cancelObservation(handlerTag, owner)
```
m-bert: This won't interfere with observers attached in other places, right?

j-piasecki: No, the registry uses the owner as key, and in this case, each owner is unique.

coado: What if the handler is never created? Is it fine to keep such observers in the registry then?

coado (commented):
It might also be worth adding some StrictMode tests.

j-piasecki:
> It might also be worth adding some StrictMode tests.

We should be able to do that with the e2e tests set up. I'm not sure how to do it currently, since it requires platform code to verify.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerModule.kt
j-piasecki: I think we're effectively leaking the closure then. I'll double-check tomorrow if we can handle it without observing and rely only on the scheduling order instead.

That said, what's the scenario where it happens (update without gesture being created/after being dropped)?

coado: No idea, just moving focus here to think if it is possible and if we should worry about it beforehand. One example is the initial issue with the strict mode but in this case it will be handled properly as we know that happens.

j-piasecki: I think that we can do it by delaying when the configure is actually called, so it runs after all the effects from the parent nodes have run - [13f12c9](https://github.com/software-mansion/react-native-gesture-handler/pull/4180/commits/13f12c951933b7b077ac3bd737a1111a381714cf)

coado (approved):
looks good 😃

PR #4207, opened 2026-05-27
[General] Fix relations not being applied in api V1

j-piasecki:
## Description

1. `configureRelations` wasn't called after creating the gesture, only after the update.
2. `filterConfig` isn't idempotent, and calling it a second time resulted in the relations arrays being cleared. Adds a simple `selectProperties` util to use instead.

## Test plan

Test legacy Multitap example


m-bert (approved):
> `filterConfig` isn't idempotent

And I think it should be, right? Anyway, that's probably not for now. Leaving ✅ so it is not blocked before release.

Also, do you know why Android worked anyway?

j-piasecki:
> And I think it should be, right?

It wouldn't hurt, but I don't know if something relies on it at this point.

> Also, do you know why Android worked anyway?

No 😞

PR #4213, opened 2026-05-31
Fix Android Pressable TalkBack activation

### LINKED ISSUE #3899: Pressable does not execute onPress on Android when TalkBack screen reader is enabled
### Description

The `Pressable` component in react-native-gesture-hander does not work on Android when the system screen reader, TalkBack, is enabled. This issue occurs with the latest version of React Native and the react-native-gesture-hander library, and is reproducible in the example app of this repo. The `Pressable` component can be focused on, but initiating a press or a long press with the screen reader does not fire the `onPress`, `onLongPress`, `onPressIn` or `onPressOut` functions. When I look at the resulting native view after layout in UI Inspector, I can see that the Pressable button does have both the "focusable" and "clickable" attributes set to true, which _should_ make it tappable. In addition, the screen reader _does_ announce that the Pressable can be tapped, due to "clickable" being true.

These issues do not occur with the `Pressable` component imported from react-native, so it does seem like a bug in RNGH. In practice this makes the Pressable from RNGH unusable on Android in apps that support the use of screen readers.

On iOS, with the VoiceOver screen reader enabled, both the React Native Pressable and the RNGH Pressable work as expected.

### Steps to reproduce

1. Install the basic-example demo app from this repository.
2. Modify the `ComponentsScreen.tsx` code by adding the two new demo cases, ReactNativePressableDemo and RNGHPressableDemo from the Gist.
3. Install on physical Android device and Enable TalkBack.
4. The React Native Pressable responds to presses, whereas the RNGH Pressable does not.

https://gist.github.com/olliFCA/7db728b0d7e3087e257cfdc3e70ee321

### Gesture Handler version

2.30.0

### React Native version

0.83.0

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

Real device

### Device model

Samsung Galaxy A55

### Acknowledgements

Yes

sorinc03:
## Description

Fixes #3899. Root-cause note: https://github.com/software-mansion/react-native-gesture-handler/issues/3899#issuecomment-4587868950

TalkBack's double-tap activation dispatches an Android accessibility click/action to the native view rather than the touch stream consumed by `Pressable`'s gesture state machine. That means the `FINALIZE -> handlePressOut` path that calls `onPress` is never reached. PR #4017 helped with touch-event ordering, but it does not cover this accessibility-action channel.

This brings `Pressable` and `LegacyPressable` to parity with the existing touchables by:

- adding default Android accessibility actions for `activate` and `longpress` when the corresponding press handlers exist;
- routing `activate` through the existing `onPressIn` -> `onPressOut` -> `onPress` lifecycle using synthetic layout-local press events;
- routing `longpress` to `onLongPress`;
- preserving user-provided accessibility actions and forwarding `onAccessibilityAction`;
- avoiding duplicate `onPress` calls for apps that already supplied their own user-handled `activate` action workaround;
- sharing the Android accessibility-action wiring between the v3 and legacy implementations.

## Test plan

- `yarn test Pressable.accessibility.test.tsx`
- `yarn test`
- `yarn ts-check`
- `yarn lint:js` (passes with existing repo warnings only)
- `git diff --check`

### DIFF
--- a/packages/react-native-gesture-handler/src/__tests__/Pressable.accessibility.test.tsx
+++ b/packages/react-native-gesture-handler/src/__tests__/Pressable.accessibility.test.tsx
@@ -4,7 +4,10 @@ import { Platform, Text } from 'react-native';
 
 import GestureHandlerRootView from '../components/GestureHandlerRootView';
 import LegacyPressable from '../components/Pressable/Pressable';
-import type { PressableProps } from '../components/Pressable/PressableProps';
+import type {
+  PressableEvent,
+  PressableProps,
+} from '../components/Pressable/PressableProps';
 import Pressable from '../v3/components/Pressable';
 
 jest.unmock('../components/Pressable/Pressable');
@@ -26,23 +29,25 @@ const implementations = [
   ],
 ] as const;
 
-beforeEach(() => {
+const setPlatform = (platform: typeof Platform.OS) => {
   Object.defineProperty(Platform, 'OS', {
     configurable: true,
-    value: 'android',
+    value: platform,
   });
+};
+
+beforeEach(() => {
+  setPlatform('android');
 });
 
 afterEach(() => {
-  Object.defineProperty(Platform, 'OS', {
-    configurable: true,
-    value: originalPlatformOS,
-  });
+  setPlatform(originalPlatformOS);
 });
 
 function renderPressable(
   Component: React.ComponentType<TestPressableProps>,
-  props: TestPressableProps
+  props: TestPressableProps,
+  layout = { x: 0, y: 0, width: 100, height: 40 }
 ) {
   const result = render(
     <GestureHandlerRootView>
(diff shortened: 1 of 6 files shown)


j-piasecki:
Can you check whether https://github.com/software-mansion/react-native-gesture-handler/pull/4222 and https://github.com/software-mansion/react-native-gesture-handler/pull/4220 solve your issues? There were multiple problems that caused TalkBack to behave incorrectly or not work.

sorinc03:
I checked the linked fixes: #4220 and #4222 are both merged. This branch has also been updated to address the implementation/test concerns from Copilot: shared `usePressableAccessibility`, layout-derived synthetic events with separate timestamps, target lookup, and disabled/non-Android coverage. The targeted `Pressable.accessibility.test.tsx` suite passes locally.

If #4220/#4222 fully cover the TalkBack issue from your side, I'm happy to close this as superseded; otherwise this PR is still ready for review as the explicit Pressable accessibility-action path.

j-piasecki:
I think those two PRs are covering the TalkBack issues. Thanks for raising awareness of this!

PR #4248, opened 2026-06-09
[Android] Do not set explicit versions in `build.gradle` 

### LINKED ISSUE #2307: [Android] The Kotlin Gradle plugin was loaded multiple times in different subprojects
### Description

Every time when I building my project, the following warning appears in the gradle output. I don't think this a problem (the project build correctly), but decided to inform about it.

My environment:

gradle 7.5.1
gradle plugin: 7.2.2

### Steps to reproduce

Start build the RN project -> the following warning appears in the gradle output:

> The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build. 
> This might happen in subprojects that apply the Kotlin plugins with the Gradle 'plugins { ... }' DSL if they specify explicit versions, even if the versions are equal.
> Please add the Kotlin plugin to the common parent project or the root project, then remove the versions in the subprojects.
> If the parent project does not need the plugin, add 'apply false' to the plugin line.
> See: https://docs.gradle.org/current/userguide/plugins.html#sec:subprojects_plugins_dsl
> The Kotlin plugin was loaded in the following projects: ':react-native-gesture-handler', ':react-native-screens'

none

### Gesture Handler version

2.8.0

### React Native version

0.70.4

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

Paper (Old Architecture)

### Build type

Release mode

### Device

Real device

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

Gesture Handler loads specific Kotlin version during build. This is unnecessary and may cause build warnings, such as one described in #2307. 

This PR removes pinning of specific kotlin/gradle versions when Gesture Handler is not root project, so we fallback to the ones declared by the root. Unfortunately we can't simply remove those lines, as `spotless` would fail.

Fixes #2307

## Test plan

Tested that android build correctly in basic-example, expo-example and standalone app.


packages/docs-gesture-handler/docs/fundamentals/installation.mdx
```diff
@@ -91,16 +91,6 @@ export function CustomModal({ children, ...rest }) {
 }
 ```
 
-Gesture Handler on Android is implemented in Kotlin. If you need to set a specific Kotlin version to be used by the library, set the `kotlinVersion` ext property in the `android/build.gradle` file and RNGH will use that version:
```
m-bert, line 94: I'm not sure if we want to remove it entirely, or add info that it is not supported anymore (e.g. starting from 3.1.0)

j-piasecki (approved):

packages/react-native-gesture-handler/android/build.gradle
```diff
@@ -2,17 +2,18 @@ import groovy.json.JsonSlurper
...
         mavenCentral()
         google()
     }
 
     dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version")
-        classpath("com.android.tools.build:gradle:8.10.1")
-        classpath("com.diffplug.spotless:spotless-plugin-gradle:7.0.4")
+        if (project == rootProject) {
+            def kotlin_version = project.properties['RNGH_kotlinVersion']
```
j-piasecki, line 12: Can you also update that version? RN is at `2.1.20`, while we have `2.0.21` set here.

m-bert: [Bumped Kotlin to 2.2.0](https://github.com/software-mansion/react-native-gesture-handler/pull/4248/commits/57309e6aaca3547d0102944fc3f4dffc9e9e28b4) as in [template](https://github.com/react-native-community/template/blob/816967c82f535cacc0f7a5c28688e884ecc3bdd9/template/android/build.gradle#L8) and [react native `main`](https://github.com/react/react-native/blob/066c0d8bd842506f029e16da28cc9fe1ce3191b1/packages/react-native/gradle/libs.versions.toml#L33)

PR #4253, opened 2026-06-10
fix(Android): apply numberOfPointers config to the pointer requirement in LongPressGestureHandler

### LINKED ISSUE #4252: LongPress `numberOfPointers` is ignored on Android — activates with a single pointer (regression in 2.26.0)
### Description

Since 2.26.0, the `numberOfPointers` config of the LongPress gesture is silently ignored on Android. A gesture configured with `Gesture.LongPress().numberOfPointers(3)`:

- **activates after a single-pointer long press** (the configured pointer count is never applied), and
- **can never activate with the configured pointer count** — as soon as the second pointer touches down, `currentPointers > numberOfPointersRequired` (2 > 1) trips and the handler fails.

iOS is unaffected (`numberOfPointers` maps to `UILongPressGestureRecognizer.numberOfTouchesRequired`).

**Root cause:** in the 2.26.0 refactor that moved handler configuration from `RNGestureHandlerModule` into the nested `Factory` classes, the LongPress factory's config assignment ended up targeting the wrong property. In [`LongPressGestureHandler.kt`](https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt#L202):

```kotlin
if (config.hasKey(KEY_NUMBER_OF_POINTERS)) {
  handler.numberOfPointers = config.getInt(KEY_NUMBER_OF_POINTERS)
}
```

`handler.numberOfPointers` resolves to the inherited `GestureHandler.numberOfPointers` — the live "pointers currently down" counter used for event payloads, which the base class overwrites with `event.pointerCount` on every motion event. The field that actually gates activation, `private var numberOfPointersRequired`, stays at its init value of `1` and is never written from config.

Before 2.26.0, the module factory called `handler.setNumberOfPointers(...)`, which correctly set `numberOfPointersRequired`.

Verified present in 2.26.0, 2.27.x, 2.28.0, 2.29.1, 3.0.0, and current `main` (by inspecting the published packages). Verified the one-line fix on-device (Android, RN 0.81.5) via a patched 2.28.0.

Fix PR incoming — routing the config value to `numberOfPointersRequired`, restoring the pre-2.26 behavior.

### Steps to reproduce

1. Render the snippet from the gist below (a full-screen `GestureDetector` with `Gesture.LongPress().numberOfPointers(3)`) in an Android app
2. Long-press with **one** finger → `onStart` fires after ~500 ms (it shouldn't)
3. Long-press with **three** fingers → `onStart` never fires (it should)
4. Run the same snippet on iOS → behaves correctly (fires only with exactly 3 fingers)

https://gist.github.com/prashanFOMO/ad838ce8a3c77c59759b8583fe8d3393

### Gesture Handler version

2.28.0 (present in 2.26.0 through 3.0.0 and current main)

### React Native version

0.81.5

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

Real device

### Device model

_No response_

### Acknowledgements

Yes

prashanFOMO:
## Description

Fixes #4252

On Android, the `numberOfPointers` config of the LongPress gesture has been silently ignored since 2.26.0: a gesture configured with `Gesture.LongPress().numberOfPointers(3)` activates after a **single-pointer** long press, and can never activate with the configured pointer count (the second pointer trips `currentPointers > numberOfPointersRequired` and fails the handler). iOS is unaffected, since the prop maps to `UILongPressGestureRecognizer.numberOfTouchesRequired`.

The cause is a one-line mix-up introduced by the factory refactor in 2.26.0. The factory's `updateConfig` assigns the config value to `handler.numberOfPointers`, which resolves to the inherited `GestureHandler.numberOfPointers` — the live pointer-count used for event payloads, overwritten with `event.pointerCount` on every motion event. The field that actually gates activation, `numberOfPointersRequired`, stays at its default of `1` and is never written from config. Before 2.26.0, the module factory called `handler.setNumberOfPointers(...)`, which set `numberOfPointersRequired` correctly.

This PR routes the config value to `numberOfPointersRequired`, restoring the pre-2.26 behavior. The private-field access from the nested `Factory` matches the existing `maxDist` assignment directly above.

## Test plan

- Applied this same one-liner to 2.28.0 via patch-package in an RN 0.81.5 (Expo 54, new architecture, Hermes) app that uses `Gesture.LongPress().numberOfPointers(3)` as a global gesture, and verified on an Android device:
  - three-pointer long press now activates the gesture (previously never fired)
  - single-pointer long press no longer activates it (previously fired after ~500 ms)
  - gestures with the default single pointer (no `numberOfPointers` set) are unaffected
- Verified iOS behavior is unchanged (the fix doesn't touch the iOS path)
- Repro snippet: https://gist.github.com/prashanFOMO/ad838ce8a3c77c59759b8583fe8d3393

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
@@ -199,7 +199,7 @@ class LongPressGestureHandler(context: Context) : GestureHandler() {
         handler.maxDist = PixelUtil.toPixelFromDIP(config.getDouble(KEY_MAX_DIST))
       }
       if (config.hasKey(KEY_NUMBER_OF_POINTERS)) {
-        handler.numberOfPointers = config.getInt(KEY_NUMBER_OF_POINTERS)
+        handler.numberOfPointersRequired = config.getInt(KEY_NUMBER_OF_POINTERS)
       }
     }
 


j-piasecki (approved):
Thanks ❤️!

PR #4264, opened 2026-06-16
[Docs] Add dynamic topbar banner

p-malecki:
## Description

Adds a topbar banner to the landing page. Content is served server-side via external service, so banners can be swapped, rotated, or pulled without any repo change or redeploy.
It won’t be displayed on the docs or search pages. The banners are currently hidden across the site.

### DIFF
--- a/packages/docs-gesture-handler/docusaurus.config.js
+++ b/packages/docs-gesture-handler/docusaurus.config.js
@@ -10,8 +10,7 @@ const webpack = require('webpack');
 const redirectsData = require('./redirects.json');
 
 import { topbarBannerReservationScript } from '@swmansion/t-rex-ui/topbar-banner';
-// @ts-expect-error -- .ts extension is intentional; not type-checked by tsc here.
-import { TOP_BAR_BANNER } from './src/components/topBarBannerConfig.ts';
+import { TOP_BAR_BANNER } from './src/components/topBarBannerConfig';
 
 const firstBannerZone = TOP_BAR_BANNER.zones[0];
 const bannerReservationHeadTags = firstBannerZone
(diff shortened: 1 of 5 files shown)


packages/docs-gesture-handler/docusaurus.config.js
```diff
@@ -9,6 +9,25 @@ const webpack = require('webpack');
 
 const redirectsData = require('./redirects.json');
 
+import { topbarBannerReservationScript } from '@swmansion/t-rex-ui/topbar-banner';
+// @ts-expect-error -- .ts extension is intentional; not type-checked by tsc here.
+import { TOP_BAR_BANNER } from './src/components/topBarBannerConfig.ts';
```
m-bert: And why do we need extension here? Seems to work without it 🤔

packages/docs-gesture-handler/src/components/topBarBannerConfig.ts
m-bert, line 1: I'm not sure if this is the right place to keep only config. What about putting it in `src/utils` instead? Or at the root?

packages/docs-gesture-handler/docusaurus.config.js
p-malecki: You're right, we can remove extension and @ts-expect-error

packages/docs-gesture-handler/src/components/topBarBannerConfig.ts
p-malecki: Good suggestion, but this config placement has already been added in several other projects. So I’d prefer to keep it in its current location for consistency

m-bert (approved):

PR #4266, opened 2026-06-16
[General] Ignore `testID` property in production environments

j-piasecki:
## Description

When in production environments, `testID` is removed from `allowedNativeProps` to prevent it from being passed to the native side. This mean that it no longer falls into the first branch here https://github.com/software-mansion/react-native-gesture-handler/blob/1473f3906e2cc297b92f693fae09b6e789068cfe/packages/react-native-gesture-handler/src/v3/hooks/utils/configUtils.ts#L86-L100

But last, which shows a warning. This PR adds `testID` to `PropsToFilter` so they fall into the second (ignore silently) branch.

## Test plan

~~AI said it's ok~~ Run app in release configuration


m-bert (approved):
> <s>AI said it's ok</s>

<s>My AI agrees</s> LGTM

PR #4274, opened 2026-06-18
[Android] Fix ConcurrentModificationException in GestureHandlerOrchestrator

kosmydel:
## Description
Fixes a `ConcurrentModificationException` in `GestureHandlerOrchestrator` when delivering and cancelling events. Event delivery now iterates over local snapshots of `gestureHandlers` instead of a shared `preparedHandlers` field and the live `asReversed()` view, both of which could be mutated mid-iteration during re-entrant state updates.

Credit to @SudoPlz for the original fix. This PR upstreams it into the main repository

> [!NOTE]
> We don't have a reliable repro for this. It surfaced in our Sentry logs as a crash, and the fix is based on inspection of the orchestrator's re-entrancy behavior.

<details><summary>Original patch confirmed to work</summary>
<p>

```
diff --git a/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt b/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
index ced298e52339d96f871144598a5ed519cf526a89..ce87cc710eea5fa98a27dcf1f94878d38c5b40e0 100644
--- a/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
+++ b/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
@@ -250,19 +250,18 @@ class GestureHandlerOrchestrator(
   }

   private fun deliverEventToGestureHandlers(event: MotionEvent) {
-    // Copy handlers to "prepared handlers" array, because the list of active handlers can change
-    // as a result of state updates
-    preparedHandlers.clear()
-    preparedHandlers.addAll(gestureHandlers)
+    // Copy handlers to local array, because the list of active handlers can change
+    // as a result of state updates. Create a local snapshot to avoid race conditions.
+    val handlersToProcess = gestureHandlers.toMutableList()

     // We want to deliver events to active handlers first in order of their activation (handlers
     // that activated first will first get event delivered). Otherwise we deliver events in the
     // order in which handlers has been added ("most direct" children goes first). Therefore we rely
     // on Arrays.sort providing a stable sort (as children are registered in order in which they
     // should be tested)
-    preparedHandlers.sortWith(handlersComparator)
+    handlersToProcess.sortWith(handlersComparator)

-    for (handler in preparedHandlers) {
+    for (handler in handlersToProcess) {
       deliverEventToGestureHandler(handler, event)
     }
   }
@@ -274,11 +273,8 @@ class GestureHandlerOrchestrator(
       handler.cancel()
     }

-    // Copy handlers to "prepared handlers" array, because the list of active handlers can change
-    // as a result of state updates
-    preparedHandlers.clear()
-    preparedHandlers.addAll(gestureHandlers)
-
+    // Use reversed() directly to create a snapshot and avoid race conditions
+    // when the list of active handlers changes as a result of state updates.
     for (handler in gestureHandlers.asReversed()) {
       handler.cancel()
     }
```

</p>
</details> 

## Test steps
1. Build and run the Android example app (`apps/basic-example`, `yarn android`).
2. Exercise screens with multiple interacting gesture handlers (e.g. nested gestures, swipeables) and confirm no regressions in gesture behavior.

Made with [Cursor](https://cursor.com)


m-bert (approved):
Thanks! ❤️

PR #4276, opened 2026-06-19
Get `uiRuntime` directly from react-native-worklets

coado:
## Description

RNGH currently obtains the UI runtime through Reanimated’s private `_WORKLET_RUNTIME` global, which stores a raw `jsi::Runtime` pointer.

This PR makes the public `react-native-worklets` Stable API the preferred way to obtain the UI runtime while preserving the existing Reanimated fallback.

### Runtime lookup

When compatible Worklets is available:

1. JavaScript calls `getUIRuntimeHolder()`.
2. The holder is temporarily exposed to native code for the synchronous installation call.
3. Android and Apple unwrap it through `worklets/Compat/StableApi.h`.
4. RNGH installs `_setGestureStateSync` on the resulting UI runtime.
5. The temporary holder is cleared afterward.

RNGH no longer includes or accesses platform-specific Worklets implementation APIs such as `worklets/android/WorkletsModule.h` or `worklets/apple/WorkletsModule.h`.

When Worklets is unavailable, too old, or holder retrieval fails, native code falls back to the existing `_WORKLET_RUNTIME` pointer provided by Reanimated.

The JavaScript initialization handles the providers independently:

- Worklets available: use the Stable API holder.
- Reanimated available without Worklets: call native installation and use `_WORKLET_RUNTIME`.
- Neither available: skip UI runtime installation.

### Optional dependency integration

`react-native-worklets` remains optional.

Native Worklets integration is enabled only for stable versions `>= 0.8.0`:

- CocoaPods resolves the package from the installation root, verifies its version and podspec, adds `RNWorklets (>= 0.8.0)`, and defines `RNGH_USE_WORKLETS=1`.
- Android detects the Worklets Gradle project and version, links its prefab target, and defines the same compile-time macro.
- `StableApi.h` is included directly when the macro is enabled, so missing or incorrectly exported headers fail compilation instead of silently selecting the fallback.

The shared runtime decorator was also split into:

- Installing RNGH bindings on a runtime that has already been resolved.
- Resolving the legacy `_WORKLET_RUNTIME` pointer when needed.

## Test plan

Added Jest coverage and tested on Runtime Decoration example on iOS and Android.


packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
```diff
@@ -44,10 +44,23 @@ endif()
...
 target_link_libraries(
   ${PACKAGE_NAME}
   ReactAndroid::reactnative
   ReactAndroid::jsi
   fbjni::fbjni
 )
+
+if(${RNGH_USE_WORKLETS})
+  target_link_libraries(${PACKAGE_NAME} react-native-worklets::worklets)
+endif()
```
j-piasecki, line 59: Can we have a single `if(${RNGH_USE_WORKLETS})` block?

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
```diff
@@ -44,10 +44,23 @@ endif()
...
 find_package(ReactAndroid REQUIRED CONFIG)
 find_package(fbjni REQUIRED CONFIG)
+if(${RNGH_USE_WORKLETS})
+  find_package(react-native-worklets REQUIRED CONFIG)
+  target_compile_definitions(${PACKAGE_NAME} PRIVATE RNGH_USE_WORKLETS=1)
+  target_include_directories(
+    ${PACKAGE_NAME}
+    PRIVATE
+    "${REACT_NATIVE_DIR}/ReactCommon/jsiexecutor"
+  )
```
j-piasecki: Do we need to do that? Maybe there's a target in react native that exposes these headers.

packages/react-native-gesture-handler/android/build.gradle
```diff
@@ -107,6 +107,10 @@ def shouldUseCommonInterfaceFromRNSVG() {
             major > 15
 }
 
+def shouldUseRuntimeFromWorklets() {
+    return rootProject.subprojects.find { it.name == 'react-native-worklets' } != null
+}
```
j-piasecki: Should we also validate worklets' version, or does it work with each one published so far?

0.8.0 may be a good cutoff candidate since it introduces a stable api: https://github.com/software-mansion/react-native-reanimated/pull/8996

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -16,6 +16,13 @@
 
 #import "RNGHRuntimeDecorator.h"
 
+#if __has_include(<worklets/apple/WorkletsModule.h>)
```
j-piasecki: Have you tried using different linking methods (static/dynamic)?

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
```diff
@@ -78,17 +82,15 @@ let Reanimated:
...
 try {
-  Reanimated = require('react-native-reanimated');
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-var-requires
-  const Worklets = require('react-native-worklets');
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const Worklets = require('react-native-worklets') as WorkletsModule;
 
   // Make sure worklets are initialized before attempting to install UI runtime bindings
-  Worklets?.scheduleOnUI(() => {
+  Worklets.scheduleOnUI?.(() => {
```
j-piasecki:
```suggestion
  Worklets?.scheduleOnUI?.(() => {
```

packages/react-native-gesture-handler/shared/runtime/RNGHRuntimeDecorator.cpp
```diff
@@ -89,25 +89,9 @@ void RNGHRuntimeDecorator::installRNRuntimeBindings(
       rnRuntime, "_RNGH_MODULE_ID", std::move(moduleIdValue));
 }
 
-bool RNGHRuntimeDecorator::installUIRuntimeBindings(
-    jsi::Runtime &rnRuntime,
-    int moduleId,
+void RNGHRuntimeDecorator::decorateUIRuntime(
```
j-piasecki: Hmm, now this is actually what installs the bindings.


```suggestion
void RNGHRuntimeDecorator::installUIRuntimeBindings(
```

packages/react-native-gesture-handler/shared/runtime/RNGHRuntimeDecorator.cpp
```diff
@@ -128,6 +112,28 @@ bool RNGHRuntimeDecorator::installUIRuntimeBindings(
 
   uiRuntime.global().setProperty(
       uiRuntime, "_setGestureStateSync", std::move(setGestureStateSync));
+}
+
+bool RNGHRuntimeDecorator::installUIRuntimeBindings(
```
j-piasecki: Maybe rename it to something like
```suggestion
bool RNGHRuntimeDecorator::tryFindUIRuntime(
```

change it to return the runtime (or null), and then in the modules, do

```
uiRuntime = tryFindUIRuntime(rnRuntime)
installUIRuntimeBindings(uiRuntime)
```

packages/react-native-gesture-handler/shared/runtime/RNGHRuntimeDecorator.cpp
j-piasecki: `installRNRuntimeBindings` already does it, so this shouldn't be needed at all.

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
coado: Yeah, sure

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
coado: RN doesn't expose a public target with this header, and worklets include it privately, so as of now, there is no way to include it transitively.

packages/react-native-gesture-handler/android/build.gradle
coado: I think we can actually change the way we access the ui runtime and use solely StableApi. That would remove direct WorkletsModuleProxy access (include WorkletsModule.h), manual `ReactCommon/jsiexecutor` include path, Apple `WorkletsModule.h` and module-registry lookup, and the `__has_include`.

packages/react-native-gesture-handler/src/handlers/gestures/reanimatedWrapper.ts
```diff
@@ -82,26 +86,33 @@ let Reanimated:
...
+      globalThis.__RNGH_UI_WORKLET_RUNTIME_HOLDER = undefined;
     }
   });
+} catch (e) {
+  // When 'react-native-worklets' is not available we want to quietly continue
+}
+
+try {
+  Reanimated = require('react-native-reanimated');
 } catch (e) {
```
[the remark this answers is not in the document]
coado: good catch but the fix is meh

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -51,11 +56,38 @@ void RNGestureHandlerModule::setGestureState(
...
+
+  if (runtimeHolder.isObject()) {
+    uiWorkletRuntime = worklets::getWorkletRuntimeFromHolder(
+        *rnRuntime_, runtimeHolder.asObject(*rnRuntime_));
+
+    if (uiWorkletRuntime) {
+      uiRuntime = &worklets::getJSIRuntimeFromWorkletRuntime(uiWorkletRuntime);
+    }
+  }
+#endif
```
j-piasecki: Why isn't this a part of `RNGHRuntimeDecorator::tryFindUIRuntime`?

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -104,16 +108,40 @@ - (void)initialize
...
+  const auto runtimeHolder = _rnRuntime->global().getProperty(*_rnRuntime, "__RNGH_UI_WORKLET_RUNTIME_HOLDER");
+
+  if (runtimeHolder.isObject()) {
+    uiWorkletRuntime = worklets::getWorkletRuntimeFromHolder(*_rnRuntime, runtimeHolder.asObject(*_rnRuntime));
+
+    if (uiWorkletRuntime != nullptr) {
+      uiRuntime = &worklets::getJSIRuntimeFromWorkletRuntime(uiWorkletRuntime);
+    }
+  }
+#endif
```
j-piasecki: And this

packages/react-native-gesture-handler/android/build.gradle
```diff
@@ -107,6 +107,25 @@ def shouldUseCommonInterfaceFromRNSVG() {
...
+    if (worklets == null) {
+        return false
+    }
+
+    def packageJson = new JsonSlurper().parseText(new File(worklets.projectDir, "../package.json").text)
+    def version = packageJson.version as String
+
+    if (version.contains('-')) {
+        return false
+    }
```
j-piasecki: Why is this needed?

Other lib checks don't have this, maybe it would be possible to fold it into `getExternalLibVersion` if necessary?

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -51,11 +56,38 @@ void RNGestureHandlerModule::setGestureState(
...
+
+  if (uiRuntime == nullptr) {
+    uiRuntime = RNGHRuntimeDecorator::tryFindUIRuntime(*rnRuntime_);
+  }
+
+  if (uiRuntime == nullptr) {
+    return false;
+  }
```
m-bert: This might be the one example when nesting these `if` reads easier

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
```diff
@@ -51,11 +56,38 @@ void RNGestureHandlerModule::setGestureState(
   method(this->javaPart_, handlerTag, state);
 }
 
-bool RNGestureHandlerModule::decorateUIRuntime() {
-  return RNGHRuntimeDecorator::installUIRuntimeBindings(
-      *rnRuntime_, getModuleId(), [&](int handlerTag, int state) {
+bool RNGestureHandlerModule::installUIRuntimeBindings() {
+  jsi::Runtime *uiRuntime = nullptr;
+
+#ifdef RNGH_USE_WORKLETS
```
m-bert: `#ifdef` checks only if the value is defined, not if it's true - I guess there's no way that this will be defined, but not `1`, right?

packages/react-native-gesture-handler/src/global.d.ts
```diff
@@ -10,4 +10,8 @@ declare global {
   var _setGestureStateAsync:
     | ((handlerTag: number, state: number) => void)
     | undefined;
+
+  // Temporary holder used to pass the Worklets UI runtime to native code.
+  // eslint-disable-next-line no-var
```
m-bert: Maybe we should disable `var` rule for whole file?

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
coado: The RNGH_USE_WORKLETS is always set to `1` here but it's better to check that explicitly.

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
coado: Yeah, we can move it to a shared method.

packages/react-native-gesture-handler/android/src/main/jni/RNGestureHandlerModule.cpp
coado: You are absolutely right.

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
coado: yes

packages/react-native-gesture-handler/src/global.d.ts
coado: yes

packages/react-native-gesture-handler/android/build.gradle
coado: We should enable worklets integration only for stable releases. This checks for versions like `0.8.0-rc.1` because `getExternalLibVersion` strips prerelease suffix. 

We could return fourth argument `isPrerelease` from it.

j-piasecki: Yeah, let's do that

j-piasecki (approved):

PR #4281, opened 2026-06-23
Do not call `onPress` on buttons while dismissing keyboard

### LINKED ISSUE #992: BaseButton does not seem to respect `TouchableWithoutFeedback`
I have a ScrollView with `TouchableWithoutFeedback` set to `never`. If I put a `TouchableWithoutFeedback` from react native and a `TextInput` in this view tapping the `TouchableWithoutFeedback` does not trigger its on press and only blurs the `TextInput`. If I however replace the `TouchableWithoutFeedback` with a `BaseButton` from this package tapping the button both blurs the view and activates the `onPress` on `BaseButton`.

m-bert:
## Description

Currently, when buttons are inside `ScrolView` with `keyboardShouldPersistTaps={never}` they still call `onPress` even though press was meant to dismiss keyboard.

This PR fixes that behavior. It attaches listeners for keyboard state and based on that it blocks `onPress*` callbacks on JS side.

Fixes #992

## Test plan

Tested on Keyboard Should Persist Taps example

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
+++ b/packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
@@ -1,10 +1,58 @@
 import type { PropsWithChildren } from 'react';
-import React, { useCallback, useMemo, useRef } from 'react';
-import type { ScrollViewProps as RNScrollViewProps } from 'react-native';
-import { StyleSheet, View } from 'react-native';
+import React, { useCallback, useEffect, useMemo, useRef } from 'react';
+import type {
+  EmitterSubscription,
+  ScrollViewProps as RNScrollViewProps,
+} from 'react-native';
+import { Keyboard, StyleSheet, View } from 'react-native';
+
+type KeyboardShouldPersistTaps = RNScrollViewProps['keyboardShouldPersistTaps'];
+
+// The listeners are shared by every mounted ScrollView, so attach/detach is
+// reference-counted: we subscribe on the first interceptor and tear down only
+// once the last one unmounts.
+let keyboardTrackerRefCount = 0;
+let keyboardTrackerSubscriptions: EmitterSubscription[] = [];
+let isSoftKeyboardVisible = false;
+
+function subscribeToKeyboardVisibility() {
+  keyboardTrackerRefCount++;
+
+  if (keyboardTrackerRefCount > 1 || Keyboard?.addListener == null) {
+    return;
+  }
+
+  const setVisible = () => {
+    isSoftKeyboardVisible = true;
+  };
+  const setHidden = () => {
+    isSoftKeyboardVisible = false;
+  };
+
+  keyboardTrackerSubscriptions = [
+    Keyboard.addListener('keyboardDidShow', setVisible),
+    Keyboard.addListener('keyboardWillShow', setVisible),
+    Keyboard.addListener('keyboardDidHide', setHidden),
+  ];
+}
+
+function unsubscribeFromKeyboardVisibility() {
+  keyboardTrackerRefCount--;
+
+  if (keyboardTrackerRefCount > 0) {
+    return;
+  }
+
+  for (const subscription of keyboardTrackerSubscriptions) {
+    subscription.remove();
+  }
+  keyboardTrackerSubscriptions = [];
+  isSoftKeyboardVisible = false;
+}
 
 export type JSResponderContextValue = {
   isRNGHResponderEvent: React.MutableRefObject<boolean>;
+  keyboardShouldPersistTaps: KeyboardShouldPersistTaps;
 };
 
 export const JSResponderContext =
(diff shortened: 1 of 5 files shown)


packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
```diff
@@ -21,6 +69,19 @@ export function updateResponderEventValue(
...
+): boolean {
+  if (jsResponderContext == null) {
+    return false;
+  }
+
+  const mode = jsResponderContext.keyboardShouldPersistTaps;
+  const keyboardNeverPersistTaps = !mode || mode === 'never';
+
+  return keyboardNeverPersistTaps && isSoftKeyboardVisible;
+}
```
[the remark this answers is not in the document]
m-bert: [Tests](https://github.com/software-mansion/react-native-gesture-handler/pull/4281/commits/3dba8abc979b5e7dfa7a9d8cffb4a3e82ab32975)

packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
```diff
@@ -1,10 +1,58 @@
...
+
+function subscribeToKeyboardVisibility() {
+  keyboardTrackerRefCount++;
+
+  if (keyboardTrackerRefCount > 1 || Keyboard?.addListener == null) {
+    return;
+  }
+
+  const setVisible = () => {
+    isSoftKeyboardVisible = true;
```
coado: Why is that called `Soft` keyboard?

packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
```diff
@@ -1,10 +1,58 @@
...
+  };
+  const setHidden = () => {
+    isSoftKeyboardVisible = false;
+  };
+
+  keyboardTrackerSubscriptions = [
+    Keyboard.addListener('keyboardDidShow', setVisible),
+    Keyboard.addListener('keyboardWillShow', setVisible),
```
coado, line 43: There are also other things that are considered by RN to determine if the touch should be dropped:

https://github.com/react/react-native/blob/main/packages/react-native/Libraries/Components/ScrollView/ScrollView.js#L1512-L1516

packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
m-bert: `Software` 😉 But I agree that it is not obvious, removed.

packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
m-bert: Let my agent speak 😆:

Good call, dug into RN's [_handleStartShouldSetResponderCapture](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Components/ScrollView/ScrollView.js#L1512-L1516).

Took the parts that apply:

Detached keyboard — only treat it as open when endCoordinates.height > 0 (height-0/floating → let the tap through), like _softKeyboardIsDetached.
Seed from Keyboard.metrics() on mount, for a keyboard that's already open.
isKeyboardVisible rides the same keyboardDidShow/Hide events RN uses for _keyboardMetrics, so it's basically softKeyboardMayBeOpen.

Skipped the rest on purpose:

currentlyFocusedInput() — can't use it at gesture time; the ScrollView blurs the input synchronously when it grabs the responder, so it's already null by the time our recognizer fires (that race is why we track visibility instead).
target !isTextInput — moot, a Pressable/Touchable is never a TextInput.
_keyboardEventsAreUnreliable() (Android <30) — RN only uses it alongside hasFocusedTextInput; alone it'd drop taps with the keyboard closed.

coado: Okay 🤷‍♂️

packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
```diff
@@ -1,10 +1,67 @@
...
+    isKeyboardVisible = false;
+  };
+
+  // Seed from the current keyboard metrics in case it is already open when the
+  // first ScrollView mounts (mirrors ScrollView seeding from Keyboard.metrics()).
+  isKeyboardVisible = keyboardIsOpen(Keyboard.metrics?.()?.height);
+
+  keyboardTrackerSubscriptions = [
+    Keyboard.addListener('keyboardDidShow', setVisible),
+    Keyboard.addListener('keyboardWillShow', setVisible),
```
j-piasecki, line 43: Why do we subscribe to both?

m-bert: On `Android` there's no `keyboardWillShow` event. So this setup allows us to quicker react on `iOS` (and calling it twice is safe). Of course we don't have to do this and rely only on `keyboardDidShow` - that should suffice.

j-piasecki (approved):

PR #4282, opened 2026-06-24
[Android] Handle hover in `Touchable`

m-bert:
## Description

This PR brings hover effects for the `Touchable` component on Android. Previously, animations based on hover were possible only on web. 

Hovering out is handled in post-frame callback. This is due to the fact that on Android `ACTION_HOVER_EXIT` is dispatched right before `ACTION_DOWN`, so without delaying hover out effect we would be left with flickers.

## Test plan

<details>
<summary>Tested on existing Touchable example and on the code below:</summary>

```tsx
import {
  GestureHandlerRootView,
  Touchable as GHTouchable,

const SLOW = 700;

  const [pressCount, setPressCount] = React.useState(0);
  const [disabled, setDisabled] = React.useState(false);

  // Auto-toggle `disabled` every 2s so hover mask/resume can be tested with a
  // mouse held still over the button — tapping a separate control would move
  // the pointer off first. Hover and watch: the visual masks to default while
  // disabled and resumes the hover look when re-enabled, pointer unmoved.
  React.useEffect(() => {
    const id = setInterval(() => setDisabled((d) => !d), 2000);
    return () => clearInterval(id);
  }, []);

  return (
      <Text style={styles.title}>Slow hover + press</Text>
      <Text style={styles.hint}>
        Use a mouse / stylus. Hover to grow & fade, press to shrink & fade more.
        Transitions should never flicker through the default state.
      </Text>

      <View style={styles.stage}>
        <GHTouchable
          // Press (active) visuals
          defaultOpacity={1}
          defaultScale={1}
          activeOpacity={0.3}
          activeScale={0.8}
          // Hover visuals
          hoverOpacity={0.6}
          hoverScale={1.2}
          // Underlay so the change is extra visible
          underlayColor="black"
          defaultUnderlayOpacity={0}
          hoverUnderlayOpacity={0.15}
          activeUnderlayOpacity={0.35}
          // Slow everything down: in/out for both tap and hover
          animationDuration={{
            tap: { in: SLOW, out: SLOW },
            hover: { in: SLOW, out: SLOW },
          }}
          disabled={disabled}
          style={styles.button}
          onPress={() => setPressCount((c) => c + 1)}>
          <Text style={styles.buttonText}>Hover / Press me</Text>
        </GHTouchable>
      </View>

      <Text style={styles.counter}>
        {disabled ? 'DISABLED' : 'enabled'} · Presses: {pressCount}
      </Text>
    </GestureHandlerRootView>
  );
}

  root: {
    flex: 1,
    backgroundColor: '#ecf0f1',
    alignItems: 'center',
    justifyContent: 'center',
    padding: 24,
  },
  title: {
    fontSize: 22,
    fontWeight: 'bold',
    color: '#2c3e50',
    marginBottom: 8,
  },
  hint: {
    fontSize: 14,
    color: '#7f8c8d',
    textAlign: 'center',
    marginBottom: 40,
  },
  stage: {
    width: 260,
    height: 260,
    alignItems: 'center',
    justifyContent: 'center',
  },
  button: {
    width: 180,
    height: 180,
    borderRadius: 24,
    backgroundColor: '#8e44ad',
    alignItems: 'center',
    justifyContent: 'center',
  },
  buttonText: {
    color: 'white',
    fontSize: 18,
    fontWeight: 'bold',
  },
  counter: {
    marginTop: 40,
    fontSize: 16,
    color: '#2c3e50',
    fontWeight: 'bold',
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -442,13 +442,16 @@ class RNGestureHandlerButtonViewManager :
     private var isPointerInsideBounds = false
     private var isHovered = false
 
-    // Resting (non-pressed) animation targets. While the pointer hovers the
-    // button, the press-out animation settles on the hover values instead of
-    // the defaults, mirroring the web priority order (pressed > hovered >
-    // default).
-    private val restingOpacity get() = if (isHovered) hoverOpacity else defaultOpacity
-    private val restingScale get() = if (isHovered) hoverScale else defaultScale
-    private val restingUnderlayOpacity get() = if (isHovered) hoverUnderlayOpacity else defaultUnderlayOpacity
+    // The hover visual is masked while disabled
+    private val effectivelyHovered get() = isHovered && isEnabled
+
+    // Resting (non-pressed) animation targets. While the pointer effectively
+    // hovers the button, the press-out animation settles on the hover values
+    // instead of the defaults, mirroring the web priority order (pressed >
+    // hovered > default).
+    private val restingOpacity get() = if (effectivelyHovered) hoverOpacity else defaultOpacity
+    private val restingScale get() = if (effectivelyHovered) hoverScale else defaultScale
+    private val restingUnderlayOpacity get() = if (effectivelyHovered) hoverUnderlayOpacity else defaultUnderlayOpacity
 
     // When non-null the ripple is drawn in dispatchDraw (above background, below children).
     // When null the ripple lives on the foreground drawable instead.
@@ -722,7 +725,7 @@ class RNGestureHandlerButtonViewManager :
         return
       }
 
-      if (isEnabled && isHovered) {
+      if (effectivelyHovered) {
         animateTo(hoverOpacity, hoverScale, hoverUnderlayOpacity, hoverAnimationInDuration.toLong())
       } else {
         animateTo(defaultOpacity, defaultScale, defaultUnderlayOpacity, hoverAnimationOutDuration.toLong())
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -18,6 +18,7 @@ import android.graphics.drawable.shapes.RectShape
 import android.os.Build
 import android.os.SystemClock
 import android.util.TypedValue
+import android.view.Choreographer
 import android.view.KeyEvent
 import android.view.MotionEvent
 import android.view.View
@@ -333,6 +334,31 @@ class RNGestureHandlerButtonViewManager :
     view.activeUnderlayOpacity = activeUnderlayOpacity
   }
 
+  @ReactProp(name = "hoverOpacity")
+  override fun setHoverOpacity(view: ButtonViewGroup, hoverOpacity: Float) {
+    view.hoverOpacity = hoverOpacity
+  }
+
+  @ReactProp(name = "hoverScale")
+  override fun setHoverScale(view: ButtonViewGroup, hoverScale: Float) {
+    view.hoverScale = hoverScale
+  }
+
+  @ReactProp(name = "hoverUnderlayOpacity")
+  override fun setHoverUnderlayOpacity(view: ButtonViewGroup, hoverUnderlayOpacity: Float) {
+    view.hoverUnderlayOpacity = hoverUnderlayOpacity
+  }
+
+  @ReactProp(name = "hoverAnimationInDuration")
+  override fun setHoverAnimationInDuration(view: ButtonViewGroup, value: Int) {
+    view.hoverAnimationInDuration = if (value > 0) value else 0
+  }
+
+  @ReactProp(name = "hoverAnimationOutDuration")
+  override fun setHoverAnimationOutDuration(view: ButtonViewGroup, value: Int) {
+    view.hoverAnimationOutDuration = if (value > 0) value else 0
+  }
+
   @ReactProp(name = ViewProps.POINTER_EVENTS)
   override fun setPointerEvents(view: ButtonViewGroup, pointerEvents: String?) {
     view.pointerEvents = when (pointerEvents) {
@@ -383,6 +409,14 @@ class RNGestureHandlerButtonViewManager :
     var defaultOpacity: Float = 1.0f
     var activeScale: Float = 1.0f
     var defaultScale: Float = 1.0f
+    var hoverAnimationInDuration: Int = 50
+    var hoverAnimationOutDuration: Int = 100
+    var hoverOpacity: Float = -1f
+      get() = if (field < 0f) defaultOpacity else field
+    var hoverScale: Float = -1f
+      get() = if (field < 0f) defaultScale else field
+    var hoverUnderlayOpacity: Float = -1f
+      get() = if (field < 0f) defaultUnderlayOpacity else field
     var underlayColor: Int? = null
       set(color) = withBackgroundUpdate {
         field = color
@@ -404,7 +438,17 @@ class RNGestureHandlerButtonViewManager :
     private var underlayDrawable: PaintDrawable? = null
     private var pressInTimestamp = 0L
     private var pendingPressOut: Runnable? = null
+    private var pendingHoverOut: Choreographer.FrameCallback? = null
     private var isPointerInsideBounds = false
+    private var isHovered = false
+
+    // Resting (non-pressed) animation targets. While the pointer hovers the
+    // button, the press-out animation settles on the hover values instead of
+    // the defaults, mirroring the web priority order (pressed > hovered >
+    // default).
+    private val restingOpacity get() = if (isHovered) hoverOpacity else defaultOpacity
+    private val restingScale get() = if (isHovered) hoverScale else defaultScale
+    private val restingUnderlayOpacity get() = if (isHovered) hoverUnderlayOpacity else defaultUnderlayOpacity
 
     // When non-null the ripple is drawn in dispatchDraw (above background, below children).
     // When null the ripple lives on the foreground drawable instead.
@@ -503,6 +547,12 @@ class RNGestureHandlerButtonViewManager :
       val eventTime = event.eventTime
       val action = event.action
 
+      if (event.actionMasked == MotionEvent.ACTION_DOWN ||
+        event.actionMasked == MotionEvent.ACTION_POINTER_DOWN
+      ) {
+        cancelPendingHoverOut()
+      }
+
       if (touchResponder != null && touchResponder !== this && touchResponder!!.exclusive) {
         if (isPressed) {
           setPressed(false)
@@ -550,6 +600,17 @@ class RNGestureHandlerButtonViewManager :
       return false
     }
 
+    override fun onHoverEvent(event: MotionEvent): Boolean {
+      if (isEnabled) {
+        when (event.actionMasked) {
+          MotionEvent.ACTION_HOVER_ENTER -> animateHoverIn()
+          MotionEvent.ACTION_HOVER_EXIT -> animateHoverOut()
+        }
+      }
+
+      return super.onHoverEvent(event)
+    }
+
     private fun getAnimatorDurationScale(): Float = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
       ValueAnimator.getDurationScale()
     } else {
@@ -564,10 +625,10 @@ class RNGestureHandlerButtonViewManager :
     }
 
     private fun applyStartAnimationState() {
-      if (activeOpacity != 1.0f || defaultOpacity != 1.0f) {
+      if (activeOpacity != 1.0f || defaultOpacity != 1.0f || hoverOpacity != 1.0f) {
         alpha = defaultOpacity
       }
-      if (activeScale != 1.0f || defaultScale != 1.0f) {
+      if (activeScale != 1.0f || defaultScale != 1.0f || hoverScale != 1.0f) {
         scaleX = defaultScale
         scaleY = defaultScale
       }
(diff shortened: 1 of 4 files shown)


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -647,20 +756,20 @@ class RNGestureHandlerButtonViewManager :
 
       if (longPressMs >= 0 && elapsed >= longPressMs) {
         // Long-press release - use the configured long-press out duration.
-        animateTo(defaultOpacity, defaultScale, defaultUnderlayOpacity, longPressOutMs)
+        animateTo(restingOpacity, restingScale, restingUnderlayOpacity, longPressOutMs)
       } else if (elapsed >= tapInMs) {
         // Press-in animation fully finished — release with the configured out duration.
```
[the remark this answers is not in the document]
m-bert: [Pointer up outside](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/1e69e68d8f3df597d501d450717fe4caca01f530)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -550,6 +600,17 @@ class RNGestureHandlerButtonViewManager :
...
+    override fun onHoverEvent(event: MotionEvent): Boolean {
+      if (isEnabled) {
+        when (event.actionMasked) {
+          MotionEvent.ACTION_HOVER_ENTER -> animateHoverIn()
+          MotionEvent.ACTION_HOVER_EXIT -> animateHoverOut()
+        }
+      }
+
+      return super.onHoverEvent(event)
+    }
```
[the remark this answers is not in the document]
m-bert: [Handle enabled](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/6be95a3954ed4e749efbeacf3a83801909c0dc32)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -520,14 +570,34 @@ class RNGestureHandlerButtonViewManager :
...
+        // super.onTouchEvent so a press-out it triggers (release, or
+        // cancel-on-leave) settles on the correct resting visual — hover while
+        // still over the view, default once the pointer leaves.
+        when (event.actionMasked) {
+          MotionEvent.ACTION_MOVE,
+          MotionEvent.ACTION_UP,
+          MotionEvent.ACTION_POINTER_UP,
+          ->
+            if (isHoveringPointer(event)) {
+              isHovered = isWithinBounds(event)
```
j-piasecki, line 589: What about styluses without hover support?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -637,6 +717,61 @@ class RNGestureHandlerButtonViewManager :
...
+      cancelPendingHoverOut()
+
+      // We have to defer hover-out due to MotionEvent orders.
+      // In case of a press, the hover-out is sent before the press-in,
+      // so we need to wait for the next frame to animate back to the default state.
+      val callback = Choreographer.FrameCallback {
+        pendingHoverOut = null
+        isHovered = false
+        applyHoverState()
+      }
```
j-piasecki, line 754: Same as on iOS - do we need to schedule this to the next frame? Aren't hover-out and press-down happening at the same time?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: [Stylus w/o hover](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/f1221006706a47e6eabcf6247aa6542b8d59f449)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -404,7 +438,17 @@ class RNGestureHandlerButtonViewManager :
...
     private var isPointerInsideBounds = false
+    private var isHovered = false
+
+    // Resting (non-pressed) animation targets. While the pointer hovers the
+    // button, the press-out animation settles on the hover values instead of
+    // the defaults, mirroring the web priority order (pressed > hovered >
+    // default).
+    private val restingOpacity get() = if (isHovered) hoverOpacity else defaultOpacity
+    private val restingScale get() = if (isHovered) hoverScale else defaultScale
+    private val restingUnderlayOpacity get() = if (isHovered) hoverUnderlayOpacity else defaultUnderlayOpacity
```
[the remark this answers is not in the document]
m-bert: [Add effective hover](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/20050b223e60a2b9a0f78143e35e1394e80f92a5)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: > do we need to schedule this to the next frame?

Unfortunately yes 😞 While those events have the same timestamp, Android dispatches hover event first. That results in first running hover out animation and then going to pressed state - flickers are visible (presented offline).

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -637,6 +730,58 @@ class RNGestureHandlerButtonViewManager :
...
+        animateTo(defaultOpacity, defaultScale, defaultUnderlayOpacity, hoverAnimationOutDuration.toLong())
+      }
+    }
+
+    private fun animateHoverIn() {
+      cancelPendingHoverOut()
+
+      if (isHovered) {
+        return
+      }
```
j-piasecki, line 734: Doesn't this have a chance to leave the button in the hovered state indefinitely?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -404,7 +438,28 @@ class RNGestureHandlerButtonViewManager :
...
+    // Whether a real hover was active when the current press began. A
+    // hover-capable pointer (mouse, or a stylus that supports hover) fires
+    // ACTION_HOVER_ENTER before the press, so `isHovered` is already true at
+    // ACTION_DOWN; a stylus without hover support never does. The touch-stream
+    // derivation below uses this to only *maintain* an existing hover, never
+    // fabricate one for a non-hovering tool.
+    private var hoverActiveAtPressStart = false
+
+    // The hover visual is masked while disabled
+    private val effectivelyHovered get() = isHovered && isEnabled
```
j-piasecki: I'm not a fan of the name here.

> Effectively primarily means "for all practical purposes" or "in actual fact," even if it is not officially or theoretically true. It is also used to describe doing something successfully to produce an intended result.

Here, it's the opposite - the view is hovered, but since it's disabled, we don't want to show it, so "effectively not hovered"

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -564,20 +656,21 @@ class RNGestureHandlerButtonViewManager :
...
     }
 
     private fun animateTo(opacity: Float, scale: Float, underlayOpacity: Float, durationMs: Long) {
-      val hasOpacity = activeOpacity != 1.0f || defaultOpacity != 1.0f
-      val hasScale = activeScale != 1.0f || defaultScale != 1.0f
-      val hasUnderlay = activeUnderlayOpacity != defaultUnderlayOpacity && underlayDrawable != null
+      val hasOpacity = activeOpacity != 1.0f || defaultOpacity != 1.0f || hoverOpacity != 1.0f
+      val hasScale = activeScale != 1.0f || defaultScale != 1.0f || hoverScale != 1.0f
+      val hasUnderlay = underlayDrawable != null &&
+        (activeUnderlayOpacity != defaultUnderlayOpacity || hoverUnderlayOpacity != defaultUnderlayOpacity)
```
j-piasecki: I think those can be getter properties to match ios

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -637,6 +730,58 @@ class RNGestureHandlerButtonViewManager :
       animateTo(activeOpacity, activeScale, activeUnderlayOpacity, tapAnimationInDuration.toLong())
     }
 
+    private fun applyHoverState() {
```
j-piasecki:
```suggestion
    private fun animateHoverState() {
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -637,6 +730,58 @@ class RNGestureHandlerButtonViewManager :
...
+      }
+
+      if (effectivelyHovered) {
+        animateTo(hoverOpacity, hoverScale, hoverUnderlayOpacity, hoverAnimationInDuration.toLong())
+      } else {
+        animateTo(defaultOpacity, defaultScale, defaultUnderlayOpacity, hoverAnimationOutDuration.toLong())
+      }
+    }
+
+    private fun animateHoverIn() {
```
j-piasecki:
```suggestion
    private fun onHoverIn() {
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -637,6 +730,58 @@ class RNGestureHandlerButtonViewManager :
...
+
+      if (isHovered) {
+        return
+      }
+
+      isHovered = true
+      applyHoverState()
+    }
+
+    private fun animateHoverOut() {
```
j-piasecki:
```suggestion
    private fun onHoverOut() {
```

j-piasecki (commented):
I think the comments are a bit overkill there

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: I don't think so. For it to happen `isHovered` would have to be stuck at `true`. This flag is set to `false` in two cases - scheduled in `onHoverOut` and in `ACTION_CANCEL`.

`onHoverIn` cancels scheduled block, but in the next frame. Also this is necessary since we don't want a flicker of hover animation. So in practice we have it set to `false` in two cases - when hover ends and we get `HOVER_EXIT` event (and this along with `HOVER_ENTER` come in pair) and we don't click the view.  The second one when interaction gets canceled.

That said, I don't think it will get stuck, but if you see how it would happen then let me know.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: [Renames](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/f1d0fc4f3c8d2e976e8eff456887ba4a651078d5)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: [Renames](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/f1d0fc4f3c8d2e976e8eff456887ba4a651078d5)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: [Renames](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/f1d0fc4f3c8d2e976e8eff456887ba4a651078d5)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: [Renames](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/f1d0fc4f3c8d2e976e8eff456887ba4a651078d5)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: [Convert to getters](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/70fbd36d702ebd2db763639b979678ae4f59e126)

m-bert:
> I think the comments are a bit overkill there

Could be 😅  [Trim comments](https://github.com/software-mansion/react-native-gesture-handler/pull/4282/commits/2d316e6622adc4bfe5ba289fa632f37a0fbc115a)

j-piasecki (approved):

PR #4289, opened 2026-07-01
[docs] Add guide for transformations example

m-bert:
## Description

Originally we thought about adding `useTransformGesture` hook. However, we decided to drop the idea and describe important caveats in docs guide.

I've also updated transformations examples to use 3x3 matrices instead of 4x4.

## Test plan

- Transformations examples
- Read docs 🤓


packages/docs-gesture-handler/docs/guides/transformations.mdx
```diff
@@ -0,0 +1,344 @@
...
+So keep the result so far in a single matrix, and let the in-progress gesture stay as plain `scale`, `rotation` and `translation` — that's what the callbacks hand you as the fingers move. Each frame you combine the two to draw the view, and when the gesture ends you fold its values into the matrix and reset them, so the next gesture picks up where it left off.
+
+## Keep the origin stable
+
+Scaling and rotation pivot around the origin, but the user expects them to pivot around the point between their fingers. To move the pivot, wrap the transform between two translations — shift that point to the origin, apply the scale or rotation, then shift it back. Do this while building the matrix, once for the scale and once for the rotation:
+
+```ts
+matrix = multiply(matrix, translate(origin.x, origin.y));
+matrix = multiply(matrix, scale(scaleValue, scaleValue));
+matrix = multiply(matrix, translate(-origin.x, -origin.y));
```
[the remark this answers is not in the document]
m-bert: `3` suffix was removed on purpose so it is not confusing at this specific place in docs

packages/docs-gesture-handler/docs/guides/transformations.mdx
```diff
@@ -0,0 +1,344 @@
...
+
+function invert2(m: number[]) {
+  'worklet';
+  const a = m[0];
+  const b = m[1];
+  const c = m[2];
+  const d = m[3];
+  const det = a * d - b * c;
+
+  return [d / det, -b / det, -c / det, a / det];
```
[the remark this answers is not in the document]
m-bert: [Add check for det=0](https://github.com/software-mansion/react-native-gesture-handler/pull/4289/commits/b8bbdea01ed2232a952878695f2ba41885517f42)

packages/docs-gesture-handler/docs/guides/transformations.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+This guide assumes you're using [Reanimated](https://docs.swmansion.com/react-native-reanimated/) to run the transformation on the UI thread. The [full example](#full-example) is at the end of this page, and a version of it can be found in the [example app](https://github.com/software-mansion/react-native-gesture-handler/tree/main/apps/common-app/src/new_api/complicated/transformations).
+
+## Use a matrix, not separate transforms
+
+The obvious approach — keeping separate `scale`, `rotation` and `translation` values and passing them to a `transform` array — falls apart once the gestures combine. The array applies its entries in a fixed order, so you can't decide whether scaling happens before or after rotation, and there's nowhere to accumulate the result of one gesture so the next can build on it.
```
coado: I don't have much experience writing docs, but wouldn't it be better if the user first sees how it should be done, and later, if they are curious, they can read about other approaches and their disadvantages?

j-piasecki: > The array applies its entries in a fixed order, so you can't decide whether scaling happens before or after rotation

You can, and that's exactly why it's an array of objects and not a simple object with all the transform props.

The issue is that every little move/scale/rotation would need to be added to the array, and React would build a new matrix from those entries on each render.

m-bert: The opening paragraph has link to the example. There's also "Full example" section and I think it would look strange if we put it earlier. 

> if they are curious, they can read about other approaches and their disadvantages?

My point when creating this guide was to explain how to do it correctly (and why other approach is bad) and then show example, not the other way around. Of course we can change it, but I think this approach fits better.

m-bert: Let me know what you think ([8bff978](https://github.com/software-mansion/react-native-gesture-handler/pull/4289/commits/8bff978ba50a68f9932c5ed83e20c0ce14a20a7e))

j-piasecki: > My point when creating this guide was to explain how to do it correctly (and why other approach is bad) and then show example, not the other way around.

It is the other way around - it starts with why the other approach is bad, then explains how to do it correctly.

m-bert: > It is the other way around - it starts with why the other approach is bad, then explains how to do it correctly.

Okay, I think we are focused on 2 different things 😅 That's why I used word "and". By the "other way around" I meant first showing the example. 

Anyway, I think I know what you mean by that. I'll rephrase it and get back to you.

m-bert: Flipped order in [b575ce5](https://github.com/software-mansion/react-native-gesture-handler/pull/4289/commits/b575ce56dd485a5b1caac4a4756b5c73c641c207)

j-piasecki (approved):

PR #4291, opened 2026-07-02
fix: remove UIPointerInteraction from the view when unbinding the hover handler

### LINKED ISSUE #4290: Hover handler never removes its UIPointerInteraction on unbind, leaking hover effects onto recycled views (iOS)
### Description

When a `Gesture.Hover()` handler is attached on iOS, `RNHoverGestureHandler` adds a `UIPointerInteraction` to the view in `bindToView:`. When the handler is detached, `unbindFromView` tries to remove it:

```objc
- (void)unbindFromView
{
#if CHECK_TARGET(13_4)
  if (@available(iOS 13.4, *)) {
    [super unbindFromView];
    [self.recognizer.view removeInteraction:_pointerInteraction];
  }
#endif
}
```

`[super unbindFromView]` removes the gesture recognizer from the view, which sets `self.recognizer.view` to nil. The next line then sends `removeInteraction:` to nil, so it does nothing. The `UIPointerInteraction` is never removed from the `UIView`.

Once the handler is dropped, the recognizer is deallocated. `UIPointerInteraction.delegate` is weak, and an interaction with a nil delegate applies the system default pointer effect over the whole view. Confirmed consequences on a real iPad running iPadOS 26 (see the repro):

1. A view whose detector swaps its gesture from Hover to another type keeps showing the native hover effect even though it no longer has any hover gesture (the highlight is very prominent with the Liquid Glass style).
2. Fabric recycles the freed native views of unmounted hover views into unrelated components, which then show the native hover effect. In the repro this is exact: unmounting 10 hover views and adding plain `View`s makes precisely the first 10 recycled targets light up.
3. After hovering such a poisoned recycled view, unmounting it aborts the app: SIGABRT from the Fabric assertion in `-[RCTViewComponentView unmountChildComponentView:index:]` (via `RCTPerformMountInstructions`). The system hover effect changes the hierarchy around the poisoned view and the next unmount transaction finds a state Fabric does not expect. So this is not only cosmetic.

Expected instead: no native hover effect on views without a hover gesture, and no crash.

Note for reproducing: Expo Go cannot show the visible symptom. It bundles RNGH 2.28.0, where the hover recognizer was still created with `initWithTarget:gestureHandler` (changed to `self` later), which keeps the interaction's delegate alive and a live delegate returning nil suppresses the effect. The interaction still leaks there, it is just invisible. Use a development build with RNGH 2.29+.

There is a one-line fix in #4291 (capture the view before calling `[super unbindFromView]`). It was verified against the repro: hover callbacks keep working, no stray native hover effects appear, and the crash is gone.

### Steps to reproduce

Run the repro app on an iPad with Apple Pencil hover or a trackpad pointer, then either test:

Deterministic (three taps, no recycling involved):
1. Hover the yellow box that has `Gesture.Hover()` (it reacts via a JS callback).
2. Tap "Swap to tap gesture" (the same mounted view's detector now has `Gesture.Tap()`, so RNGH drops the hover handler and re-attaches).
3. Hover the box again. It shows the native hover effect although it has no hover gesture anymore.

Recycling:
1. Tap "Unmount hover views" (10 views with bare `Gesture.Hover()` unmount).
2. Tap "Add 30 targets" twice (plain `View`s with no gestures, only added, never removed, so the recycle pool drains).
3. Hover the numbered targets: the first 10, which received the recycled views, show the native hover effect.
4. Tap "Clear" after hovering them: the app crashes.

https://github.com/petterikorpimaa/rngh-hover-leak-repro

### Gesture Handler version

2.32.0 (the code is unchanged on `main`)

### React Native version

0.86.0

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

Real device

### Device model

iPad with Apple Pencil hover (iPadOS 26.2)

### Acknowledgements

Yes

petterikorpimaa:
## Description

Fixes #4290.

`RNHoverGestureHandler` adds a `UIPointerInteraction` to the view in `bindToView:`, but never actually removes it. In `unbindFromView`, `[super unbindFromView]` runs first and detaches the gesture recognizer, which sets `self.recognizer.view` to nil. The following `removeInteraction:` call is then a message to nil and does nothing, so the interaction stays on the `UIView` for the rest of its life.

With Fabric view recycling, that leaked interaction later belongs to an unrelated component. Its delegate (the recognizer) is gone by then, and a `UIPointerInteraction` without a delegate applies the system default pointer effect over the whole view. The visible result is native hover effects (very prominent with the Liquid Glass style on iPadOS 26) appearing on random elements that never had a hover gesture, accumulating as hover handlers unmount and remount.

This PR captures the view reference before calling `[super unbindFromView]` and removes the interaction from the captured reference.

## Test plan

Tested in an app that uses `Gesture.Hover()` on most of its toolbar and menu items, on a real iPad (iPadOS 26) with Apple Pencil hover and a trackpad:

- Before the change: after a few mount/unmount cycles of hover-enabled views, hovering over unrelated elements shows the native hover effect on random views.
- With the change (applied to the app as a package patch): the stray hover effects no longer appear, and hover gestures keep working as before.
- The `RNGestureHandler` pod compiles cleanly with the change.

### DIFF
--- a/packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
+++ b/packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
@@ -122,8 +122,9 @@ API_AVAILABLE(ios(13.4))
 {
 #if CHECK_TARGET(13_4)
   if (@available(iOS 13.4, *)) {
-    [super unbindFromView];
+    // Remove the interaction before [super unbindFromView] detaches the recognizer and nils recognizer.view.
     [self.recognizer.view removeInteraction:_pointerInteraction];
+    [super unbindFromView];
   }
 #endif
 }


petterikorpimaa:
Standalone repro, with results detailed in #4290: https://github.com/petterikorpimaa/rngh-hover-leak-repro

Verified this fix against it on a real iPad (iPadOS 26, RN 0.86.0, RNGH 2.32.0). Without the fix: a deterministic three-tap repro shows the leaked native hover effect on a still-mounted view, recycled views spread it to unrelated components, and unmounting a poisoned view after hovering it aborts the app in a Fabric assertion (`unmountChildComponentView:index:`). With the fix: hover callbacks keep working, no stray native hover effects, no crash.

packages/react-native-gesture-handler/apple/Handlers/RNHoverHandler.m
```diff
@@ -122,8 +122,11 @@ - (void)unbindFromView
 {
 #if CHECK_TARGET(13_4)
   if (@available(iOS 13.4, *)) {
+    // Capture the view before [super unbindFromView] detaches the recognizer
+    // and sets recognizer.view to nil, otherwise the interaction is never removed.
+    UIView *view = self.recognizer.view;
     [super unbindFromView];
-    [self.recognizer.view removeInteraction:_pointerInteraction];
+    [view removeInteraction:_pointerInteraction];
```
j-piasecki: Wouldn't simply swapping the order also fix that?

petterikorpimaa: Good call, it does. Verified on an iPad and PR updated 👍

j-piasecki (approved):
Thanks ❤️!

PR #4298, opened 2026-07-06
Fix Easing import in LongPress example

haxonadora:
## Description

`Easing` should be imported from `react-native-reanimated` instead of `react-native`. `Easing` from `react-native` is not a worklet like the one from `react-native-reanimated`. 

Using `Easing` from `react-native` here throws an error `[Worklets] Tried to synchronously call a Remote Function. Called "BezierEasing" on the UI Runtime`


j-piasecki (approved):
Thanks!

PR #4303, opened 2026-07-07
[Web] Make `getEffectiveBoundingRect` recursive

m-bert:
## Description

Currently `getEffectiveBoundingRect` takes bounding box either from view, or children if element has `display: contents;`. However, in case of nested detectors this brakes, as can be seen when running the example code below.

This PR makes `getEffectiveBoundingRect` recursive, so it will not stop on the first layer of children.

## Test plan

<details>

```tsx
import {
  GestureDetector,
  useHoverGesture,
  useNativeGesture,

// Web-only repro for the getEffectiveBoundingRect fix.
//
// The OUTER hover gesture has a `hitSlop`, which makes its handler run
// checkHitSlop() → measureView() → getEffectiveBoundingRect() on the detector's
// `display:contents` host. The INNER GestureDetector nests a SECOND
// `display:contents` host between that host and the real box (the same shape as
// a hover detector wrapping a Touchable). Before the fix, the outer host's only
// child was itself `display:contents`, so its measured rect collapsed to 0×0 at
// the origin and checkHitSlop rejected every real pointer — hover never fired.
// After the fix, getEffectiveBoundingRect recurses through the inner
// `display:contents` child to the real box, so hover activates normally.
  const [hovered, setHovered] = useState(false);

  const hover = useHoverGesture({
    manualActivation: true,
    disableReanimated: true,
    hitSlop: { top: 10, bottom: 10, left: 10, right: 10 },
    onBegin: () => setHovered(true),
    onFinalize: () => setHovered(false),
  });

  // Any inner gesture works — it exists only to nest a second detector host.
  const inner = useNativeGesture({});

  return (
      <GestureDetector gesture={hover}>
        <GestureDetector gesture={inner}>
          <View style={[styles.box, hovered && styles.boxHovered]}>
            <Text style={styles.label}>
              {hovered ? 'hovered ✅' : 'hover me (web)'}
            </Text>
          </View>
        </GestureDetector>
      </GestureDetector>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
  },
  box: {
    width: 220,
    height: 130,
    borderRadius: 16,
    backgroundColor: '#a78bfa',
    alignItems: 'center',
    justifyContent: 'center',
  },
  boxHovered: {
    backgroundColor: '#34d399',
  },
  label: { fontSize: 18, fontWeight: '700', color: '#1a1a2e' },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
@@ -7,6 +7,7 @@ import type IGestureHandler from '../handlers/IGestureHandler';
 import type { SVGRef } from '../interfaces';
 import {
   getEffectiveBoundingRect,
+  hasDisplayContents,
   isPointerInBounds,
   isRNSVGElement,
 } from '../utils';
@@ -143,7 +144,7 @@ export class GestureHandlerWebDelegate
 
     const localView =
       this.gestureHandler.usesNativeOrVirtualDetector() &&
-      this.view.style.display === 'contents'
+      hasDisplayContents(this.view)
         ? (this.view.children[0] as HTMLElement)
         : this.view;
 
@@ -345,11 +346,7 @@ export class GestureHandlerWebDelegate
   ): void {
     this.ensureView(this.view);
 
-    const hasDisplayContents =
-      this.view.style.display === 'contents' ||
-      getComputedStyle(this.view).display === 'contents';
-
-    if (hasDisplayContents) {
+    if (hasDisplayContents(this.view)) {
       for (const child of Array.from(this.view.children)) {
         if (child instanceof HTMLElement) {
           // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access
--- a/packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
+++ b/packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
@@ -6,6 +6,7 @@ import { SingleGestureName } from '../../v3/types';
 import type IGestureHandler from '../handlers/IGestureHandler';
 import type { SVGRef } from '../interfaces';
 import {
+  firstNonContentsView,
   getEffectiveBoundingRect,
   hasDisplayContents,
   isPointerInBounds,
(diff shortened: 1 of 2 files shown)


packages/react-native-gesture-handler/src/web/utils.ts
```diff
@@ -3,7 +3,8 @@ import { PointerType } from '../PointerType';
 import type { GestureHandlerRef, Point, SVGRef } from './interfaces';
 
 // For display: contents elements (like the gesture detector wrapper), getBoundingClientRect
-// returns all zeros since the element has no box. Compute the bounding box from children instead.
+// returns all zeros since the element has no box. Derive the bounds from the children instead
+// (recurse until we reach elements that actually have a box).
 export function getEffectiveBoundingRect(view: HTMLElement): DOMRect {
   if (view.style.display === 'contents' && view.children.length > 0) {
```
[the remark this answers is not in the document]
m-bert: [Add helper](https://github.com/software-mansion/react-native-gesture-handler/pull/4303/commits/087bc4061293d6c00ce8e7f65dce62bc8566009c)

packages/react-native-gesture-handler/src/web/tools/GestureHandlerWebDelegate.ts
```diff
@@ -143,7 +144,7 @@ export class GestureHandlerWebDelegate
 
     const localView =
       this.gestureHandler.usesNativeOrVirtualDetector() &&
-      this.view.style.display === 'contents'
+      hasDisplayContents(this.view)
         ? (this.view.children[0] as HTMLElement)
         : this.view;
```
j-piasecki: Why does this work? Now you're checking for `display: contents` recursively, but you're still taking the first child when it's set.

m-bert: Well, this one (`absoluteToLocal`) is used only in `Pinch` and `Rotation` to calculate correct `anchor` and `focal` points. I have not tried degenerate example of nesting these handlers with other `display: contents;` view. Problems with `Hover` were discovered only because of our on-going `Pressable` refactor. 

Anyway, given that it is better to use correct view rather than assuming correct hierarchy, I've changed that ([bd1fa6e](https://github.com/software-mansion/react-native-gesture-handler/pull/4303/commits/bd1fa6e2b3e7b000b49b1e5a9af587f97e8b0c4e)).

Let me know what you think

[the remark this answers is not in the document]
m-bert: Also changed in [bd1fa6e](https://github.com/software-mansion/react-native-gesture-handler/pull/4303/commits/bd1fa6e2b3e7b000b49b1e5a9af587f97e8b0c4e)

j-piasecki: Ideally, I think we would calculate the bounding box of all non-contents nodes, but this is fine for now.

j-piasecki (approved):

PR #4309, opened 2026-07-09
Add Imperative Gesture Handler Testing API

coado:
## Description

The goal is to make gesture lifecycle tests easier to write when we want to assert application state after each gesture step, without requiring users to manually construct RNGH state events with `state`, `oldState`, or `handlerTag`.

Instead of this lower-level style:

```ts
fireGestureHandler(panGesture, [
  { oldState: State.UNDETERMINED, state: State.BEGAN },
  { oldState: State.BEGAN, state: State.ACTIVE },
  { oldState: State.ACTIVE, state: State.END },
]);
```

tests can now use imperative controller:
```ts
const gesture = createGestureController(panGesture);

gesture.begin();
expect(onBegin).toHaveBeenCalled();

gesture.activate();
expect(onActivate).toHaveBeenCalled();

gesture.update({ translationX: 50 });
expect(onUpdate).toHaveBeenCalledWith(
  expect.objectContaining({ translationX: 50 })
);

gesture.end();
expect(onFinalize).toHaveBeenCalledWith(
  expect.objectContaining({ canceled: false })
);
```

## Hook gesture rerenders

Hook-based gestures can be recreated when their callbacks or configuration change during a React rerender while retaining the same handler tag. The controller now resolves the latest registered gesture before every lifecycle operation. This ensures that subsequent steps use the newest callback closures and configuration, including the current enabled value.

For example, if a rerender occurs between `begin()` and `activate()`, `activate()` invokes the callback from the latest render rather than the callback captured when the controller was created.

## Reusing a controller for another stream

A controller can now run multiple gesture streams.

After `end()`, `fail()`, or `cancel()`, the terminal state remains available for assertions through `getState()`. Calling `begin()` again resets the finished controller internally and starts a new stream.

```ts
controller.begin();
controller.activate();
controller.end();

expect(controller.getState()).toBe(State.END);

controller.begin();
expect(controller.getState()).toBe(State.BEGAN);
```

## Test plan

Added tests using new API.


packages/docs-gesture-handler/docs/guides/testing.mdx
```diff
@@ -62,7 +62,94 @@ In order to load mocks provided by RNGH, add the following to your jest config:
...
+
+  controller.activate();
+  expect(onActivate).toHaveBeenCalledTimes(1);
+
+  controller.update({ translationX: 50 });
+  expect(onUpdate).toHaveBeenCalledWith(
+    expect.objectContaining({ translationX: 50 })
+  );
+
+  controller.end();
```
j-piasecki, line 165: I assume this will also call `onDeactivate` in this scenario, right?

packages/docs-gesture-handler/docs/guides/testing.mdx
```diff
@@ -62,7 +62,94 @@ In order to load mocks provided by RNGH, add the following to your jest config:
...
+
+| Method       | Behavior                                                                                                         |
+| ------------ | ---------------------------------------------------------------------------------------------------------------- |
+| `begin()`    | Starts a stream and calls `onBegin`. If the previous stream finished, the controller resets it before beginning. |
+| `activate()` | Activates a begun stream and calls `onActivate`.                                                                 |
+| `update()`   | Dispatches an update for an active stream and calls `onUpdate`. It can be called multiple times.                 |
+| `end()`      | Ends a begun or active stream. Calls `onDeactivate` if active, then `onFinalize` with `canceled: false`.         |
+| `fail()`     | Fails a begun or active stream. Calls `onDeactivate` if active, then `onFinalize` with `canceled: true`.         |
+| `cancel()`   | Cancels a begun or active stream. Calls `onDeactivate` if active, then `onFinalize` with `canceled: true`.       |
+| `getState()` | Returns the controller's current state without dispatching an event.                                             |
```
j-piasecki: Do we need this when the internal states are no longer exposed to users from the hooks API?

packages/docs-gesture-handler/docs/guides/testing.mdx
```diff
@@ -62,7 +62,94 @@ In order to load mocks provided by RNGH, add the following to your jest config:
...
+- [`createGestureController`](#creategesturecontroller) to imperatively control a gesture lifecycle one step at a time.
+- [`fireGestureHandler`](#firegesturehandler) to dispatch a complete event stream.
+- [`getByGestureTestId`](#getbygesturetestid) to find a gesture by its test ID.
+
+### createGestureController
+
+```ts
+import { createGestureController } from 'react-native-gesture-handler/jest-utils';
+
+createGestureController: (componentOrGesture) => GestureController;
```
j-piasecki, line 80: Does `GestureController` work with older APIs or only V3?

packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
```diff
@@ -24,19 +24,29 @@ export function registerGesture<
...
 
 export function unregisterGesture(handlerTag: number) {
+  if (!isTestEnv()) {
+    return;
+  }
+
   const gesture = hookGestures.get(handlerTag);
 
-  if (gesture && isTestEnv() && gesture.config.testID) {
+  if (gesture?.config.testID) {
```
j-piasecki:
```suggestion
  if (gesture?.config?.testID) {
```

packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
```diff
@@ -491,6 +491,277 @@ type ExtractConfig<T> =
...
+    Object.entries(State).find(([, value]) => value === state)?.[0] ??
+    String(state)
+  );
+}
+
+function validateControllerEvent(event: Record<string, unknown>) {
+  for (const field of FORBIDDEN_CONTROLLER_EVENT_FIELDS) {
+    invariant(
+      !hasProperty(event, field),
+      `createGestureController manages '${field}' internally. Pass only gesture event payload fields.`
```
j-piasecki:
```suggestion
      `GestureController manages '${field}' internally. Pass only gesture event payload fields.`
```

packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
```diff
@@ -491,6 +491,277 @@ type ExtractConfig<T> =
...
+  if (typeof target === 'string') {
+    return getByGestureTestId(target);
+  }
+
+  if (isGesture(target)) {
+    return target;
+  }
+
+  if (isHookGesture(target)) {
+    return findGesture(target.handlerTag) ?? target;
```
j-piasecki, line 572: What's the case where `findGesture(target.handlerTag)` returns something else than `target`

packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
```diff
@@ -491,6 +491,277 @@ type ExtractConfig<T> =
...
+  }
+
+  public begin(event: GestureControllerEvent<TEventPayload> = {}) {
+    const handlerData = this.resolveHandlerData();
+
+    if (!this.isEnabled(handlerData)) {
+      return;
+    }
+
+    this.resetIfFinished();
```
j-piasecki, line 594: What if not finished? I.e. `begin` called on a gesture in `ACTIVE` state.

packages/react-native-gesture-handler/src/__tests__/gestureController.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+  panController.begin({ x: 1 });
+  panController.update({ translationX: 1, velocityX: 2 });
+  // @ts-expect-error unknown pan payload field
+  panController.update({ translationXX: 1 });
+  // @ts-expect-error raw state-machine fields are managed internally
+  panController.begin({ state: State.BEGAN });
+
+  const tapController = createGestureController(tap);
+  tapController.begin({ x: 1 });
+  // @ts-expect-error tap payloads do not include pan translation fields
```
j-piasecki, line 51: This is only a type error, right? Will runtime allow this?

packages/docs-gesture-handler/docs/guides/testing.mdx
```diff
@@ -62,7 +62,94 @@ In order to load mocks provided by RNGH, add the following to your jest config:
...
+
+- A Gesture Handler component found using a Jest query such as `getByTestId`.
+- A gesture object.
+- A gesture test ID string.
+
+When a gesture object is passed directly, the event payload type is inferred from
+the gesture. Every lifecycle method accepts an optional partial event payload and
+fills omitted handler-specific properties with defaults.
+
+The controller exposes these methods:
```
m-bert:
```suggestion
The controller exposes the following methods:
```

packages/docs-gesture-handler/docs/guides/testing.mdx
m-bert: I'd go even further. If the user imperatively manipulates gesture states by calling `begin()`, etc., then do we even need to check state? If we don't support relations then we know exactly at which state we are after calling given lifecycle method. Or have I missed something?

packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
```diff
@@ -24,19 +24,29 @@ export function registerGesture<
   handlerTag: number,
   gesture: SingleGesture<TConfig, THandlerData, TExtendedHandlerData>
 ) {
-  if (isTestEnv() && gesture.config.testID) {
-    hookGestures.set(handlerTag, gesture);
+  if (!isTestEnv()) {
+    return;
+  }
+
+  hookGestures.set(handlerTag, gesture);
```
m-bert, line 31: Do we want to remove the necessity of `testID`?

packages/docs-gesture-handler/docs/guides/testing.mdx
coado: Technically it should work with older APIs but not sure if we need to add test examples for them.

packages/react-native-gesture-handler/src/__tests__/gestureController.test.tsx
coado: Yes, almost all keywords are accepted (modulo `handlerTag`, `nativeEvent`, `oldState`, and `state`) and forwarded into the event.

packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
coado: We need to have each gesture object in the registry because they might become stale on rerenders.

packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
coado: On rerender previous target goes stale.

packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
coado: Then it throws an error.

packages/docs-gesture-handler/docs/guides/testing.mdx
coado: Yes, we can add that to the test as well.

packages/docs-gesture-handler/docs/guides/testing.mdx
coado: @j-piasecki I would say it is not necessary to have it in the API. 
@m-bert I think we should enforce only legal state changes. If we didn't do that, users would be able to test their implementation logic against impossible state flows. I see that it might be annoying to always start with, for instance, `begin()` and `activate()` if the intention is to test `onUpdate` behavior, so maybe we could also think about adding short-path methods that would internally run the necessary state transitions to get into the desired state directly?

m-bert: > I think we should enforce only legal state changes.

With that I agree, true. However, if we are talking about `getState` it doesn't enforce anything, unless user has to manually check if they're in the correct state before calling next lifecycle method, which doesn't make much sense.

But I guess that

>then do we even need to check state?

Could sound like we don't have to do this and that's not what I had in mind 😅

packages/react-native-gesture-handler/src/handlers/handlersRegistry.ts
m-bert: Okay. This might be useful if we decide to get back to key-based relations 😅

packages/docs-gesture-handler/docs/guides/testing.mdx
coado: That's why I said that is it not necessary to have that in the public API. This method was only used in internal tests which I changed already to not depend on it.

packages/docs-gesture-handler/docs/guides/testing.mdx
j-piasecki: No need I think

j-piasecki:
LGTM, please fix the CI failing, and I'll leave it to @m-bert in case he has some more comments.

packages/docs-gesture-handler/docs/guides/testing.mdx
```diff
@@ -62,7 +62,95 @@ In order to load mocks provided by RNGH, add the following to your jest config:
...
+- [`createGestureController`](#creategesturecontroller) to imperatively control a gesture lifecycle one step at a time.
+- [`fireGestureHandler`](#firegesturehandler) to dispatch a complete event stream.
+- [`getByGestureTestId`](#getbygesturetestid) to find a gesture by its test ID.
+
+### createGestureController
+
+```ts
+import { createGestureController } from 'react-native-gesture-handler/jest-utils';
+
+createGestureController: (componentOrGesture) => GestureController;
```
m-bert, line 80: What about making this block `CollapsibleCode` and adding `GestureComponent` type below like we do in other cases (e.g. [events section](https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/callbacks-events#events))?

packages/react-native-gesture-handler/src/jestUtils/jestUtils.ts
```diff
@@ -491,6 +491,272 @@ type ExtractConfig<T> =
...
+}
+
+class GestureControllerImpl<
+  TEventPayload extends Record<string, unknown> = Record<string, unknown>,
+> implements GestureController<TEventPayload>
+{
+  private state: State = State.UNDETERMINED;
+
+  // eslint-disable-next-line no-useless-constructor
+  constructor(private resolveHandlerData: () => HandlerData) {}
```
m-bert, line 585: 🫨

m-bert (commented):
One more thing before 🟢 Now we have to set `disableReanimated` every time we create gesture in tests. That may be annoying, but more important, it can break with components. The solution would be to set it on our side when we are in test environment.

packages/docs-gesture-handler/docs/guides/testing.mdx
coado: Not sure if that's what you meant https://github.com/software-mansion/react-native-gesture-handler/pull/4309/changes/72bdad4019d87e238a09fd829f87c1308e920e30. Could you please check if it is ok?

m-bert: Yes, that's it 😄

packages/react-native-gesture-handler/src/v3/hooks/utils/configUtils.ts
```diff
@@ -36,6 +36,10 @@ export function resolveInternalConfigProps<
   THandlerData,
   TExtendedHandlerData extends THandlerData,
 >(config: BaseGestureConfig<TConfig, THandlerData, TExtendedHandlerData>) {
+  if (isTestEnv() && config.disableReanimated === undefined) {
+    config.disableReanimated = true;
+  }
```
m-bert, line 41: Hmm, is the second condition necessary? Maybe we want to override it anyway? (cc @j-piasecki)

j-piasecki: I think we talked about this, but yeah - I think `disableReanimated` should always be true in test envs.

m-bert (approved):

PR #4316, opened 2026-07-15
Fix fatal crash `Cannot read property 'translationX' of undefined` when a touch event is serialized without `allTouches`

huextrat:
> [!NOTE]
> This PR was written with AI assistance (Claude), based on a production crash investigated from Sentry.

## Description

Fixes a fatal, unhandled production crash in apps using the v3 API (`usePanGesture` + `GestureDetector`) with Reanimated on Android (Fabric / New Architecture):

```
TypeError: Cannot read property 'translationX' of undefined
    at diffCalculator
    at getChangeEventCalculator
    at handleUpdateEvent
    at eventHandler
```

Since the error is thrown inside a worklet running on the UI thread, it results in a native `CppException` caught by the `UncaughtExceptionHandler` — a hard crash, not a JS redbox.

**Observed trigger:** very fast repeated taps on a screen with a `GestureDetector` + pan gesture (reproduced in production on a low-end Android 15 device, but the underlying race is not device-specific).

### Root cause

The failure is a chain across the Android event serialization and the v3 event classification:

1. **Kotlin — `GestureHandler.kt`**: `dispatchTouchEvent()` only guards on `changedTouchesPayload != null`. In `dispatchTouchUpEvent()`, `extractAllPointersData()` runs *before* the changed pointer is re-added to `trackedPointers`. If the tracked pointers were already cleared at that point (e.g. `cancelPointers()` fired by a rapid succession of taps racing with the touch-up dispatch), `allTouchesPayload` is `null` while `changedTouchesPayload` is still populated — so the event is dispatched anyway.

2. **Kotlin — `RNGestureHandlerTouchEvent.kt`**: the serializer omitted the key entirely when the payload was `null`:

   ```kotlin
   handler.consumeAllTouchesPayload()?.let { putArray("allTouches", it) }
   ```

   → a touch event can reach JS **without the `allTouches` key**. Note that iOS always serializes `allTouches`/`changedTouches` as (possibly empty) arrays (`RNGestureHandlerManager.mm` initializes `.allTouches = {}`), so this was also a platform inconsistency.

3. **TS — `src/v3/hooks/utils/eventUtils.ts`**: touch events are discriminated by key presence:

   ```ts
   export function isTouchEvent(...) { 'worklet'; return 'allTouches' in event; }
   ```

   Key absent → the touch event is **misclassified as an update event**.

4. **TS — `src/v3/hooks/callbacks/eventHandler.ts`**: the event is routed to `handleUpdateEvent()` → `getChangeEventCalculator()` reads `current.handlerData` (`undefined` for a touch event) and passes it to the gesture's `diffCalculator`, which reads `current.translationX` → `TypeError` in a UI-thread worklet → fatal crash.

### Fix

**JS (defensive, worklet-safe, no behavior change for valid events):**

- `eventHandler()` now drops events that are neither state-change events, nor touch events, nor carry `handlerData`, instead of treating them as update events. (`handlerData` is present on all well-formed update events on every platform: Android `createNativeEventData`, web `GestureHandler.ts`, and `jestUtils`.)
- `getChangeEventCalculator()` returns the event unchanged when `handlerData` is `undefined` instead of calling the diff calculator with undefined data.

**Android (root cause):**

- `RNGestureHandlerTouchEvent.createEventData()` always serializes `allTouches` and `changedTouches`, falling back to empty arrays instead of omitting the keys — matching the iOS implementation.

## Test plan

- New regression test in `src/__tests__/api_v3.test.tsx`: fires a malformed touch event (no `allTouches`, no `oldState`, no `handlerData`) through a pan gesture's `jsEventHandler`. Without the JS fix it reproduces the exact production error (`Cannot read properties of undefined (reading 'translationX')` through `diffCalculator`); with the fix the event is dropped without invoking any callback, and subsequent valid update events still compute `changeX`/`changeY` correctly.
- New unit tests for `getChangeEventCalculator` in `src/__tests__/utils.test.tsx`: change payload computed for well-formed events; event returned untouched when `handlerData` is missing.
- Full Jest suite passes (80/80), `yarn ts-check` clean, `yarn lint:js` clean.
- Android: `spotlessCheck` passes, library compiles via `apps/basic-example` (`:react-native-gesture-handler:compileDebugKotlin` — BUILD SUCCESSFUL).


m-bert (approved):
Hi @huextrat! Thank you for this PR! I took the liberty of slightly modifying it - merging `main` and removing some comments which were not necessary. Hope you don't mind!

huextrat:
> Hi @huextrat! Thank you for this PR! I took the liberty of slightly modifying it - merging `main` and removing some comments which were not necessary. Hope you don't mind!

Not at all, thanks for cleaning it up! You know the codebase best. Happy to help anytime.

PR #4327, opened 2026-07-22
Fix `minVelocity` props behavior

m-bert:
## Description

`Pan`'s velocity-based activation criteria (`minVelocity`, `minVelocityX`, `minVelocityY`) had three
long-standing problems:

1. **Android: `minVelocityY` tested the horizontal velocity.** `shouldActivate()` declared
   `val vy = velocityY` but compared `vx` in both comparisons of the Y branch, so a pan configured
   with only `minVelocityY` activated on fast horizontal movement and never on vertical movement.
   The typo predates the 2022 repo restructure (#2270).

2. **All platforms: per-axis thresholds were compared with sign.** A positive `minVelocityY` only
   activated on downward movement, negative only on upward. This is surprising — `minVelocityY: 100`
   reads as "vertical speed of at least 100", not "drag down". The signed logic appears to have been
   copied from the offset-range checks (`activeOffsetX/Y`), where signed semantics actually make
   sense. The checks now compare absolute values: `abs(velocity) >= abs(threshold)`, so movement in
   either direction along the axis activates. (`minVelocity` already compared the velocity vector
   magnitude and is unchanged.)

3. **Web: `minVelocity` meant something different than on native.** It was mapped onto the per-axis
   X/Y thresholds ("either axis exceeds the value") instead of the vector magnitude like Android and
   iOS compute it. The `minVelocitySq` field existed for this but was never assigned from the config.
   It is now wired up, matching native behavior (e.g. a diagonal drag at 600 pt/s with
   `minVelocity: 500` now activates on web as it does on native).

None of these props were documented anywhere. This PR adds them to the docs (new API + 2.x pages)
and to the JSDoc of all three API layers, described as speed "expressed in points per second".

> [!WARNING]
> I've used [`grep.app`](https://grep.app/) to check whether these props are used and seems that they're not, so it _should be safe_ to introduce these changes.

## Test plan

<details>

```tsx
// Repro for Pan minVelocityY issues:
// 1. Android compared vx (horizontal velocity) in the minVelocityY branch,
//    so a fast horizontal swipe wrongly activated the pan.
// 2. All platforms compared the velocity with its sign, so a positive
//    minVelocityY only activated on downward movement. The checks now use
//    absolute values: either direction along the axis activates.
//
// Setup: minDistance is set huge so distance can never activate the pan —
// only the velocity criteria can. Expected: swiping DOWN or UP faster than
// 500 activates; a fast horizontal swipe never does.
import {
  GestureDetector,
  GestureHandlerRootView,
  usePanGesture,

  const [log, setLog] = useState<string[]>([]);

  const append = (entry: string) =>
    setLog((prev) => [...prev.slice(-8), entry]);

  const pan = usePanGesture({
    minVelocityY: 800,
    runOnJS: true,
    onActivate: (e) => {
      append(
        `ACTIVATED vx=${Math.round(e.velocityX)} vy=${Math.round(e.velocityY)}`
      );
    },
    onFinalize: (e) => {
      if (e.canceled) {
        append('finished without activation');
      }
    },
  });

  return (
      <Text style={styles.title}>minVelocityY = 500, minDistance = 10000</Text>
      <Text style={styles.hint}>
        Fast horizontal swipe should NOT activate.{'\n'}
        Fast vertical swipe (up OR down) SHOULD activate.
      </Text>
        <View style={styles.box} testID="panBox">
          <Text style={styles.boxLabel}>swipe here</Text>
        </View>
      </GestureDetector>
      <View style={styles.log}>
        {log.map((entry, i) => (
          <Text key={i} style={styles.logLine}>
            {entry}
          </Text>
        ))}
      </View>
    </GestureHandlerRootView>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    paddingTop: 60,
  },
  title: {
    fontSize: 16,
    fontWeight: 'bold',
  },
  hint: {
    textAlign: 'center',
    marginVertical: 12,
    color: '#666',
  },
  box: {
    width: 300,
    height: 300,
    backgroundColor: 'tomato',
    borderRadius: 16,
    justifyContent: 'center',
    alignItems: 'center',
  },
  boxLabel: {
    color: 'white',
    fontSize: 18,
  },
  log: {
    marginTop: 20,
    minHeight: 200,
    alignSelf: 'stretch',
    paddingHorizontal: 24,
  },
  logLine: {
    fontFamily: 'monospace',
    fontSize: 13,
  },
});
```

</details>

### DIFF
--- a/packages/docs-gesture-handler/docs/gestures/use-pan-gesture.mdx
+++ b/packages/docs-gesture-handler/docs/gestures/use-pan-gesture.mdx
@@ -125,6 +125,30 @@ minDistance: number | SharedValue<number>;
 
 Minimum distance the finger (or multiple fingers) need to travel before the gesture activates. Expressed in points.
 
+### minVelocity
+
+```ts
+minVelocity: number | SharedValue<number>;
+```
+
+Minimum speed the pointer has to reach in order to activate the gesture. Expressed in points per second.
+
+### minVelocityX
+
+```ts
+minVelocityX: number | SharedValue<number>;
+```
+
+Minimum speed along X axis the pointer has to reach in order to activate the gesture. Expressed in points per second.
+
+### minVelocityY
+
+```ts
+minVelocityY: number | SharedValue<number>;
+```
+
+Minimum speed along Y axis the pointer has to reach in order to activate the gesture. Expressed in points per second.
+
 ### minPointers
 
 ```ts
--- a/packages/docs-gesture-handler/versioned_docs/version-2.x/gesture-handlers/pan-gh.md
+++ b/packages/docs-gesture-handler/versioned_docs/version-2.x/gesture-handlers/pan-gh.md
@@ -52,6 +52,18 @@ See [set of properties inherited from base handler class](/docs/2.x/gesture-hand
 
 Minimum distance the finger (or multiple finger) need to travel before the handler [activates](/docs/2.x/under-the-hood/state#active). Expressed in points.
 
+### `minVelocity`
+
+Minimum speed the pointer has to reach in order for the handler to [activate](/docs/2.x/under-the-hood/state#active). Expressed in points per second.
+
+### `minVelocityX`
+
+Minimum speed along X axis the pointer has to reach in order for the handler to [activate](/docs/2.x/under-the-hood/state#active). Expressed in points per second.
+
+### `minVelocityY`
+
+Minimum speed along Y axis the pointer has to reach in order for the handler to [activate](/docs/2.x/under-the-hood/state#active). Expressed in points per second.
+
 ### `minPointers`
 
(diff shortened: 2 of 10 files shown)


j-piasecki (approved):
This change looks good, but one thing I've noticed on pan gesture: https://github.com/software-mansion/react-native-gesture-handler/blob/13d0070c0df937fe7badee2d2f7188b0e34cb83c/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/PanGestureHandler.kt#L366-L370

when using shared value in config, any key will reset `minDistance`. Similartly, anything that doesn't require `hasCustomActivationCriteria` will reset it to false. Can you handle that in a follow-up?

PR #4331, opened 2026-07-23
Move Interceptor on `ScrollView`, not its content

### LINKED ISSUE #4328: [v3] Regression: FlatList content doesn't scroll when stickyHeaderIndices is set, does not affect LegacyFlatlist
### Description

I updated react-native-gesture-handler from version 2.32.0 to 3.1.0 and noticed that some Flatlists in my app stopped scrolling, both in Android and iOS. The issue seems to be related to the property stickyHeaderIndices. LegacyFlatlists do not suffer from this issue, so this appears to be either a regression in the new implementation or a migration step that isn't documented or that I may have overlooked.

Android Emulator video

https://github.com/user-attachments/assets/24fa83f7-ebf4-4a1d-a1c3-2dcdba4c56ef

### Steps to reproduce

1. Download the `flatlist-no-scroll` branch from the example repo
2. Run the app in either IOS or Android
3. Scroll both lists provided in the example
4. You will notice the `Flatlist` scroll indicator will update but not the content of the list, `LegacyFlatlist` should work as expected.

https://github.com/gitChaoslord/rn-gesture-handler-issues
Branch in question `example/flatlist-no-scroll`
https://github.com/gitChaoslord/rn-gesture-handler-issues/tree/example/flatlist-no-scroll

### Gesture Handler version

3.1.0

### React Native version

0.86.0

### Platforms

Android, iOS

### JavaScript runtime

Hermes

### Workflow

React Native (without Expo)

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

Android emulator

### Device model

_No response_

### Acknowledgements

Yes

m-bert:
## Description

#4158 introduced `ScrollViewResponderInterceptor`, which wraps all `ScrollView` children in a single logical-responder `View`. RN's `ScrollView.render()` resolves `stickyHeaderIndices` against `React.Children.toArray(this.props.children)` and wraps the child at each sticky index in `ScrollViewStickyHeader`, which counter-translates that child against the scroll offset. With the content collapsed into one child, index `0` is the entire content — so everything gets pinned. Indices > 0 were a silent no-op (they pointed past the single child).

To fix this, we now move the wrapper outside of `ScrollView` and handle keyboard dismissal manually.

Fixes #4328 

## Test plan

<details>

```tsx
import React from 'react';
import { FlatList, LegacyFlatList } from 'react-native-gesture-handler';

// Repro for https://github.com/software-mansion/react-native-gesture-handler/issues/4328
// v3 FlatList with stickyHeaderIndices: scroll indicator moves, content doesn't.
// LegacyFlatList behaves correctly.

const items = Array.from({ length: 50 }, (_, i) => ({
  id: i.toString(),
  text: `Text content for ${i}`,
}));

  const renderItem = ({ item }: { item: (typeof items)[0] }) => (
    <Text style={styles.item}>{item.text}</Text>
  );

  return (
      <View style={styles.flex1}>
        <FlatList
          data={items}
          keyExtractor={(item) => item.id}
          contentContainerStyle={styles.contentContainer}
          stickyHeaderHiddenOnScroll={false}
          ListHeaderComponentStyle={styles.headerContainer}
          stickyHeaderIndices={[0]}
          ListHeaderComponent={<Text>FlatList (v3) Header</Text>}
          renderItem={renderItem}
        />
      </View>
      <View style={styles.flex1}>
        <LegacyFlatList
          data={items}
          keyExtractor={(item) => item.id}
          contentContainerStyle={styles.contentContainer}
          stickyHeaderHiddenOnScroll={false}
          ListHeaderComponentStyle={styles.headerContainer}
          stickyHeaderIndices={[0]}
          ListHeaderComponent={<Text>LegacyFlatList Header</Text>}
          renderItem={renderItem}
        />
      </View>
    </View>
  );
}

  container: {
    flex: 1,
  },
  flex1: {
    flex: 1,
  },
  headerContainer: {
    paddingTop: 24,
    paddingBottom: 12,
    backgroundColor: '#f6f6f6',
  },
  contentContainer: {
    paddingHorizontal: 24,
  },
  item: {
    paddingVertical: 4,
  },
});
```

</details>


packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
```diff
@@ -110,44 +110,53 @@ const ScrollViewResponderInterceptor = ({
     return () => unsubscribeFromKeyboardVisibility();
   }, []);
 
+  return (
+    <JSResponderContext value={contextValue}>{children}</JSResponderContext>
+  );
```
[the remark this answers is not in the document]
m-bert: I don't think React <19 is a concern

packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
```diff
@@ -110,44 +110,58 @@ const ScrollViewResponderInterceptor = ({
...
-      <View
-        collapsable={false}
-        onStartShouldSetResponderCapture={resetRNGHResponderEvent}
-        onStartShouldSetResponder={handleStartShouldSetResponder}
-        pointerEvents="box-none"
-        style={styles.logicalResponder}>
-        {children}
-      </View>
-    </JSResponderContext>
+    <View
```
j-piasecki: You're rendering a new non-flattenable view for each ScrollView child. This will measurably impact performance.

j-piasecki (changes requested):

packages/react-native-gesture-handler/src/v3/components/ScrollViewResponderInterceptor.tsx
```diff
@@ -115,29 +143,67 @@ const ScrollViewResponderInterceptor = ({
...
+      // Unhandled tap — claim to dismiss the keyboard on release, like RN
+      // ScrollView's 'handled' claim would.
+      const shouldClaim =
+        keyboardIsDismissible() &&
+        event.target !== TextInput.State.currentlyFocusedInput();
+      claimedForKeyboardDismissal.current = shouldClaim;
+      return shouldClaim;
+    },
+    []
+  );
```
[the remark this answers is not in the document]
m-bert: My Claude says:

> The second argless call can't reach event.target — keyboardIsDismissible() is evaluated first and returns false when no input is focused (always the case in the jest env), so the && short-circuits before event is touched. That's also why this exact call pattern passes in the suite today. Same applies to onResponderRelease, which is additionally gated on a prior claim. I'd rather keep the handler signatures identical to the RN ScrollView.js handlers they mirror (which have the same argless-call behavior) than make the event optional; if a future test exercises the dismissible path, it should pass a stub event instead.

And I'm with him 😄

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -79,26 +79,34 @@ export const ScrollView = (
...
-      }>
-      <ScrollViewResponderInterceptor
-        keyboardShouldPersistTaps={keyboardShouldPersistTaps}>
+    <ScrollViewResponderProvider
+      keyboardShouldPersistTaps={keyboardShouldPersistTaps}>
+      <GHScrollView
+        {...rest}
+        ref={props.ref}
+        keyboardShouldPersistTaps={keyboardShouldPersistTaps}
+        // In 'handled' mode the provider above owns the keyboard dismissal, so
```
j-piasecki, line 88: What about `always` and `never`?

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
```diff
@@ -79,26 +79,34 @@ export const ScrollView = (
...
-          : undefined
-      }>
-      <ScrollViewResponderInterceptor
-        keyboardShouldPersistTaps={keyboardShouldPersistTaps}>
+    <ScrollViewResponderProvider
+      keyboardShouldPersistTaps={keyboardShouldPersistTaps}>
+      <GHScrollView
+        {...rest}
+        ref={props.ref}
+        keyboardShouldPersistTaps={keyboardShouldPersistTaps}
```
j-piasecki, line 87: Does this work with the scrollview's responder disabled? From what I can see in the implementation almost all references to it are in responder callbacks.

j-piasecki (commented):
The title seems incorrect after the changes

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
m-bert: In `never` RN grabs the touch during capture, before children even see it, so there's no conflict with our marks (press suppression comes from `isKeyboardDismissingTap`, #4281). Disabling the responder there would only break the dismissal.

In `always` RN does nothing keyboard-related, so there's nothing to intercept. Only `handled` has the bubble claim that conflicted with the marks, which is why both the flag and the provider's handlers are gated on it.

packages/react-native-gesture-handler/src/v3/components/GestureComponents.tsx
m-bert: Yeah, it works — the flag only kills the handled bubble claim and its blur-on-release, which is exactly what the provider reimplements. Everything else that reads `keyboardShouldPersistTaps` is either the `never`-only capture path (no-op in `handled`) or the tap-while-decelerating claim, which sits before the disable check and keeps working.

My Claude instrumented it to be sure: with the flag on, a tap during deceleration still hits `isAnimating=true`, the `ScrollView` claims it, keyboard stays open, identical to plain RN `ScrollView` next to it. `never`/`always` don't set the flag, so they're fully stock.

Only known gap: our `onResponderTerminate` stand-in for `_observedScrollSinceBecomingResponder` doesn't cover a tap landing in the tiny window right after deceleration ends - stock would skip the blur there, we wouldn't. Couldn't trigger it on purpose though.

j-piasecki (approved):

PR #4339, opened 2026-07-23
Update `AGENTS.md`

m-bert:
## Description

Sometimes agents create reproductions or example codes using old APIs. This PR updates `AGENTS.md` to enforce using v3 API by default.

## Test plan

We will see (I guess)


coado (approved):
yeah

PR #4342, opened 2026-07-24
[Android] Refactor `Touchable` not to rely on `GestureDetector`

j-piasecki:
## Description

Part 1/3 of the `Touchable` refactor (Android; iOS and web follow in stacked PRs). Previously, `Touchable` wrapped the button in a `NativeDetector` and derived `onPress`/`onPressIn`/`onPressOut`/`onLongPress` from gesture callbacks in JS, adding unnecessary overhead.

This PR moves the whole press pipeline into the button on Android:

- `RNGestureHandlerButtonViewManager` now creates, attaches, and drops a `NativeViewGestureHandler` itself, driven by a new `handlerTag` prop. The handler is attached with the new `ACTION_TYPE_NONE`, so it dispatches no gesture events to JS; its config is kept in sync via the `cancelOnLeave`, `gestureTestID`, `gestureHitSlop`, and `enabled` props (cached on the view, since prop application order isn't guaranteed).
- `NativeViewGestureHandlerHook` gains `onHandlerUpdate`/`onHandlerStateChange` hooks, which `ButtonViewGroup` uses to run a press state machine natively — including the long-press timer (`hasLongPressHandler` + `longPressDuration`).
- The resulting `press`/`pressIn`/`pressOut`/`longPress`/`interactionFinished` events are dispatched as direct events (`RNGestureHandlerButtonEvent`) carrying pointer data.
- The new `Touchable.android.tsx` renders `GestureHandlerButton` directly — no `NativeDetector`, no JS state machine — and only forwards the native events, keeping the keyboard-dismissing-tap suppression in JS.

## Test plan

Existing tests in the Example app

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
@@ -436,6 +436,7 @@ class RNGestureHandlerButtonViewManager :
     private var underlayDrawable: PaintDrawable? = null
     private var pressInTimestamp = 0L
     private var pendingPressOut: Runnable? = null
+    private var pendingLongPress: Runnable? = null
     private var pendingHoverOut: Choreographer.FrameCallback? = null
     private var isPointerInsideBounds = false
     private var isHovered = false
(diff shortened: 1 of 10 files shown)


packages/react-native-gesture-handler/src/v3/components/Touchable/Touchable.android.tsx
```diff
@@ -0,0 +1,200 @@
...
+} from '../../../components/GestureHandlerButton';
+import { getTVProps } from '../../../components/utils';
+import { getNextHandlerTag } from '../../../handlers/getNextHandlerTag';
+import {
+  isKeyboardDismissingTap,
+  JSResponderContext,
+} from '../ScrollViewResponderInterceptor';
+import type { AnimationDuration, TouchableProps } from './TouchableProps';
+
+const isAndroid = Platform.OS === 'android';
```
m-bert, line 16: Is there a case when this is not `true`, if we are inside `Touchable.android.tsx`?

packages/react-native-gesture-handler/src/v3/components/Touchable/Touchable.android.tsx
```diff
@@ -0,0 +1,200 @@
...
+    onPress,
+    onPressIn,
+    onPressOut,
+    children,
+    disabled = false,
+    cancelOnLeave = true,
+    ref,
+    ...rest
+  } = props;
+  const [handlerTag] = useState(() => getNextHandlerTag());
```
m-bert, line 87: Maybe we should also do the same in [useGesture](https://github.com/software-mansion/react-native-gesture-handler/blob/6b0ca7bd7a9fba740bd10e834e40f492cdb24c88/packages/react-native-gesture-handler/src/v3/hooks/useGesture.ts#L32), wdyt?

m-bert (commented):
Great job!

packages/react-native-gesture-handler/src/v3/components/Touchable/Touchable.android.tsx
m-bert: Ok, I can see this file is changed in the follow-up 😅

m-bert (approved):

packages/react-native-gesture-handler/src/v3/components/Touchable/Touchable.android.tsx
j-piasecki: Yes, we should

packages/react-native-gesture-handler/src/v3/components/Touchable/TouchableProps.ts
```diff
@@ -78,17 +90,22 @@ export type TouchableProps = Omit<
     /**
      * Called when the component gets pressed.
      */
-    onPress?: ((event: CallbackEventType) => void) | undefined;
+    onPress?: ((event: ButtonEvent) => void) | undefined;
+
+    /**
+     * Called when the component gets long pressed.
+     */
+    onLongPress?: ((event: ButtonEvent) => void) | undefined;
```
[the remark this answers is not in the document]
j-piasecki: Added a tsignore for now so the typecheck passes, all three prs should land before any actual releases.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -63,6 +68,90 @@ class RNGestureHandlerButtonViewManager :
...
+    view.managedHandlerHitSlop = gestureHitSlop
+    if (gestureHitSlop != null) {
+      updateManagedHandlerConfig(
+        view,
+        Arguments.createMap().apply {
+          putMap("hitSlop", gestureHitSlop)
+        },
+      )
+    }
+  }
```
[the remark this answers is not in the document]
j-piasecki: Changed hitslop declaration in js to
```
  const hitSlop =
    typeof props.hitSlop === 'number' || props.hitSlop == null
      ? {
          top: props.hitSlop ?? 0,
          left: props.hitSlop ?? 0,
          bottom: props.hitSlop ?? 0,
          right: props.hitSlop ?? 0,
        }
      : props.hitSlop;
```

so undefined hitslop gets reset to 0.

PR #4344, opened 2026-07-24
[Web] Refactor `Touchable` not to rely on `GestureDetector`

j-piasecki:
## Description

- `ActionType.NONE` is added on the JS side; the web `GestureHandler` skips `sendEvent` dispatch for handlers attached with it.
- `GestureHandlerButton.web.tsx` now creates and attaches its own `NativeViewGestureHandler` when `handlerTag` is set (mirroring the native platforms) and drops it on unmount.
- The web `NativeViewGestureHandler` runs the same press state machine (including the long-press timer) when it manages a button, dispatching press events as DOM `CustomEvent`s (`gh:buttonPress` etc.). The button listens for these and forwards them to the `onButton*` callbacks; managed buttons no longer drive pressed state from their own pointer handlers.
- `Touchable.web.tsx` is deleted — web now uses the shared `Touchable` implementation.
- Also fixes the web `GestureHandler` registering event listeners for initially disabled handlers (config is applied before attachment, so the enabled-change path never ran) — covered by a new unit test.

## Test plan

Existing tests in the Example app


packages/react-native-gesture-handler/src/web/handlers/__tests__/GestureHandler.test.ts
m-bert, line 1: I've mixed feelings about putting this test file here 🤔

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -187,6 +276,103 @@ export const ButtonComponent = ({
...
+    };
+    const handleInteractionFinished = (event: Event) => {
+      onButtonInteractionFinished?.(wrapEvent(event));
+    };
+
+    node?.addEventListener(ButtonEventName.Press, handlePress);
+    node?.addEventListener(ButtonEventName.PressIn, handlePressIn);
+    node?.addEventListener(ButtonEventName.PressOut, handlePressOut);
+    node?.addEventListener(ButtonEventName.LongPress, handleLongPress);
+    node?.addEventListener(
```
m-bert: These are added for each button, including old buttons and Pressable, right? If so, maybe we could gaurd it at the beginning of the hook so that they won't be attached when not needed? 

```tsx
useIsomorphicLayoutEffect(() => {
  const node = viewRef.current;
  if (handlerTag === undefined || node === null) {
    return;
  }
  // ...
}, [
  handlerTag,
  onButtonInteractionFinished,
  // ...
]);
```

I know they won't be fired anyway, so I'll leave that up to you

m-bert (approved):

packages/react-native-gesture-handler/src/web/handlers/__tests__/GestureHandler.test.ts
j-piasecki: Moved to `__tests__` directory at the same level in [Move test file](https://github.com/software-mansion/react-native-gesture-handler/pull/4344/commits/05e3e2d0cfe1d0f559ee284838ef9519d2922b50). Is that ok?

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
j-piasecki: Done in [Don't attach press event listeners on legacy buttons](https://github.com/software-mansion/react-native-gesture-handler/pull/4344/commits/7348115d1c1001e4af332368a42ae01b582af086)

packages/react-native-gesture-handler/src/components/GestureHandlerButton.web.tsx
```diff
@@ -69,6 +116,37 @@ export const ButtonComponent = ({
...
+    enabled,
+    shouldCancelWhenOutside: cancelOnLeave,
+    shouldActivateOnStart: false,
+    disallowInterruption: true,
+    yieldsToContinuousGestures: true,
+    testID: gestureTestID,
+    hitSlop: gestureHitSlop,
+    hasLongPressHandler,
+    longPressDuration,
+  });
```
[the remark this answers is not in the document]
j-piasecki: That's known & pre-existing

packages/react-native-gesture-handler/src/web/handlers/__tests__/GestureHandler.test.ts
m-bert: It is, thanks!

PR #4367, opened 2026-07-30
[Android] Properly handle `requestDisallowInterceptTouchEvent` for v3

j-piasecki:
## Description

On Android, when a native view calls `requestDisallowInterceptTouchEvent` (e.g. a pager starting a swipe), `RNGestureHandlerRootView` reacted by cancelling **all** handlers registered on the root. `react-native-pager-view` calls it eagerly on touch down, so any v3 gesture rendered inside a pager (e.g. inside material top tabs) was cancelled before it could activate — a long press nested in top tabs never activated at all.

This PR makes the cancellation targeted:

- The root view now cancels only legacy handlers (v1/v2 action types) via the new `GestureHandlerOrchestrator.cancelAllLegacyHandlers`, instead of the old trick of activating the internal `RootViewGestureHandler`. The root handler is now attached with `ACTION_TYPE_NONE` so it's excluded from that sweep (and no longer sends dead events to JS).
- v3 handlers are cancelled by new `requestDisallowInterceptTouchEvent` overrides on `RNGestureHandlerDetectorView` and `ButtonViewGroup`. Since the request only bubbles upward from the requesting view, only handlers attached to its ancestors are cancelled — handlers below the requester (like the long press under the pager) keep working.
- Cancellation is skipped while the orchestrator is delivering events (`isHandlingTouch`), mirroring the existing `passingTouch` guard, so disallow requests caused by RNGH's own event delivery don't cancel gestures.
- `findGestureHandlerRootView` now returns the nearest *enabled* root view so the checks above consult the orchestrator that actually manages the subtree.

## Test plan

Tested on reproducer from https://github.com/software-mansion/react-native-gesture-handler/issues/2383

https://github.com/user-attachments/assets/4d01dff7-5268-42ff-befa-5c971d39bb4a

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
@@ -46,6 +46,28 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
     super.onDetachedFromWindow()
   }
 
+  override fun requestDisallowInterceptTouchEvent(disallowIntercept: Boolean) {
+    super.requestDisallowInterceptTouchEvent(disallowIntercept)
+    if (!disallowIntercept) {
+      return
+    }
+
+    val orchestrator = RNGestureHandlerRootView.findGestureHandlerRootView(this)?.orchestrator ?: return
+
+    if (!orchestrator.isHandlingTouch) {
+      val currentHandlers = attachedHandlers.mapNotNull { tag ->
+        RNGestureHandlerModule.registries[this.moduleId]?.getHandler(tag)
+      }.filter {
+        // The handler has been recorded
+        it.view != null
+      }
+
+      currentHandlers.forEach {
+        it.cancel()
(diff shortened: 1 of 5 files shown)


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerDetectorView.kt
```diff
@@ -46,6 +46,28 @@ class RNGestureHandlerDetectorView(context: Context) : ReactViewGroup(context) {
...
+      val currentHandlers = attachedHandlers.mapNotNull { tag ->
+        RNGestureHandlerModule.registries[this.moduleId]?.getHandler(tag)
+      }.filter {
+        // The handler has been recorded
+        it.view != null
+      }
+
+      currentHandlers.forEach {
+        it.cancel()
+      }
```
m-bert: Can't we do it simpler?

```suggestion
val registry = RNGestureHandlerModule.registries[moduleId] ?: return
for (tag in attachedHandlers) {
  registry.getHandler(tag)?.takeIf { it.view != null }?.cancel()
}
```

m-bert (approved):

PR #4368, opened 2026-07-30
[Docs] Add `Touchable` section to landing page

m-bert:
## Description

Adds `Touchable` section to docs

https://github.com/user-attachments/assets/ff5cb001-33b4-4f98-adec-002578a66f37

## Test plan

Read docs 🤓

### DIFF
--- a/packages/docs-gesture-handler/src/components/TouchableShowcase/index.tsx
+++ b/packages/docs-gesture-handler/src/components/TouchableShowcase/index.tsx
@@ -1,0 +1,83 @@
+import BrowserOnly from '@docusaurus/BrowserOnly';
+import clsx from 'clsx';
+import React from 'react';
+import { StyleSheet, Text } from 'react-native';
+import {
+  GestureHandlerRootView,
+  Touchable,
+} from 'react-native-gesture-handler';
+
+import styles from './styles.module.css';
+
+function TouchableDemo() {
+  return (
+    <div className={styles.card}>
+      <GestureHandlerRootView style={rnStyles.root}>
+        <Touchable
+          style={rnStyles.button}
+          activeOpacity={0.8}
+          underlayColor="var(--swm-purple-dark-140)"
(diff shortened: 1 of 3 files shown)


packages/docs-gesture-handler/src/components/TouchableShowcase/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <p className={styles.description}>
+          One fully customizable component meant to supersede Pressable and
+          Buttons. It comes with native scale, opacity, underlay, and Android
+          ripple feedback built in.
+        </p>
+        <a
+          className={clsx(styles.link, styles.stat)}
+          href="https://x.com/swmansion/status/2082488629167841415"
+          target="_blank"
+          rel="noreferrer">
```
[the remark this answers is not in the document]
m-bert: It includes `noopener` behavior ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noreferrer))

packages/docs-gesture-handler/src/components/TouchableShowcase/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <span aria-hidden="true">⚡</span> ~300ms faster than React Native
```
j-piasecki: This is pointless without context - a percentage would be much better, but also not that impressive.

j-piasecki (approved):

packages/docs-gesture-handler/src/components/TouchableShowcase/index.tsx
m-bert: Removed in [69d2f75](https://github.com/software-mansion/react-native-gesture-handler/pull/4368/commits/69d2f75fd33f71f505c10d49e3fa0842bd9555a1)

PR #4378, opened 2026-08-04
[CI] Upload application artifacts

j-piasecki:
## Description

Updates the CI:
- build-android, build-ios are now reusable workflows that upload the app binary once finished
  - the apps are now built in release variant
- adds android, ios workflows that build expo example (tests can be plugged here)
- adds android-basic, ios-basic workflows that build basic example (they don't depend on the library JS code)

## Test plan

CI run

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,38 @@
+name: Test Android
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/android.yml
+      - .github/workflows/android-build.yml
+      - packages/react-native-gesture-handler/android/**
+      - packages/react-native-gesture-handler/shared/**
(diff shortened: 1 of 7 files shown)


m-bert (approved):

.github/workflows/android.yml
```diff
@@ -0,0 +1,33 @@
+name: Test Android
```
m-bert: Shouldn't it be 

```yml
name: Build Android (expo-example)
```

.github/workflows/ios.yml
```diff
@@ -0,0 +1,35 @@
+name: Test iOS
```
m-bert: Same as above

m-bert (approved):

PR #4385, opened 2026-08-05
Chore: ref forwarding to child in wrap v2

tshmieldev:
This will help with strictmode adoption, as using refs can help avoid the erroring path of findHostInstance / getNodeHandle

This PR aims to be non-breaking

### DIFF
--- a/packages/react-native-gesture-handler/src/hostInstance.ts
+++ b/packages/react-native-gesture-handler/src/hostInstance.ts
@@ -1,0 +1,37 @@
+export function isHostInstance(instance: unknown) {
+  return (
+    (instance as { __internalInstanceHandle?: unknown } | null | undefined)
+      ?.__internalInstanceHandle !== undefined
+  );
+}
+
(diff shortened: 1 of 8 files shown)


packages/react-native-gesture-handler/src/hostInstance.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (instance === null || instance === undefined || isHostInstance(instance)) {
+    return instance;
+  }
+
+  const nativeRef = (
+    instance as { getNativeScrollRef?: () => unknown }
+  ).getNativeScrollRef?.();
+
+  return isHostInstance(nativeRef) ? nativeRef : instance;
+}
```
j-piasecki, line 38: These have already been added in https://github.com/software-mansion/react-native-gesture-handler/pull/4369

tshmieldev: addressed, thanks

j-piasecki (approved):
Thanks!

Issue #4066, opened 2026-04-08
ReferenceError: Cannot access 'tagMessage' before initialization on web in dev mode (v3 beta)


tobisamuel:
### Description

Any use of the v3 hook API (`useTapGesture`, etc.) on **web in dev mode** (`expo start --web`) crashes immediately:

```
ReferenceError: Cannot access 'tagMessage' before initialization
  at factory (node_modules/react-native-gesture-handler/lib/commonjs/utils.js:11:22)
  at loadModuleImplementation (.../@expo/cli/build/metro-require/require.js:248:36)
  at factory (node_modules/react-native-gesture-handler/lib/commonjs/handlers/handlersRegistry.js:17:14)
  at factory (node_modules/react-native-gesture-handler/lib/commonjs/handlers/gestures/eventReceiver.js:12:25)
```

The crash does **not** happen with `expo export --platform web` (production build) — only in the dev server. It also does not happen on native platforms.

A couple of things that seem relevant:

- `tagMessage` in `src/utils.ts` has a `'worklet'` directive, new in v3. The pre-compiled CommonJS output at `lib/commonjs/utils.js` has `exports.tagMessage = tagMessage;` at the top of the module (before the function declaration), and `function tagMessage()` with `'worklet'` lower down.
- On web, Metro resolves `main` → `lib/commonjs/index.js` (pre-compiled CJS). The worklets Babel plugin processes the `'worklet'` directive and appears to convert the function declaration into a `const` declaration, which eliminates hoisting and causes the TDZ error on the earlier `exports.tagMessage = tagMessage` assignment.

### Steps to reproduce

1. Create a new Expo project
2. Install `react-native-gesture-handler@3.0.0-beta.2`
3. Use any v3 gesture hook in a component (e.g. `useTapGesture` + `GestureDetector`)
4. Run `npx expo start --web`
5. Observe the crash

### A link to a [Gist](https://gist.github.com/), an [Expo Snack](https://snack.expo.io/) or a link to a repository based on [this template](https://github.com/react-native-community/reproducer-react-native) that reproduces the bug.

https://github.com/tobisamuel/web-repro

### Gesture Handler version

3.0.0-beta.2

### React Native version

0.83.4

### Platforms

Web

### JavaScript runtime

Hermes

### Workflow

Using Expo Go

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

None

### Acknowledgements

Yes

m-bert:
[...]

PR #4386, opened 2026-08-05
Add an `exports` map and re-enable the CommonJS build behind the `require` condition

m-bert:
## Description

This PR adds an `exports` map to the package and brings back the `commonjs` build target (removed in #4069), exposed only through the `require` export condition.

The key order in the map is the mechanism: a resolver picks the first key present in its condition set. Jest presets (`@react-native/jest-preset`, `jest-expo`) resolve with `['require', 'react-native']`, so test suites now load ready-made CommonJS with zero config — no more `transformIgnorePatterns` tweaking. Metro resolves native platforms with `['react-native']` only and web with `['browser']`, so app bundles keep getting the `src` chain (native) or the ESM build (web) and structurally cannot reach the CommonJS artifact. This narrows rather than reverses #4069: the worklet breakage from #4066 required CommonJS inside a Metro bundle, which the condition gate makes unresolvable.

A `"./*": "./*"` catch-all keeps every historical deep path working, since the package never had an `exports` field and all paths were de-facto public.

Defense in depth for the one remaining edge (a consumer Metro config manually adding `"require"` to `unstable_conditionNames`): the CommonJS entry starts with a guard that emits a `console.warn` with an issue link when that build is evaluated inside a React Native app. The build flavor is marked at build time (`scripts/mark-commonjs-build.js`), because Metro wraps all module formats in CommonJS-style wrappers, making runtime detection impossible.

`jestSetup.js` gains `lib/commonjs` twins of the existing mocks, plus mocks for the three codegen component specs — those ship as raw TypeScript (codegen parses them from source), which the untransformed CommonJS chain cannot require.

## Test plan

- Repo gates: `yarn test` (110), `yarn ts-check`, `yarn lint-js`, madge — all green.
- Packed the real tarball and verified consumers:
  - jest with stock `@react-native/jest-preset` (RN 0.86) and stock `jest-expo` (SDK 57): package resolves to `lib/commonjs`, the full barrel loads, jestSetup mocks apply, `jest-utils` works — zero config, which fails on current `main`.
  - Metro bundle (bare RN 0.86) and `expo export` (SDK 55, SDK 57): sourcemap audit shows all bundled modules come from `src/`, zero from either `lib` flavor.
  - `expo export --platform web`, prod and dev mode (the #4066 scenario): all modules resolve to `lib/module` ESM, zero CommonJS.
  - Node: subpath resolution (root, `jest-utils`, `jestSetup`, `package.json`, deep paths), guard silent under Node and jest, warns when `navigator.product` is spoofed to `ReactNative`.
- On-device (iOS simulator, example app): v3 pan gesture with `onUpdate` writing shared values — runs as a worklet on the UI thread (`_WORKLET === true`), box follows the drag.

### DIFF
--- a/packages/react-native-gesture-handler/jestSetup.js
+++ b/packages/react-native-gesture-handler/jestSetup.js
@@ -14,6 +14,15 @@ jest.mock('./src/components/touchables', () =>
 jest.mock('./src/v3/detectors/HostGestureDetector', () =>
   require('./src/mocks/hostDetector')
 );
+jest.mock('./src/specs/RNGestureHandlerButtonNativeComponent', () =>
+  require('./src/mocks/nativeComponentSpec')
+);
+jest.mock('./src/specs/RNGestureHandlerDetectorNativeComponent', () =>
+  require('./src/mocks/nativeComponentSpec')
+);
+jest.mock('./src/specs/RNGestureHandlerRootViewNativeComponent', () =>
+  require('./src/mocks/nativeComponentSpec')
+);
 
 jest.mock('./lib/module/RNGestureHandlerModule', () =>
   require('./lib/module/mocks/module')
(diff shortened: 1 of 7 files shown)


m-bert:
Closing this as changes in `jest` config are acceptable

PR #4388, opened 2026-08-05
Move hitSlop normalization to JS

coado:
## Description

This moves the parsing and validation of `hitSlop` into one shared JS module and reduces the platforms to reading a fixed shape.

JS now sends a normalized [left, top, right, bottom, width, height] array, where null marks an unspecified edge. That layout matches what the platforms already stored internally (FloatArray(6) on Android, RNGHHitSlop on Apple), so the native parsers shrink to mapping null → NaN, plus the DIP→px conversion on Android.

Applied at all four producers: filterConfig (v1 and v2), prepareConfigForNativeSide (v3), bindSharedValues (v3's UI-thread path) and the button's web wrapper.

## Test plan

Tested on Android, iOS and Web

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
@@ -11,7 +11,6 @@ import android.view.View
 import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.bridge.ReadableMap
-import com.facebook.react.bridge.ReadableType
 import com.facebook.react.bridge.UiThreadUtil
 import com.facebook.react.bridge.WritableArray
 import com.facebook.react.uimanager.PixelUtil
@@ -183,18 +182,6 @@ open class GestureHandler {
     hitSlop!![HIT_SLOP_BOTTOM_IDX] = bottomPad
     hitSlop!![HIT_SLOP_WIDTH_IDX] = width
     hitSlop!![HIT_SLOP_HEIGHT_IDX] = height
-    require(!(hitSlopSet(width) && hitSlopSet(leftPad) && hitSlopSet(rightPad))) {
-      "Cannot have all of left, right and width defined"
-    }
-    require(!(hitSlopSet(width) && !hitSlopSet(leftPad) && !hitSlopSet(rightPad))) {
-      "When width is set one of left or right pads need to be defined"
-    }
-    require(!(hitSlopSet(height) && hitSlopSet(bottomPad) && hitSlopSet(topPad))) {
-      "Cannot have all of top, bottom and height defined"
-    }
-    require(!(hitSlopSet(height) && !hitSlopSet(bottomPad) && !hitSlopSet(topPad))) {
-      "When height is set one of top or bottom pads need to be defined"
-    }
   }
 
   fun setHitSlop(padding: Float?) {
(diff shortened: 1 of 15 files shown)


packages/react-native-gesture-handler/src/utils.ts
```diff
@@ -43,6 +43,28 @@ export function tagMessage(msg: string) {
...
+  'worklet';
+
+  const runtimeKind: number | undefined = globalThis.__RUNTIME_KIND;
+
+  if (runtimeKind !== undefined) {
+    return runtimeKind !== RUNTIME_KIND_REACT_NATIVE;
+  }
+
+  return globalThis._WORKLET === true;
+}
```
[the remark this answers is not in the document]
coado: There is a dev dependency on `react-native-reanimated`

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -983,68 +970,37 @@ open class GestureHandler {
...
-            GestureHandler.HIT_SLOP_NONE,
-          )
+          return
+        }
+
+        val hitSlop = config.getArray(KEY_HIT_SLOP)!!
+
+        fun edge(index: Int) = if (hitSlop.isNull(index)) {
+          GestureHandler.HIT_SLOP_NONE
         } else {
```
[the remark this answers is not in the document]
coado: If we have type safety in TS for payload that is send to native, then we also have type safety in native. For reference, we do the same for relations:

```
val array = config.getArray(key)!!
return IntArray(array.size()).also { ... }
```

packages/react-native-gesture-handler/src/__tests__/hitSlop.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    beforeEach(() => {
+      consoleError = jest.spyOn(console, 'error').mockImplementation();
+    });
+
+    afterEach(() => {
+      globalThis.__RUNTIME_KIND = initialRuntimeKind;
+      globalThis._WORKLET = undefined;
+      consoleError.mockRestore();
+    });
```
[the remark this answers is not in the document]
coado: This is the same as above. We have dev dependency on reanimated

packages/react-native-gesture-handler/src/handlers/hitSlop.ts
```diff
@@ -0,0 +1,146 @@
...
+): CanonicalHitSlop | null | undefined {
+  'worklet';
+
+  if (hitSlop === undefined || hitSlop === null) {
+    return hitSlop;
+  }
+
+  if (Array.isArray(hitSlop)) {
+    return hitSlop;
+  }
```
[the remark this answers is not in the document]
coado: The only accepted array by types is the one from `CanonicalHitSlop` which is exactly what the method returns so there is no need for runtime check.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt
```diff
@@ -983,68 +970,37 @@ open class GestureHandler {
...
         }
+
+        handler.setHitSlop(
+          edge(HIT_SLOP_LEFT_IDX),
+          edge(HIT_SLOP_TOP_IDX),
+          edge(HIT_SLOP_RIGHT_IDX),
+          edge(HIT_SLOP_BOTTOM_IDX),
+          edge(HIT_SLOP_WIDTH_IDX),
+          edge(HIT_SLOP_HEIGHT_IDX),
+        )
```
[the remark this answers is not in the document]
coado: The method returns and accepts `CanonicalHitSlop` array only so I don't think we have to introduce additional runtime checks

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -33,7 +33,28 @@ - (RNGestureHandler *)gestureHandler
...
+  RNGHHitSlopIndexWidth,
+  RNGHHitSlopIndexHeight,
+  RNGHHitSlopIndexCount,
+};
+
+static CGFloat RNGHHitSlopEdge(NSArray *hitSlop, RNGHHitSlopIndex index)
+{
+  if (index >= hitSlop.count) {
+    return NAN;
+  }
```
m-bert: Is this necessary?

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -166,32 +187,18 @@ - (void)updateConfig:(NSDictionary *)config
     _cancelsJSResponder = [RCTConvert BOOL:prop];
   }
 
+  // The `width`/`height` combinations are validated on the JS side, before the value gets here.
```
m-bert:
```suggestion
```

packages/react-native-gesture-handler/src/handlers/hitSlop.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  // On the UI runtime this runs inside a shared value listener, where throwing
+  // would tear down the listener and leave the shared value and the native
+  // config out of sync. Report instead, and let the invalid value take the same
+  // path it would take in a release build — validation is purely diagnostic and
+  // must not change what the pipeline does.
+  if (isWorkletRuntime()) {
+    console.error(tagMessage(error));
+    return;
+  }
```
m-bert: I don't think it actually makes sense, wouldn't simple `throw new Error` be enough?

packages/react-native-gesture-handler/src/handlers/hitSlop.ts
m-bert:
```suggestion
  if (hitSlop === undefined || hitSlop === null || Array.isArray(hitSlop)) {
    return hitSlop;
  }
```

packages/react-native-gesture-handler/src/v3/hooks/utils/configUtils.ts
```diff
@@ -108,8 +107,14 @@ export function prepareConfigForNativeSide<
   for (const [key, value] of Object.entries(config)) {
     // @ts-ignore That's the point, we want to see if key exists in the whitelists
     if (allowedNativeProps.has(key) || handlerPropsWhiteList.has(key)) {
+      const unpackedValue = Reanimated?.isSharedValue(value)
+        ? value.value
+        : value;
```
m-bert: We have [`maybeUnpackValue`](https://github.com/software-mansion/react-native-gesture-handler/blob/2b37b3cf478ca373049ed47f09c14262f82a7835/packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts#L115). I know it was there so if it is because of circular dependency then it's fine, otherwise we can swap that.

packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
```diff
@@ -48,14 +50,17 @@ export function bindSharedValues<
...
+        updateGestureHandlerConfig(handlerTag, {
+          dispatchesReanimatedEvents: shouldUseReanimatedDetector && !value,
+        });
+        return;
+      }
+
+      updateGestureHandlerConfig(handlerTag, {
+        [configKey]:
+          configKey === 'hitSlop' ? normalizeHitSlop(value as HitSlop) : value,
+      });
```
m-bert: Maybe `switch(configKey)` or another `if`? That way all options would be handled in the same manner. But that's not strong opinion and I'm also fine with what it is now.

packages/react-native-gesture-handler/src/utils.ts
```diff
@@ -35,18 +35,40 @@ export function hasProperty(object: object, key: string) {
...
+
+/**
+ * Whether the caller is executing on a Worklet runtime (the UI runtime or a
+ * worker) rather than on the React Native one.
+ *
+ * Worklets deprecated `_WORKLET` in favour of `__RUNTIME_KIND`, but Reanimated
+ * is an optional dependency here and older versions only expose the former, so
+ * both are consulted.
+ */
+export function isWorkletRuntime(): boolean {
```
m-bert: Shouldn't this be in [`reanimatedUtils`](https://github.com/software-mansion/react-native-gesture-handler/blob/2b37b3cf478ca373049ed47f09c14262f82a7835/packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts)?

packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
```diff
@@ -48,14 +50,19 @@ export function bindSharedValues<
...
+        });
+        return;
+      }
+
+      if (configKey === 'hitSlop') {
+        updateGestureHandlerConfig(handlerTag, {
+          hitSlop: normalizeHitSlop(value as HitSlop),
+        });
+        return;
+      }
```
m-bert, line 65: Doesn't it lack 

```ts
updateGestureHandlerConfig(handlerTag, { [configKey]: value })
```

at the end?

m-bert (changes requested):

packages/react-native-gesture-handler/src/v3/hooks/utils/reanimatedUtils.ts
coado: yes, good catch

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -167,31 +184,16 @@ - (void)updateConfig:(NSDictionary *)config
   }
 
   prop = config[@"hitSlop"];
-  if ([prop isKindOfClass:[NSNumber class]]) {
-    _hitSlop.left = _hitSlop.right = _hitSlop.top = _hitSlop.bottom = [prop doubleValue];
```
j-piasecki, line 171: Can we leave the single-number path in place on both platforms? Handling this is a matter of building a rect using the number, and passing it is much more efficient than the wrappers RN is using for objects (at least on Android).

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
```diff
@@ -167,31 +184,16 @@ - (void)updateConfig:(NSDictionary *)config
...
-    if (!isnan(_hitSlop.width) && !isnan(_hitSlop.left) && !isnan(_hitSlop.right)) {
-      RCTLogError(@"Cannot have all of left, right and width defined");
-    }
-    if (isnan(_hitSlop.top) && isnan(_hitSlop.bottom) && !isnan(_hitSlop.height)) {
-      RCTLogError(@"When height is set one of top or bottom pads need to be defined");
-    }
-    if (!isnan(_hitSlop.height) && !isnan(_hitSlop.top) && !isnan(_hitSlop.bottom)) {
-      RCTLogError(@"Cannot have all of top, bottom and height defined");
-    }
+    // An explicit `null` clears the hit slop; a missing key leaves the previous value alone.
```
j-piasecki: Won't explicit `null` be translated to `nil`?

coado: yes, that is true

packages/react-native-gesture-handler/apple/RNGestureHandler.mm
coado: sure thing

m-bert:
@coderabbitai can this be merged?

m-bert (approved):

m-bert (approved):

PR #4399, opened 2026-08-06
[Docs] Document Touchable hover callbacks

j-piasecki:
## Description

Documents `onHoverIn`/`onHoverOut` on the `Touchable` page, together
with the behaviour that isn't obvious from the signature: how the two
pair up, what `disabled` does to an open hover, which platforms report
the crossings during a press, tvOS driving hover from focus, and hover
state not surviving an unmount or a detach.

## Test plan

`yarn start` in `packages/docs-gesture-handler`.

### DIFF
--- a/packages/docs-gesture-handler/docs/components/touchable.mdx
+++ b/packages/docs-gesture-handler/docs/components/touchable.mdx
@@ -373,13 +373,16 @@ label="Show composed types definitions"
 expandedLabel="Hide composed types definitions"
 lineBounds={[0, 1]}
 src={`
-onPressIn?: (e: GestureEvent<NativeHandlerData>) => void;
+onPressIn?: (e: ButtonEvent) => void;
 
-type GestureEvent<NativeHandlerData> = {
-    handlerTag: number;
+type ButtonEvent = {
+    pointerInside: boolean;
+    x: number;
+    y: number;
+    absoluteX: number;
+    absoluteY: number;
     numberOfPointers: number;
     pointerType: PointerType;
-    pointerInside: boolean;
 }
 
 enum PointerType {
@@ -400,13 +403,16 @@ label="Show composed types definitions"
 expandedLabel="Hide composed types definitions"
 lineBounds={[0, 1]}
 src={`
-onPressOut?: (e: GestureEvent<NativeHandlerData>) => void;
+onPressOut?: (e: ButtonEvent) => void;
 
-type GestureEvent<NativeHandlerData> = {
-    handlerTag: number;
+type ButtonEvent = {
+    pointerInside: boolean;
+    x: number;
+    y: number;
+    absoluteX: number;
+    absoluteY: number;
     numberOfPointers: number;
     pointerType: PointerType;
-    pointerInside: boolean;
 }
 
 enum PointerType {
@@ -427,13 +433,16 @@ label="Show composed types definitions"
 expandedLabel="Hide composed types definitions"
 lineBounds={[0, 1]}
 src={`
-onPress?: (e: GestureEvent<NativeHandlerData>) => void;
+onPress?: (e: ButtonEvent) => void;
 
-type GestureEvent<NativeHandlerData> = {
-    handlerTag: number;
+type ButtonEvent = {
+    pointerInside: boolean;
+    x: number;
+    y: number;
+    absoluteX: number;
+    absoluteY: number;
     numberOfPointers: number;
     pointerType: PointerType;
-    pointerInside: boolean;
 }
 
 enum PointerType {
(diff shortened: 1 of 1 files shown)


packages/docs-gesture-handler/docs/components/touchable.mdx
```diff
@@ -373,13 +373,16 @@ label="Show composed types definitions"
...
+type ButtonEvent = {
+    pointerInside: boolean;
+    x: number;
+    y: number;
+    absoluteX: number;
+    absoluteY: number;
     numberOfPointers: number;
     pointerType: PointerType;
-    pointerInside: boolean;
 }
```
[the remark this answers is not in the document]
j-piasecki: that would just obfuscate the docs imo

packages/docs-gesture-handler/docs/components/touchable.mdx
```diff
@@ -449,12 +458,102 @@ Triggered when the button gets pressed (analogous to `onPress` in `Pressable` fr
...
+enum PointerType {
+    TOUCH,
+    STYLUS,
+    MOUSE,
+    KEY,
+    OTHER,
+}
+`}/>
+
+Triggered when a non-touch pointer — a mouse, a trackpad cursor, or a hovering stylus — moves over the button (analogous to `onHoverIn` in `Pressable` from RN core). Touch pointers never report hover.
```
m-bert:
```suggestion
Triggered when a non-touch pointer - a mouse, a trackpad cursor, or a hovering stylus - moves over the button (analogous to `onHoverIn` in `Pressable` from RN core). Touch pointers never report hover.
```

packages/docs-gesture-handler/docs/components/touchable.mdx
```diff
@@ -449,12 +458,102 @@ Triggered when the button gets pressed (analogous to `onPress` in `Pressable` fr
...
+    STYLUS,
+    MOUSE,
+    KEY,
+    OTHER,
+}
+`}/>
+
+Triggered when the hovering pointer leaves the button (analogous to `onHoverOut` in `Pressable` from RN core).
+
+The two callbacks are paired — an `onHoverIn` is followed by exactly one `onHoverOut` once the pointer leaves. Because hover is only reported while the button is not `disabled`, disabling a hovered button reports `onHoverOut`, and re-enabling it while the pointer is still inside reports `onHoverIn` again.
```
m-bert:
```suggestion
The two callbacks are paired - an `onHoverIn` is followed by exactly one `onHoverOut` once the pointer leaves. Because hover is only reported while the button is not `disabled`, disabling a hovered button reports `onHoverOut`, and re-enabling it while the pointer is still inside reports `onHoverIn` again.
```

packages/docs-gesture-handler/docs/components/touchable.mdx
```diff
@@ -449,12 +458,102 @@ Triggered when the button gets pressed (analogous to `onPress` in `Pressable` fr
...
+
+On Android, iOS, and macOS the crossings are reported during a press as well: dragging a held pointer off the button and back on reports `onHoverOut` and `onHoverIn`, provided a hover was already open when the press started — which for a stylus needs hardware that reports stylus hover at all. On web nothing is reported while a stylus is in contact, because the browser captures the pointer for the duration of the press.
```
m-bert:
```suggestion
On Android, iOS, and macOS the crossings are reported during a press as well: dragging a held pointer off the button and back on reports `onHoverOut` and `onHoverIn`, provided a hover was already open when the press started - which for a stylus needs hardware that reports stylus hover at all. On web nothing is reported while a stylus is in contact, because the browser captures the pointer for the duration of the press.
```

packages/docs-gesture-handler/docs/components/touchable.mdx
```diff
@@ -449,12 +458,102 @@ Triggered when the button gets pressed (analogous to `onPress` in `Pressable` fr
...
+
+On tvOS hover follows focus rather than a pointer. Focusing the button reports `onHoverIn` and moving focus away reports `onHoverOut`, with the event describing the center of the button and a `pointerType` of `OTHER`.
+
+Unmounting a hovered component does not report `onHoverOut` — the callback goes away with the component. On Android, iOS, and macOS the hover state is also not carried across a detach: if the component moves in the tree while hovered and lands away from the pointer, the pending `onHoverOut` is not delivered, and the next hover over it is treated as a continuation rather than a new `onHoverIn`, resynchronizing on the following `onHoverOut`. On web the component is rebuilt instead, so the next hover reports a fresh `onHoverIn`. Use these callbacks to drive presentation, not to hold state that has to be unwound.
```
m-bert:
```suggestion
Unmounting a hovered component does not report `onHoverOut` - the callback goes away with the component. On Android, iOS, and macOS the hover state is also not carried across a detach: if the component moves in the tree while hovered and lands away from the pointer, the pending `onHoverOut` is not delivered, and the next hover over it is treated as a continuation rather than a new `onHoverIn`, resynchronizing on the following `onHoverOut`. On web the component is rebuilt instead, so the next hover reports a fresh `onHoverIn`. Use these callbacks to drive presentation, not to hold state that has to be unwound.
```

m-bert (changes requested):
Tbh I don't like the fact that this long hover explanation is below `onHoverOut`. I mean I know that it is in the right section, but some parts sound like they belong to `onHoverIn` rather than `onHoverOut`. Maybe we could duplicate it? Also this is now one big section amid prop descriptions, while it covers both props at the same time.

j-piasecki:
@m-bert actually, I'm not sure whether describing it in such detail is needed (contrasts with the existing sections). I trimmed it in [9089315](https://github.com/software-mansion/react-native-gesture-handler/commit/90893158fbe796dc96f5234afe683a56cc2beb9c) and removed tvOS block ~~moved the tvOS note to a `:::note` block~~. Let me know what you think.

m-bert (approved):

PR #4402, opened 2026-08-07
[Android] Fix handlers cancelled while awaiting leaking in the orchestrator

### LINKED ISSUE #4401: Cancelled-while-awaiting handler is never cleaned up, permanently blocking all gestures (Android)
### Description

On Android, if a gesture handler is cancelled while it is `isAwaiting` (i.e. it is waiting for
another handler in a `Gesture.Exclusive`/`requireExternalGestureToFail` relation), the orchestrator
never resets or removes it. It stays in `gestureHandlers` and `awaitingHandlers` for the lifetime of
the process, and every gesture that would have to wait for it never activates.

The practical effect is that **all** gesture-handler based touchables stop responding app-wide —
`TouchableOpacity`/`RectButton`/`BaseButton` from RNGH, and any `GestureDetector`. Nothing recovers
it: unmounting and remounting the screens does not help, only restarting the app does. There is no
warning or error logged.

The most common way to hit this is unmounting a `GestureDetector` while one of its gestures is
awaiting: detaching the view cancels the handler, and
`RNGestureHandlerRegistry.detachHandlerInternal` explicitly does this so that the orchestrator drops
its reference (see the comment on line 135) — but for an awaiting handler that cleanup silently
never happens.

## Diagnosis

Instrumenting `RNGestureHandlerRootHelper.dispatchTouchEvent` to dump the orchestrator's handler
list shows the leak clearly.

Healthy tap:

```
handlers=2[NativeViewGestureHandler#13(state=2,active=false,awaiting=false,view=3942)
           RootViewGestureHandler#-8(state=2,active=false,awaiting=false,view=8)]
after UP: handlers=0[]
```

After the unmount-while-awaiting:

```
handlers=2[TapGestureHandler#24(state=3,active=false,awaiting=true,view=4182)
           RootViewGestureHandler#-8(state=2,active=false,awaiting=false,view=8)]  awaiting=1
```

`state=3` is `STATE_CANCELLED`. That handler is present on every subsequent touch and is never
removed, because both cleanup paths in `GestureHandlerOrchestrator` skip awaiting handlers:

```kotlin
// GestureHandlerOrchestrator.kt:103
if (isFinished(handler.state) && !handler.isAwaiting) { handler.reset() … }

// GestureHandlerOrchestrator.kt:113
gestureHandlers.removeAll { isFinished(it.state) && !it.isAwaiting }
```

`isAwaiting` is cleared in `makeActive`, and in `onHandlerStateChange` when the orchestrator itself
kills an awaiting handler because the awaited one ended (`GestureHandlerOrchestrator.kt:189`).
It is *not* cleared when the handler is cancelled directly — which is exactly what
`RNGestureHandlerRegistry.detachHandlerInternal` does on view detach
(`RNGestureHandlerRegistry.kt:134-138`):

```kotlin
if (handler.view != null) {
  // Handler is in "prepared" state which means it is registered in the orchestrator and can
  // receive touch events. This means that before we remove it from the registry we need to
  // "cancel" it so that orchestrator does no longer keep a reference to it.
  UiThreadUtil.runOnUiThread { handler.cancel() }
}
```

So the intent documented there — cancel so the orchestrator drops the reference — does not hold for
awaiting handlers.

Note that `dropGestureHandler` only calls `registry.dropHandler(handlerTag)`; the orchestrator is
never told to purge the handler, so the cancel is the only cleanup mechanism.

## Suggested fix

Clearing `isAwaiting` when a handler reaches a terminal state that can never be resolved makes the
existing cleanup collect it. `STATE_END` must stay excluded, since the orchestrator deliberately
keeps those pinned so `makeActive` can send their synthetic events.

```kotlin
// GestureHandlerOrchestrator.onHandlerStateChange, right after `handlingChangeSemaphore += 1`
if (handler.isAwaiting &&
  (newState == GestureHandler.STATE_CANCELLED || newState == GestureHandler.STATE_FAILED)
) {
  handler.isAwaiting = false
}
```

We have been running this as a `patch-package` patch and it resolves the problem: after the same
repro the handler list is back to `awaiting=0` with no leftover handler, and all touchables work.
No regressions observed so far in normal gesture use (taps, scroll, pinch, the double/single tap
`Exclusive` composition itself).

An alternative would be to have `dropHandler`/`detachHandlerInternal` purge the handler from the
orchestrator's lists directly, which would also cover handlers stuck in other states.

## Environment

- react-native-gesture-handler: **3.1.0** (also reproduces on 3.0.2)
- React Native: 0.86.0-1 (`react-native-tvos`), New Architecture enabled, Hermes disabled
- Platform: Android only (not tested on iOS)
- Device: OnePlus CPH2653, Android 16 (SDK 36)
- Build type: debug

### Steps to reproduce

Verified path (in a real app):

1. Render a `GestureDetector` with `Gesture.Exclusive(Gesture.Tap().numberOfTaps(2), Gesture.Tap().numberOfTaps(1))`.
2. Single-tap it. The single-tap handler goes into the awaiting state, waiting for the double tap to fail.
3. Unmount the detector while that wait is still pending (in our app the component tree is unmounted
   when the activity enters Android picture-in-picture, which happens right after a tap).
4. Return to the app and tap any RNGH touchable anywhere in the app.

Expected: the touchable responds.
Actual: no gesture ever begins again — no press feedback, no `onBegin`, no `onPress`, on any screen,
until the app process is restarted.

I have not built a standalone Snack, but the condition is not PiP specific — any unmount of a
detector while a handler of it is awaiting should reproduce it (e.g. unmounting from a `setTimeout`
fired within the double-tap window, or navigating away in that window).

-

### Gesture Handler version

3.1.0

### React Native version

0.86.0.-1

### Platforms

Android

### JavaScript runtime

JSC

### Workflow

React Native (without Expo)

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

Real device

### Device model

OnePlus CPH2653, Android 16 (SDK 36)

### Acknowledgements

Yes

m-bert:
## Description

On Android, cancelling a handler while it is awaiting another one (e.g. the single tap in `Exclusive(doubleTap, singleTap)` waiting for the double tap to fail) leaves it in the orchestrator forever. Both cleanup paths in `cleanupFinishedHandlers` skip handlers with `isAwaiting` set, and the rescue loop in `onHandlerStateChange` never reaches it because `dropGestureHandler` drops interaction relations on the JS thread before the posted cancel runs on the UI thread, so `shouldHandlerWaitForOther` no longer matches.

The leaked handler stays in `gestureHandlers`, which makes `ButtonViewGroup.shouldBeginWithRecordedHandlers` return `false` on every subsequent touch. As a result all button-based touchables (`Pressable`, `RectButton`, `BaseButton`, `Touchables`) stop responding app-wide until the app process is restarted. The most common trigger is unmounting a `GestureDetector` during the wait window.

This change clears `isAwaiting` when a handler reaches `STATE_CANCELLED` or `STATE_FAILED`, since such a handler can never be resolved by the one it was waiting for, letting the existing cleanup collect it. `STATE_END` stays pinned, as `makeActive` relies on it to send synthetic events. Going through `onHandlerStateChange` also covers cancel paths that never touch the registry, e.g. `tryActivate` cancelling an awaiting handler via `shouldBeCancelledByFinishedHandler`.

Fixes #4401

## Test plan

<details>
<summary>Tested on the following code</summary>

```tsx
import React, { useRef, useState } from 'react';
import {
  Pressable as RNPressable,
  StyleSheet,
  Text,
  View,
} from 'react-native';
import {
  GestureDetector,
  Pressable,
  RectButton,
  useExclusiveGestures,
  useTapGesture,

// Repro for https://github.com/software-mansion/react-native-gesture-handler/issues/4401
// (Android): cancelling a handler while it is awaiting (Exclusive single tap
// waiting for double tap to fail) leaves it in the orchestrator forever.
//
// Steps:
// 1. Single-tap the purple box. 120ms later (inside the double-tap window,
//    while the single-tap handler is awaiting) the detector unmounts itself.
// 2. Try the probe buttons below. According to the issue, ALL RNGH-based
//    touchables should now be dead app-wide until app restart.

function ExclusiveBox({ onGone }: { onGone: () => void }) {
  const timer = useRef<ReturnType<typeof setTimeout> | null>(null);

  const doubleTap = useTapGesture({
    runOnJS: true,
    numberOfTaps: 2,
    onActivate: () => console.log('[repro] double tap activated'),
  });

  const singleTap = useTapGesture({
    runOnJS: true,
    requireToFail: doubleTap,
    onActivate: () => console.log('[repro] single tap activated'),
    onTouchesUp: () => {
      // Unmount while the single-tap handler is awaiting double-tap failure
      if (timer.current == null) {
        timer.current = setTimeout(() => {
          console.log('[repro] unmounting detector while awaiting');
          onGone();
        }, 120);
      }
    },
  });

  const exclusive = useExclusiveGestures(doubleTap, singleTap);

  return (
    <GestureDetector gesture={exclusive}>
      <View style={[styles.box, { backgroundColor: 'rebeccapurple' }]}>
        <Text style={styles.boxLabel}>
          SINGLE-TAP ME{'\n'}(unmounts in 120ms)
        </Text>
      </View>
    </GestureDetector>
  );
}

  const [mounted, setMounted] = useState(true);
  const [detectorTaps, setDetectorTaps] = useState(0);
  const [pressableTaps, setPressableTaps] = useState(0);
  const [rectTaps, setRectTaps] = useState(0);
  const [rnTaps, setRnTaps] = useState(0);

  const probeTap = useTapGesture({
    runOnJS: true,
    onActivate: () => {
      console.log('[probe] GestureDetector tap');
      setDetectorTaps((n) => n + 1);
    },
  });

  return (
      {mounted ? (
        <ExclusiveBox onGone={() => setMounted(false)} />
      ) : (
        <RNPressable
          style={[styles.box, { backgroundColor: 'gray' }]}
          onPress={() => setMounted(true)}>
          <Text style={styles.boxLabel}>DETECTOR GONE — tap to remount</Text>
        </RNPressable>
      )}

      <GestureDetector gesture={probeTap}>
        <View style={[styles.probe, { backgroundColor: 'darkorange' }]}>
          <Text style={styles.boxLabel}>Probe detector: {detectorTaps}</Text>
        </View>
      </GestureDetector>

      <Pressable
        style={[styles.probe, { backgroundColor: 'seagreen' }]}
        onPress={() => {
          console.log('[probe] RNGH Pressable');
          setPressableTaps((n) => n + 1);
        }}>
        <Text style={styles.boxLabel}>RNGH Pressable: {pressableTaps}</Text>
      </Pressable>

      <RectButton
        style={[styles.probe, { backgroundColor: 'steelblue' }]}
        onPress={() => {
          console.log('[probe] RectButton');
          setRectTaps((n) => n + 1);
        }}>
        <Text style={styles.boxLabel}>RectButton: {rectTaps}</Text>
      </RectButton>

      <RNPressable
        style={[styles.probe, { backgroundColor: 'dimgray' }]}
        onPress={() => {
          console.log('[probe] RN core Pressable');
          setRnTaps((n) => n + 1);
        }}>
        <Text style={styles.boxLabel}>RN core Pressable: {rnTaps}</Text>
      </RNPressable>
    </View>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    gap: 16,
    padding: 24,
  },
  box: {
    width: 260,
    height: 110,
    borderRadius: 12,
    justifyContent: 'center',
    alignItems: 'center',
  },
  probe: {
    width: 260,
    height: 56,
    borderRadius: 12,
    justifyContent: 'center',
    alignItems: 'center',
  },
  boxLabel: {
    color: 'white',
    fontWeight: 'bold',
    textAlign: 'center',
  },
});
```

</details>

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
@@ -161,6 +161,12 @@ class GestureHandlerOrchestrator(
   fun onHandlerStateChange(handler: GestureHandler, newState: Int, prevState: Int) {
     handlingChangeSemaphore += 1
 
+    if (handler.isAwaiting &&
+      (newState == GestureHandler.STATE_CANCELLED || newState == GestureHandler.STATE_FAILED)
+    ) {
+      handler.isAwaiting = false
+    }
+
     if (isFinished(newState)) {
       // We have to loop through copy in order to avoid modifying collection
       // while iterating over its elements


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -161,6 +161,12 @@ class GestureHandlerOrchestrator(
   fun onHandlerStateChange(handler: GestureHandler, newState: Int, prevState: Int) {
     handlingChangeSemaphore += 1
 
+    if (handler.isAwaiting &&
```
j-piasecki, line 164: What about folding `if (isAwaiting) isAwaiting = false` into `if (isFinished(newState))` below?

j-piasecki (approved):

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
m-bert: `isFinished` also contains `STATE_END` right? Then I think it should not be moved

j-piasecki: Yes, I don't think this would be reachable in the END transition, but conceptually I still think it matches - a finished handler should not be awaiting anything, regardless of how it finished.

m-bert: > I don't think this would be reachable in the END transition

It is reachable - it happens on every single `tap` in `Exclusive(double, single)`. A discrete handler activates and ends immediately while still awaiting: on touch up the `tap` goes `ACTIVE -> END` with `isAwaiting` still `true`, so `onHandlerStateChange` runs with `newState == END`. If the flag gets cleared there, `cleanupAwaitingHandlers` removes the handler from `awaitingHandlers` within the same call, and when the double tap fails at the end of its window the rescue loop no longer finds it - `makeActive` never runs and the synthetic `BEGAN -> ACTIVE -> END` events are never sent.

j-piasecki: And this is why awaiting should be a state 😄.

PR #4420, opened 2026-08-14
[Web] Activate ScrollView's native gesture on real scroll instead of pointer distance

m-bert:
## Description

On web, `NativeViewGestureHandler` activated after ~`15px` of pointer movement in any direction, even though the browser does the scrolling itself. A vertical `ScrollView` would claim horizontal drags, and once active, `InteractionManager` failed any `Pan` whose activation criteria (`minDistance`, `activeOffsetX`, ...) delayed activation past the slop - such pans could never activate inside a `ScrollView` or `FlatList`.

This PR adds `ScrollEventManager` which delivers the view's `scroll` events to handlers via a new `onScroll` hook. Handlers with the `ScrollView` role now activate only when the view actually scrolls, with a `2px` pointer travel requirement that ignores momentum-scroll ticks after a touch meant to stop a fling.

> [!IMPORTANT]
> This covers only new, hook based API

## Test plan

- Unit tests for scroll-driven activation, the momentum guard and the unchanged legacy path
- Pan activation criteria screen: all boxes activate per their criteria inside the ScrollView, negatives stay inactive, scrolling works
- Buttons in FlatList screen: scrolling from a button doesn't fire a press, taps do; tap-to-stop-momentum fires nothing and the next tap works

<details>

```tsx
import React, { useRef, useState } from 'react';
import { Button, StyleSheet, Text, View } from 'react-native';
import {
  GestureDetector,
  ScrollView,
  usePanGesture,
import Animated, {
  interpolateColor,
  useAnimatedStyle,
  useSharedValue,
  withTiming,

import type { FeedbackHandle } from '../../../common';
import { COLORS, commonStyles, Feedback } from '../../../common';

type PanConfig = Parameters<typeof usePanGesture>[0];

type DraggableBoxProps = {
  label: string;
  comment: string;
  config?: PanConfig;
  onActivated: (label: string) => void;
};

function DraggableBox({ label, comment, config, onActivated }: DraggableBoxProps) {
  const translateY = useSharedValue(0);

  const panGesture = usePanGesture({
    minDistance: config?.minDistance,
    minVelocity: config?.minVelocity,
    activeOffsetX: config?.activeOffsetX,
    maxPointers: config?.maxPointers,
    runOnJS: true,
    onActivate: () => {
      colorProgress.value = withTiming(1, { duration: 100 });
      onActivated(label);
    },
    onUpdate: (event) => {
      translateX.value = event.translationX;
      translateY.value = event.translationY;
    },
    onFinalize: () => {
      colorProgress.value = withTiming(0, { duration: 100 });
      translateX.value = withTiming(0);
      translateY.value = withTiming(0);
    },
  });

    transform: [
      { translateX: translateX.value },
      { translateY: translateY.value },
    ],
    backgroundColor: interpolateColor(
      colorProgress.value,
      [0, 1],
      [COLORS.NAVY, COLORS.GREEN]
    ),
  }));

  return (
    <View style={[commonStyles.subcontainer, styles.entry]}>
        <Animated.View style={[styles.box, animatedStyle]}>
          <Text style={styles.label}>{label}</Text>
        </Animated.View>
      </GestureDetector>
      <Text style={commonStyles.instructions}>{comment}</Text>
    </View>
  );
}

export default function PanActivationCriteriaExample() {
  const [maxPointers, setMaxPointers] = useState(1);
  const feedbackRef = useRef<FeedbackHandle>(null);

  const onActivated = (label: string) =>
    feedbackRef.current?.showMessage(`Activated: ${label}`);

  return (
      <ScrollView style={styles.scroll}>
        <Text style={commonStyles.instructions}>
          Each box turns green the moment its pan activates. Drag each one and
          verify the activation criteria are respected.
        </Text>
        <DraggableBox
          label="minDistance: 100"
          comment="Should activate only after the finger travels 100pt in any direction."
          config={{ minDistance: 100 }}
          onActivated={onActivated}
        />
        <DraggableBox
          label="minVelocity: 800"
          comment="Should activate only on a fast drag (over 800pt/s), regardless of direction. Slow drags must never activate."
          config={{ minVelocity: 800 }}
          onActivated={onActivated}
        />
        <DraggableBox
          label="activeOffsetX: ±60"
          comment="Should activate only after moving 60pt horizontally. Vertical drags must not activate."
          config={{ activeOffsetX: [-60, 60] }}
          onActivated={onActivated}
        />
        <View style={styles.updateSection}>
          <DraggableBox
            label={`minDistance: 120\nmaxPointers: ${maxPointers}`}
            comment="Explicit minDistance combined with another prop updated at runtime. After pressing the button below, activation must still require 120pt of travel — partial config updates must not reset minDistance."
            config={{ minDistance: 120, maxPointers }}
            onActivated={onActivated}
          />
          <Button
            title="Update unrelated prop (maxPointers)"
            onPress={() => setMaxPointers((prev) => (prev === 1 ? 2 : 1))}
          />
        </View>
      </ScrollView>
      <View style={styles.feedbackOverlay} pointerEvents="none">
        <Feedback ref={feedbackRef} duration={2000} />
      </View>
    </View>
  );
}

  container: {
    flex: 1,
  },
  scroll: {
    paddingVertical: 24,
  },
  feedbackOverlay: {
    position: 'absolute',
    bottom: 20,
    alignSelf: 'center',
  },
  entry: {
    paddingVertical: 24,
    gap: 12,
  },
  box: {
    width: 150,
    height: 150,
    borderRadius: 20,
    justifyContent: 'center',
    alignItems: 'center',
  },
  label: {
    color: 'white',
    fontWeight: '600',
    textAlign: 'center',
  },
  updateSection: {
    paddingBottom: 32,
  },
});
```

</details>


packages/react-native-gesture-handler/src/__tests__/webNativeViewGestureHandler.test.ts
```diff
@@ -0,0 +1,219 @@
...
+import NativeViewGestureHandler from '../web/handlers/NativeViewGestureHandler';
+import type { AdaptedEvent } from '../web/interfaces';
+import { EventTypes, NativeGestureRole } from '../web/interfaces';
+import type { GestureHandlerDelegate } from '../web/tools/GestureHandlerDelegate';
+import GestureHandlerOrchestrator from '../web/tools/GestureHandlerOrchestrator';
+import ScrollEventManager from '../web/tools/ScrollEventManager';
+
+// The Jest environment is node — provide the minimal DOM surface the handler
+// touches (canUseDOM, instanceof HTMLElement, getComputedStyle).
+class FakeHTMLElement {
```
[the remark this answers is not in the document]
m-bert: [Copilot review](https://github.com/software-mansion/react-native-gesture-handler/pull/4420/commits/648473f88159390954175ffce3d0776b4c44a3fb)

packages/react-native-gesture-handler/src/web/tools/ScrollEventManager.ts
```diff
@@ -0,0 +1,35 @@
...
+      x: rect.x - this.view.scrollLeft,
+      y: rect.y - this.view.scrollTop,
+      offsetX: -this.view.scrollLeft,
+      offsetY: -this.view.scrollTop,
+      pointerId: -1,
+      eventType: EventTypes.MOVE,
+      pointerType: PointerType.OTHER,
+      time: event.timeStamp,
+    };
+  }
```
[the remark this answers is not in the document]
m-bert: [Copilot review](https://github.com/software-mansion/react-native-gesture-handler/pull/4420/commits/648473f88159390954175ffce3d0776b4c44a3fb)

packages/react-native-gesture-handler/src/web/tools/ScrollEventManager.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  public unregisterListeners(): void {
+    this.view.removeEventListener('scroll', this.scrollCallback);
+  }
+
+  protected mapEvent(event: Event): AdaptedEvent {
+    // Synthesize a pointer glued to the scrolled content - deltas between
+    // events equal the scrolled distance. getBoundingClientRect would only
+    // add a constant offset at the cost of a layout read on every event.
```
j-piasecki, line 24: Doesn't the DOM event contain the layout info by chance?

packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
```diff
@@ -144,6 +148,9 @@ export default class NativeViewGestureHandler extends GestureHandler {
       return;
     }
 
+    this.scrollDetected = false;
+    this.isScrollDriven = this.computeIsScrollDriven();
```
j-piasecki: Why recalculate this on every pointer down? It doesn't change until the gesture is reattached to a different view.

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
```diff
@@ -398,6 +399,9 @@ export default abstract class GestureHandler implements IGestureHandler {
   protected onWheel(_event: AdaptedEvent): void {
     // Used only by pan gesture handler
   }
+  protected onScroll(_event: AdaptedEvent): void {
+    // Used only by native view gesture handler
+  }
```
j-piasecki, line 404: Would be great to refactor this so only gestures actually needing the callbacks register listeners. That's for a follow-up, though.

packages/react-native-gesture-handler/src/__tests__/webNativeViewGestureHandler.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    view.setAttribute(
+      NATIVE_GESTURE_ROLE_ATTRIBUTE,
+      NativeGestureRole.ScrollView
+    );
+    const handler = createHandler(view);
+
+    handler.pointerDown(touchEvent(100, 100, EventTypes.DOWN));
+    handler.pointerMove(touchEvent(100, 130, EventTypes.MOVE));
+
+    view.dispatchEvent('scroll');
```
j-piasecki, line 146: Won't the browser dispatch `pointercancel` around that time, since it's taking over? That would mean the gesture gets canceled, no?

packages/react-native-gesture-handler/src/web/handlers/GestureHandler.ts
m-bert: In that case the follow-up should also cover onWheel I suppose

packages/react-native-gesture-handler/src/web/tools/ScrollEventManager.ts
m-bert: Unfortunately, scroll event is a generic event, so it doesn't have this info ([source](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll_event#event_type))

packages/react-native-gesture-handler/src/__tests__/webNativeViewGestureHandler.test.ts
m-bert: It does - `pointercancel` maps to `cancel()`, and the scroll events that follow are ignored since the tracker is empty. That's the same outcome as before this PR (the browser takes over below the 15px slop). However, the spec doesn't define the ordering between `pointercancel` and scroll, so the handler must also handle scroll arriving while the pointer is still tracked - that's the case this test covers.

packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
m-bert: [Set ScrollDriven once](https://github.com/software-mansion/react-native-gesture-handler/pull/4420/commits/b1e461241ea3ab31308ba36a3ea0643bdcb796e3)

packages/react-native-gesture-handler/src/__tests__/webNativeViewGestureHandler.test.ts
j-piasecki: Doesn't that mean that now there are cases where the scroll view native handler is never able to activate? If it receives `pointercancel` before `onscroll`, then the cancel will always happen before activation, no?

m-bert: Correct. Though it was almost as broken before: the distance-based activation raced the browser's scroll takeover, so the gesture typically activated and then got cancelled by the same `pointercancel` a moment later. We might consider not cancelling `Native` on `pointercancel` (treating it as the scroll claiming the touch), but then it's not obvious when to actually end the gesture, probably on `scrollend` with a timeout fallback. I wouldn't do it in this PR though. If you think it's worth having, I'd do it in a follow-up. WDYT?

j-piasecki: Just wanted to make sure I understand the change correctly.

> We might consider not cancelling Native on pointercancel

That doesn't sound like a good idea. How do we know what triggered it?

> I wouldn't do it in this PR though. If you think it's worth having, I'd do it in a follow-up.

Unless there's a use case for that, I don't think we need it now.

j-piasecki (approved):

packages/react-native-gesture-handler/src/__tests__/webNativeViewGestureHandler.test.ts
m-bert: > Unless there's a use case for that, I don't think we need it now.

I don't think there is. It's been like that for a long time and seems to do the job. We might consider changing it if it will be actually needed, that's true.

PR #4422, opened 2026-08-14
Remove `REACT_NATIVE_MINOR_VERSION` build flag

m-bert:
## Description

This PR removes the `REACT_NATIVE_MINOR_VERSION` build flag and fixes the `target_compile_reactnative_options` call in the Android `CMakeLists`.

The only consumer of `REACT_NATIVE_MINOR_VERSION` was a >= 81 check in `RNGHRuntimeDecorator.cpp`, which is always true on supported react-native versions. If a version check becomes necessary in the future, react-native ships `<cxxreact/ReactNativeVersion.h>` with the `REACT_NATIVE_VERSION_MINOR` macro, available on all platforms and build systems.

The `target_compile_reactnative_options` block added in #3688 never executed: it ran before `find_package(ReactAndroid)` defines `ReactAndroid_VERSION_MINOR`, referenced an undefined `LIB_TARGET_NAME`, and was missing the include of `react-native-flags.cmake` that defines the function. This PR fixes it to match how reanimated and worklets use it. Note that this applies RN_SERIALIZABLE_STATE and HERMES_V1_ENABLED=1 to our JNI target for the first time, matching how ReactAndroid itself is built.

## Test plan

- basic-example builds and launches on iOS (pod install + yarn ios)
- basic-example builds on Android (assembleDebug, -Wall -Werror, no new warnings)
- compile_commands.json confirms RN_SERIALIZABLE_STATE and HERMES_V1_ENABLED=1 are applied to all translation units
- basic-example builds under the experimental SwiftPM setup from react-native 0.87
- pod ipc spec evaluates the modified podspec cleanly

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
+++ b/packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
@@ -1,17 +1,8 @@
 project(GestureHandler)
 cmake_minimum_required(VERSION 3.9.0)
 
-string(
-  APPEND
-  CMAKE_CXX_FLAGS
-  " -DREACT_NATIVE_MINOR_VERSION=${REACT_NATIVE_MINOR_VERSION}")
-
 set(CMAKE_VERBOSE_MAKEFILE ON)
-if(${REACT_NATIVE_MINOR_VERSION} GREATER_EQUAL 73)
-    set(CMAKE_CXX_STANDARD 20)
-else()
-    set(CMAKE_CXX_STANDARD 17)
-endif()
+set(CMAKE_CXX_STANDARD 20)
 
 set(PACKAGE_NAME "gesturehandler")
 set(RNGH_DIR "${CMAKE_SOURCE_DIR}/../../../../")
@@ -22,6 +13,8 @@ file(GLOB_RECURSE gesture_handler_shared_SRCS CONFIGURE_DEPENDS "${RNGH_DIR}/sha
 
 include(${REACT_ANDROID_DIR}/cmake-utils/folly-flags.cmake)
 add_compile_options(${folly_FLAGS})
+
+include(${REACT_NATIVE_DIR}/ReactCommon/cmake-utils/react-native-flags.cmake)
 
 add_library(${PACKAGE_NAME}
   SHARED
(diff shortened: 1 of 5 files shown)


packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
```diff
@@ -1,17 +1,8 @@
 project(GestureHandler)
 cmake_minimum_required(VERSION 3.9.0)
 
-string(
-  APPEND
-  CMAKE_CXX_FLAGS
-  " -DREACT_NATIVE_MINOR_VERSION=${REACT_NATIVE_MINOR_VERSION}")
-
 set(CMAKE_VERBOSE_MAKEFILE ON)
-if(${REACT_NATIVE_MINOR_VERSION} GREATER_EQUAL 73)
```
j-piasecki, line 10: 🙈

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
```diff
@@ -38,13 +31,11 @@ target_include_directories(
...
 )
 
-if(ReactAndroid_VERSION_MINOR GREATER_EQUAL 80)
-  target_compile_reactnative_options(${LIB_TARGET_NAME} PRIVATE)
-endif()
-
 find_package(ReactAndroid REQUIRED CONFIG)
 find_package(fbjni REQUIRED CONFIG)
 
+target_compile_reactnative_options(${PACKAGE_NAME} PRIVATE)
```
j-piasecki, line 37: Why move it down?

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
```diff
@@ -23,6 +14,8 @@ file(GLOB_RECURSE gesture_handler_shared_SRCS CONFIGURE_DEPENDS "${RNGH_DIR}/sha
 include(${REACT_ANDROID_DIR}/cmake-utils/folly-flags.cmake)
 add_compile_options(${folly_FLAGS})
 
+include(${REACT_NATIVE_DIR}/ReactCommon/cmake-utils/react-native-flags.cmake)
```
j-piasecki, line 17: Where does it come from?

m-bert: It ships with React Native (`ReactCommon/cmake-utils`) since 0.80 and defines `target_compile_reactnative_options`. Without it `target_compile_reactnative_options` fails with "Unknown CMake command". Reanimated and Worklets include it the same way.

packages/react-native-gesture-handler/android/src/main/jni/CMakeLists.txt
m-bert: The old spot was above `find_package(ReactAndroid)`, so `ReactAndroid_VERSION_MINOR` was undefined and the guard was always `false`. Moved next to `target_link_libraries` to match Reanimated.

j-piasecki (approved):

PR #4425, opened 2026-08-14
[Tests] Add basic long press e2e test

j-piasecki:
## Description

Updates the basic long press example and adds an e2e argent flow test

## Test plan

`argent flow run basic-long-press-test`

### DIFF
--- a/apps/common-app/src/new_api/simple/longPress/index.tsx
+++ b/apps/common-app/src/new_api/simple/longPress/index.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import { View } from 'react-native';
+import React, { useCallback, useState } from 'react';
+import { Text, View } from 'react-native';
 import {
   GestureDetector,
   useLongPressGesture,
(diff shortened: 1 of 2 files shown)


apps/common-app/src/new_api/simple/longPress/index.tsx
```diff
@@ -57,7 +72,11 @@ export default function LongPressExample() {
   return (
     <View style={commonStyles.centerView}>
       <GestureDetector gesture={longPressGesture}>
-        <Animated.View style={[commonStyles.box, animatedStyle]} />
+        <Text style={commonStyles.header}>Long press count: {count}</Text>
+        <Animated.View
```
m-bert: `Text` should be outside of the detector I suppose

apps/common-app/src/new_api/simple/longPress/index.tsx
```diff
@@ -57,7 +72,11 @@ export default function LongPressExample() {
...
+          testID="long-press-box"
+          style={[commonStyles.box, animatedStyle]}
```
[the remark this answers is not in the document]
j-piasecki: > GestureDetector only supports a single child view

that's not true for V3, but moved it outside

apps/common-app/src/new_api/simple/longPress/index.tsx
j-piasecki: Moved it outside the detector

m-bert (approved):

PR #4429, opened 2026-08-15
Pass `reanimatedEventHandler` only when the Reanimated detector is used

### LINKED ISSUE #4428: reanimatedEventHandler is passed to the plain host detector, so a non-function ends up on a DirectEventHandler prop
### Description

`NativeDetector` always passes `reanimatedEventHandler` to the native component, even when the gesture does not use the Reanimated detector:

```tsx
// src/v3/detectors/NativeDetector.tsx
      : {
          onGestureHandlerReanimatedEvent:
            gesture.detectorCallbacks.reanimatedEventHandler,
        };
```

The issue is that two different conditions decide whether the handler exists and whether the Reanimated detector is used.

`useGestureCallbacks` builds the handler whenever `disableReanimated` is not set:

```ts
// src/v3/hooks/useGestureCallbacks.ts:53
let reanimatedEventHandler;

if (!config.disableReanimated) {
```

`shouldUseReanimatedDetector` also requires worklet callbacks:

```ts
// src/v3/hooks/utils/configUtils.ts:80
config.shouldUseReanimatedDetector =
  !config.disableReanimated &&
  Reanimated !== undefined &&
  hasWorkletEventHandlers(config) &&
  !config.dispatchesAnimatedEvents;
```

So a gesture with no worklet callbacks gets `shouldUseReanimatedDetector === false` and renders the plain `HostGestureDetector`, which forwards props as they are. But `reanimatedEventHandler` is still defined, so the host component receives Reanimated's `{ workletEventHandler }` object instead of a function.

`onGestureHandlerReanimatedEvent` is a codegen `DirectEventHandler`. React checks that the listener is a function in `getListener`, so when something dispatches that event through one of these detectors you get:

```
Expected `onGestureHandlerReanimatedEvent` listener to be a function, instead got a value of `object` type.
```

Most of the time it stays hidden. A handler on a plain detector gets `dispatchesReanimatedEvents: shouldUseReanimatedDetector && !runOnJS`, which is false, so it never emits the event. It only becomes visible when a handler ends up attached to a detector that is not its own. That is how we found it, and it threw on every gesture frame.

Composed gestures are affected the same way, since `useComposedGesture` calls `useComposedEventHandler` unconditionally. That includes RNGH's own `Pressable` and `Touchable`.

We saw it on iOS, but the branch above is the shared non-web path, so Android runs the same code.

### Suggested fix

Only pass the handler when the Reanimated detector is actually used:

```tsx
      : {
          onGestureHandlerReanimatedEvent: gesture.config
            .shouldUseReanimatedDetector
            ? gesture.detectorCallbacks.reanimatedEventHandler
            : undefined,
        };
```

The prop is part of the static view config, so `undefined` only means there is no JS listener. Native still emits the event.

Happy to open a PR if you want.

### Steps to reproduce

1. New Architecture app with Reanimated installed.
2. Render a gesture with no worklet callbacks inside a `GestureDetector`, for example `useNativeGesture()`.
3. Check what reaches the host component. `onGestureHandlerReanimatedEvent` is an object, not a function.

The repro app shows four gestures side by side and prints, for each one, which detector `NativeDetector` renders and the `typeof` of the handler it gets. The first three are wrong, the last one is a worklet gesture as a control.

https://github.com/antFrancon/rngh-reanimated-handler-repro

### Gesture Handler version

3.2.1 (also unchanged on main)

### React Native version

0.86.2

### Platforms

iOS, Android

### JavaScript runtime

Hermes

### Workflow

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

iOS simulator

### Device model

iPhone 17 (iOS 26.5)

### Acknowledgements

Yes

antFrancon:
## Description

`NativeDetector` always passes `reanimatedEventHandler` to the native component, even when the gesture does not use the Reanimated detector. Two different conditions decide whether the handler exists and whether the Reanimated detector is used.

`useGestureCallbacks` builds the handler whenever `disableReanimated` is unset:

```ts
// src/v3/hooks/useGestureCallbacks.ts
let reanimatedEventHandler;

if (!config.disableReanimated) {
```

`shouldUseReanimatedDetector` additionally requires worklet callbacks:

```ts
// src/v3/hooks/utils/configUtils.ts
config.shouldUseReanimatedDetector =
  !config.disableReanimated &&
  Reanimated !== undefined &&
  hasWorkletEventHandlers(config) &&
  !config.dispatchesAnimatedEvents;
```

So a gesture with no worklet callbacks gets `shouldUseReanimatedDetector === false`, renders the plain `HostGestureDetector`, which forwards props verbatim, and receives Reanimated's `{ workletEventHandler }` object under `onGestureHandlerReanimatedEvent`. That prop is a codegen `DirectEventHandler`, so React throws out of `getListener` instead of dispatching:

```
Expected `onGestureHandlerReanimatedEvent` listener to be a function, instead got a value of `object` type.
```

It stays latent most of the time, because a handler on a plain detector gets `dispatchesReanimatedEvents: shouldUseReanimatedDetector && !runOnJS`, which is false, so it never emits the event. It surfaces once a handler ends up attached to a detector that is not its own. The prop is invalid either way.

This passes the handler only when the Reanimated detector is actually used. The event prop is part of the static view config, so `undefined` only means there is no JS listener, and native still emits the event.

I left the web branch untouched, since I have not tested whether the same mismatch applies there. Happy to extend it if you think it does.

Fixes #4428

## Test plan

- `yarn ts-check`, `yarn lint:js` and `yarn test` in `packages/react-native-gesture-handler`.
- Repro app: https://github.com/antFrancon/rngh-reanimated-handler-repro. It renders four gestures and prints, for each, which detector `NativeDetector` picks and the `typeof` of the handler it receives.
  - Before: `useNativeGesture()`, `usePanGesture()` and `useTapGesture({ onActivate })` all render `HostGestureDetector` with `typeof reanimatedEventHandler === 'object'`.
  - After: those three receive `undefined`, and the worklet control still gets the handler on `ReanimatedNativeDetector`.
- We have been running this change as a patch in an app on Expo SDK 57, RN 0.86.2, React 19.2.3, Reanimated 4.5.3, Fabric, Hermes, iOS, with no regressions in gesture behaviour.

### DIFF
--- a/packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
@@ -51,25 +51,24 @@ export function NativeDetector<
   // On native, Reanimated handles routing internally based on the event names
   // passed to the useEvent hook. We only need to pass it once, so that Reanimated
   // can setup its internal listeners.
+  //
+  // `reanimatedEventHandler` is built whenever `disableReanimated` is unset, but
+  // `shouldUseReanimatedDetector` additionally requires worklet callbacks. When it is false we
+  // render the plain host component, which forwards props verbatim, so passing the handler would
+  // put a non-function on a codegen `DirectEventHandler` prop.
+  const reanimatedEventHandler = gesture.config.shouldUseReanimatedDetector
+    ? gesture.detectorCallbacks.reanimatedEventHandler
+    : undefined;
+
   const reanimatedHandlers =
     Platform.OS === 'web'
       ? {
-          onGestureHandlerReanimatedEvent:
-            gesture.detectorCallbacks.reanimatedEventHandler,
-          onGestureHandlerReanimatedStateChange:
-            gesture.detectorCallbacks.reanimatedEventHandler,
-          onGestureHandlerReanimatedTouchEvent:
-            gesture.detectorCallbacks.reanimatedEventHandler,
+          onGestureHandlerReanimatedEvent: reanimatedEventHandler,
+          onGestureHandlerReanimatedStateChange: reanimatedEventHandler,
+          onGestureHandlerReanimatedTouchEvent: reanimatedEventHandler,
         }
       : {
-          // `reanimatedEventHandler` is built whenever `disableReanimated` is unset, but
-          // `shouldUseReanimatedDetector` additionally requires worklet callbacks. When it is
-          // false we render the plain host component, which forwards props verbatim, so passing
-          // the handler would put a non-function on a codegen `DirectEventHandler` prop.
-          onGestureHandlerReanimatedEvent: gesture.config
-            .shouldUseReanimatedDetector
-            ? gesture.detectorCallbacks.reanimatedEventHandler
-            : undefined,
+          onGestureHandlerReanimatedEvent: reanimatedEventHandler,
         };
 
   return (
--- a/packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
+++ b/packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
@@ -62,8 +62,14 @@ export function NativeDetector<
             gesture.detectorCallbacks.reanimatedEventHandler,
         }
       : {
-          onGestureHandlerReanimatedEvent:
-            gesture.detectorCallbacks.reanimatedEventHandler,
+          // `reanimatedEventHandler` is built whenever `disableReanimated` is unset, but
+          // `shouldUseReanimatedDetector` additionally requires worklet callbacks. When it is
+          // false we render the plain host component, which forwards props verbatim, so passing
+          // the handler would put a non-function on a codegen `DirectEventHandler` prop.
+          onGestureHandlerReanimatedEvent: gesture.config
+            .shouldUseReanimatedDetector
+            ? gesture.detectorCallbacks.reanimatedEventHandler
+            : undefined,
         };
 
   return (


packages/react-native-gesture-handler/src/v3/detectors/NativeDetector.tsx
```diff
@@ -62,8 +62,14 @@ export function NativeDetector<
             gesture.detectorCallbacks.reanimatedEventHandler,
         }
       : {
-          onGestureHandlerReanimatedEvent:
-            gesture.detectorCallbacks.reanimatedEventHandler,
+          // `reanimatedEventHandler` is built whenever `disableReanimated` is unset, but
+          // `shouldUseReanimatedDetector` additionally requires worklet callbacks. When it is
+          // false we render the plain host component, which forwards props verbatim, so passing
+          // the handler would put a non-function on a codegen `DirectEventHandler` prop.
```
m-bert: I don't think we need this 

```suggestion
```

m-bert (commented):
Hi @antFrancon! Thank you for submitting this PR! Could you please also add the same guard for the web part? I know that it should be safe since we have `forReanimated`, but I'd like to keep these in sync 😅

antFrancon:
Done, pushed.

I hoisted the guard into a single `reanimatedEventHandler` const that both branches use, rather than repeating the ternary four times. Net effect is a bit less code than before.

And you are right that web is already safe: `forReanimated` is set from `config.dispatchesReanimatedEvents`, which is `shouldUseReanimatedDetector && !runOnJS`, so the handler never emits there. The guard just makes it explicit instead of load-bearing on that chain.

`ts-check`, `lint:js` and `test` all pass in `packages/react-native-gesture-handler`.

m-bert (approved):

PR #4433, opened 2026-08-17
fix(Android): cancel button handlers when a native view takes the touch lock

### LINKED ISSUE #4432: [Android] Pressable fires onPress when the touch only stops a fling
### Description

On Android, `Pressable` fires `onPress` when the touch was only meant to stop a fling.

Put a finger down on a list row while the list is still decelerating, then lift it without moving: the row under the finger receives a press. In a shop list that means opening a product page the user never tapped.

This is a regression introduced with the v3 `Pressable`. Using `LegacyPressable`, or passing any relation prop (`simultaneousWith` / `requireToFail` / `block`, which routes the component to `StatefulPressable`), makes the phantom press go away.

**Root cause**

`Pressable` without relation props renders `PressableWithTouchable`, which handles the press natively through `ButtonViewGroup`. The `NativeViewGestureHandler` the button manages for itself is attached with `ACTION_TYPE_NONE`:

https://github.com/software-mansion/react-native-gesture-handler/blob/main/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt#L484

When a native view grabs the touch lock — Android's `ScrollView` intercepts the `ACTION_DOWN` while its scroller is still running and calls `requestDisallowInterceptTouchEvent(true)` up the tree — `RNGestureHandlerRootHelper.requestDisallowInterceptTouchEvent()` reacts with `GestureHandlerOrchestrator.cancelAllLegacyHandlers()`, which only cancels handlers whose action type is `ACTION_TYPE_JS_FUNCTION_OLD_API`, `ACTION_TYPE_JS_FUNCTION_NEW_API`, `ACTION_TYPE_REANIMATED_WORKLET` or `ACTION_TYPE_NATIVE_ANIMATED_EVENT`.

The button's handler is none of those, so it survives, reaches `STATE_END` on the finger lift and dispatches the press. The framework's `ACTION_CANCEL` doesn't save it either, since RNGH delivers touches itself and ignores `onInterceptTouchEvent`.

The v2 `Pressable` was built on `GestureDetector`, so its handlers carried a JS action type and were cancelled by that same path — hence the regression.

`LegacyPressable` and `StatefulPressable` are unaffected for the same reason.

### Steps to reproduce

1. Render a list (`FlatList`, `FlashList` or a plain `ScrollView` from `react-native`) whose rows are wrapped in `Pressable` from `react-native-gesture-handler`.
2. Fling the list hard so it keeps scrolling on its own.
3. While it decelerates, put a finger down on a row to stop the scroll.
4. Lift the finger without moving it.

Expected: nothing happens, the touch only stopped the scroll.
Actual: `onPress` fires for the row under the finger.

### Repro

```tsx
import { FlatList, StyleSheet, Text, View } from 'react-native';
import { GestureHandlerRootView, Pressable } from 'react-native-gesture-handler';

const DATA = Array.from({ length: 80 }, (_, index) => `Item ${index}`);

  const [pressed, setPressed] = useState<string>('—');

  return (
      <View style={styles.banner}>
        <Text style={styles.bannerText}>last onPress: {pressed}</Text>
      </View>
      <FlatList
        data={DATA}
        keyExtractor={(item) => item}
        renderItem={({ item }) => (
          <Pressable style={styles.row} onPress={() => setPressed(item)}>
            <Text>{item}</Text>
          </Pressable>
        )}
      />
    </GestureHandlerRootView>
  );
}

  root: { flex: 1 },
  banner: { padding: 16, backgroundColor: '#eee' },
  bannerText: { fontWeight: 'bold' },
  row: { padding: 24, borderBottomWidth: 1, borderBottomColor: '#ddd' },
});
```

Fling the list, stop it with a finger, lift without moving: the banner shows the row that was under the finger. Swap the import for `LegacyPressable` and it stays at `—`.

### Gesture Handler version

3.2.1

### React Native version

0.86.2

### Platforms

Android

### JavaScript runtime

Hermes

### Workflow

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

Real device

huextrat:
## Description

Fixes #4432

On Android, a `Pressable` fires `onPress` when the touch was only meant to stop a fling: put a finger down on a list row while the list decelerates, lift it without moving, and the row under the finger gets pressed.

`Pressable` without relation props renders `PressableWithTouchable`, which presses natively through `ButtonViewGroup`. The `NativeViewGestureHandler` the button manages for itself is attached with `ACTION_TYPE_NONE`, so it isn't cancelled when a native view takes the touch lock: `RNGestureHandlerRootHelper.requestDisallowInterceptTouchEvent()` calls `cancelAllLegacyHandlers()`, which only cancels the action-driven handlers (`JS_FUNCTION_OLD_API`, `JS_FUNCTION_NEW_API`, `REANIMATED_WORKLET`, `NATIVE_ANIMATED_EVENT`).

Android's `ScrollView` intercepts the `ACTION_DOWN` while its scroller is still running and calls `requestDisallowInterceptTouchEvent(true)` up the tree, so that path does run — the button handler just isn't part of what it cancels. It then reaches `STATE_END` on the finger lift and dispatches the press. The framework's own `ACTION_CANCEL` doesn't reach the button either, since RNGH delivers touches itself and ignores `onInterceptTouchEvent`.

This cancels `NativeViewGestureHandler`s attached with `ACTION_TYPE_NONE` as well. The root view's handler shares that action type but is not a `NativeViewGestureHandler`, so it keeps running and interception is unaffected.

The v2 `Pressable` went through `GestureDetector`, so its handlers carried a JS action type and were cancelled by this very path — `LegacyPressable` and `StatefulPressable` (any relation prop) are unaffected today, which is a decent A/B when checking the fix.

I also renamed `cancelAllLegacyHandlers` to `cancelHandlersLosingToNativeGesture`, since it no longer cancels only the legacy handlers — happy to drop the rename if you'd rather keep the diff minimal.

## Test plan

Repro app: a `FlatList` (from `react-native`) whose rows are wrapped in `Pressable` from `react-native-gesture-handler`, each row logging its `onPress`.

1. Fling the list hard.
2. While it decelerates, put a finger down on a row to stop the scroll.
3. Lift the finger without moving it.

Before: `onPress` fires for the row under the finger.
After: nothing fires, and a deliberate tap on a settled list still presses normally.

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
@@ -366,9 +366,17 @@ class GestureHandlerOrchestrator(
   }
 
   /**
-   * Cancels all handlers created using API v1 and v2
+   * Cancels the handlers that lose to a native view taking the touch lock: the ones created using
+   * API v1 and v2, and the [NativeViewGestureHandler] a button manages for itself.
+   *
+   * The button's handler is attached with [GestureHandler.ACTION_TYPE_NONE] - it dispatches its
+   * events natively rather than through an action - so an action type check alone leaves it
+   * running. It has to be cancelled here as well, or a touch that a native view claims still ends
+   * the button handler and fires a press. The root view's own handler shares that action type and
+   * must keep running, hence the type check rather than a plain [GestureHandler.ACTION_TYPE_NONE]
+   * one.
    */
-  fun cancelAllLegacyHandlers() {
+  fun cancelHandlersLosingToNativeGesture() {
     val handlersToProcess = obtainHandlerList()
     handlersToProcess.addAll(gestureHandlers)
 
@@ -377,7 +385,8 @@ class GestureHandlerOrchestrator(
         if (it.actionType == GestureHandler.ACTION_TYPE_JS_FUNCTION_OLD_API ||
           it.actionType == GestureHandler.ACTION_TYPE_JS_FUNCTION_NEW_API ||
           it.actionType == GestureHandler.ACTION_TYPE_REANIMATED_WORKLET ||
-          it.actionType == GestureHandler.ACTION_TYPE_NATIVE_ANIMATED_EVENT
+          it.actionType == GestureHandler.ACTION_TYPE_NATIVE_ANIMATED_EVENT ||
+          (it is NativeViewGestureHandler && it.actionType == GestureHandler.ACTION_TYPE_NONE)
         ) {
           it.cancel()
         }
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
@@ -120,7 +120,7 @@ class RNGestureHandlerRootHelper(private val context: ReactContext, wrappedView:
     if (orchestrator != null && !passingTouch) {
       // if we are in the process of delivering touch events via GH orchestrator, we don't want to
       // treat it as a native gesture capturing the lock
-      orchestrator.cancelAllLegacyHandlers()
+      orchestrator.cancelHandlersLosingToNativeGesture()
     }
   }
 


packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -377,7 +385,8 @@ class GestureHandlerOrchestrator(
         if (it.actionType == GestureHandler.ACTION_TYPE_JS_FUNCTION_OLD_API ||
           it.actionType == GestureHandler.ACTION_TYPE_JS_FUNCTION_NEW_API ||
           it.actionType == GestureHandler.ACTION_TYPE_REANIMATED_WORKLET ||
-          it.actionType == GestureHandler.ACTION_TYPE_NATIVE_ANIMATED_EVENT
+          it.actionType == GestureHandler.ACTION_TYPE_NATIVE_ANIMATED_EVENT ||
+          (it is NativeViewGestureHandler && it.actionType == GestureHandler.ACTION_TYPE_NONE)
```
j-piasecki, line 389: This will regress https://github.com/software-mansion/react-native-gesture-handler/pull/4367 - `requestDisallowInterceptTouchEvent` would cancel all buttons globally again, instead of being scoped to the relevant subtree.

@m-bert is exploring other approaches to this problem in https://github.com/software-mansion/react-native-gesture-handler/pull/4441

huextrat: Yes, I close this PR in favor of https://github.com/software-mansion/react-native-gesture-handler/pull/4441

PR #4441, opened 2026-08-18
[Android] Fix buttons firing press events when a scroll takes over the touch

### LINKED ISSUE #4432: [Android] Pressable fires onPress when the touch only stops a fling

m-bert:
## Description

`Pressable` without relation props presses natively through `ButtonViewGroup`, whose managed `NativeViewGestureHandler` is attached with `ACTION_TYPE_NONE`. RNGH delivers touches through the orchestrator regardless of what happens in the native dispatch, so when a native `ScrollView` takes the gesture over, nothing stops the handler - it reaches `STATE_END` on lift and fires a press. This shows up in three ways:

- fling catch: the `ScrollView` intercepts `DOWN` while decelerating, the button never sees any native event, yet `onPress` fires on lift (#4432)
- drag: the `ScrollView` intercepts on `MOVE` when the finger starts scrolling from a row, and `onPress` still fires on lift ([comment](https://github.com/software-mansion/react-native-gesture-handler/issues/4432#issuecomment-5315039272))
- long press while scrolling: the content moves with the finger, so the pointer never leaves the row and the long-press timer posted on `BEGAN` fires mid-scroll (same comment)

In all three the `ScrollView` calls `requestDisallowInterceptTouchEvent(true)`, but the existing sweep (`cancelAllLegacyHandlers`) only cancels action-driven handlers, and the `ButtonViewGroup` override from #4367 never runs since the request only bubbles up from the `ScrollView`.

Cancelling button handlers directly at request time (the #4433 approach) is not valid either: an eager disallow-intercept (`react-native-pager-view`'s `NestedScrollableHost` requests it on `DOWN` whenever it's nested inside another `ViewPager`, without intercepting anything) is indistinguishable from a real interception at that moment, so every `Pressable` inside nested pagers (e.g. material top tabs in a pager) would go dead - the regression class #4367 fixed.

The two can be told apart by when the grab happened and whether the native dispatch still reached the button:

- `ButtonViewGroup` tracks `receivedNativeDown` - set in `dispatchTouchEvent` (handler delivery bypasses it), reset on `BEGAN`, which the orchestrator dispatches before the native dispatch of the same `DOWN`.
- `RNGestureHandlerRootHelper` records the disallow request, and once the root view finishes `super.dispatchTouchEvent` runs `cancelHandlersOnNativeTouchGrab`, cancelling handlers whose hook opts in.
- The hook decides via `shouldCancelOnNativeTouchGrab(grabbedMidGesture) = grabbedMidGesture || !receivedNativeDown`: a grab on any pass after `DOWN` means actual dragging (cancel, matching what the legacy `Pressable` and RN's `Pressable` do), while a grab during the `DOWN` pass spares a button that received that `DOWN` (a defensive disallow lets the event through).

Only `ButtonViewGroup` opts into the hook, so handlers attached to detectors, scrollables and text inputs are unaffected. The cost on passes without a disallow request is a single boolean check.

Fixes #4432
Supersedes #4433

## Test plan

Repro below: a `SectionList` with `Pressable` rows (`onPress` + `onLongPress`), a `Pressable` and a long-press `GestureDetector` inside nested `PagerView`s (the eager-disallow setup from #2383), and an engine toggle (v3 / `LegacyPressable` / RN `Pressable`). All runs on the same emulator, main vs this PR:

| scenario | main | this PR |
| --- | --- | --- |
| fling the list, touch a row to stop it, lift | phantom `onPress` | nothing |
| put a finger on a row and drag-scroll, lift | phantom `onPress` | nothing |
| hold a row while drag-scrolling past 500 ms | phantom `onLongPress` | nothing |
| tap a row on a settled list | `onPress` | `onPress` |
| stationary long press on a row | `onLongPress` | `onLongPress` |
| tap the `Pressable` inside nested pagers | `onPress` | `onPress` |
| long press the detector box inside nested pagers (#2383) | activates | activates |

`LegacyPressable` behaves the same in the list scenarios; inside nested pagers it doesn't fire on main either - its handlers are cancelled on any disallow-intercept request, which is the pre-existing legacy behavior this PR doesn't change. RN's `Pressable` doesn't go through RNGH and is clean everywhere.

<details>
<summary>Repro</summary>

```tsx
import {
  Pressable as RNPressable,
  SectionList,
  StyleSheet,
  Text,
  View,
} from 'react-native';
import PagerView from 'react-native-pager-view';
import {
  GestureDetector,
  LegacyPressable,
  Pressable,
  useLongPressGesture,

const SECTIONS = Array.from({ length: 8 }, (_, section) => ({
  title: `Section ${section}`,
  data: Array.from({ length: 10 }, (_, index) => `Item ${section}-${index}`),
}));

const ENGINES = ['Pressable (v3)', 'LegacyPressable', 'RN Pressable'] as const;
const COMPONENTS = [Pressable, LegacyPressable, RNPressable] as const;

function LongPressBox({ onLongPress }: { onLongPress: () => void }) {
    runOnJS: true,
    onActivate: onLongPress,
  });

  return (
      <View style={styles.gestureBox} />
    </GestureDetector>
  );
}

  const [engine, setEngine] = useState(0);
  const [lastEvent, setLastEvent] = useState('none');
  const [eventCount, setEventCount] = useState(0);

  const Row = COMPONENTS[engine] as typeof Pressable;

  const report = (kind: string, item: string) => {
    setLastEvent(`${kind} ${item}`);
    setEventCount((count) => count + 1);
  };

  return (
    <View style={styles.root}>
      <View style={styles.banner}>
        <Text style={styles.bannerText}>engine: {ENGINES[engine]}</Text>
        <Text style={styles.bannerText}>
          last: {lastEvent} (count: {eventCount})
        </Text>
        <Pressable
          style={styles.toggle}
          onPress={() => {
            setEngine((current) => (current + 1) % ENGINES.length);
            setLastEvent('none');
            setEventCount(0);
          }}>
          <Text style={styles.toggleText}>Toggle engine</Text>
        </Pressable>
      </View>
      {/* Nested pagers: the inner pager's NestedScrollableHost calls
          requestDisallowInterceptTouchEvent(true) on ACTION_DOWN only when it
          sits inside another ViewPager2 — the eager-disallow case from #4367. */}
      <PagerView style={styles.pager} initialPage={0}>
        <View key="outer-a" style={styles.page}>
          <PagerView style={styles.innerPager} initialPage={0}>
            <View key="a" style={[styles.page, styles.pageRow]}>
              <Row
                style={styles.pagerButton}
                onPress={() => report('press', 'pager-button')}>
                <Text style={styles.toggleText}>Pager button</Text>
              </Row>
              {/* The #2383 setup: a long-press gesture inside nested pagers
                  (material top tabs are pager-view underneath). */}
              <LongPressBox onLongPress={() => report('gesture', 'pager-box')} />
            </View>
            <View key="b" style={styles.page}>
              <Text>Page B</Text>
            </View>
          </PagerView>
        </View>
        <View key="outer-b" style={styles.page}>
          <Text>Outer page B</Text>
        </View>
      </PagerView>
      <SectionList
        sections={SECTIONS}
        keyExtractor={(item) => item}
        renderSectionHeader={({ section }) => (
          <Text style={styles.sectionHeader}>{section.title}</Text>
        )}
        renderItem={({ item }) => (
          <Row
            style={styles.row}
            onPress={() => report('press', item)}
            onLongPress={() => report('longPress', item)}>
            <Text>{item}</Text>
          </Row>
        )}
      />
    </View>
  );
}

  root: {
    flex: 1,
  },
  banner: {
    padding: 16,
    gap: 8,
    backgroundColor: '#eee',
  },
  bannerText: {
    fontWeight: 'bold',
  },
  toggle: {
    alignSelf: 'flex-start',
    paddingVertical: 8,
    paddingHorizontal: 16,
    borderRadius: 8,
    backgroundColor: 'steelblue',
  },
  toggleText: {
    color: 'white',
  },
  sectionHeader: {
    paddingHorizontal: 24,
    paddingVertical: 8,
    fontWeight: 'bold',
    backgroundColor: '#ddd',
  },
  row: {
    padding: 24,
    borderBottomWidth: 1,
    borderBottomColor: '#ddd',
  },
  pager: {
    height: 110,
    borderBottomWidth: 2,
    borderBottomColor: '#bbb',
  },
  page: {
    alignItems: 'center',
    justifyContent: 'center',
  },
  pageRow: {
    flexDirection: 'row',
    gap: 16,
  },
  gestureBox: {
    width: 64,
    height: 44,
    borderRadius: 8,
    backgroundColor: 'crimson',
  },
  innerPager: {
    alignSelf: 'stretch',
    flex: 1,
  },
  pagerButton: {
    paddingVertical: 12,
    paddingHorizontal: 24,
    borderRadius: 8,
    backgroundColor: 'darkorange',
  },
});
```

</details>


m-bert:
@coderabbitai review again

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
```diff
@@ -365,20 +365,13 @@ class GestureHandlerOrchestrator(
     event.recycle()
   }
 
-  /**
-   * Cancels all handlers created using API v1 and v2
-   */
-  fun cancelAllLegacyHandlers() {
+  private inline fun cancelTrackedHandlers(predicate: (GestureHandler) -> Boolean) {
```
j-piasecki:
```suggestion
  private inline fun cancelHandlersMatching(predicate: (GestureHandler) -> Boolean) {
```

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
```diff
@@ -120,10 +121,27 @@ class RNGestureHandlerRootHelper(private val context: ReactContext, wrappedView:
...
+      nativeTouchGrabRequested = true
       orchestrator.cancelAllLegacyHandlers()
     }
   }
 
+  /**
+   * A disallow-intercept request may mean a real interception, but it may also be just a
+   * defensive call from a view that lets the event through (e.g. a nested pager). The two can only
+   * be told apart after the native dispatch completes, so cancellation runs here, not at request time.
+   */
```
j-piasecki, line 147: What does that mean? That a view may call `requestDisallowInterceptTouchEvent` outside of processing the event stream? If so, why cannot `passingTouch` be used?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
```diff
@@ -744,6 +748,8 @@ class RNGestureHandlerButtonViewManager :
       val localLastEventWasInside = lastEventWasInside
 
       if (newState == GestureHandler.STATE_BEGAN) {
+        // Reset for the new gesture - BEGAN precedes the native dispatch of DOWN that sets the flag.
+        receivedNativeDown = false
```
j-piasecki: Wouldn't it make more sense to reset this in the `end/failed/canceled` branch?

Currently, the flag is set to true after the gesture finishes until a new one is started, instead of for the gesture duration.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootView.kt
```diff
@@ -62,7 +62,9 @@ class RNGestureHandlerRootView(context: Context?) : ReactViewGroup(context) {
     return if (rootViewEnabled && rootHelper!!.dispatchTouchEvent(event)) {
       true
     } else {
-      super.dispatchTouchEvent(event)
+      val handled = super.dispatchTouchEvent(event)
+      rootHelper?.onNativeDispatchEnd(event)
```
j-piasecki, line 67: This is cleared only in the `dispatchTouchEvent` chain. If something were to call `requestDisallowInterceptTouchEvent` outside of it (I think RN does), `nativeTouchGrabRequested` would stay set to true for the next dispatch.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
m-bert: `passingTouch` is set to `true` for the orchestrator pass. When we need this, it would already be `false`.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: It has been done this way to ensure that the flag is reset before the next gesture needs it, even if it for some reason misses the terminal states. We can change it if you prefer to.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
```diff
@@ -120,10 +121,23 @@ class RNGestureHandlerRootHelper(private val context: ReactContext, wrappedView:
     if (orchestrator != null && !passingTouch) {
       // if we are in the process of delivering touch events via GH orchestrator, we don't want to
       // treat it as a native gesture capturing the lock
```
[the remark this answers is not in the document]
m-bert: [f7c0be9](https://github.com/software-mansion/react-native-gesture-handler/pull/4441/commits/f7c0be9a56f7faf10df858738af238910d6889ca)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootView.kt
m-bert: [d6fe536](https://github.com/software-mansion/react-native-gesture-handler/pull/4441/commits/d6fe536c4775bb96ab948f1ecbf7238a5f7e4aac)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/GestureHandlerOrchestrator.kt
m-bert: [1e89b32](https://github.com/software-mansion/react-native-gesture-handler/pull/4441/commits/1e89b3215d5eb024fa4792ffecb35eddb1008fdd)

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
j-piasecki: > even if it for some reason misses the terminal states

How would that happen? If it does, it should fail very loudly IMO.

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
j-piasecki: Fair, but

> What does that mean?

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
m-bert: This might happen if the gesture is `disabled`, since the flag is updated on the `DOWN` event. I'm not sure if we care that much about this case though

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerRootHelper.kt
m-bert: Rephrased in [e9feb1f](https://github.com/software-mansion/react-native-gesture-handler/pull/4441/commits/e9feb1f912b1efa3e7eb4a2c1cb757400df66aa6), let me know if that's better

packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/react/RNGestureHandlerButtonViewManager.kt
j-piasecki: When it's disabled, wouldn't it simply not change state? So clearing in `BEGIN` also wouldn't work?

m-bert: When `disabled` - yes, correct. But if you then switch it to `enabled`, then we would be left with stale value, as the terminal state has not been reached and there's no reset in `BEGIN` in that case.

m-bert: Let me know if [this comment](https://github.com/software-mansion/react-native-gesture-handler/pull/4441/commits/b51a16d18aa53d9fcc19eae32b389826d3c0c70b) satisfies you, or does it need deeper explanation 😅

j-piasecki (approved):

PR #4457, opened 2026-08-20
[iOS] Remove a stale TODO in `State.ts` and the unused constants export

m-bert:
## Description

Removes two leftovers that reference each other:

- The `TODO use State from RNModule` comment in `src/State.ts`, added in the TS rewrite (#1327). The idea was to source the `State` object from the native module's exported constants instead of duplicating the values in JS. That's no longer viable: the TurboModule spec has no `getConstants`, web and Jest have no native module to ask, and `as const` literal types have to exist at compile time anyway.

- The `constantsToExport` method in `RNGestureHandlerModule.mm` (the `State` and `Direction` dictionaries, exported since 2017) together with the now-unused `RNGestureHandlerDirection.h` import. Nothing on the JS side reads these - without `getConstants` in the spec they are unreachable on the new architecture.

`requiresMainQueueSetup` stays, as module initialization inserts into the static `_managers` map that component views read on the main thread.

No behavior change.

## Test plan

- Built and ran `basic-example` on the iOS simulator; the app renders and a tap on the `VirtualGestureDetector` text logs `Tapped on first part!`.
- Verified nothing references the exported constants: no `getConstants` in `NativeRNGestureHandlerModule.ts`, no `Module.State` / `Module.Direction` reads in `src/`.

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
@@ -17,7 +17,6 @@
 #import "RNGHRuntimeDecorator.h"
 
 #import "RNGestureHandler.h"
-#import "RNGestureHandlerDirection.h"
 #import "RNGestureHandlerState.h"
 
 #import "RNGestureHandlerButton.h"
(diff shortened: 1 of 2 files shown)


packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
```diff
@@ -278,28 +277,6 @@ - (void)addOperationBlock:(GestureHandlerOperation)operation
...
-      @"FAILED" : @(RNGestureHandlerStateFailed),
-      @"END" : @(RNGestureHandlerStateEnd)
-    },
-    @"Direction" : @{
-      @"RIGHT" : @(RNGestureHandlerDirectionRight),
-      @"LEFT" : @(RNGestureHandlerDirectionLeft),
-      @"UP" : @(RNGestureHandlerDirectionUp),
-      @"DOWN" : @(RNGestureHandlerDirectionDown)
-    }
-  };
```
j-piasecki, line 300: Android module didn't export it?

j-piasecki (approved):

packages/react-native-gesture-handler/apple/RNGestureHandlerModule.mm
m-bert: It did, [until you removed it](https://github.com/software-mansion/react-native-gesture-handler/pull/3489/changes#diff-fb02679eef83b54be0c840ae5985197f817a7aa4e2521f8991d1377fd6617d21L191) 😄

PR #4458, opened 2026-08-20
[Docs] Add JSON-LD and llms.txt for AI search

halskiszymon:
Two SEO changes for the docs site. Nothing changes the rendered page.

- **JSON-LD**: `Organization` with the same `@id` as swmansion.com, plus `SoftwareSourceCode` for the library, derived from `siteConfig`. The identical `@id` is what makes engines read the docs and the company site as one entity instead of two unrelated ones.
- **`llms.txt`**, generated at build — no new dependency. Lists every docs page. The docs build now fails if the file is missing or lists nothing.

Both come from one local plugin, `plugins/swm-geo.js`, registered with a single line. The same change is going out across the docs repos; reference: software-mansion/react-native-reanimated#10332.

**Check:** CI prints `llms.txt lists N pages`.


.github/workflows/docs-check.yml
```diff
@@ -33,3 +33,17 @@ jobs:
...
+          if [ ! -s "$file" ]; then
+            echo "::error::$file is missing or empty"
+            exit 1
+          fi
+          count=$(grep -cE '^- \[[^]]+\]\(https://docs\.swmansion\.com/react-native-gesture-handler/[^)]+\)' "$file" || true)
+          if [ "$count" -eq 0 ]; then
+            echo "::error::$file lists no pages"
+            exit 1
+          fi
+          echo "llms.txt lists $count pages"
```
m-bert: Do we need this last `echo`? Also, lacks newline


```suggestion
          echo "llms.txt lists $count pages"
          
```

m-bert (approved):

.github/workflows/docs-check.yml
halskiszymon: not needed, removed

PR #4463, opened 2026-08-21
Remove stale TODOs

m-bert:
## Description

Removes TODO comments that describe deliberate, final behavior as unfinished work. No logic changes.

- `android/.../react/RNGestureHandlerRootView.kt` - drops the `resettable lateinit` TODO from the 2021 Kotlin conversion; `rootHelper` is assigned at most once and never reset, so the nullable var is the intended shape
- `src/web/handlers/NativeViewGestureHandler.ts` - replaces the activation TODO with a note that `shouldActivateOnStart` is documented as Android-only and on web is honored internally only for the Button role, other roles activate on real interaction on purpose
- `android/.../react/events/RNGestureHandlerStateChangeEvent.kt` - drops two `coalescing` TODOs; state transitions are discrete and each one must reach JS, so they are intentionally never coalesced
- `apple/RNGestureHandlerEvents.mm` - drops the same coalescing TODO from `RNGestureHandlerStateChange` for consistency with Android

## Test plan

Comment-only changes, CI passing is sufficient.

### DIFF
--- a/packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
@@ -32,7 +32,8 @@ export default class NativeViewGestureHandler extends GestureHandler {
   public override readonly isContinuous = true;
   private role: NativeGestureRole | null = null;
 
-  // TODO: Implement logic for activation on start properly
+  // Documented as Android-only; on web it's used internally for the Button
+  // role, while other roles intentionally activate on real interaction.
   private shouldActivateOnStart = false;
   private disallowInterruption = false;
   private yieldsToContinuousGestures = false;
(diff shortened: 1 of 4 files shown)


packages/react-native-gesture-handler/src/web/handlers/NativeViewGestureHandler.ts
```diff
@@ -32,7 +32,8 @@ export default class NativeViewGestureHandler extends GestureHandler {
   public override readonly isContinuous = true;
   private role: NativeGestureRole | null = null;
 
-  // TODO: Implement logic for activation on start properly
+  // Documented as Android-only; on web it's used internally for the Button
+  // role, while other roles intentionally activate on real interaction.
```
j-piasecki, line 36: This literally reads like an LLM response

m-bert: Indeed 

But as discussed ofline, this comment actually makes sense 🤷

j-piasecki (approved):

PR #4470, opened 2026-08-24
Fix ReanimatedDrawerLayout animation speed after rerender

### LINKED ISSUE #4469: ReanimatedDrawerLayout uses stale animationSpeed after rerender
### Description

`ReanimatedDrawerLayout` keeps using the `animationSpeed` value from its initial render after the prop changes.

`animateDrawer` reads `animationSpeedProp`, but the callback does not include it in its dependency list. Since `openDrawer` and `closeDrawer` capture `animateDrawer`, programmatic operations after a prop-only rerender continue to calculate the spring `mass` from the previous prop value.

This reproduces on current `main` at `5de6d2358`.

### Steps to reproduce

1. Render `ReanimatedDrawerLayout` with `animationSpeed={2}` and retain its imperative ref.
2. Call `openDrawer()` without an explicit per-call `animationSpeed`; the spring uses `mass: 0.5`.
3. Rerender the same component with `animationSpeed={4}`.
4. Call `openDrawer()` again without an explicit per-call speed.
5. Observe that the second spring still uses `mass: 0.5` rather than `0.25`.

### Reproduction

https://gist.github.com/ngocdevv/04e609cfedd9b0125ad2bca28aa3e78a

The deterministic Jest regression test fails on current `main` with:

```text
Expected mass: 0.25
Received mass: 0.5
```

### Expected behavior

The second operation uses the latest prop and configures the spring with `mass: 0.25`.

### Actual behavior

The second operation continues using the initial prop and configures `mass: 0.5`.

### Root cause

`animationSpeedProp` is referenced inside `animateDrawer`, but is missing from its `useCallback` dependency array in `src/components/ReanimatedDrawerLayout.tsx`.

### Gesture Handler version

Current `main` at `5de6d2358` (`package.json`: `2.29.0`).

### React Native version

`0.87.0`

### Related library versions

- `react-native-reanimated`: `4.6.0-nightly-20260811-248dee712`
- `react-native-worklets`: `0.12.0-nightly-20260810-fb9cb5596`

### Platforms

Android, iOS, and web (the stale React callback is shared code).

### JavaScript runtime

Hermes

### Workflow

React Native library repository / Jest reproduction

### Architecture

New Architecture (Fabric)

### Build type

Debug / test

### Device

Automated Jest reproduction; no device required.

### Acknowledgements

Yes

ngocdevv:
## Description

Fixes #4469.

`ReanimatedDrawerLayout` memoized `animateDrawer` without `animationSpeedProp`, so changing the prop did not affect later programmatic `openDrawer()` or `closeDrawer()` calls. The callback now tracks the prop and the imperative methods receive the latest default spring speed after a rerender.

## Test plan

- `yarn workspace react-native-gesture-handler test --runInBand` — 159 tests passed
- `yarn workspace react-native-gesture-handler ts-check`
- `yarn workspace react-native-gesture-handler lint-js` — no errors (existing warnings remain)
- `yarn workspace react-native-gesture-handler build`


packages/react-native-gesture-handler/src/__tests__/reanimatedDrawerLayoutAnimationSpeed.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    withSpring: jest.fn((toValue: unknown) => toValue),
+  };
+});
+
+jest.mock('react-native-worklets', () => ({
+  scheduleOnRN: jest.fn(
+    (callback: (...args: unknown[]) => void, ...args: unknown[]) =>
+      callback(...args)
+  ),
+}));
```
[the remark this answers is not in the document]
ngocdevv: Updated in b858ef858. The regression test now uses the complete official `react-native-worklets/src/mock`; the imperative call is wrapped in async `act` and flushes the mock’s RN microtask so scheduled state updates stay inside the test boundary. Focused Jest, TypeScript, ESLint, and Prettier checks pass.

packages/react-native-gesture-handler/src/__tests__/reanimatedDrawerLayoutAnimationSpeed.test.tsx
m-bert, line 1: I don't think this is needed

m-bert (changes requested):
Hi @ngocdevv! Thank you for subimitting this PR! While this fix looks good, I don't think we need this big test file just for this specific behavior.

packages/react-native-gesture-handler/src/__tests__/reanimatedDrawerLayoutAnimationSpeed.test.tsx
ngocdevv: Removed the regression test file in 73f97c33b, leaving the PR with only the one-line callback dependency fix. I also updated the PR description and re-ran the package checks on the branch after syncing with current main: 159 tests passed, TypeScript passed, lint completed with 0 errors, and the package build passed.

m-bert (approved):

PR #4472, opened 2026-08-26
Don't pass dependencies to Reanimated hooks on native

m-bert:
## Description

Since Reanimated 4.6 (software-mansion/react-native-reanimated#10009) the native implementations of `useAnimatedStyle`, `useDerivedValue` and other hooks log a dev warning whenever a dependencies argument is passed, since dependencies are only relevant on web:

```
WARN  [Reanimated] dependencies should only be used in web implementation.
```

We pass dependencies in two places:

- `ReanimatedSwipeable` passes `[appliedTranslation, rowState]` to `useAnimatedStyle`. The dependencies are still needed on web, where bundlers resolve the compiled `lib` output that isn't processed by the Reanimated babel plugin, so they are now passed only when `Platform.OS === 'web'`.
- `ReanimatedDrawerLayout` passed an empty array to `useDerivedValue`. An empty array does nothing on any platform (the web fallback only reads non-empty dependencies), so it's simply removed.

## Test plan

- Open the Swipeable and Drawer examples in the example app on native and check that the warning is no longer logged.
- Check that Swipeable still animates correctly in the example app running on web.

### DIFF
--- a/packages/react-native-gesture-handler/src/components/ReanimatedDrawerLayout.tsx
+++ b/packages/react-native-gesture-handler/src/components/ReanimatedDrawerLayout.tsx
@@ -343,7 +343,7 @@ const DrawerLayout = function DrawerLayout(
 
   useDerivedValue(() => {
     onDrawerSlide && scheduleOnRN(onDrawerSlide, openValue.value);
-  }, []);
+  });
 
   const isDrawerOpen = useSharedValue(false);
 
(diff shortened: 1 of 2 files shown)


j-piasecki (approved):

erkia:
Why add such noise in your code, instead of removing the warning?

PR #4479, opened 2026-09-03
Reset `numberOfPointers` in `LongPressGestureHandler`

giaBaoJS:
## Description

`LongPressGestureHandler` accepts `numberOfPointers` but never resets it, so the value survives a config update that no longer sets it.

`setConfig` rebuilds a handler's whole config as `resetConfig()` followed by `updateConfig()`:

* web: `src/web/handlers/GestureHandler.ts:790`
* Apple: `apple/RNGestureHandler.mm:131`
* Android: `android/src/main/java/com/swmansion/gesturehandler/core/GestureHandler.kt:911`

`updateConfig` only assigns properties that are present in the incoming config, so every property it can write has to be cleared by `resetConfig`. `numberOfPointers` was written but never cleared:

* `src/web/handlers/LongPressGestureHandler.ts:127` writes it, `:76` did not reset it
* `android/.../core/LongPressGestureHandler.kt:197` writes it, `:34` did not reset it
* `apple/Handlers/RNLongPressHandler.m:255` writes it, `:232` did not reset it

Every sibling handler already resets its pointer count, which is what made this stand out: `Tap` resets `minNumberOfPointers`, `Fling` resets `numberOfPointersRequired`, `Pan` resets `minPointers`/`maxPointers`.

### What breaks

`GestureDetector` re-sends the full config on every update (`src/handlers/gestures/GestureDetector/updateHandlers.ts:68`), and the config only contains `numberOfPointers` when the gesture actually sets it. So a long press that stops requesting a multi pointer press keeps the stale requirement:

```jsx
const longPress = useLongPressGesture(
  twoFingerMode ? { numberOfPointers: 2 } : {}
);
```

After `twoFingerMode` flips back to `false`, the handler still requires 2 pointers. On web `tryActivate` returns early because `trackedPointersCount !== numberOfPointers` (`src/web/handlers/LongPressGestureHandler.ts:163`), so a normal one finger long press never activates again. The same holds for the v2 `Gesture.LongPress().numberOfPointers(2)` builder and the v1 `<LongPressGestureHandler numberOfPointers={2}>` prop.

The fix resets the value on all three platforms, using a named default in the two places that had a bare literal.

## Test plan

Added `src/web/handlers/__tests__/LongPressGestureHandler.test.ts`, following the existing `GestureHandler.test.ts` / `webNativeViewGestureHandler.test.ts` pattern. One test drives the regression (config with `numberOfPointers: 2`, then a config without it, then a single pointer press), and one control test confirms `numberOfPointers` still applies while it is in the config.

Counterfactual, run in this checkout. With the fix:

```
$ yarn jest src/web/handlers/__tests__/LongPressGestureHandler.test.ts
PASS src/web/handlers/__tests__/LongPressGestureHandler.test.ts
  LongPressGestureHandler config reset
    v a config without numberOfPointers restores the single pointer default (2 ms)
    v numberOfPointers still applies while it stays in the config
```

Then reverting only `src/web/handlers/LongPressGestureHandler.ts` and keeping the test:

```
$ git show HEAD:packages/.../src/web/handlers/LongPressGestureHandler.ts > packages/.../src/web/handlers/LongPressGestureHandler.ts
$ yarn jest src/web/handlers/__tests__/LongPressGestureHandler.test.ts
  x a config without numberOfPointers restores the single pointer default

    expect(received).toBe(expected) // Object.is equality

    Expected: 4
    Received: 2

    > 74 |     expect(handler.state).toBe(State.ACTIVE);

Tests: 1 failed, 1 passed, 2 total
```

`4` is `State.ACTIVE`, `2` is `State.BEGAN`: the handler stayed in `BEGAN` because it was still waiting for a second pointer. Restoring the file turns it green again. The control test passes in both directions, so the failure is specific to the reset.

Checks in this checkout:

* `yarn workspace react-native-gesture-handler test` -> 20 suites, 163 tests passing (162 before this change)
* `yarn workspace react-native-gesture-handler ts-check` -> clean
* `yarn eslint --ext '.js,.ts,.tsx' src/` -> 0 errors, 0 warnings on the touched files
* `yarn prettier --check './src/**/*.{js,jsx,ts,tsx}'` -> all matched files use Prettier code style
* `./android/gradlew -p android spotlessCheck -q` -> exit 0
* `clang-format --style=file` on `RNLongPressHandler.m` -> no diff on the changed lines

The Android and Apple changes mirror the web one and are not covered by the Jest suite; I did not run them on a device.

### DIFF
--- a/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
+++ b/packages/react-native-gesture-handler/android/src/main/java/com/swmansion/gesturehandler/core/LongPressGestureHandler.kt
@@ -29,13 +29,14 @@ class LongPressGestureHandler(context: Context) : GestureHandler() {
     val systemDefaultMaxDist = DEFAULT_MAX_DIST_DP * context.resources.displayMetrics.density
     defaultMaxDist = systemDefaultMaxDist
     maxDist = defaultMaxDist
-    numberOfPointersRequired = 1
+    numberOfPointersRequired = DEFAULT_NUMBER_OF_POINTERS_REQUIRED
   }
 
   override fun resetConfig() {
     super.resetConfig()
     minDurationMs = DEFAULT_MIN_DURATION_MS
     maxDist = defaultMaxDist
+    numberOfPointersRequired = DEFAULT_NUMBER_OF_POINTERS_REQUIRED
     shouldCancelWhenOutside = DEFAULT_SHOULD_CANCEL_WHEN_OUTSIDE
   }
 
@@ -211,5 +212,6 @@ class LongPressGestureHandler(context: Context) : GestureHandler() {
     private const val DEFAULT_SHOULD_CANCEL_WHEN_OUTSIDE = true
     private const val DEFAULT_MIN_DURATION_MS: Long = 500
     private const val DEFAULT_MAX_DIST_DP = 10f
+    private const val DEFAULT_NUMBER_OF_POINTERS_REQUIRED = 1
   }
 }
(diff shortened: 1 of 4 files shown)


j-piasecki (approved):
Thanks!

PR #4480, opened 2026-09-04
Reset `enableTrackpadTwoFingerGesture` in web `PanGestureHandler`

giaBaoJS:
## Description

On web, `PanGestureHandler.updateGestureConfig` writes `enableTrackpadTwoFingerGesture`, but `resetConfig` never restores it.

`setGestureConfig` is a full config replace (`resetConfig` then `updateGestureConfig`), and it is what `updateHandlers` / `useGesture` call whenever the gesture config changes. So once a pan gesture has been configured with `enableTrackpadTwoFingerGesture: true`, a later config that no longer carries the prop keeps two-finger trackpad panning enabled, and a wheel event from a trackpad still activates the gesture.

Every other field the web `PanGestureHandler` reads from the config is restored in `resetConfig`; this one was missed. iOS is already correct: `RNPanHandler`'s `resetConfig` sets `recognizer.allowedScrollTypesMask = 0`, which is the same property. Android does not support the prop, so this is web only.

## Test plan

Added `src/web/handlers/__tests__/PanGestureHandler.test.ts` with two cases:

- configure the handler with `enableTrackpadTwoFingerGesture: true`, then apply a config without it, and feed a touchpad wheel event: the handler must stay `UNDETERMINED`.
- configure it with `enableTrackpadTwoFingerGesture: true` and feed the same event: the handler must reach `ACTIVE`, so the flag itself keeps working.

Without the one-line change in `resetConfig`, the first test fails with `Expected: 0 / Received: 4` (the gesture activates from stale config). The second one passes both ways.

`yarn test`, `yarn lint-js` and `yarn ts-check` are green in `packages/react-native-gesture-handler`.

### DIFF
--- a/packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
@@ -170,6 +170,7 @@ export default class PanGestureHandler extends GestureHandler {
     this.maxPointers = DEFAULT_MAX_POINTERS;
 
     this.activateAfterLongPress = 0;
+    this.enableTrackpadTwoFingerGesture = false;
 
     this.hasCustomActivationCriteria = false;
   }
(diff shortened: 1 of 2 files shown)


j-piasecki (approved):
Thanks!

PR #4484, opened 2026-09-06
Reset accumulated wheel delta in web `WheelEventManager`

giaBaoJS:
## Description

`WheelEventManager` has no pointer to follow, so it synthesizes coordinates by accumulating `deltaX`/`deltaY` on top of each wheel event's client coordinates. Its `resetManager` override calls only `super.resetManager()` and never clears `wheelDelta`, unlike `PointerEventManager`, which clears its own bookkeeping there.

`resetManager` runs on every handler reset (`GestureHandler.reset` -> `delegate.reset` -> `manager.resetManager`), which the orchestrator triggers once a gesture reaches `END`. So when a trackpad pan ends, the whole scroll distance of that gesture stays in `wheelDelta`.

To reproduce, put a `Pan` gesture with `enableTrackpadTwoFingerGesture` on a view, then do two two-finger trackpad pans in a row without moving the cursor in between. The second gesture reports `absoluteX`/`absoluteY` offset by the first gesture's total scroll. A two-finger scroll does not move the cursor, so the `pointermove` listener that clears the delta does not necessarily fire, and the offset keeps growing with each gesture.

Fix is to clear `wheelDelta` in `resetManager`, matching what `PointerEventManager` already does.

## Test plan

Added `src/web/tools/__tests__/WheelEventManager.test.ts` with two cases:

- deltas still accumulate across wheel events within one gesture (`y` is 130 after deltas of 100 and 30), so the accumulation behaviour is not lost.
- after `resetManager`, the next wheel event reports only its own delta (`y` is 30, not 130).

The second test fails on `main` with `Expected: 30, Received: 130` and passes with the fix. `yarn jest src/web`, `yarn ts-check` and `eslint`/`prettier` on the touched files all pass.

### DIFF
--- a/packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts
@@ -334,14 +334,13 @@ export default class PanGestureHandler extends GestureHandler {
     }
   }
 
-  private scheduleWheelEnd(event: AdaptedEvent) {
+  private scheduleWheelEnd() {
     clearTimeout(this.endWheelTimeout);
 
     this.endWheelTimeout = setTimeout(() => {
       if (this.state === State.ACTIVE) {
         this.end();
-        this.tracker.removeFromTracker(event.pointerId);
-        this.state = State.UNDETERMINED;
+        this.reset();
       }
 
       this.wheelDevice = WheelDevice.UNDETERMINED;
@@ -363,7 +362,7 @@ export default class PanGestureHandler extends GestureHandler {
           : WheelDevice.MOUSE;
 
       if (this.wheelDevice === WheelDevice.MOUSE) {
-        this.scheduleWheelEnd(event);
+        this.scheduleWheelEnd();
         return;
       }
 
@@ -383,7 +382,7 @@ export default class PanGestureHandler extends GestureHandler {
     this.updateVelocity(event.pointerId);
 
     this.tryToSendMoveEvent(false, event);
-    this.scheduleWheelEnd(event);
+    this.scheduleWheelEnd();
   }
 
   private shouldActivate(): boolean {
--- a/packages/react-native-gesture-handler/src/web/handlers/__tests__/PanGestureHandler.test.ts
+++ b/packages/react-native-gesture-handler/src/web/handlers/__tests__/PanGestureHandler.test.ts
@@ -3,14 +3,23 @@ import { PointerType } from '../../../PointerType';
 import { State } from '../../../State';
 import type { AdaptedEvent } from '../../interfaces';
 import { EventTypes } from '../../interfaces';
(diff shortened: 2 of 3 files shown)


m-bert (changes requested):
Hi @giaBaoJS! Thank you for submitting this PR! Unfortunately seems that it is not enough. The good news is, replacing 

```tsx
        this.end();
        this.tracker.removeFromTracker(event.pointerId);
        this.state = State.UNDETERMINED;
      }
```

with 

```tsx
        this.end();
		this.reset();
      }
```

in [`PanGestureHandler`](https://github.com/software-mansion/react-native-gesture-handler/blob/ebc923b2bd4f66ed0108a6d8a64b3c80517c11b4/packages/react-native-gesture-handler/src/web/handlers/PanGestureHandler.ts#L342) does the job. 

The test passes as it calls `manager.resetManager` manually. Given that - I don't think we need this test file.

giaBaoJS:
Done, `scheduleWheelEnd` now calls `reset()`. Its `event` parameter became unused, so I dropped it.

The `WheelEventManager.resetManager` change is still needed alongside it. `reset()` reaches the wheel delta only through the delegate, and that path ends at `resetManager()`:

```ts
// src/web/tools/GestureHandlerWebDelegate.ts:184
reset(): void {
  this.eventManagers.forEach((manager: EventManager<unknown>) =>
    manager.resetManager()
  );
}
```

With only the `PanGestureHandler` change, `WheelEventManager.resetManager()` is still just `super.resetManager()` and `wheelDelta` survives. I measured each half separately against the new test: reverting either one leaves the next gesture's first event at `y: 130` instead of `30`.

On why `reset()` never ran before: the old block assigned `State.UNDETERMINED` synchronously, so by the time the orchestrator's `cleanupFinishedHandlers` microtask looked at the handler, `isFinished(handler.state)` was false and `cleanHandler` (which calls `handler.reset()`) skipped it.

`WheelEventManager.test.ts` is gone. The case now lives in the existing `web/handlers/__tests__/PanGestureHandler.test.ts`: it dispatches real `wheel` events at a real `WheelEventManager` attached via `handler.attachEventManager`, advances the 30ms end timer, then checks the coordinates the next gesture reports. Nothing in the test calls `resetManager`; the delegate stub forwards `reset()` to the attached managers the way `GestureHandlerWebDelegate` does.

One note on the swap itself: `reset()` does more than the two lines it replaces. `resetTracker()` clears every tracked pointer and the velocity tracker rather than one pointer, and `onReset()` clears a pending `activateAfterLongPress` timeout. Both look correct for a gesture that has just ended, and the wheel path only tracks the synthetic `-1` pointer anyway.

m-bert (approved):
> The WheelEventManager.resetManager change is still needed alongside it.

I'm aware, that's why I said "it is not enough", not that it is wrong 😄 

Thank you once again for this PR

PR #4486, opened 2026-09-07
[Web] Reset NativeViewGestureHandler config on a full config replace

j-piasecki:
## Description

The web NativeViewGestureHandler did not override resetConfig, so its props kept old values after setGestureConfig dropped them. It also forced shouldCancelWhenOutside to true in init, overriding an explicit false. The defaults now live in resetConfig, matching Android.

## Test plan

Added tests in `webNativeViewGestureHandler.test.ts`.


m-bert (approved):
Love the branch name

PR #4495, opened 2026-09-08
[iOS] Keep button hit testing within its own subtree

### LINKED ISSUE #4494: [iOS] Disabled Touchable crashes inside a SwiftUI-backed React Native blur view
### Description

Tapping the disabled area of a Gesture Handler Touchable inside a SwiftUI-backed React Native blur view crashes my iOS app with a native stack overflow.

Expected: the disabled area should ignore the tap without crashing, while an enabled nested Touchable remains usable.

React Native's experimental enableSwiftUIBasedFilters flag is explicitly enabled through a native override. The view uses filter: [{ blur: 0 }]; in this React Native version, radius 0 still creates the SwiftUI container.

Additional versions: Expo SDK 57 (expo 57.0.20), React 19.2.3, Reanimated 4.6.0, and Worklets 0.12.1.

### **Device observations**
Tapping the disabled outer area crashes.
Adding the native boundary guard below prevents the crash without disabling pointer events.
Setting pointerEvents="none" also prevents the crash.
An enabled nested Touchable responds, but tapping the surrounding disabled area crashes without the guard.
These observations are from the test screen in my original app. The linked repository extracts that screen and has built successfully on EAS; standalone device results have not yet been recorded.

### **Crash details**
The crash report contains EXC_BAD_ACCESS / SIGSEGV and Thread stack size exceeded due to excessive recursion.

The stack includes:

Repeated SwiftUI _UIHostingView._hitTest(with:) calls
-[RNGestureHandlerButton shouldHandleTouch:atPoint:]
-[RNGestureHandlerButton hitTest:withEvent:]
-[RNGestureHandlerButtonComponentView hitTest:withEvent:]
React Native RCTViewComponentView hit testing

### Suspected cause and proposed fix

The hit-test loop can walk beyond the disabled button into its ancestors. I suspect returning a SwiftUI hosting ancestor causes the recursive hit testing.

Stopping at the button boundary prevents the crash on my device:

```objc
while (inner && ![self shouldHandleTouch:inner atPoint:point]) {
  if (inner == self) {
    return nil;
  }
  inner = inner.superview;
}
```
Eligible child controls can still be found before reaching the boundary, which appears consistent with the nested-button behaviour introduced in https://github.com/software-mansion/react-native-gesture-handler/pull/1991

**AI assistance**
Used AI (Codex) to help investigate the native code, create the reproduction project, and prepare this report and proposed patch. I personally tested the behaviour on my iPhone. May need further investigation

### Steps to reproduce

**1.** Follow the reproduction repository's instructions to build and install the iOS Debug development client. The repository includes the native SwiftUI filter flag override.

**2.** Start Metro and open the app. Select Original case, leaving Pointer-events workaround off.

**3.** Select Enable test area and tap the disabled outer Touchable.

**4.** Reopen the app and select Patched case. Leave the pointer-events workaround off, enable the test area, and tap the same area to compare.

**5.** To check nesting, turn on Enabled inner Touchable and re-enable the test area. Compare taps on the inner button and the disabled outer area around it.

https://github.com/rileysay/rngh-swiftui-hit-test-repro

### Gesture Handler version

3.2.1

### React Native version

0.86.3

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

### Architecture

New Architecture (Fabric)

### Build type

Debug mode

### Device

Real device

### Device model

Iphone 16 Pro

### Acknowledgements

Yes

rileysay:
## Description

Fixes #4494.

When a disabled `RNGestureHandlerButton` has no eligible touch target, the hit-test loop can continue past the button and return an ancestor outside its subtree.

This change returns `nil` when the search reaches the button itself without finding an eligible target. Eligible descendants can still be selected before reaching that boundary.

The guard prevents the SwiftUI-backed blur crash in my device testing. The suspected connection between returning a hosting ancestor and recursive hit testing is described in the linked issue.

## Test plan

Tested the same boundary guard in my app’s test screen on an iPhone 16 Pro running iOS 26.5.2, using an Expo SDK 57 development build with React Native 0.86.3 and Gesture Handler 3.2.1.

React Native’s experimental `enableSwiftUIBasedFilters` flag was enabled.

- Original behaviour: tapping the disabled outer area crashes.
- Boundary guard enabled, pointer-events workaround off: tapping the same area no longer crashes.

Reproduction project and build instructions:
https://github.com/rileysay/rngh-swiftui-hit-test-repro

The standalone project has built successfully on EAS. The device observations above were recorded using the test screen in my original app.

AI assistance: I used Codex to help investigate the native code, prepare the reproduction project, and propose this patch. I personally verified the device observations above.

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerButton.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerButton.mm
@@ -1193,6 +1193,11 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
 
 - (void)mouseDown:(NSEvent *)event
 {
+  if (!_userEnabled) {
+    return;
+  }
+
+  
   _isTouchInsideBounds = YES;
   [self handleAnimatePressIn];
   [super mouseDown:event];
@@ -1200,6 +1205,10 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
 
 - (void)mouseUp:(NSEvent *)event
 {
+  if (!_userEnabled) {
+    return;
+  }
+  
   NSPoint locationInView = [self convertPoint:[event locationInWindow] fromView:nil];
   _isHovered = NSPointInRect(locationInView, self.bounds);
   [self recordHoverSampleForMouseEvent:event];
@@ -1212,6 +1221,10 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
 
 - (void)mouseDragged:(NSEvent *)event
 {
+  if (!_userEnabled) {
+    return;
+  }
+  
   NSPoint locationInWindow = [event locationInWindow];
   NSPoint locationInView = [self convertPoint:locationInWindow fromView:nil];
   BOOL currentlyInside = NSPointInRect(locationInView, self.bounds);
@@ -1244,6 +1257,10 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
 
 - (BOOL)beginTrackingWithTouch:(UITouch *)touch withEvent:(UIEvent *)event
 {
+  if (!_userEnabled) {
+    return NO;
+  }
+  
   _isTouchInsideBounds = YES;
   // A pencil's hover-out arrives just before touch-down but only schedules the
   // clear, so `_isHovered` still reflects the open hover. Under Reduce Motion
@@ -1469,10 +1486,7 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
   }
 
   RNGHUIView *inner = [super hitTest:point withEvent:event];
-  while (inner && ![self shouldHandleTouch:inner atPoint:point]) {
-    if (inner == self) {
-      return nil;
-    }
+  while (inner && inner != self && ![self shouldHandleTouch:inner atPoint:point]) {
     inner = inner.superview;
   }
   return inner;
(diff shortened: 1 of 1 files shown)


m-bert (changes requested):
Hi @rileysay! Thank you for this PR!

The change seems ok, but there's one problem - it now allows siblings of disabled buttons to take over the touch, even though they shouldn't. 

<details>
<summary>Here is the example of the failing structure</summary>

```tsx
 <View style={{ height: 120 }}>
	<Touchable
	  onPress={() => console.log('outer')}
	  style={{
	    position: 'absolute',
	    inset: 0,
	    backgroundColor: '#9fd8a8',
	  }}
	/>
	<Touchable
	  disabled
	  style={{
	    position: 'absolute',
	    left: 40,
	    top: 20,
	    width: 200,
	    height: 80,
	    backgroundColor: '#9a9a9a',
	  }}
	/>
</View>
```

</details>

To fix this, we can slightly change the approach:

1. Return `self` instead of `nil`, so that button still can become a touch target:

```diff
- while (inner && ![self shouldHandleTouch:inner atPoint:point]) {
+ while (inner && inner != self && ![self shouldHandleTouch:inner atPoint:point]) {
```

2. Early return from [`beginTrackingWithTouch:withEvent:`](https://github.com/software-mansion/react-native-gesture-handler/blob/a95f00f9605cdf2503647dc60ffa380b8cc02e44/packages/react-native-gesture-handler/apple/RNGestureHandlerButton.mm#L1245) when button is disabled:

```obj-c
- (BOOL)beginTrackingWithTouch:(UITouch *)touch withEvent:(UIEvent *)event
{
  if (!_userEnabled) {
    return NO;
  }
  ...
```

3. Early return from point 2 should also be applied to `mouseDown`, `mouseUp` and `mouseDragged` methods to keep `macOS` behavior correct.

After that everything should work fine 😅

rileysay:
@m-bert Thanks for the review and the example. I’ve applied the suggested changes and retested the revised patch on iOS and it fixed the sibiling issue. 

I’ve also added the guards to the macOS mouse methods, but I don’t have a Mac, so I’ll leave that part to the big leagues to test 😅

j-piasecki (approved):
LGTM! I'll leave it up for @m-bert

m-bert (approved):
LGTM! I do agree with coderabbit about the macOS part, but that's what we will cover in a follow-up. Changes in this are sufficient to fix the issue.

Thank you again! ❤️

PR #4506, opened 2026-09-09
[iOS] Preserve disabled button targets under box-none parents

rileysay:
## Description

Follow-up to #4495.

When an RNGH `Touchable`/`Pressable` uses `pointerEvents="box-none"`, a disabled
RNGH button returned from one of its child hit tests is rejected by
`shouldHandleTouch:` because `userEnabled` is false.

The `box-none` branch then continues checking lower siblings, which allows an
enabled sibling underneath the disabled control to receive the touch.

This is inconsistent with the behavior preserved in #4495, where a disabled
button can still remain the hit target so controls underneath don't take over
the touch, while the disabled button itself doesn't activate.

The `box-none` path now preserves `RNGestureHandlerButton` child hits instead
of discarding them based on their enabled state.

## Device observations

Reproduced on iOS with both RNGH `Touchable` and RNGH `Pressable`.

With an RNGH parent using `pointerEvents="box-none"`:

- tapping the disabled foreground triggers the enabled sibling underneath
- tapping the exposed background also triggers the background, as expected

Changing only the parent to a React Native `View` with the same
`pointerEvents="box-none"`:

- tapping the disabled foreground does nothing
- tapping the exposed background triggers the background

<details>
<summary>Test example</summary>

```tsx
import {
  Pressable as RNPressable,
  StyleSheet,
  Text,
  View,
} from 'react-native';
import {
  Pressable as RNGHPressable,
  Touchable,

type ButtonType = 'touchable' | 'pressable';

export default function BoxNoneDisabledTargetExample() {
  const [buttonType, setButtonType] = useState<ButtonType>('touchable');
  const [useRNParent, setUseRNParent] = useState(false);
  const [backgroundPresses, setBackgroundPresses] = useState(0);

  const Button =
    buttonType === 'touchable'
      ? Touchable
      : RNGHPressable;

  const children = (
    <>
      <Button
        onPress={() => setBackgroundPresses((count) => count + 1)}
        style={[StyleSheet.absoluteFillObject, styles.background]}>
        <Text>Enabled background</Text>
      </Button>

      <Button
        disabled
        style={styles.disabledForeground}>
        <Text>Disabled foreground</Text>
      </Button>
    </>
  );

  return (
      <RNPressable
        onPress={() =>
          setButtonType((current) =>
            current === 'touchable' ? 'pressable' : 'touchable'
          )
        }>
        <Text>Button type: {buttonType}</Text>
      </RNPressable>

      <RNPressable onPress={() => setUseRNParent((value) => !value)}>
        <Text>
          Parent: {useRNParent ? 'React Native View' : 'RNGH button'}
        </Text>
      </RNPressable>

      {useRNParent ? (
        <View pointerEvents="box-none" style={styles.frame}>
          {children}
        </View>
      ) : (
        <Touchable pointerEvents="box-none" style={styles.frame}>
          {children}
        </Touchable>
      )}

      <Text>Background presses: {backgroundPresses}</Text>
    </View>
  );
}

  container: {
    padding: 16,
    gap: 12,
  },
  frame: {
    height: 150,
  },
  background: {
    backgroundColor: '#9fd8a8',
    justifyContent: 'flex-end',
    padding: 8,
  },
  disabledForeground: {
    position: 'absolute',
    left: 40,
    right: 24,
    top: 20,
    height: 85,
    backgroundColor: '#9a9a9a',
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

</details>

## Test plan

Test on a physical iOS device with both RNGH `Touchable` and RNGH `Pressable`.

For each component:

1. Use an RNGH parent with `pointerEvents="box-none"`.
2. Tap well inside the disabled foreground.
3. Verify the enabled background does **not** receive `onPress`.
4. Tap the exposed background area.
5. Verify the background does receive `onPress`.
6. Replace only the parent with a React Native `View` using
   `pointerEvents="box-none"` and repeat.

Expected:

| Parent | Disabled foreground | Exposed background |
| --- | --- | --- |
| RNGH `box-none` parent | no press | background press |
| RN `View` `box-none` parent | no press | background press |

Before this change, the RNGH-parent case incorrectly sends the disabled
foreground tap to the background sibling.

## AI assistance

Used AI to help inspect the surrounding hit-test paths and prepare the test
case and patch. I personally reproduced the issue on device.

### DIFF
--- a/packages/react-native-gesture-handler/apple/RNGestureHandlerButton.mm
+++ b/packages/react-native-gesture-handler/apple/RNGestureHandlerButton.mm
@@ -1197,7 +1197,6 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
     return;
   }
 
-  
   _isTouchInsideBounds = YES;
   [self handleAnimatePressIn];
   [super mouseDown:event];
@@ -1208,7 +1207,7 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
   if (!_userEnabled) {
     return;
   }
-  
+
   NSPoint locationInView = [self convertPoint:[event locationInWindow] fromView:nil];
   _isHovered = NSPointInRect(locationInView, self.bounds);
   [self recordHoverSampleForMouseEvent:event];
@@ -1224,7 +1223,7 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
   if (!_userEnabled) {
     return;
   }
-  
+
   NSPoint locationInWindow = [event locationInWindow];
   NSPoint locationInView = [self convertPoint:locationInWindow fromView:nil];
   BOOL currentlyInside = NSPointInRect(locationInView, self.bounds);
@@ -1260,7 +1259,7 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
   if (!_userEnabled) {
     return NO;
   }
-  
+
   _isTouchInsideBounds = YES;
   // A pencil's hover-out arrives just before touch-down but only schedules the
   // clear, so `_isHovered` still reflects the open hover. Under Reduce Motion
@@ -1426,7 +1425,8 @@ static CATransform3D RNGHCenterScaleTransform(NSRect bounds, CGFloat scale)
 {
   if ([view isKindOfClass:[RNGestureHandlerButton class]]) {
     RNGestureHandlerButton *button = (RNGestureHandlerButton *)view;
-    return button.userEnabled;
+    return button.userEnabled && button.pointerEvents != RNGestureHandlerPointerEventsBoxNone &&
+        button.pointerEvents != RNGestureHandlerPointerEventsNone;
   }
 
   // Certain subviews such as RCTViewComponentView have been observed to have disabled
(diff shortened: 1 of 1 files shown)


rileysay:
@m-bert  I believe this is all good? 🫡

m-bert:
HI @rileysay! I'm currently investigating one edge case that may be a consequence of this PR, I'll let you know once we decide 😅

m-bert:
Hi @rileysay! I'm back and I think I have what we need!

So the edge case I was talking about looks like this. Given the nested buttons structure, let's say `A -> B -> C`, where `B` has pointer events set to `box-none`, currently clicking on the `C` results in press being passed to `A`. After changes in this PR, presses are passed to `B`. I'd say that this is wrong as `B` has `box-none`, but feel free to correct me if you think otherwise. 

<details>
<summary>Here you can see example structure:</summary>

```tsx
<Touchable onPress={() => log('A')}>
  <Touchable pointerEvents="box-none" onPress={() => log('B')}>
    <Touchable disabled onPress={() => log('C')}>
      <Text>C</Text>
    </Touchable>
  </Touchable>
</Touchable>
```

</details>

Seems like this simple change in [`RNGestureHandlerButton`](https://github.com/software-mansion/react-native-gesture-handler/blob/c5db1fcfffa18a61835f377ff3874c14cfa2dcab/packages/react-native-gesture-handler/apple/RNGestureHandlerButton.mm#L1457) should be enough:

```diff
  if ([view isKindOfClass:[RNGestureHandlerButton class]]) {
    RNGestureHandlerButton *button = (RNGestureHandlerButton *)view;

-   return button.userEnabled;
+   return button.userEnabled && button.pointerEvents != RNGestureHandlerPointerEventsBoxNone &&
+       button.pointerEvents != RNGestureHandlerPointerEventsNone;
  }

```

I'd be glad if you could confirm that it works 😅 Also remember to run `yarn format:apple`

rileysay:
@m-bert Ah yep, you’re right 😅 I cooked up a little pointer-events matrix and indeed see the edge-case. I’ve applied your suggested change and don’t see any issues with it.

m-bert (approved):
Thank you! ❤️

PR #4508, opened 2026-09-09
fix: stop circular GestureStateManagerType import on web

roryabraham:
## Description

`moduleSuffixes: [".web", ""]` makes `import type { GestureStateManagerType } from './gestureStateManager'` in `gestureStateManager.web.ts` resolve to this file itself. That circular self-import left the type unresolvable (`any`), so every `on*` gesture callback's `stateManager` parameter was untyped. Consumers (e.g. Expensify App) then hit `@typescript-eslint/no-unsafe-call` on `state.activate()` / `state.fail()`.

A `./gestureStateManager.ts` extension import would skip the suffix, but this package's tsconfig does not set `allowImportingTsExtensions`, so the type is declared locally in the web file to match the native `GestureStateManagerType`.

Related consumer patch: https://github.com/Expensify/App/pull/99495

## Test plan

1. Type-only change — no runtime behavior change.
2. `tsc` with `moduleSuffixes: [".web", ""]` should resolve `GestureStateManagerType` from the web file instead of collapsing it to `any`.
3. Gesture `on*` callbacks should type `state.activate()` / `state.fail()` / `state.begin()` / `state.end()` without `no-unsafe-call`.


j-piasecki:
I took the liberty of also updating the v3 implementation. Since v2 is not that well organized, I think that the duplicated type can stay there. I'll leave it for @m-bert to double-check.

m-bert (approved):
> I took the liberty of also updating the v3 implementation.

And I took the liberty of renaming the file to match convention in this directory 😅 (cc @j-piasecki)

Other than that looks good! ✅

roryabraham:
sorry for the slop PR. I was moving quickly and it was a type-only change to fix circular import issues, so I just sent it. Thanks for polishing it up for me.

PR #4510, opened 2026-09-11
[iOS] add SPM build support

kacperzolkiewski:
## Description

This PR adds

- support for Swift Package Manager on iOS
- ios-spm-build-test.yml workflow to test SPM build on CI

CocoaPods support is preserved.

### CocoaPods

Run basic-example app and test if it builds and works properly.

### SPM

Manual verification with a blank RN **0.87.x** app:

1. `npx @react-native-community/cli@latest init MyApp`
2. `cd packages/react-native-gesture-handler && yarn && yarn build && npm pack`
3. Go to `MyApp` project
4. copy generated `react-native-gesture-handler-2.29.0.tgz` into it
5. add `react-native-gesture-handler: './react-native-gesture-handler-2.29.0.tgz'` dependency inside `package.json`
6. `npm install`
7. `npx react-native spm scaffold --deintegrate --yes`
8. `npm run ios`
9. MyApp should build and work properly.
10. You can add some basic `react-native-gesture-handler` usage to AppContent to check if everything is linked properly at runtime.

#### Worklets + GH test
11. install `"react-native-worklets": "0.13.0-nightly-20260913-5336bb7f4"`
12. repeat steps 7-8
13. You can add some basic `react-native-worklets` usage to AppContent to check if everything is linked properly at runtime.

Build iOS (SPM) ci job should pass

### DIFF
--- a/.github/workflows/ios-spm-build-test.yml
+++ b/.github/workflows/ios-spm-build-test.yml
@@ -1,0 +1,97 @@
+name: Build iOS (SPM)
+
+on:
+  pull_request:
+    paths:
+      - packages/react-native-gesture-handler/Package.swift
+      - packages/react-native-gesture-handler/package.json
+      - packages/react-native-gesture-handler/react-native.config.js
+      - packages/react-native-gesture-handler/package.json
+      - packages/react-native-gesture-handler/RNGestureHandler.podspec
+      - packages/react-native-gesture-handler/apple/**
+      - packages/react-native-gesture-handler/shared/**
+      - packages/react-native-gesture-handler/src/**
+      - rnrepo.config.json
+      - yarn.lock
+      - '!packages/react-native-gesture-handler/src/**/*.test.*'
+      - .github/workflows/ios-spm-build-test.yml
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+concurrency:
+  group: ios-spm-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ios-spm-isolated:
+    if: github.repository == 'software-mansion/react-native-gesture-handler'
+
+    runs-on: macos-26-xlarge
+    timeout-minutes: 60
+
+    env:
+      APP_NAME: GHSpmTest
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+
+      - name: Select Xcode
+        run: |
+          XCODE_APP="/Applications/Xcode_26.4.1.app"
+          if [ ! -d "$XCODE_APP" ]; then
+            echo "Xcode 26.4.1 is not installed on this runner. Available:" >&2
+            ls -d /Applications/Xcode*.app >&2
+          fi
+          echo "Using $XCODE_APP"
+          echo "DEVELOPER_DIR=$XCODE_APP/Contents/Developer" >> "$GITHUB_ENV"
+          DEVELOPER_DIR="$XCODE_APP/Contents/Developer" xcodebuild -version
+
+      - name: Use Node.js 24
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: yarn
+
+      - name: Read React Native version
+        id: rn-version
+        run: |
+          RN_VERSION=$(node -p "require('./packages/react-native-gesture-handler/package.json').devDependencies['react-native']")
+          echo "version=$RN_VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Install monorepo dependencies
+        run: yarn --immutable
+
+      - name: Build and pack library
+        working-directory: packages/react-native-gesture-handler
+        run: |
+          yarn build
+          mkdir -p /tmp/packs
+          npm pack --pack-destination /tmp/packs
+          echo "GH_TARBALL=$(ls /tmp/packs/react-native-gesture-handler-*.tgz)" >> "$GITHUB_ENV"
+
+      - name: Create blank RN app
+        run: |
+          npx @react-native-community/cli@latest init "${{ env.APP_NAME }}" \
+            --version "${{ steps.rn-version.outputs.version }}"
+
+      - name: Install GH from tarball
+        working-directory: ${{ env.APP_NAME }}
+        run: npm install "${{ env.GH_TARBALL }}"
+
+      - name: SPM scaffold
+        working-directory: ${{ env.APP_NAME }}
+        run: npx react-native spm scaffold --deintegrate --yes
+
+      - name: Build iOS (SPM)
+        working-directory: ${{ env.APP_NAME }}/ios
+        run: |
+          xcodebuild build \
+            -project "${{ env.APP_NAME }}.xcodeproj" \
+            -scheme "${{ env.APP_NAME }}" \
+            -configuration Debug \
+            -sdk iphonesimulator \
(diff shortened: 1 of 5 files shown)


.github/workflows/ios-spm-build-test.yml
```diff
@@ -0,0 +1,173 @@
...
+
+concurrency:
+  group: ios-spm-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ios-spm-isolated:
+    if: github.repository == 'software-mansion/react-native-gesture-handler'
+
+    runs-on: macos-26-xlarge
```
m-bert: Do we need `xlarge`? Those are not free and I'd like to avoid adding them if unnecessary

.github/workflows/ios-spm-build-test.yml
```diff
@@ -0,0 +1,173 @@
...
+            -scheme "${{ env.APP_NAME }}" \
+            -configuration Debug \
+            -sdk iphonesimulator \
+            -destination 'generic/platform=iOS Simulator' \
+            -quiet
+
+  ios-spm-with-worklets:
+    if: github.repository == 'software-mansion/react-native-gesture-handler'
+
+    runs-on: macos-26-xlarge
```
m-bert: Same here

.github/workflows/ios-spm-build-test.yml
```diff
@@ -0,0 +1,173 @@
...
+        run: |
+          xcodebuild build \
+            -project "${{ env.APP_NAME }}.xcodeproj" \
+            -scheme "${{ env.APP_NAME }}" \
+            -configuration Debug \
+            -sdk iphonesimulator \
+            -destination 'generic/platform=iOS Simulator' \
+            -quiet
+
+  ios-spm-with-worklets:
```
m-bert: Actually, do we need this? I think the first one is enough. I'm not against having it, I just don't think we need another CI for this specific case (cc @j-piasecki)

.github/workflows/ios-spm-build-test.yml
```diff
@@ -0,0 +1,173 @@
...
+      - name: Build iOS (SPM)
+        working-directory: ${{ env.APP_NAME }}/ios
+        run: |
+          xcodebuild build \
+            -project "${{ env.APP_NAME }}.xcodeproj" \
+            -scheme "${{ env.APP_NAME }}" \
+            -configuration Debug \
+            -sdk iphonesimulator \
+            -destination 'generic/platform=iOS Simulator' \
+            -quiet
```
m-bert:
```suggestion
            -quiet

```

New line 😅

packages/react-native-gesture-handler/include/.gitkeep
m-bert, line 1: Is this the only way to handle the build failure?

.github/workflows/ios-spm-build-test.yml
m-bert: Also if we decide to keep it, wouldn't using 

```yml
matrix:
        worklets: [false, true]
```

be cleaner approach?

packages/react-native-gesture-handler/Package.swift
```diff
@@ -0,0 +1,83 @@
...
+import Foundation
+import PackageDescription
+
+let packageDirectory = Context.packageDirectory
+
+func rnWorkletsPackageExists() -> Bool {
+    let url = URL(fileURLWithPath: packageDirectory)
+        .appendingPathComponent("../RNWorklets/Package.swift")
+    return FileManager.default.fileExists(atPath: url.path)
+}
```
m-bert, line 13: During my investigation with Claude I've found out that using `'../RNWorklets/Package.swift'` is wrong and it should use `.deletingLastPathComponent()` instead:

```swift
func rnWorkletsPackageExists() -> Bool {
    let url = URL(fileURLWithPath: packageDirectory)
        .deletingLastPathComponent()
        .appendingPathComponent("RNWorklets/Package.swift")
    return FileManager.default.fileExists(atPath: url.path)
}
```

otherwise Worklets are not correctly linked and using StateManager throws error. Could you look more into it?

.github/workflows/ios-spm-build-test.yml
m-bert: And yet another thing that may be useful if we decide to keep that, maybe we should add one block that actually checks if Worklets were linked correctly? For more context check my `Package.Swift` comment

```yml
- name: Assert worklets are linked into RNGestureHandler
  working-directory: ${{ env.APP_NAME }}
  run: |
    swift package --package-path ios/build/generated/autolinking show-dependencies --format dot \
      | grep 'libs/RNGestureHandler" -> ".*libs/RNWorklets"'
```

packages/react-native-gesture-handler/Package.swift
```diff
@@ -0,0 +1,83 @@
...
+
+let useWorklets = rnWorkletsPackageExists()
```
m-bert, line 15: On main we also check for Worklets version, I assume the drop is intentional, right?

m-bert (commented):
Thank you ❤️ 

Left few comments, please address them

packages/react-native-gesture-handler/Package.swift
kacperzolkiewski: Interesting. I tested it in a blank app with Worklets, and it builds and works properly. I'll look into that.

.github/workflows/ios-spm-build-test.yml
kacperzolkiewski: Yes you're right, there is no need to use xlarge, fixed

.github/workflows/ios-spm-build-test.yml
kacperzolkiewski: Done :D

packages/react-native-gesture-handler/include/.gitkeep
kacperzolkiewski: With the empty `include` directory, we're aligned with SPM's default behavior, so I don't think there's anything wrong with it. I tried setting `apple` as the `publicHeadersPath`, for example, but then I got an error like this:
`umbrella header found at .../apple/RNGestureHandler.h, but directories exist next to it: .../Handlers, .../RNGestureHandler.xcodeproj`

packages/react-native-gesture-handler/Package.swift
kacperzolkiewski: Yes that is intentional. I'm not sure if there's a reason to check whether Worklets is` >= 0.8.0`, because even if we use Gesture Handler with an older version of Worklets, the build would still fail since Worklets doesn't support SPM yet. SPM support will probably be available in one of the next versions, 0.13.0 or 0.12.3 so it is deffinitely higher than 0.8.0.

packages/react-native-gesture-handler/Package.swift
kacperzolkiewski: Nice catch! It worked when I was using features directly from `react-native-worklets`, but when using something like `StateManager` that is using worklets on its own, as you pointed out, it was broken. Fixed, thanks!

.github/workflows/ios-spm-build-test.yml
kacperzolkiewski: I applied suggested changes with matrix and linking check

m-bert (approved):
Thank you ❤️

PR #4518, opened 2026-09-15
[Web] Ignore repeated keydown while an activation key is held

giaBaoJS:
## Description

On web, `KeyboardEventManager` translates Enter/Space presses into pointer events so gestures can be driven from the keyboard.

`onKeyUp` guards on `isPressed`, so a key release emits at most one pointer up. `keyDownCallback` has no matching guard, so it emits a pointer down for *every* `keydown` it sees:

- Browsers auto-repeat `keydown` while a key is held. Holding Enter on a view with a gesture attached produces one pointer down per repeat (roughly 30 per second after the initial delay).
- The two activation keys are independent. Pressing Enter and then Space while Enter is still down produces two pointer downs, and the following key release produces a single pointer up.

Minimal repro on `main`: focus a view with a `Tap` gesture on web and hold Enter. `TapGestureHandler.updateState` takes the `State.BEGAN` branch on each repeated down and calls `startTap()`, which clears the pending timeout and schedules a fresh `maxDurationMs` fail timer. The tap therefore never fails on duration no matter how long the key is held, and it activates on release.

Separately, `isPressed` survives `resetManager()`. A handler disabled mid-press resets and unregisters its managers; once re-enabled, the eventual key release still passes the `!this.isPressed` guard and emits a pointer up with no matching pointer down.

## Why it happens

`packages/react-native-gesture-handler/src/web/tools/KeyboardEventManager.ts` on current `main`:

- L27-L40 `keyDownCallback` dispatches `EventTypes.DOWN` for any activation key with no check on `isPressed`, unlike `onKeyUp` at L43-L51.
- `resetManager` is not overridden, so `isPressed` is never cleared. `PointerEventManager` (L242) and `WheelEventManager` both override it to clear their own per-gesture state.

## What the fix does

Replaces the `isPressed` flag with the set of currently held activation keys:

- `keyDownCallback` only dispatches a pointer down when the set is empty; otherwise it records the key and returns. The check sits after the `Tab` cancelation branch so Tab still cancels an ongoing press.
- `onKeyUp` dispatches a pointer up only for a key that is actually in the set, and only once the set is empty. A flag alone is not enough here: Enter down, Space down, Space up would clear it and let the still held Enter reopen the gesture on its next repeat.
- `resetManager` clears the set, matching `PointerEventManager` and `WheelEventManager`.

No public API or config changes.

## Test plan

New `src/web/tools/__tests__/KeyboardEventManager.test.ts`, run by the existing `run-jest-tests` workflow. It builds the minimal DOM surface the manager touches (the same pattern as `webNativeViewGestureHandler.test.ts`) and covers:

- three `keydown`s for a held Enter produce one pointer down (3 before the fix)
- a release lets the next press start a new gesture (2 downs, 1 up)
- Space pressed and released while Enter is held emits no pointer up, Enter's next repeat emits no second pointer down, and releasing Enter emits exactly one pointer up (2 downs and an early up before the fix)
- Tab still cancels an ongoing press
- after `resetManager()` a key release emits no pointer up, and the next press starts a fresh gesture (a stray up before the fix)

Package suite: 177 passed before, 182 after. `yarn lint-js` and `yarn ts-check` clean.

## Boundary

`isPressed` is cleared by the `Tab` cancelation branch, so a key that is still physically held after a Tab cancel will start a new gesture on its next auto-repeat. Repro: hold Enter, press Tab, keep holding Enter. Distinguishing that from a genuine new press needs `KeyboardEvent.repeat`, which is a separate behaviour change, so I left it alone.


m-bert:
Hi! We decided to take another approach and change `KeyboardEventManager` semantics a bit. Could you look at #4519 and confirm that it solves issue you've found? 😅

giaBaoJS:
Yes, #4519 covers every case my tests cover, and it closes a gap I had left open.

Checked against the three scenarios in #4518:

1. Held `Enter`. The first keydown sets `keyPointer`, every auto-repeat returns at the `keyPointer !== null` branch, and `event.repeat` guards it a second time. One pointer down.
2. `Enter` down, `Space` down, `Space` up. The `Space` keydown returns because `keyPointer` is already set, and the `Space` keyup does not match `keyPointer`, so the press survives. This is the case I had to fix in a second revision of #4518.
3. `resetManager()` clears `keyPointer`, so a handler reset mid press no longer leaks a later pointer up.

It also fixes the boundary I named as out of scope in #4518: hold `Enter`, press `Tab`, keep holding `Enter`. After the cancel, `event.repeat` stops the still held key from opening a new gesture. That needed `KeyboardEvent.repeat`, which I had deliberately left alone, so #4519 is the better change.

One behaviour difference worth putting on the record, in case it is not intended. With a single `keyPointer`, holding `Enter`, then `Space`, then releasing `Enter` sends the pointer up while `Space` is still physically held. Tracking the set of held activation keys keeps it down until both are released. The single pointer is the more defensible model to me, since one gesture should own one pointer, so I am not asking for a change.

#4519 carries no tests. #4518 has five that cover the three cases above plus `Tab` cancelation and plain keyup delivery, and each one is red on current `main` for its own behavioural reason. If it is useful, I can reduce #4518 to just the test file rebased on #4519, so the semantics change lands with coverage. Tell me which you prefer.

m-bert:
Thank you for your response! In that case I'm closing this PR in favor of #4519

PR #4526, opened 2026-09-17
[Web] Detect composed gesture changes in the detector

m-bert:
## Description

The web `HostGestureDetector` memoizes its handler-tag set with the tags spread into the dependency array. React compares dependencies only up to the shorter length, so a tag appended or removed at the end of `handlerTags` was never seen. Switching a detector from a single gesture to a composition containing it (`gesture={cond ? composed : single}`) turns `[pan]` into `[pan, tap]`: the tap was never attached and never fired. Switching back left the tap attached in the web engine. React also logged "The final argument passed to useMemo changed size between renders" on every switch.

The memo is now keyed on the tags joined into a string, so any change in membership, including a change in length, rebuilds the set and resyncs the subscriptions, while unchanged tags keep the same set across renders.

## Test plan

<details>

```tsx
import {
  GestureDetector,
  Pressable,
  usePanGesture,
  useSimultaneousGestures,
  useTapGesture,

import { COLORS } from '../../../common';

// Web only. Each box switches its detector between a single pan and a
// composition of the same pan plus a tap. Both hooks are always called, only
// the value handed to the detector changes, which turns the detector's
// `handlerTags` from `[pan]` into `[pan, tap]` and back.
//
// Expected: with the composition the box counts taps, with the single pan it
// does not. Bug: the web detector memoizes its tag set on the spread array and
// React compares only the common prefix when the length changes, so:
// - box A (starts as pan only): the tap never attaches after switching to
//   the composition,
// - box B (starts as pan + tap): the tap stays attached after switching back
//   to the single pan.

function Case({
  name,
  initialWithTap,
}: {
  name: string;
  initialWithTap: boolean;
}) {
  const [withTap, setWithTap] = useState(initialWithTap);
  const [panCount, setPanCount] = useState(0);
  const [tapCount, setTapCount] = useState(0);

  const pan = usePanGesture({
    activeOffsetX: [-10, 10],
    onActivate: () => {
      setPanCount((n) => n + 1);
    },
  });

  const tap = useTapGesture({
    onActivate: () => {
      setTapCount((n) => n + 1);
    },
  });

  const composed = useSimultaneousGestures(pan, tap);

  return (
    <View style={styles.caseContainer}>
      <Text style={styles.label} testID={`${name}-status`}>
        {name}: {withTap ? 'pan + tap' : 'pan only'} (taps expected to{' '}
        {withTap ? 'count' : 'be ignored'})
      </Text>
      <Text style={styles.label} testID={`${name}-counters`}>
        pan activations: {panCount}, tap activations: {tapCount}
      </Text>

      <Pressable
        style={styles.button}
        onPress={() => setWithTap((v) => !v)}
        testID={`${name}-toggle`}>
        <Text style={styles.buttonLabel}>
          {withTap ? 'Use pan only' : 'Use pan + tap'}
        </Text>
      </Pressable>

      <GestureDetector gesture={withTap ? composed : pan}>
        <View style={styles.box} testID={`${name}-box`}>
          <Text style={styles.boxLabel}>
            {name}:{' '}
            {withTap ? 'pan + tap, tap or drag here' : 'pan only, drag here'}
          </Text>
        </View>
      </GestureDetector>
    </View>
  );
}

export default function WebComposedMembershipExample() {
  return (
    <View style={styles.content}>
      <Case name="A" initialWithTap={false} />
      <Case name="B" initialWithTap />
    </View>
  );
}

  content: {
    flex: 1,
    flexDirection: 'row',
    justifyContent: 'space-evenly',
    padding: 16,
  },
  caseContainer: {
    alignItems: 'center',
  },
  label: {
    fontSize: 14,
    color: COLORS.NAVY,
    marginVertical: 4,
    textAlign: 'center',
  },
  button: {
    marginVertical: 8,
    paddingHorizontal: 20,
    paddingVertical: 10,
    borderRadius: 8,
    backgroundColor: COLORS.PURPLE,
  },
  buttonLabel: {
    color: 'white',
    fontSize: 16,
  },
  box: {
    marginTop: 12,
    width: 260,
    height: 200,
    borderRadius: 16,
    backgroundColor: COLORS.GREEN,
    justifyContent: 'center',
    alignItems: 'center',
  },
  boxLabel: {
    color: 'white',
    fontSize: 16,
  },
});
```

</details>


packages/react-native-gesture-handler/src/v3/detectors/HostGestureDetector.web.tsx
```diff
@@ -225,7 +225,15 @@ function teardown(refs: DetectorRefs) {
 const HostGestureDetector = (props: GestureHandlerDetectorProps) => {
   const { handlerTags, children } = props;
 
-  const handlerTagsSet = useMemo(() => new Set(handlerTags), [...handlerTags]);
+  // Spreading the array into the deps does not work: React compares
+  // deps only up to the shorter length, so a tag appended or removed at the end
+  // was never picked up.
+  const handlerTagsKey = handlerTags.join(',');
+  const handlerTagsSet = useMemo(
```
j-piasecki: This set is used only as a dependency of an effect below. Can't it be simplified so we don't memoize the set and instead move the handlerTagsKey to the effect's dependency array?

m-bert: Done in [63f0b76](https://github.com/software-mansion/react-native-gesture-handler/pull/4526/commits/63f0b769540c053c4e0582bca4348b16b5f60787)

